PR #1, opened 2017-01-31
Fix reducers not being functions in `update` description

evgenykochetkov:


jbucaran:
@evgenykochetkov Thank you :)

PR #3, opened 2017-02-01
Removed unused parameter + minor optimization

tzellman:


jbucaran:
@tzellman Thanks. The bundle size was also reduced by 1 byte!

PR #17, opened 2017-02-04
Tree needs to be an array for JSX

maraisr:
With the follow example using JSX the code will fail, hence the need to cast the `tree` argument as an array if not already.

```jsx

const model = {};

const update ={};

const view = (model, msg) => (<div><h1>Test</h1></div>);

app({model, view, update});
```

### DIFF
--- a/h.js
+++ b/h.js
@@ -3,6 +3,8 @@ module.exports = function (tag, data, tree) {
         svg(tag, data, tree)
     }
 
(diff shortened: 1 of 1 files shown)


jbucaran:
Thanks @maraisr! 🚀

jbucaran:
https://github.com/hyperapp/hyperapp/releases/tag/0.0.8

PR #18, opened 2017-02-04
JSX tree needs to default the data, and tree to an array.

maraisr:
May as well just default these in the params.


maraisr:
Hang tight, will have to modify the PR so that we can build out es6 stuff through uglifyjs.

jbucaran:
@maraisr I think that will break in some older versions of node.

maraisr:
Either way think we need to default the `data` to an empty object when it is likely to be sent in as null through babel's jsx transform, same with the `tree` param. It will be sent in as `undefined` hence, we need to test for undefiend and create an empty array rather than creating an array of empty.

maraisr:
On top of that, do you think we should start writing some tests? Ill create a PR for testing, you happy to use Jest?

jbucaran:
@maraisr Tests, yes! I'm not familiar with Jest. I'll have a look. What do you think about https://github.com/substack/tape?

maraisr:
@jbucaran yeah man, that'll be sick.. Tape is nice, IMO jest just has a much nicer cli output I've created a new pr #19 kinda screwed up git in this pr...

maraisr:
@jbucaran #21 is the new PR...

jbucaran:
Tape is a format agnostic, so if we wanted a nice reporter, we'd just pipe tape's output to a reporter :)

PR #20, opened 2017-02-04
Added JSX example to readme

maraisr:

PR #21, opened 2017-02-04
Support es2015 in source, as well as fix jsx bugs

maraisr:


maraisr:
From my #20 pr the example won't actually render, will investigate further.

jbucaran:
@maraisr I think my comments got lost in your previous PR.

I'd prefer to not support es2015 in the source as that would require using something like babel to compile HyperApp. If you don't mind, let's fix that.

* https://github.com/addyosmani/es6-equivalents-in-es5#rest-parameters

jbucaran:
One more thing, I'll take care of bundling the library, so you don't have to npm run build yourself locally :)

maraisr:
Cool mate - that's fine, you're the product owner, I'll code to that! Will amend this PR, with a fix 😅🙈

selfup:
Another way to take in the arguments with ES5:

```javascript
var someFunc = function() {
  var args = [].slice.call(arguments);
  return args;
};

var exampleArgs = someFunc(1, 2, 3, 4);

console.log(exampleArgs);
```

From: https://github.com/addyosmani/es6-equivalents-in-es5#array-like-object-to-array

package.json
```diff
@@ -17,14 +17,19 @@
         "vanilla"
     ],
     "scripts": {
-        "build": "browserify app.js -s app -g uglifyify -p bundle-collapser/plugin | uglifyjs > dist/app.min.js && browserify html.js -s html -g uglifyify -p bundle-collapser/plugin | uglifyjs > dist/html.min.js && browserify h.js -s h -g uglifyify -p bundle-collapser/plugin | uglifyjs > dist/h.min.js",
+        "build:app": "browserify app.js -s app -g uglifyify -p bundle-collapser/plugin -t [ babelify --presets [ es2015 ] ] | uglifyjs > dist/app.min.js",
+        "build:html": "browserify html.js -s html -g uglifyify -p bundle-collapser/plugin -t [ babelify --presets [ es2015 ] ] | uglifyjs > dist/html.min.js",
+        "build:h": "browserify h.js -s h -g uglifyify -p bundle-collapser/plugin -t [ babelify --presets [ es2015 ] ] | uglifyjs > dist/h.min.js",
+        "build": "npm run build:app && npm run build:html && npm run build:h",
```
tunnckoCore: yea. definitely looks better than one whole line

jbucaran:
> @selfup  Is the html tagged template string transpiled already when hyperx is pulled in?

Yes. Browserify will transform the string literals into something like this.

```js
var _templateObject = _taggedTemplateLiteral(["\n<h1>Hi.</h1>\n"], ...)
```

Using a browserify transform like yoyoify we should be able to those even further into `h` calls, skpping Hyperx HTML tokenizing.

Alternatively, using Babel we can achieve the same, but using JSX instead of Hyperx.

selfup:
@jbucaran Ahh that is very interesting (the `yoyoify` part). Thanks for the detailed explanation!

PR #23, opened 2017-02-04
Fixed #22

### LINKED ISSUE #22: Errors for input tags in template strings
A simple demo: 

```
app({
  model:  'hello',
  view: model => html`<div><input type="text" value="anything"></div>`
})
```

... errors out with:

`TypeError: undefined is not an object (evaluating 't.data')`

See here [http://codepen.io/anon/pen/mRKJdZ](http://codepen.io/anon/pen/mRKJdZ)

It can be simplified even further: just a plain `<input>` with no attributes has same result. Also trying to close it various ways ( `<input />`, `<input></input>` ... ) makes no difference.

Looks like it happens in the hyperx part of the code. However, I'm not familiar enough with hyperx to debug it or know how to submit a bug report to them. Or perhaps I'm doing something fundamentally wrong?

tzellman:
Only create/append child nodes if they are present; this could be due to a bug in upstream hyperx, creating undefined nodes in the input tree


jbucaran:
@tzellman Let's keep this one open, as it might be necessary when we bring correct JSX support.

tzellman:
Yeah, I actually reverted that specific change just now and updated h.js:

`tree = tree ? (Array.isArray(tree) ? tree : [tree]) : []`

I see that since making the PR you reverted that line of code, so I would say we ignore/close this PR.

dodekeract:
> bring correct JSX support

Will this be optional?

jbucaran:
@dodekeract Yes. Using the `html` function is also optional too by the way.

dodekeract:
@jbucaran How would this work without `html`?

jbucaran:
Using the `h` function directly.

```js
app({
    model: 0,
    update: {
        add: model => model + 1,
        sub: model => model - 1
    },
    view: (model, msg) => h("div", {}, [
      h("button", { onclick: msg.add }, ["+"]),
      h("h1", {}, [model]),
      h("button", { disabled: model <= 0 }, onclick: msg.sub }, ["-"])
    ])
})
```

[**CodePen**](http://codepen.io/jbucaran/pen/jyKypX?editors=0010)

jbucaran:
@dodekeract 

Using the `h` function directly.

```js
app({
    model: 0,
    update: {
        add: model => model + 1,
        sub: model => model - 1
    },
    view: (model, msg) => h("div", {}, [
      h("button", { onclick: msg.add }, ["+"]),
      h("h1", {}, [model]),
      h("button", { disabled: model <= 0 , onclick: msg.sub }, ["-"])
    ])
})
```

[**CodePen**](http://codepen.io/jbucaran/pen/jyKypX?editors=0010)

PR #26, opened 2017-02-05
Use addEventListener for popstate/hashchange events

itrelease:
With this approach if user will set his own handler on `popstate/hashchange` event it will not reset library behaviour


maraisr:
Nice pick up @itrelease mate! Consider this merged! :100:

jbucaran:
addEventListener is not available in IE8, which is why I had it like that, but @itrelease brought a good point which I had not considered.

Thanks!

maraisr:
We should look at settling on a "this is what we're going to support", so I don't merge stuff again....

But you're right in that regard.

jbucaran:
@maraisr That's alright, you made the right choice!

PR #27, opened 2017-02-05
Set inline handler for non-function values

itrelease:


itrelease:
For things like ```<div onclick="alert(1)">whoa</div>```

jbucaran:
> @itrelease For things like `<div onclick="alert(1)">whoa</div>`

Ah, I get it. 🤔 
Seems like a pointless thing to do that would just increase the bundle size, when you can simply do instead:

```js
<div onclick=${_ => alert(1)}>Whoa</div>
```

itrelease:
@jbucaran fair point I guess

jbucaran:
@itrelease Awesome! Thanks.

PR #28, opened 2017-02-05
Render vdom to string

itrelease:


jbucaran:
@itrelease Interesting PR. I'll take some time to review this. 

I can imagine renderToString is useful mostly for rendering on the server, but how does it work and what is it doing exactly?

A few questions:

* Can we provide this functionality as a separate package/module?
* Can we remove all the dependencies?

itrelease:
@jbucaran for now you can get string with this code:
```
app({
  view: ...,
  model: ...
  renderToString: true
});
```
but yeah I will try to move it to separate module and maybe remove deps

SkaterDad:
I put together an attempt at rendering to string also,  including some tests and plenty of comments. 

https://github.com/SkaterDad/hyperapp/tree/ssr?files=1

I was going to rewrite the tests to use raw 'h'  function calls instead of the hyperx tagged templates. 

Feel free to use whatever code you want from there.   I'm pretty satisfied with the renderToString function.

jbucaran:
@SkaterDad Awesome!

@itrelease Can we leave a record here that explains the purpose of renderToString? Like, why would I want to use it and where will I use it?

What purpose does it serve to use it in the dev console for example? Debugging?

itrelease:
@jbucaran I think the main reason to have such feature is server side rendering

maraisr:
Thanks for the PR mate! Love the initiative. @jbucaran might disagree, but in my opinion, we're slowly moving away from the "1kb selling point".

@itrelease do you think we can do something like what RxJs does, in that if you want the `map` operator, you have to import it. Something like:

```js
import {app} from 'hyperapp';
import 'hyperapp/ssr';

app({model, view, update, renderToString: true});
```

That way you're not bloating the build file for those that won't ever use SSR.

I've done something similar with pulling out the SVG support "straight out of the box" for those that won't ever deal with SVG's, and if they do it's as simple as `import 'hyperapp/namespace/svg';`

itrelease:
@maraisr I absolutely agree with you, I've created separate `renderToString.js` module so main bundle won't affect in size.

jbucaran:
> @maraisr  we're slowly moving away from the "1kb selling point"

There are plenty of >1kb-large libraries out there. Part of this project's value is in its small size and the freedom that comes with it. 

Not about faster bundle downloads, but the fact you can keep the entire system in your head.

Take React or Angular, no doubt they're open source, but the source code is so obfuscated they might as well be closed source.

terkelg:
> the fact you can keep the entire system in your head.

Word. It's something especially [new developers](https://github.com/hyperapp/hyperapp/issues/59) appreciate. It's hard to write elegant and self-documenting code, and @jbucaran has done an amazing job with that.

I'm a fan of isomorphic rendering too and need it in my own projects. I can look into making a tiny module for this without introducing extra dependencies.

jbucaran:
@terkelg Please 🙏

itrelease:
@jbucaran in case you've missed I moved `renderToString` to separate module so client bundle won't bloat in size and deps are not that critical for server side. But I still think that pursuing no dependencies is not that good. I mean for example current virtual dom implementation is kinda naive. It doesn't handle cursor position in input fields as well as selection etc. I think it would be better to pick some mature implementation where already fixed many bugs and have community that focused on making it better

tunnckoCore:
What about https://github.com/tunnckoCore/mich-to-html :P Pretty simple and I believe enough. It seems it outputs everything correct.

The needed change here, would be changing some names of properties in the virtual dom implementation :) Like `data` -> `properties` and `tag` -> `tagName`. It is internal change, so it won't affect anyone.

jbucaran:
Just like we provide a router out of the box, we should provide this toString functionality out of the box too.

@SkaterDad @itrelease You folks have time to look at this?

jbucaran:
@itrelease We've created a repo here: https://github.com/hyperapp/server to move forward with SSR. See you there!

PR #29, opened 2017-02-05
Modernify: More ES6 idioms, use webpack to create a UMD bundle 📡 🛰

maraisr:


jbucaran:
@maraisr Thanks for the initiative, but I am failing to see what webpack is bringing to the table for HyperApp.

maraisr:
@jbucaran please see my comment over at https://github.com/hyperapp/hyperapp/issues/30#issuecomment-277568413

jbucaran:
@maraisr What's the advantage of having an UMD module?

maraisr:
Following your comment https://github.com/hyperapp/hyperapp/issues/30#issuecomment-277571278

>So, we still need to provide each separately.
Actually not, within a node environment, you'd do something like `const {app, h, html} = require('hyperapp');` right, which points to the `src/index.js` file which exports all three of those as an object. 

And with tree-shaking, your bundler should be stripping out anything it's not using. Thanks [RollUp](https://github.com/rollup/rollup) who started the trend.

jbucaran:
@maraisr Yeah, index.js is provided for node environments.

maraisr:
>What's the advantage of having an UMD module?

So people using commonjs can import using require, people using requirejs can import using the define method, or global, or on window. (So people can grab the build file, and use `hyperapp.app` straight in a script tag without needing to bundle)

jbucaran:
@maraisr Wait, I still don't get it.

People using node, can use:

```js
const { app, html } = require("hyperapp")
```

And people grabbing the build file directly in the browser don't do anything since `app`, `html` or `h` are the global namespace exported in the browserified bundles (app.min.js, html.min,js, h.min,hs, etc.)

maraisr:
To put it simply, the only reason the `dist/` folder exists is so we can host our lib on cdnjs, or rawgit, etc... Everyone else should be using the src folder because everyone has their own method of bundling. Some people run their app in an enterprise environment, would use chrome 54 so es6 is supported natively so why serve them es3?

maraisr:
> And people grabbing the build file directly in the browser don't do anything since

Yes, but what I'm saying is there is no reason to serve these files separately... when the `app` bundle contains all they'll need in 1 call, that is far smaller than the "bundle" of three files.

maraisr:
UMD: requirejs, commonjs, all people using browser side bundling and not through a bundler (node).

See http://requirejs.org/docs/start.html

jbucaran:
> @maraisr Yes, but what I'm saying is there is no reason to serve these files separately... when the app bundle contains all they'll need in 1 call, that is far smaller than the "bundle" of three files.

By the way, in a real production environments, users will likely be using also https://www.npmjs.com/package/yo-yoify to transform html templates into raw h calls or babel to transform JSX into raw h calls.

So, bundling `html.min.js` into a single unified module is total madness.

maraisr:
> So, bundling html.min.js into a single unified module is madness.

Yes but if you're using tree-shaking (which you should be in 2017), you'd be going `import {app, h} from 'hyperapp';`, and then your bundler should be statically analysing your source, seeing the un-used export of `html`, and thus pulling it out.

maraisr:
> By the way, in a real production environments, users will likely be using:

Real production environments, you wouldn't be using the `dist/` folder bundle, and therefore you wouldn't have the bloat of a `html` import being a problem in the first place, because you're only importing what you're needing.

jbucaran:
@maraisr Good. What about the CDN?

maraisr:
CDN: You use the `dist/` folder, which is UMD so you've got every bundler supported out of the box, requirejs, commonjs, webpack externals, etc....

maraisr:
I guess what I'm more or less leaning toward is, that this project has a claim to be a "modern ui framework" yet failing to use any modern technologies. Like really... Who in 2017 would start a greenfield project and consider using browserfy?

jbucaran:
@maraisr For environments where the user is bundling their application I think you've gotten your point across good, so I see introducing webpack, rollup or similar will likely be happening. But what about users grabbing the script via CDN?

Do we have to force them to download the entire bundle when they can just pick which one they want?

maraisr:
> Do we have to force them to download the entire bundle when they can just pick which one they want?

In short no. We can go down the route that `firebase` did, in that you can download the entire bundle, or you can grab a `base` + `html` so 2 urls... Which webpack can solve too, using the `CommonChunks` plugin.

jbucaran:
@maraisr Using html is really just for prototyping or small-ish projects, in **production** you **always** want to transform those `html <h1>Hi</h1>` into raw `h` calls, so only app.min.js and h.min.js is what most people will actually be using in the real world.

One more thing, using the unified bundle means that users grabbing directly will need to do:


```
const { app, html } = require("hyperapp")
```

Right?

maraisr:
@jbucaran correct. Which "simulates" what they'll be doing in node anyway.

jbucaran:
@maraisr Makes sense, since that's what you'd do in a node environment, right. **Ace** ✋

jbucaran:
Close in favor of #37.

PR #31, opened 2017-02-05
Add tests into workflow

maraisr:
This is still a work in progress, will ping you here when this is ready to role.


jbucaran:
@maraisr I think I'm going to go with [tape](https://github.com/substack/tape) for tests.

maraisr:
Tape doesn't offer a headless browser not unless you use `tape-run`, but even then using electron and speed were all just too cumbersome. Jest supports all this out of the box. And having Facebook behind it, means it's well supported. Don't think going with tape would be a good move.

jbucaran:
@maraisr That sounds great, but what's the catch then? 🤔

maraisr:
@jbucaran  catch? None! haha :100:

jbucaran:
@maraisr Interesting. I don't think we'll introduce webpack though, so can you keep this PR focused on tests only? 🙏

jbucaran:
@maraisr Think we can close this?

maraisr:
Closed in favour of #52

PR #36, opened 2017-02-06
Prevent default browser behaviour compliant with specs.

terkelg:
As specified in the DOM 2 Events specification.
I also find this more transparent and understandable.

It's a minor thing, and the downside is a few more characters.
Feel free to close RR if you disagree.

_Thanks for your work on HyperApp. 🥇_

### DIFF
--- a/app.js
+++ b/app.js
@@ -58,7 +58,7 @@ module.exports = function (options) {
 
                 } else {
                     msg.setLocation(target.pathname)
-                    return false
+                    e.preventDefault()
                 }
             }
         }


jbucaran:
LGTM.

Should we call `e.stopPropagation` too? or is it safe to ignore that one?

jbucaran:
@terkelg LGTM. Should we call e.stopPropagation too? or is it safe to ignore that one?

terkelg:
Good question. I think it’s enough, since the goal is to just prevent the default browser behaviour, that's what you want in most situations. 
I can’t think of a reason to stop propagation, it's default event behaviour, so I assume it's safe to ignore ```e.stopPropagation```.

jbucaran:
Thank you @terkelg!

terkelg:
NP, it's a super interesting project. Keep up the good work!

PR #37, opened 2017-02-06
Modernize

jbucaran:
* More ES6 idioms, and use webpack to create a UMD bundle.

See also: #38


tunnckoCore:
```js
const Wrapper = require("wrapper-webpack-plugin")
```

unused

Also, why set all options of uglify to true explicitly, they defaults to true - only `warnings` is true by defaults.

jbucaran:
Forgot to remove it! Thanks @tunnckoCore.

PR #42, opened 2017-02-06
Use let instead of var, since we are committed to ES6/babel

tzellman:
Also, removed extraneous console.log

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -1,11 +1,10 @@
-console.log("DEV")
 module.exports = options => {
 	const defer = (fn, data) => {
 		setTimeout(_ => fn(data), 0)
 	}
 
 	const merge = (a, b) => {
-		var obj = {}, key
+		let obj = {}, key
 
 		if (isPrimitive(typeof b) || Array.isArray(b)) {
 			return b
@@ -39,10 +38,10 @@ module.exports = options => {
 			}
 
 			if (index >= 0) {
-				var element = parent.childNodes[index]
+				let element = parent.childNodes[index]
 
 				if (oldNode && oldNode.data) {
-					var hook = oldNode.data.onremove
+					let hook = oldNode.data.onremove
 					if (hook) {
 						defer(hook, element)
 					}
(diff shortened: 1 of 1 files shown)


jbucaran:
@tzellman Because of the way let is transpiled, this increases the bundle size like 200 bytes, which is really uninspiring, so I can't merge this one, but I' certainly getting rid of that console.log haha! 😳

tzellman:
No worries - I noticed the increase in size as well.

jbucaran:
Thanks! 🙇

dodekeract:
@jbucaran Can you share why this happens? Why does a simple `var` -> `let` replacement increase bundle size?

tzellman:
Transpiling `let` to `var` causes extra temporary variables to be created in order to enforce that vars within a functional scope clash. These extra declarations can increase the file size.

[Here](https://babeljs.io/repl/#?babili=false&evaluate=false&lineWrap=true&presets=es2015%2Creact%2Cstage-0&targets=&browsers=&builtIns=false&code=function%20f()%20%7B%0A%20%20%20%20var%20i%3B%0A%20%20%20%20for%20(let%20i%20%3D%200%3B%20i%20%3C%205%3B%20%2B%2Bi)%7B%7D%0A%20%20%20%20for%20(var%20i%20%3D%200%3B%20i%20%3C%205%3B%20%2B%2Bi)%7B%7D%0A%20%20%20%20for%20(let%20i%20%3D%200%3B%20i%20%3C%205%3B%20%2B%2Bi)%7B%7D%0A%7D) is a link that might help show how babel creates temporary variables.

jbucaran:
Probably transforming something like this:

```js
for (let i = 0; i < 10; i++) {
 a[i] = ...
}
```

creates a separate function like:

```
var X = function(i) {
   a[i] = ...
}

for (let i = 0; i < 10; i++) X(i)
```

See [variable scoping and its pitfalls](http://www.2ality.com/2011/02/javascript-variable-scoping-and-its.html).

tunnckoCore:
ha! Cool. It seems that's the same using buble. Specifically if `let` is used in for loop, instead of var - it extracts the body of for loop outside of it so it becomes something like that

```js
  for (var route in routes) {
    var returned = loop( route );

    if ( returned ) return returned.v;
  }
```

Good catch.

PR #44, opened 2017-02-06
Converted remaining vars to const or let.

rbiggs:
For better block scoping, converted remaining occurrences of var to
const or let where reassignment was necessary.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -4,7 +4,8 @@ module.exports = options => {
 	}
 
 	const merge = (a, b) => {
-		var obj = {}, key
+		const obj = {}
+		let key
 
 		if (isPrimitive(typeof b) || Array.isArray(b)) {
 			return b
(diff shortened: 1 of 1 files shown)


rbiggs:
Nice job in converting app.js to ES6 syntax. 👍 I noticed there were a lot of vars still hanging around. For better block scoping in converted the remaining ones to const unless reassignment was necessary.   This builds and passes your jest test. I also dropped it into a local copy of your examples and they all work.

jbucaran:
Hi @rbiggs! Thank you, but this is the same as https://github.com/hyperapp/hyperapp/pull/42.

tunnckoCore:
This won't happen using Rollup + Buble. But there another thing would happen that will increase the bundle with 50-100 bytes (because the `route` function).

rbiggs:
Closing

PR #46, opened 2017-02-06
Use standard style

danigb:
Add standard code style. This PR includes:

- Change the all code (src, test and webpack configuration) to match standard style
- Add standard dependency to package.json 
- Run standard eslinter after running tests
- Add standard style badge to README

As a side effect, 5 bytes are saved from the dist/hyperapp.min.js file 😂 

EDIT: I removed the `wrapper-webpack-plugin` since (as the standard eslint reported) was not used.


terkelg:
I prefer the standard linter too. However, why not do the lint before running the tests?
There's no reason to wait for unit testing when the linter might fail - this is especially true when using a CI.

dodekeract:
Tabs -> spaces is a step in the wrong direction IMHO. Makes it hard to read if you don't coincidentally prefer the same indentation as the person that wrote it.

tzellman:
Ultimately it is up to @jbucaran - everyone has their own preferences (e.g. some people prefer 4-space soft-tabs) - in the end it doesn't affect file size. I knew this was going to come up at some point.. Maybe throwing a `.editorconfig` in will help enforce the spacing preference.

https://i.ytimg.com/vi/oBkMztuHlVg/hqdefault.jpg

tzellman:
As a matter of fact, @jbucaran used to use 4-space soft-tabs (see [here](https://github.com/hyperapp/hyperapp/blob/7c1b35fc465455c18230e39ab42875984816ae34/app.js)). Everything was updated to use tabs sometime in the last day..

SkaterDad:
I'm a big fan of 4 space soft tabs also.  It's so much easier to read than
2 spaces.  Ultimately it doesn't matter, though.  Linters can autofix that,
right?

tunnckoCore:
> Run standard eslinter after running tests

Not make sense. Lintings should always be the first thing.

As about tabs/spaces. I hate tabs and has chrome extension that converts to 2 spaces. Using tabs to me is still fucking ridiculous and looks awful. Don't know from where it comes but last months it is like a trend which I don't get, because 2 spaces is convention for years and most of the community use it.

dodekeract:
@tunnckoCore How does it look/act any different for you when you set the width to 2? Only thing it changes is that people that prefer 4-wide can get 4-wide without changing the files.

IMO forcing indentation-width is the same as hard-coding the syntax theme/colors into every file. How code gets represented should be only decided by the person looking at it, not the file.

tunnckoCore:
> How does it look/act any different for you when you set the width to 2?

Sometimes i see such code.

I'm not against anything. That's why `eslint` and `editorconfig` are awesome things.

tunnckoCore:
Actually, I'm using [TabSize on Github](https://chrome.google.com/webstore/detail/tab-size-on-github/ofjbgncegkdemndciafljngjbdpfmbkn?utm_source=chrome-app-launcher-info-dialog), so yea I see 4 spaces which isn't good to me.

danigb:
For sure there are good reasons to use tabs instead of spaces. What it's interesting about a linter is:

- Help to find bugs (and therefore, it raises the code quality)
- Makes the codebase more homogenous and easy to read when are many contributors.

I don't mind what style we choose. I used `standard` because it's easy to setup and the actual code is quite similar. 

So please, let's be constructive here and propose and alternative if `standard` is not an option. I will follow the path. As soon as we get a linter, the better.

danigb:
@terkelg said:

> However, why not do the lint before running the tests? This makes most sense to me.

The reason is that eslinter takes time. When you run tests a lot, I prefer to check if they are working, and then check the style (in fact, I'm using a eslint editor integration, so I see the style errors on my editor first). Also it's unlikely that eslint fails on the ci if works on local, and I assume everybody call `npm test` before commit ;-)

tunnckoCore:
I'm not against `standard` I just said my prefs about tab/spaces. I use `standard` too from its beginning and its most of my ~300 packages.

dodekeract:
@danigb [This](https://github.com/jlongster/prettier) looks interesting. It's not really a linter, it's more like a formatter that parses the original source into an AST and pretty-prints it.

There's also a [fork that uses tabs](https://github.com/arijs/prettier-with-tabs).

jbucaran:
@dodekeract Yes, I wanted to try prettier once, but then forgot about it.

jbucaran:
@danigb I was sloppy and didn't check that standard has no configuration (which is good, right), but they use single quotes, which require a bit of effort to type in my JPN layout keyboard. Shift+7.

I'm willing to consider prettier though!

rbiggs:
Hey guys, I guess you're not aware that you can use tabs or spaces and let git do the conversion for you when you check out and in. It can convert two spaces to tabs when you check out, then convert tabs back to two spaces when you check in. I had to do that a couple of years ago on a project. You basically set up a `.git/info/attributes` file in your local hyperapp repository with this:
`*.js  filter=tabspace`
Most likely the attributes file isn't in the hyperapp repo, so you'll need to create it and add the previous code to it. That's all. Next you'll  have to run some git commands. Umm... If I remember right it's something like this:
```
git config --global filter.indent.smudge  'expand -t 2'
git config --global filter.indent.clean  'unexpand -t 2'
```
That should convert two spaces to tabs and back. Now, as I said, I have used this in four years but I'm pretty sure that is what I was using. I'd test it out on a small test repo first. Smudge is the checkout part and clean is the check if I remember right. If this doesn't work, look up `filter.indent.smudge` in the git docs. I could be going senile.

jbucaran:
@rbiggs I had no idea about this. TIL.

rbiggs:
Yup, well test it out first. I haven't used in in a few years. Try it on some test repo. If it works you should make that as a requirement for collaboration. Put the setup in your docs for collaboration. That way people can use the tabs/spaces they want and you get the two spaces you want. By the way, I with you on the two spaces thing. Guess all those years of Node dev have me set in my ways. LOL

dodekeract:
@jbucaran TIL `git` solved the holy war between developers.

jbucaran:
Ok, let's close this issue since I don't think we'll be introducing standard, but I am open to suggestions wrt to `prettier`. If anyone wants to give it a shot and report the results, even better! :) 👋

Issue #38, opened 2017-02-06
Modernize


jbucaran:
⚠️ **A decision was made. See the [aftermath](https://github.com/hyperapp/hyperapp/pull/48#issuecomment-277988927).**

---

@maraisr Summarizing the discussion in #29, #30, #31 and #32 we'd like to modernize the code base to use more ES6 idioms and introduce webpack to bundle the distribution.

### Rationale

* Using more ES6 idioms may encourage more contributions.

* Using more ES6 idioms means a smaller bundle for users targetting really modern browsers. For example, in production users may consume our `src` directory directly with their application and don't babelify (or maybe babelify a _some_ things via plugins) and use something like babili to minify modern JavaScript. This means even smaller bundles.

* Other than prototyping, examples, presentations and maybe even real small-ish apps, users will want to bundle their apps using a bundler like browserify, webpack, rollup, etc.

jbucaran:
@SkaterDad There was no consideration since I'm not versed particularly in either rollup or webpack , but I'll take whichever can produce smaller bundles.

tunnckoCore:
Btw, I don't think `iife` is the right thing.

SkaterDad:
@tunnckoCore I used `iife` in that config since @jbucaran had Webpack outputting something similar.

Since that bundle is for the browser, I think it's correct (per [this tutorial](https://code.lengstorf.com/learn-rollup-js/)).

I've heard it said before that Rollup is a great choice for library authors.  Vue.js uses it with Buble (instead of Babel).

jbucaran:
@tunnckoCore We've actually got a smaller bundle, so I'd like to think the current setup is a step in the right direction. @SkaterDad I'd really like to give rollup a try though. Can you help with this?

tunnckoCore:
@jbucaran @SkaterDad I use Rollup + Buble everywhere every day too. It is just amazing combo. But in some cases it is not cool, because Buble adds more bytes as needed - I'm fighting that today and end up just to not use buble haha and write vanilla es5. In my current case 50-100 bytes is important - because the promo slogan. Here if use Rollup + Buble the case would be the same, because of the [app.js#L152-L169](https://github.com/hyperapp/hyperapp/blob/master/src/app.js#L152-L169).

jbucaran:
@tunnckoCore Totally! Try to make the bundle smaller 🙏

SkaterDad:
[...]

PR #48, opened 2017-02-06
Rollup integration

maraisr:


tunnckoCore:
You can remove `babelrc` and `babel` devDep. It is not used.

As about the rest, seems good workaround. I'm working on @rolldown (built on Rollup) which tries to fix some things and adds more features. So in the near days we can use rolldown cli with more simpler config. :)

maraisr:
@tunnckoCore no it is used, Jest needs it.

tunnckoCore:
damn.. ;d

SkaterDad:
Looks like this saved a few hundred bytes per minified/gzipped file?  👍

maraisr:
@SkaterDad thanks for the suggestion.

jbucaran:
@maraisr Something is off.

```
MODULENAME OUTPUTFILE                SIZE  
app        ./dist/app.min.js         3.01kB
app        ./dist/app.min.js.gz      1.37kB
html       ./dist/html.min.js        276B  
html       ./dist/html.min.js.gz     198B  
h          ./dist/h.min.js           246B  
h          ./dist/h.min.js.gz        177B  
hyperapp   ./dist/hyperapp.min.js    3.26kB
hyperapp   ./dist/hyperapp.min.js.gz 1.49kB
```

html.min.js doesn't make sense or rollup is using some alien gzip technology from the future.

jbucaran:
It seems rollup does not bundle non-ES6 modules.

SkaterDad:
Can this help? 
https://github.com/rollup/rollup-plugin-commonjs

maraisr:
@SkaterDad already using that mate... 

@jbucaran fixed mate.. Rollup doesnt accept the `require` syntax, so just converted it to a `import` block.

terkelg:
By switching to ES2015 import/export syntax, we break node support and potential future isomorphic rendering. It will no longer be possible to run the project without using build tools.

I feel this conflicts with the core principles of HyperApp. The idea behind using Hyperx to turn template literals into vNodes is to avoid unnecessary build tools, and take advantage built in language features for quick prototyping. (Compared to JSX which requires transpiling).

Now that the project require more complex build tools (transpile etc) we might as well switch to use JSX and remove the only dependency left. However, in my opinion that's the wrong direction.

SkaterDad:
Rollup can output multiple formats, so in addition to the browser builds, a Node compatibile build could be made also. 

Maybe check how Vue set up their build?  I would guess that codebase is full ES2015, but they support SSR.

maraisr:
@terkelg what you fail to understand is, that this is purely for cdn purposes. To have a build dist ready to go for quick prototyping... In real production use cases you'd use the import syntax as its spec, to ecma. And you'd end up using yo-yoify for template literals or jsx all directly yo h calls.

maraisr:
@SkaterDad sure - technically speaking we could produce a "node" version as well. Probably best we do that. Seeing as we do have a few optimizers and things going on.

jbucaran:
@maraisr Thanks for this PR. I'll take over now and see if I can get a simpler configuration, using rollup, but creating a UMD bundle.

jbucaran:
@terkelg I agree with you 💯, but if we can generate the same bundle we previously had using browserify, but smaller, what's to lose? Sure, we introduce a compilation step in this repo, but that's a burden we authors surely can bear.

* Those of us that want to use go on and use tagged literals + HyperApp will bundle our apps using `browserify -t hyperxify` or yoyofy, etc., with the rollup generated UMD hyperapp.min.js.

* Those of us who want to use JSX + HyperApp will bundle our apps using webpack or rollup with the rollup generated UMD app.min.js and h.min.js or perhaps we can roll out a hyperapp.lean.min.js or something like that.

maraisr:
What's wrong with this solution?

maraisr:
I mean it's as simple as using "UMD" as the format type?

jbucaran:
@maraisr Nothing. I'm using this as a base of course.

terkelg:
Sure I get your point and I perfectly understand that. It's just that there's been added a lot of complexity and extra dependencies over the last two days with very little benefits.

If the only thing needed was an UMD build, it could have been done by appending a single [browserify transform](https://www.npmjs.com/package/browserify-umdify) to the build command.

@SkaterDad I know, but it still requires a build tool to get that file.

jbucaran:
@terkelg I think we were (or only myself) misunderstanding the whole thing. If we produce a UMD bundle, then HyperApp consumers can build their app using browserify -t hyperxify if they so prefer to.

terkelg:
UMD is a bundle for production which definitely should be a of the build process and It's perfect for consumers.We can even host a CDN directly from NPM using https://unpkg.com/.
I totally agree that it's a good idea to provide different builds for end consumers.

However, that was possible two days ago too with Browserify. It was just a matter of adding a new line to ```package.json``` called ```build:umd``` for example and use a Browserify transform for that.

It wasn't necessary to convert the whole code base to ES2015, break node support, remove the possibility to use hyperxify out of the box and add extra dev dependencies.

But to use Rollup to build a UMD bundle, and then use Browserify to build that bundle makes no sense to me. Then it's better to just stick with one build step, and find a solution to the hyperxify problem.

tunnckoCore:
@terkelg what extra deps? All changes are just completely for the development flow and improving HyperApp and some automation.

The whole thing that we should do is

1. `dist/hyperapp.es.js` - `pkg.module` - points to the ES2015 **bundle** (rollup's `es` format) - not minified
2. `dist/hyperapp.umd.js` - `pkg.main` - points to the UMD bundle (rollup's `umd` format) - not minified

It is a bit and kinda tricky. The `iife` format is the smallest, but it is not cool and not flexible. So best is the umd format for both browser and node.

The umd is must for few reasons.

1) If you expose `iife` format to `pkg.main` it is good for Unpkg CDN, but not for browserify, nor nodejs
2) If you expose `iife` format to `pkg.browser` - okey for unpkg and node, but not for browserify
3) If above (2) and `umd` format to `pkg.main`, then it is okey for node and unpkg, but it is not clear what browserify will get (I just don't know, didn't dive deep)

So best is UMD on `pkg.main` and don't set `pkg.browser` in anyway - just for sure. Because if you set umd on `pkg.main`, es on `pkg.module` and iife on `pkg.browser`, then when someone use browserify over `require('hyperapp')` he will get the iife, which won't work.

Just played that game today, while building https://github.com/tunnckoCore/gibon

maraisr:
@terkelg  if you want to be stuck in 2015 then sure, stick to those. But for the rest of us wanting to use modern technologies, and code using modern languages. Then yeah rollup / es2015+ would be the way to go.

tunnckoCore:
Ooh.. and on more thing. The IIFE bundle is good just for the purpose to see REAL bytes (minus the sourcemap comment ~30-40bytes) on `app`, output that info and remove the bundle.

tunnckoCore:
> EDIT: I am just trying to see if it's possible to remove the build.js in favor of npm script only solution.

@jbucaran as about that.. no, it's best possible currently. Maybe we can remove `zopfli` and let it be as before few days. Or just that bash

```sh
_compress () {
  zopfli -h 2>/dev/null
  if [ $? -eq 0 ]; then
    zopfli "$1" -i1000 -c
  else
    gzip -c "$1"
  fi
}

build_gz () {
  _compress "dist/$1.min.js" > "dist/$1.gz"
}

build_gz "$@"
```


btw, @maraisr why you use `zlib` compression instead of `gzip` in the zopfli?

terkelg:
I think it's a healthy discussion, no reason to be harsh. I'm fine with ES2015, and I use both that and Rollup in some of my [own](https://github.com/terkelg/math-toolbox) projects. I also think now that it's included that we should go all in and make use of stuff like ```jsnext:main```and ```cjs:main``` in the package.json.

But I'm not a fan of converting every single function to arrow functions just because you can, and thereby make the code more cryptic for new developers. ES6 arrow functions are not even about saving keystrokes or readability, but lexical scope. They remove some of the dynamic features of the language.
 
I know how to use the build tools, but what about new developers? ES6 only increases your burden because it gives you more tools which you will need to learn to properly use HyperApp.

jbucaran:
Can you clarify?

> @terkelg  remove the possibility to use hyperxify out of the box

tunnckoCore:
> But I'm not a fan of converting every single function to arrow functions just because you can

I mentioned it https://github.com/hyperapp/hyperapp/issues/38#issuecomment-277895726

> which you will need to learn to properly use HyperApp.

Anyone can use it with just include the old school `script` tag or `require`. I can't get what they will need to learn.

Multiple builds are just plus. And for Rollup is just enough to see `export default`, so it wasn't need the whole convert to ES2015.

jbucaran:
> ES6 only increases your burden because it gives you more tools which you will need to learn to properly use HyperApp.

I don't understand what you mean. Users will consume the UMD bundle, either grabbing the CDN or on node. They are not forced to write ES6 anywhere.

terkelg:
@jbucaran You're right, and UMD is just that a final build for end users. I'm not talking about end users, I'm talking about developer experience in this project. Developers interested in contributing to HyperApp or read the code to learn how a framework works. This project is perfect for that since it's so simple and elegant. That's why I'm concerned with the added complexity.

jbucaran:
I've got some preliminary results of what I am thinking of, names are subject to change, but this LGTM.

@maraisr @rbiggs @terkelg @tunnckoCore @SkaterDad @danigb 
```fish 
7.5K hyperapp.html.min.js             # app+html+h, for browserify/hyperxify/yoyoify
3.1K hyperapp.html.min.js.gz        # app+html+h, for browserify/hyperxify/yoyoify
29K hyperapp.html.min.js.map     
3.5K hyperapp.min.js                     # app+h, for rollup/webpack/babel/JSX
1.6K hyperapp.min.js.gz                # app+h, for rollup/webpack/babel/JSX
13K hyperapp.min.js.map
```

jbucaran:
@terkelg Thank your for clarifying, now I see what you meant. My changes over this PR will drastically reduce the complexity, so hang on.

jbucaran:
@maraisr Can you create the branch on this repo instead of your fork? That way we can both push to the same branch, otherwise I'd have to send a PR to your fork.

maraisr:
There is another branch sitting on this repo. @jbucaran

jbucaran:
@maraisr I am still at work. When I get off I'll start wrapping this up for the 0.0.12 release.

maraisr:
Thank you again @terkelg we will take your input into consideration.

maraisr:
@jbucaran cool mate - let me have a look before we merge. Tend to agree with @terkelg we do overuse the consts / arrow functions. Anyway... All this "arguing" is doing my head in, it's all very subjective, and with my experience with OSS everyone has their opinions of how things ought to be done, which is fair enough...

But we just got to realize out of all of this - we're all humans, we all come from different parts of the world, Australia, Japan, and America! All three have different trends floating around in that part of the globe.

And this lib being in its infancy, it doesnt matter how "solid" we have things just yet. I mean Vue was re-written, Webpack was re-written, Angular is in its 3rd phase of a re-write. So like, just pick a bloody bundler and get it over and done with. 

The only thing I'll stand by and argue is - this lib is going to be written in es6+. This "cryptic" and nonsense around "new developers" is all just garbage. Almost all of the latest tutorial / help forums these days are all es6+ examples.

Anyway. Lets keep things positive boys.

jbucaran:
Thanks for the writeup @maraisr. I'm glad you're going to take it easy with this. 

Now, I am sorry I can't please everyone, but it's obvious we need to make a decision to move forward and start answering the other issues, write the manual and tests.

So, here is what I've settled on:

### `require`

**Because** it lets you prototype and require the source directly. In my value scale that scores higher than the other option.

I understand ES6 modules _is_ the spec, but last time I check that's not in node yet. When the time comes, worry not, I hope I can be the first to replace those couple of `module.exports` in HyperApp with `import/export`.

### Rollup

**Because** it beats browserified bundles in size considerably. Here's the no brainer config.

```js
import buble from "rollup-plugin-buble"
import uglify from "rollup-plugin-uglify"
import cjs from "rollup-plugin-commonjs"
import resolve from "rollup-plugin-node-resolve"
import zopfli from "rollup-plugin-zopfli"

export default {
    plugins: [
        buble(),
        cjs(),
        resolve(),
        uglify(),
        zopfli()
    ]
}
```

and npm scripts:

```json
"b": "mkdir -p dist && npm run b:all > /dev/null",
"b:all": "npm run b:hyperapp && npm run b:hyperapp:hx",
"b:hyperapp": "rollup -n hyperapp -cmf umd -o dist/hyperapp.min.js -- index.js",
"b:hyperapp:hx": "rollup -n hyperapp -cmf umd -o dist/hyperapp.hx.min.js -- hx.js",
```

### hyperapp.min.js
This is for JSX use and includes app and h.

### hyperapp.hx.min.js
This is fox Hyperx use and includes app, h and `hyperx(h)`. 

> I didn't come up with the `hx` prefix. It's what the author often uses as well. [See](https://github.com/substack/hyperx#virtual-dom-node-example).

### jest

I was going to go with tape, but I find jest really approachable and if I don't get out of my circle of trust I'm never going to learn new things either.

PR #52, opened 2017-02-07
Add more tests

danigb:
Add some tests. coverage: 65.36%
Work in progress... but some help needed (I've placed TODO and FIXME marks inside the test code)


SkaterDad:
I've noticed that all of the tests in this PR and in `master` define their views with the `hyperx` tagged templates.

I think it would be a good idea to duplicate those tests using raw `h` function calls also, to isolate any potential bugs that could be introduced by `hyperx`.  While messing around with SVGs the other day, I noticed some bugs from `hyperx` that would not occur if defining the v-dom nodes by hand.

I think this is especially important if JSX is being officially supported, since it all gets transpiled down to `h` calls.

Thoughts?

jbucaran:
@SkaterDad Makes sense.

@danigb You'll want to rebase this after I merge #54.

jbucaran:
@danigb Now that 0.0.12 has landed, we can make this the next priority.

danigb:
I've just rebased with master. Some tests are skipped (see comments).

jbucaran:
Seems you'll need to rebase again! Sorry! 🙏

danigb:
No problem. Here you go.

maraisr:
Thanks @danigb  ive re-directed this PR to point to a tests branch, where i've cleaned up a few stylistic things.

danigb:
Hi @jbucaran & @maraisr 

I've just rebased again against tests branch. You had replaced `var` with `const` but that make the test failing. I've used `let` instead. 

> Can you please sync your branch with this new upstream branch, and resolve conflicts with a -x Theirs

What is `-x Theirs`? I'm afraid I didn't do it... 🤕

test/app.spec.js
```diff
@@ -24,6 +24,11 @@ describe("App", () => {
     app({ model: {}, view: () => (html`<div>Hi</div>`) })
   })
 
+  // FIXME: if fails... this is not true
+  it.skip("allow all params to be optional", () => {
+    app()
```
jbucaran: What would be the point of doing this though?

jbucaran:
> I've noticed that all of the tests in this PR and in master define their views with the hyperx tagged templates.

> I think it would be a good idea to duplicate those tests using raw h function calls also, to isolate any potential bugs that could be introduced by hyperx. While messing around with SVGs the other day, I noticed some bugs from hyperx that would not occur if defining the v-dom nodes by hand.

> I think this is especially important if JSX is being officially supported, since it all gets transpiled down to h calls.

I thought about it, and I think using we need to test for both.

The `html` function we use to create virtual nodes with HyperApp, is ***not*** hyperx. I was confused about this too at some point, so forgive me if you already understand this.

Let me explain the difference between `html` and Hyperx, so that we can be on the same page.

Hyperx is a [template tagged literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_template_literals) function **factory**. 

Hyperx is not a tagged literal function, but rather _returns_ one, let's call it `hx`. 

Hyperx also accepts a function `h` and other options. This function `h` we hand to Hyperx will be invoked via `hx` as the Hyperx tokenizer parses the HTML.

```js
const hx = hyperx(h)
hx`<h1>Hi</h1>` // h( h1, {}, [ 'Hi.' ] )
```

In HyperApp, we call `hx`, **`html`** instead, for convenience. It also aids syntax highlighters in some editors, GitHub's markdown parser, etc.

---

I think we want our tests to check that the function get got from Hyperx, `html` is correctly calling `h`. HyperApp's implementation, due to its tiny nature has a tightly coupled parts, so I believe [functional](http://stackoverflow.com/questions/2741832/unit-tests-vs-functional-tests) tests are also important to have.

test/app.spec.js
danigb: Because that's what the README says... We have to 1) change the README and specify what params are not optional or 2) change the code...

test/app.spec.js
```diff
@@ -97,6 +97,22 @@ describe("App", () => {
   });
 })
 
+describe("Views", () => {
+  it("allows inline styles as object with properties in camelCase", () => {
```
jbucaran: @danigb Can you explain this test? HyperApp is not really doing anything here, since we're calling setattribute, the DOM expects the property to be camelCase anyway.

danigb:
Ok, I've added a couple of tests more. 

> I want to rework some of those tests where you're using let, the ones testing lifecycle events, etc. I think we can do even better.

Ok, go ahead. I'll stop here until you apply those changes. Maybe you can merge this PR and work in your own branch.

jbucaran:
> Because that's what the README says... We have to 1) change the README and specify what params are not optional or 2) change the code...

Oh, you're right, lol.

@danigb I was thinking about working on them after merging this PR. 

You done adding tests? What's missing?

danigb:
The coverage is right now at 70%. Run the tests and open `hyperapp/coverage/lcov-report/app.js.html` to see what is missing.

I had problems with the routing tests. They are skipped (take a look at them).

test/app.spec.js
danigb: Yes, probably the name is not well written. I just want to cover https://github.com/hyperapp/hyperapp/blob/master/app.js#L194 lines...

PR #54, opened 2017-02-07
Stability

jbucaran:
Require, rollup, UMD bundle for JSX/Hyperx and jest. 

See https://github.com/hyperapp/hyperapp/pull/48#issuecomment-277988927 for details.


jbucaran:
@maraisr The third tests passes locally, but fails on Travis. Do you have any idea why?

jbucaran:
@tunnckoCore Thanks!

tunnckoCore:
What about to set a convention for both builds - "lite" and "full", it more clear to mind?

- `dist/hyperapp.lite.js` - JSX Pragma / Hyperx / Browserify users - just `app, h` - like _~1.5kb_ - currently it is bundled from `index.js`
- `dist/hyperapp.full.js` - full build - `app, html, h` - like _~3kb_ - currently it is bundled from `hx.js`

Don't think it make sense to have `.min` ext.

jbucaran:
I agree about the .min extension being useless. I'll check what others are doing too.

🤔 Hmm, I am not sure. Lite would suggest that the distribution is a lightweight version of HyperApp, not really obvious that it's for JSX either.

I think that most people will probably end up using JSX anyway, so it makes sense to name that bundle just hyperapp. 

Using template literals is via Hyperx, which is often abbreviated to hx, so I found hyperapp.hx.js very logical.

Notice that in node, when using Hyperx, you'll need to import:

```js
const { app, html } = require("hyperapp/hx")
```

PR #58, opened 2017-02-08
Updated packagejson to remove extra spaces

maraisr:


tzellman:
Not that I care, but the commit message doesn't completely match the change.

PR #60, opened 2017-02-08
Tests

jbucaran:


test/app.spec.js
```diff
@@ -1,53 +1,308 @@
-/* global describe, test, expect */
+/* global beforeEach, describe, it, expect */
 
-const { app, html } = require("../hx")
+const { app, html, h } = require("../hx")
```
dodekeract: Shouldn't this be `const {app, html, h} = require('../')`?

PR #64, opened 2017-02-08
Create the default root element and initial render after the document is ready

tzellman:
It seems safest to wait until the document is ready before we attempt to:

* create the default root element
* initially call `render`

### DIFF
--- a/app.js
+++ b/app.js
@@ -5,14 +5,6 @@ module.exports = function (options) {
 	var reducers = options.update || {}
 	var effects = options.effects || {}
 
-	var subscriptions = function (subs) {
-		// `subs` will be deprecated in favor of only `subscriptions` >= 0.0.13
-		subs = options.subscriptions || options.subs
-		for (var key in subs) {
-			subs[key](model, msg, hooks.onError)
-		}
-	}
-
 	var hooks = merge({
 		onAction: Function.prototype,
 		onUpdate: Function.prototype,
@@ -22,7 +14,7 @@ module.exports = function (options) {
 	}, options.hooks)
 
 	var node
-	var root = options.root || document.body.appendChild(document.createElement("div"))
+	var root = options.root
 	var view = options.view || function () {
 		return root
 	}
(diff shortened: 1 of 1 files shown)


tzellman:
I might need to make a slight mod to rebase with the changes made in 4126833.

jbucaran:
> It seems safest to wait until the document is ready before we attempt to:

Playing devil's advocate here: why is it safest? and also, how come we were not getting errors?

jbucaran:
> @tzellman It seems safest to wait until the document is ready before we attempt to:

Playing devil's advocate here: why is it safest? and also, how come we were not getting any errors until now? That could be indicating a deeper problem.

dodekeract:
@jbucaran If that's helpful I definitely already had that issue in `0.0.7`. Took a while to understand while the `subs` didn't work, but made sense that the event never fired when executing `app` on `load`.

jbucaran:
@dodekeract Minor correction, it's `subscriptions` now.

tzellman:
The reason I made the change was strictly due to potential timing issues. If your JS runs in the `<head>` tag, then it is possible to attempt to create the default root element before the DOM is ready. Same goes for the `render` method, in case you passed in a truthy value for root.

dodekeract:
@jbucaran You mean #69. 🙂

jbucaran:
Yup. 😅

tzellman:
Just posted a comment with example in #69.

jbucaran:
What do other frameworks do in this instance? Does anybody know?

dodekeract:
@jbucaran React definitely doesn't work, because you have to manually give `ReactDOM.render()` a DOMElement. I guess it's a bit more low-level than hyperapp in that regard.

jbucaran:
@dodekeract Try `app({ root: document.body })`.

jbucaran:
Root elements, subscriptions and initial render (incl. router support) landed in master. 
* https://github.com/hyperapp/hyperapp/blob/7e248b2fbeeed757030a8ec96a520af624810a44/src/app.js#L43. 

Thank you @tzellman for bringing this up!

PR #71, opened 2017-02-09
Add .editorconfig with tab indent_style

tzellman:
This should define clearly what the preferred indent style is for the project, and make it easier for (most) editors to adjust automatically.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = tab
+
+[*.{diff,md}]
+trim_trailing_whitespace = false
+
+[*.json]
+indent_style = space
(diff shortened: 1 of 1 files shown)


dodekeract:
Is `indent_size = 4` optional? As the `indent_style` is set to `tab` there is no need to forcefully overwrite the user's setting for that. 🙂

.editorconfig
```diff
@@ -0,0 +1,12 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = tab
+indent_size = 4
```
dodekeract: Is `indent_size = 4` optional? As the `indent_style` is set to tab there is no need to forcefully overwrite the user's setting for that. 🙂

tzellman: Good point - I removed it.

jbucaran:
@tzellman Thanks!

Issue #8, opened 2017-02-01
Example with child components? 🍼


joelnet:
Are child components possible? Are there any examples with child components?

dodekeract:
The [routing example](https://github.com/hyperapp/hyperapp#routing) utilizes an `anchor` child.

PetrSnobelt:
It will be great if there is more advanced example - something like dynamic list of counters - it can show how to use child components, how to dynamically add elements and how to model updates.

jbucaran:
@joelnet @PetrSnobelt 

Here is a dynamic list of counters.

```js

const counter = (count, add, sub) => html`
    <div>
        <h1>${count}</h1>
        <button onclick=${add}>+</button>
        <button onclick=${sub} disabled=${count <= 0}>-</button>
    </div>
`

const counterReduce = (model, index, reduce) =>
    model.map((item, i) => index === i ? reduce(item) : item)

app({
    model: [],
    update: {
        create: model => model.concat(0),
        remove: model => model.slice(0, -1),
        add: (model, index) => counterReduce(model, index, count => count + 1),
        sub: (model, index) => counterReduce(model, index, count => count - 1),
    },
    view: (model, msg) => html`
        <div>
            <button onclick=${msg.create}>Add Counter</button>
            <button onclick=${msg.remove}>Remove Counter</button>
            ${model.map((item, i) => counter(item, _ => msg.add(i), _ => msg.sub(i)))}
        </div>
    `
})
```
[**CodePen**](http://codepen.io/jbucaran/pen/vgRJeP?editors=0010)

Also, check out the TodoMVC implementation.

* [Live](https://hyperapp-todomvc.gomix.me/)
* [Source](https://gomix.com/#!/project/hyperapp-todomvc)

nichoth:
For reference here is using multiple stateful components, whereas in the above `counter` is a static function.

```js
var hyperapp = require('../')
var app = hyperapp.app
var html = hyperapp.html

function CounterList () {
    return [Counter()]
}
CounterList.update = {
    addCounter: model => model.concat([Counter()]),
    removeCounter: (model, i) => {
        var newState = [].concat(model)
        newState.splice(i, 1)
        return newState
    },
    counter: function (model, ev) {
        var newState = [].concat(model)
        var c = newState[ev.index]
        newState[ev.index] = Counter.update[ev.type](c)
        return newState
    }
}

CounterList.view = function (model, msg) {
    var counters = model.map(function (counter, i) {
[...]

Issue #2, opened 2017-01-31
multiple `app()`s / component approach / return dom


tunnckoCore:
So, it is just amazing lib. A was working on some idea like that, code name `mitak` ;d Based on events with just router, pub/sub system. For now it is 450bytes gzip. But yea, anyway.

So the problem of return nothing just bumped while I tried to make a Todo List in components way.

So lets get started. We have 3 `app()`s - TodoItem, TodoList and main.

- main - main entry, place for header & footer for example
- TodoItem - component for each todo, e.g. `<li></li>`
- TodoList - component holding all `TodoItem`s, actions and view which is just like `<ul>TodoItem()</ul>`

main

```js
const { app, html } = require('hyperapp')

const main = app({
  // root: document.body,
  model: {
    title: 'Todo List Example',
    link: 'https://github.com/hyperapp/hyperapp'
  },
  view: (state) => html`<main id="HyperApp">
    <h1>${state.title}</h1>
    ${TodoList()}
    <footer>Powered by <a data-no-routing href="${state.link}">HyperApp</a></footer>
  </main>`
})
```

TodoList

```js
const TodoList = () => app({
  model: {
    todos: [],
    count: 0,
    input: '',
    placeholder: 'Add new todo'
  },
  update: {
    toggle: (state, data) => ({
      todos: state.todos.map((item) => {
        return data.id === item.id
          ? Object.assign({}, item, { done: !data.done })
          : item
      })
    }),
    remove: (state, data) => ({
      todos: state.todos.map((todo) => todo.id !== data.id)
    }),
    add: (state) => ({
      todos: state.todos.concat({
        id: state.count++,
        done: false,
        text: state.input
      })
    }),
    input: (state, data) => ({ input: data.value })
  },
  view: (state, action) => html`<section>
    <ul>${state.todos.map((item) => TodoItem(item, action))}</ul>
    <p>
      <input
        class="add"
        type="text"
        onkeyup=${e => e.keyCode === 13 ? action.add() : ''}
        oninput=${e => action.input({ value: e.target.value })}
        placeholder=${state.placeholder}
      />
      <button onclick=${action.add}>add</button>
    </p>
  </section>`
})
```

TodoItem

```js
const TodoItem = (state, actions) => app({
  model: state,
  update: actions,
  view: (item, action) => html`<li onclick=${e => action.toggle(item)}>${item.text}</li>`
})
```

All is okey, but it adds Todo to DOM body two times

[...]

PR #77, opened 2017-02-10
Support nodes whose tag is a function.

jbucaran:
HyperApp's architecture is largely based in Flux/Redux and/or the Elm Architecture. There's a single state shared by all the views. Another way to put that, views can't have a local state. 

But what about components? The answer is stateless components.

```jsx
<MyComponent props=...></MyComponent>
```

**Related**
* https://github.com/reactjs/redux/issues/1385
* https://github.com/hyperapp/hyperapp/issues/2
* https://github.com/hyperapp/hyperapp/issues/8


danigb:
That's a really nice addition 👏👏

tunnckoCore:
Cool!

But i believe it would be better to pass `data` directly and append the `children`, it is how is done in React and etc? I played with this and `mich-h` stuff and kinda don't liked to add `children` "special property" to the props.

```js
	if (typeof tag === "function") {
		tree = tree || []
		data.children = tree
		return tag(data, tree)
	}
```

So

```js
const Foo = (props) => <div id={props.id} className={props.cls}>{props.children}</div>

const Foo2 = (props, children) => <div id={props.id} className={props.cls}>{children}</div>

const foo = <Foo cls="baz qux" id="hi">
  <strong>hello</strong>
</Foo>
```

tunnckoCore:
Hm... No, no, nevermind. Kinda like more your approach when rethink it. It is more clean and explicit.

```jsx
const Foo2 = ({ props, children }) =>
  <div id={props.id} className={props.cls}>{children}</div>
```

jbucaran:
@tunnckoCore Yeah, but who cares? I think some people would. I find this approach cleaner too, but

> It won't be react compatible directly,

But what if it was? That'd be another reason to give HyperApp a chance.

tunnckoCore:
Don't know, I like them either way. Current no-react-compatible is more cleaner.

PR #79, opened 2017-02-11
Remove the router

jbucaran:


dodekeract:
So how are we supposed to do anything with hyperapp now?

jbucaran:
The router will live [here](https://github.com/hyperapp/router).

jrop:
I was just playing around with this because I needed a hash-router (due to constraints, I cannot use pushState), and came up with this:

Code:
https://glitch.com/edit/#!/brainy-ferryboat

Preview:
https://brainy-ferryboat.glitch.me/

lukejacksonn:
Nice! I feel like someone has already had a go at a hash based router, or adapted the hyperapp router to work with hashes 🤔 can't remember who though

zaceno:
There's this: https://github.com/zaceno/hyperapp-hash-router

... which is basically just a search-and-replace of hyperapps built in router. (doesn't even have a package.json so you can import it though)

PR #96, opened 2017-02-18
Add CDNJS version badge

PeterDaveHello:
Sometimes there may be a delay caused by directory/files structure changes so that the bot can recognize if the new version is acceptable and need a human to take a look, I'll be good to have a version badge just like the npm version badge to let you know that what's the latest version on CDNJS 😄


jbucaran:
LGTM

PR #97, opened 2017-02-19
Use ES6 rest parameter syntax

terkelg:
The code is simpler and easier to read, since you don't have to convert the array-like ```arguments``` object into a real array and shift the two first arguments. 

Besides that, it's less code.

### DIFF
--- a/src/h.js
+++ b/src/h.js
@@ -1,10 +1,6 @@
-export default function (tag, data) {
+export default function (tag, data, ...children) {
 	data = data || {}
 
-	var children = []
-	children.push.apply(children, arguments)
-	children.shift()
-	children.shift()
 	var head = children[0]
 
 	for (var name in data) {


jbucaran:
@terkelg Yup yup, we had something like this in an earlier sketch. I'd love to merge this, but:
* http://caniuse.com/#search=rest%20param

terkelg:
I see, but isn't the build being compiled with Babel?

jbucaran:
@terkelg Nope! We're pretty much ES3 except for ES6 modules. We're only using ES6 modules so that users building their app with Rollup or Webpack can enjoy tree-shaking.

terkelg:
Nice one! 👍

tunnckoCore:
In anyway, why not just `[].slice.call(arguments, 2)` instead of two shifts? Performance?

jbucaran:
Maybe, but really I have no better answer, since that's what worked for me.

Issue #109, opened 2017-02-24
Router catch all? handling 404s.


traducer:
When routes aren't found is there a way to define a catch all route ```"*"``` in the router?

I want to handle 404 pages for any routes or sub-routes not defined in my application.

So if someone goes to ```/this/should/404``` I can catch it and have the server render a 404 page.

jbucaran:
@traducer Currently `/` acts as the wildcard when no other route matches. 

If that's not good enough, we can make `/` match only `/` and `*` match everything when no other route matches.

What do you think?

traducer:
@jbucaran, I think a default route that a user can specify at app creation would be a wise decision. ```/``` should strictly be the indexRoute. If a user wants it to be the default route, they should have the freedom to choose what that default route is. This is basically what [sheet-router](https://github.com/yoshuawuyts/sheet-router) does for unmatched routes. 

So for example, when creating the routes structure:

```javascript
const { h, router, app } = require('hyperapp')

app({
  router: router,
  view: {
    '/': _ => <Home />,
    '/profile': _ => <Profile />,
    '/404': _ => <NotFound />,
    'defaultRoute': '/404' //this can also be a function 

  }
})
```


My suggestions would be instead of returning [```routes['/']```](https://github.com/hyperapp/hyperapp/blob/master/src/router.js#L60)

return the route the user defined at app creation, something like:

```javascript
var dft = routes['defaultRoute']

/* if a default route exists check if it's a string, if it is return the function mapped to that string.
** Ff it isn't a string return the function mapped to defaultRoute. If it doesn't exist at all, return routes['/'] 
*/
return !dft ? routes['/'] : (typeof dft === 'string') ? routes[dft] : dft 
```

jbucaran:
@traducer Yeah, I like that. Now that the router is an optional dependency, I'd love to make the router better.

What about using "*" as a wild card instead of :404? I think :404 is rather confusing.

traducer:
@jbucaran yes an asterisk is probably the best character for a wild card. I'm not sure if there would be any implications in the route matching since you are using regular expressions to match routes.

lukejacksonn:
I'm going to try and implement this tonight, that is.. make `/` an explicit route the default route `*` not had a look at the code yet but how hard can it be 😅

jbucaran:
@traducer Implemented by defining a route with a key: `*`. Not properly yet documented I think.

PR #113, opened 2017-02-26
Makes * the fallback route if no matching route is found

lukejacksonn:
As a response to https://github.com/hyperapp/hyperapp/issues/109. Actions taken:

- match function now returns `routes["*"]` when no route is found
- regexify function now returns a regex that will match `.*` when it is given the path `*` 
- Added tests for just `/` route 
- Added `*` route to both tests `view` object

Tests are passing but I would like to discuss if any of the logic could be tidied up!

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -61,7 +61,11 @@ export default function (render, options) {
             }
         }
 
-        return routes["*"] || function(m,a) { return "404 Not Found" }
+        if (!routes["*"]) {
+          throw 'No matching route for ' + path
+        }
+
+        return routes["*"]
     }
 
     function regexify(path) {
--- a/src/router.js
+++ b/src/router.js
@@ -41,6 +41,10 @@ export default function (render, options) {
         for (var route in routes) {
             var re = regexify(route), params = {}, match
 
+            if (route === "*") {
+                continue
+            }
+
             path.replace(new RegExp(re.re, "g"), function () {
                 for (var i = 1; i < arguments.length - 2; i++) {
                     params[re.keys.shift()] = arguments[i]
@@ -57,13 +61,11 @@ export default function (render, options) {
             }
         }
 
-        return routes["*"]
+        return routes["*"] || function(m,a) { return "404 Not Found" }
     }
 
     function regexify(path) {
-        var keys = []
-        var re = path === '*' ?
-            '.*' : "^" + path
+        var keys = [], re = "^" + path
             .replace(/\//g, "\\/")
             .replace(/:([A-Za-z0-9_]+)/g, function (_, key) {
                 keys.push(key)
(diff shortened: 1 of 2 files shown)


jbucaran:
@lukejacksonn Awesome! Let me review this.

Does the router still work for `/`?

lukejacksonn:
yeh because now `/` just gets treated as any other route!

lukejacksonn:
It does demand that the objects passed to `views` has a `*` route though now otherwise it would break.. we could easily add a default view though like a default `404` message

jbucaran:
@lukejacksonn But we could avoid having to specify a `*` property correct?

```jsx
        for (var name in routes) {
            if (name === "*") {
                continue
            }
        ...
      }
```

lukejacksonn:
then what gets returned from `match()`?

lukejacksonn:
If you make line 60 look like this:
`return routes["*"] || function(m,a) { return "404 Not Found" }`

Then the user doesn't need to specify any route at all (this would simply result in "404 Not Found" being rendered)

lukejacksonn:
Adding your snippet has two advantages:

- Means that the order of the routes in the object passed to view doesn't matter (that is the `*` route can be at the top, bottom or somewhere in between), currently it has to be the last route.
- It means that a regex for `*` will not need to be created, so the the logic in `regexify` does not need to be altered. Currently it has been.

This combined with a default 404 message seems like a more robust approach to me.

jbucaran:
@lukejacksonn So, which one should we use?

lukejacksonn:
Both 🎉  I will push it up now

lukejacksonn:
How do you feel about the message `404 Not found`? Something like `No route found for /path` might be more useful for a beginner.

jbucaran:
@lukejacksonn I think we should just throw, no_

lukejacksonn:
ahh yeh, could do

jbucaran:
Thanks a lot @lukejacksonn! I'll take it from here. I'll pull your branch locally, do some mods and ping you again, perhaps you can review it tomorrow? 🙏

lukejacksonn:
Sure! Can't wait 💯

tunnckoCore:
meh.. throwing text..

jbucaran:
(╯°□°）╯︵ ┻━┻

lukejacksonn:
I'm going to close this as it has been merged right?

jbucaran:
@lukejacksonn Not yet, as I am still working on the new plugin architecture. 🙏

lukejacksonn:
Yeh sorry I just wanted rid of those extra commits really! They have been removed now 😅

jbucaran:
The new router is almost ready, but the regex matcher and logic "/" or "*" stuff will not change, so your PR is still quite relevant!

PR #115, opened 2017-02-27
Implement Recursive Actions (Fix #73)

### LINKED ISSUE #73: Nested Update/Effects
Have you considered this:

```javascript
app({
	update: {
		example: {
			attempt: () => alert(1)
		}
	}
})
```

It should then be possible to call `msg.example.attempt()`.

---

<details>
<summary>Meanwhile I ended up using this</summary>

```javascript
const convert = (actions, prefix = '') => Object
	.keys(actions)
	.map(name => typeof actions[name] === 'function'
		? ({[`${prefix[0].toLowerCase()}${prefix.slice(1)}${name[0].toUpperCase()}${name.slice(1)}`]: actions[name]})
		: (convert(actions[name], `${prefix}${name[0].toUpperCase()}${name.slice(1)}`))
	)
	.reduce((a, b) => Object.assign(a, b), {})

app({
	update: convert({
		example: {
			attempt: () => alert(1)
		}
	})
})
```

This maps `example.attempt` to `exampleAttempt`.

</details>

dodekeract:
Makes this possible:

```jsx
const { h, app } = hyperapp
/** @jsx h */

app({
	model: 0,
	reducers: {
		counter: {
			save: (_, data) => data
		}
	},
	effects: {
		counter: {
			sub: (model, actions) => actions.counter.save(model - 1)
			add: (model, actions) => actions.counter.save(model + 1)
		}
	},
	view: (model, {counter}) => (
		<div>
			<button onclick={counter.add}>
				+
			</button>
			<h1>{model}</h1>
			<button onclick={counter.sub} disabled={model <= 0}>
				-
			</button>
		</div>
	)
})
```

---

## Meta

Issue: #73

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -21,23 +21,25 @@ export default function (options) {
 		}
 	}, options.hooks)
 
-	for (var name in merge(reducers, effects)) {
-		(function (name) {
-			actions[name] = function (data) {
-				hooks.onAction(name, data)
-
-				var effect = effects[name]
-				if (effect) {
-					return effect(model, actions, data, hooks.onError)
-				}
-
-				var update = reducers[name], _model = model
-				render(model = merge(model, update(model, data)), view, node)
-
-				hooks.onUpdate(_model, model, data)
-			}
-		}(name))
+	function recurseActions (isReducer, pointer, actionPointer, _name) {
+		for (var key in pointer) {
+			if (!actionPointer[key]) actionPointer[key] = {}
+			var name = _name ? _name + '.' + key : key
+			if (pointer[key] instanceof Function) {
+				(function (key, name) {
+					actionPointer[key] = function (data) {
+						hooks.onAction(name, data)
+						if (isReducer) {
+							hooks.onUpdate(model, model = merge(model, pointer[key](model, data)), data)
+							render(model, view, node)
+						} else return pointer[key](model, actions, data, hooks.onError)
+					}
+				})(key, name)
+			} else recurseActions(isReducer, pointer[key], actionPointer[key], name)
+		}
 	}
+	recurseActions(0, effects, actions, '')
+	recurseActions(1, reducers, actions, '')
 
 	ready(function () {
 		root = options.root || document.body.appendChild(document.createElement("div"))
--- a/src/app.js
+++ b/src/app.js
@@ -38,8 +38,8 @@ export default function (options) {
 			} else recurseActions(isReducer, pointer[key], actionPointer[key], name)
 		}
 	}
-	recurseActions(0, effects, actions, '')
-	recurseActions(1, reducers, actions, '')
+	recurseActions(0, effects, actions)
+	recurseActions(1, reducers, actions)
 
 	ready(function () {
 		root = options.root || document.body.appendChild(document.createElement("div"))
--- a/src/app.js
+++ b/src/app.js
@@ -22,21 +22,20 @@ export default function (options) {
 	}, options.hooks)
 
 	function recurseActions (isReducer, pointer, actionPointer, _name) {
-		for (var key in pointer) {
+		Object.keys(pointer).forEach(function (key) {
 			if (!actionPointer[key]) actionPointer[key] = {}
 			var name = _name ? _name + '.' + key : key
-			if (pointer[key] instanceof Function) {
-				(function (key, name) {
-					actionPointer[key] = function (data) {
-						hooks.onAction(name, data)
-						if (isReducer) {
-							hooks.onUpdate(model, model = merge(model, pointer[key](model, data)), data)
-							render(model, view, node)
-						} else return pointer[key](model, actions, data, hooks.onError)
-					}
-				})(key, name)
-			} else recurseActions(isReducer, pointer[key], actionPointer[key], name)
-		}
+			var reference = pointer[key]
+			if (typeof reference === 'function') {
+				actionPointer[key] = function (data) {
+					hooks.onAction(name, data)
+					if (isReducer) {
+						hooks.onUpdate(model, model = merge(model, reference(model, data)), data)
+						render(model, view, node)
+					} else return reference(model, actions, data, hooks.onError)
+				}
+			} else recurseActions(isReducer, reference, actionPointer[key], name)
+		})
 	}
 	recurseActions(0, effects, actions)
 	recurseActions(1, reducers, actions)


src/app.js
```diff
@@ -21,23 +21,25 @@ export default function (options) {
...
-
-				var effect = effects[name]
-				if (effect) {
-					return effect(model, actions, data, hooks.onError)
-				}
-
-				var update = reducers[name], _model = model
-				render(model = merge(model, update(model, data)), view, node)
-
-				hooks.onUpdate(_model, model, data)
```
dodekeract: Oh, I also fixed `onUpdate` not firing if `render` crashes. That's probably the time where you need `onUpdate` most.

src/app.js
```diff
@@ -21,23 +21,25 @@ export default function (options) {
...
+	function recurseActions (isReducer, pointer, actionPointer, _name) {
+		for (var key in pointer) {
+			if (!actionPointer[key]) actionPointer[key] = {}
+			var name = _name ? _name + '.' + key : key
+			if (pointer[key] instanceof Function) {
+				(function (key, name) {
+					actionPointer[key] = function (data) {
+						hooks.onAction(name, data)
+						if (isReducer) {
+							hooks.onUpdate(model, model = merge(model, pointer[key](model, data)), data)
```
dodekeract: `_model` is not needed anymore. It's now implicit.

src/app.js
```diff
@@ -21,23 +21,25 @@ export default function (options) {
...
+							render(model, view, node)
+						} else return pointer[key](model, actions, data, hooks.onError)
+					}
+				})(key, name)
+			} else recurseActions(isReducer, pointer[key], actionPointer[key], name)
+		}
 	}
+	recurseActions(0, effects, actions, '')
+	recurseActions(1, reducers, actions, '')
```
dodekeract: Could be `false/true`, but `0/1` leads to -2 bytes.

jbucaran:
@dodekeract Interesting! 😄 

Can you clarify the objective is to be able to accomplish namespaces for actions?

dodekeract:
@jbucaran yes.

jbucaran:
What if you want to use namespaces for some actions, but not for others?

dodekeract:
@jbucaran Still works, of course. Namespaces can be as deep as you want.

(tests don't fail, which proofs this claim)

jbucaran:
@dodekeract I'm going through the code, in the meantime, could you try to think how the documentation will need to be rewritten or adapted?

dodekeract:
@jbucaran Honestly, since this affects both reducers and effects I'd add a new topic `actions` to the Wiki. Then explain that both will be combined and can have arbitrary depth.

I'd also suggest to adapt one or more examples to use this, so that it's "advertised" to users that don't read the Wiki.

jbucaran:
Okay, I'm done, very nice job @dodekeract! I'll ship this with the next release.

If you can help with the documentation a bit, we could ship this fater! 🤘

src/app.js
jbucaran: @dodekeract What do you mean by "implicit"?

src/app.js
```diff
@@ -21,23 +21,25 @@ export default function (options) {
...
-
-				var update = reducers[name], _model = model
-				render(model = merge(model, update(model, data)), view, node)
-
-				hooks.onUpdate(_model, model, data)
-			}
-		}(name))
+	function recurseActions (isReducer, pointer, actionPointer, _name) {
+		for (var key in pointer) {
+			if (!actionPointer[key]) actionPointer[key] = {}
```
dodekeract: It creates empty objects for every key it'll later try to assign a function to. You can't just assign it to nothing.

src/app.js
```diff
@@ -21,23 +21,25 @@ export default function (options) {
...
+			var name = _name ? _name + '.' + key : key
+			if (pointer[key] instanceof Function) {
+				(function (key, name) {
+					actionPointer[key] = function (data) {
+						hooks.onAction(name, data)
+						if (isReducer) {
+							hooks.onUpdate(model, model = merge(model, pointer[key](model, data)), data)
+							render(model, view, node)
```
dodekeract: We can. I just wanted to keep the scope to a single change.

src/app.js
dodekeract: Well, `_model` got replaced by moving the assignment into the `hooks.onUpdate` call.

src/app.js
```diff
@@ -21,23 +21,25 @@ export default function (options) {
...
-				render(model = merge(model, update(model, data)), view, node)
-
-				hooks.onUpdate(_model, model, data)
-			}
-		}(name))
+	function recurseActions (isReducer, pointer, actionPointer, _name) {
+		for (var key in pointer) {
+			if (!actionPointer[key]) actionPointer[key] = {}
+			var name = _name ? _name + '.' + key : key
+			if (pointer[key] instanceof Function) {
```
dodekeract: Not sure, but works too. Fine with changing this.

src/app.js
```diff
@@ -21,23 +21,25 @@ export default function (options) {
...
-					return effect(model, actions, data, hooks.onError)
-				}
-
-				var update = reducers[name], _model = model
-				render(model = merge(model, update(model, data)), view, node)
-
-				hooks.onUpdate(_model, model, data)
-			}
-		}(name))
+	function recurseActions (isReducer, pointer, actionPointer, _name) {
```
dodekeract: I'm really sure that we can't avoid that. The signature is also as good as possible AFAIK.

I agree that the signature isn't the most readable one, but try removing anything. There is nothing that isn't necessary.

src/app.js
```diff
@@ -21,23 +21,25 @@ export default function (options) {
...
+		for (var key in pointer) {
```
dodekeract: Might work. I personally don't like `forEach` without arrow functions though. I'll see what the size difference is.

dodekeract: Oh, we actually can't get rid of the closure since name still depends on it.

src/app.js
jbucaran: Can we get rid of the `if` and just assign it to {}?

src/app.js
jbucaran: No worries. I'll take care of it.

jbucaran:
@dodekeract Thanks for your comments. I'm taking care of it now.

dodekeract:
@jbucaran Not so fast, I just saved another 6 Bytes. 🎉

src/app.js
dodekeract: @jbucaran No, that would overwrite the actions we previously assigned to the same part of the actions tree.

jbucaran:
Which was smaller, the one with the forEach or for loop?

dodekeract:
@jbucaran Last commit is smallest. `forEach` enabled some further optimizations due to scoping.

jbucaran:
👍 Thanks @dodekeract.

PR #118, opened 2017-02-27
Optimize isPrimitive

dodekeract:
Should be ~~the same speed~~ faster, ~~but 9~~ and 20 bytes less after minification/build.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -70,7 +70,7 @@ export default function (options) {
 	function merge(a, b) {
 		var obj = {}, key
 
-		if (isPrimitive(typeof b) || Array.isArray(b)) {
+		if (isPrimitive(b) || Array.isArray(b)) {
 			return b
 		}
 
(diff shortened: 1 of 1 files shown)


jbucaran:
@dodekeract Is this a new GitHub feature? I'm actually seeing the difference in bytes in the UI.

dodekeract:
@jbucaran Quite sure that's just my commit message. :smile:

jbucaran:
@dodekeract You are right 😆. Fail.

Yeah, this looks perfect.

dodekeract:
@jbucaran Sorry if the last commit is too hacky. Was wondering if even less is possible. Turns out all 3 commits make it (9 + 11 + 9 =) 29 Bytes smaller.

jbucaran:
@dodekeract Haha, yep, that's too hacky. Let's avoid that.

jbucaran:
@dodekeract Let's go back to the original commit.

PR #119, opened 2017-02-27
Improve Namespace Tests

dodekeract:


jbucaran:
@dodekeract Thanks, now I understand what the guard was for! 👍 

Are you done? I'd like to merge this right away.

dodekeract:
Yes. I don't think I have to change anything here.

PR #122, opened 2017-03-02
Fixes example event listener attributes, all to lower case

lukejacksonn:


jbucaran:
Thanks a lot @lukejacksonn!

PR #124, opened 2017-03-02
Introduce plugin architecture.

jbucaran:
Plugins are used like this:

```jsx
app({
  ...,
  plugins: [PluginFoo, PluginBar, PluginBaz]
})
```

In a nutshell, if you know how to make an app with HyperApp you already know how to make a plugin.

A plugin is a function that exports an object with the same props as the object you pass into `app(options)` (except views, plugins don't export views at the moment, but you can listen to the onRender hook to alter this process, see Router).

For example, a plugin that adds a new subscription to your app:

```jsx
const MyPlugin = app => ({
    subscriptions: [ _ => console.log("Hi, I'm a plugin") ],
})
```

A plugin function is called by HyperApp during your app's init step (before rendering a view). The `app` argument is the same options object you pass into `app(options)`. Plugins don't need to modify this object, instead they return an object that is merged back into your app's options. 

Since this is JavaScript, plugins can totally modify this object's properties, so they have both the power to bring joy and devastation to the world.

I'll be updating the docs later, once this ships.

---

Also on this branch:

* Entire test spec rewritten.
* New Router based in the new plugin architecture.
* New and faster `h`.
  * As a result, smaller app.js since we no longer need to handle bools/nulls in the vdom engine.


jbucaran:
@tunnckoCore @lukejacksonn Could you review this PR too? 🙏

lukejacksonn:
Sure, I will take a look this afternoon! 👍

src/router.js
```diff
@@ -1,73 +1,73 @@
...
+		subscriptions: [
+			function (model, actions) {
+				window.addEventListener("popstate", function () {
+					actions.router.setLocation(location.pathname)
+				})
+			}
+		],
+		hooks: {
+			onRender: function (model, view) {
+				console.log(model.router.location)
```
lukejacksonn: console.log here 🔥

src/router.js
```diff
@@ -1,73 +1,73 @@
...
+					return routes[route](model, actions, params)
+				}
+			})
+
+			if (match) {
+				return match
+			}
+		}
+
+		return routes["/"]
```
lukejacksonn: Were we going to be implementing the changes made in https://github.com/hyperapp/hyperapp/pull/113 to this PR?

jbucaran: Better after this is landed 👍.

tunnckoCore:
Too much set in one commit.. but i'll try. In first glance i have some notes.

src/router.js
jbucaran: Haha.

src/router.js
```diff
@@ -1,73 +1,73 @@
...
+	function match(routes, path) {
+		for (var route in routes) {
+			var re = regexify(route), params = {}, match
+
+			path.replace(new RegExp(re.re, "g"), function () {
+				for (var i = 1; i < arguments.length - 2; i++) {
+					params[re.keys.shift()] = arguments[i]
+				}
+
+				match = function (model, actions) {
```
tunnckoCore: Move function out of the replace? Move it inside the outside `if`. It would be faster for sure. It may sounds like a super micro optimization, but not only that won't redefined that function on each found match, but it would looks a bit better and natural.

jbucaran: Done.

src/app.js
```diff
@@ -1,48 +1,47 @@
...
 	var node
 	var root
 	var batch = []
 
-	var hooks = merge({
-		onAction: Function.prototype,
-		onUpdate: Function.prototype,
-		onError: function (err) {
+	var onError = function (err) {
+		if (!hook("onError", function (fn) {
```
jbucaran: That's not going to work, because we pass onError to effects and subscriptions.

src/router.js
```diff
@@ -1,81 +1,76 @@
...
+			for (var i = 1; i < arguments.length - 2; i++) {
+				params[re.keys.shift()] = arguments[i]
+			}
 
-            if (match) {
-                return match
-            }
-        }
+			match = function (model, actions) {
+				return routes[route](model, actions, params)
```
dodekeract: Shouldn't `params` be in the model now?

dodekeract: This should actually also improve performance since `match` doesn't need to be executed on **every** render and instead just on every `go`. I'd not be surprised if this gives us `-0.1` or more relative performance.

src/app.js
```diff
@@ -53,32 +52,72 @@ export default function (options) {
...
+
+		reducers = merge(reducers, app.reducers)
+		effects = merge(effects, app.effects)
+
+		if (app.subscriptions) {
+			subscriptions = subscriptions.concat(app.subscriptions)
+		}
+
+		if (app.hooks) {
+			hooks.push(app.hooks)
```
dodekeract: Instead of pushing objects into an array I would iterate through the objects and construct this data-structure:

```javascript
var hooks = {
	onRender: [f1, f2, f3],
	onError: [f4]
}
```

This has the advantage that access is fast and easy. Basically it should be handled the same as `subscriptions`.

src/app.js
```diff
@@ -53,32 +52,72 @@ export default function (options) {
...
+		}
+	}
+
+	function hook(name, fn) {
+		for (var key in hooks) {
```
dodekeract: Don't use `for in` for array iteration. It's slow and potentially buggy.

src/app.js
```diff
@@ -1,48 +1,47 @@
...
-	}, options.hooks)
+	}
 
-	initializeActions(actions, effects, false)
-	initializeActions(actions, reducers, true)
+	use(app)
 
-	domContentLoaded(function () {
-		root = options.root || document.body.appendChild(document.createElement("div"))
+	for (var key in plugins) {
```
dodekeract: Don't use `for ... in ...` for array iteration.

src/app.js
```diff
@@ -1,48 +1,47 @@
...
-			}, options)
-		}
+	load(function () {
+		root = app.root || document.body.appendChild(document.createElement("div"))
 
-		for (var key in subs) {
-			subs[key](model, actions, hooks.onError)
+		render(model, view)
+
+		for (var key in subscriptions) {
```
dodekeract: `for ... in`

src/app.js
```diff
@@ -53,32 +52,72 @@ export default function (options) {
 
 			if (typeof action === "function") {
 				container[key] = function (data) {
-					hooks.onAction(name, data)
+					hook("onAction", function (fn) {
```
dodekeract: if you implement the data structure I suggested this would be

```javascript
hooks.onAction.forEach(hook => hook(name, data))
```

src/index.js
```diff
@@ -1,5 +1,5 @@
 import h from "./h"
 import app from "./app"
-import router from "./router"
+import Router from "./router"
```
dodekeract: Why is this TitleCase? It's not a class.

jbucaran: > It's not a class.
I don't care about this convention. I think it sucks to limit this convention for a construct I rarely use in JavaScript anyway. I don't use Classes/Prototypes.

Router is more distinctive. It's not a component, like in React, but they also use Router, which is familiar.

dodekeract: @jbucaran the thing is that it implies that `router` has to be used like `new Router()`.

src/router.js
jbucaran: Not sure I understand what you are saying 🤔.

> Shouldn't params be in the model now?

Not sure about this. `location` is in the model not because it's the only way for the Router to control when the view is rendered without using a hack like passing a `render` function as we were doing before.

src/app.js
jbucaran: On the other hand for loop is more verbose and we'd need to check if it's empty or not with .length. I'll experiment with `for` and see how that affects size/performance. 👍

src/app.js
jbucaran: https://github.com/hyperapp/hyperapp/pull/124#discussion_r104085615

src/app.js
jbucaran: https://github.com/hyperapp/hyperapp/pull/124#discussion_r104085615

src/app.js
dodekeract: Size might be more, but `for in` is basically a hack when used on arrays.

src/app.js
jbucaran: @dodekeract Can you provide some pseudocode?

src/app.js
jbucaran: I see, this sounds good. Let me see how I can do that.

src/router.js
dodekeract:
```javascript
model.router = {
	matched: '/some/route/:id',
	url: '/some/route/34593458',
	params: {
		id: 34593458
	}
}
```

src/index.js
jbucaran: I don't use new or this; as an extension, neither does HyperApp or will ever do. So, why should Router imply that?

src/app.js
dodekeract:
```javascript
var hooks = {
	onRender: []
}
function addToHooks(hook, name) {
	hooks[name].push(hook)
}
function handlePluginHooks (hooks) {
	for (var name in hooks) {
		addToHooks(hooks[name], name)
	}
}
```

src/index.js
dodekeract: Because it's TitleCase. For someone familiar with javascript that usually means it has to be instantiated.

src/router.js
dodekeract: I think it makes more sense and it should also be faster, since right now we call `match` **every time the model is updated**. That function looks quite expensive.

src/index.js
dodekeract: It's not important though since hyperapp isn't even aware of that name. I just think it's better for examples since it follows the usual conventions.

src/router.js
jbucaran: @dodekeract Ah, I think I see what you are saying now! It doesn't make sense to call match every time the view is rendered 👍.

src/app.js
tunnckoCore: wat? no, no. I not meant that. It's no matter how the var is named. The point is to not be inside the if condition

```js
var foobar = function (fn) { fn(err) }
if (!hook("onError", foobar)) {
  throw err
}
```

and that (as currently) are the same

```js
if (!hook("onError", function (fn) { fn(err) })) { // collapsed in one line
  throw err
}
```

src/router.js
tunnckoCore: Nope. It isn't (looking at the latest commit) [L54-56](https://github.com/hyperapp/hyperapp/blob/3da2f333259a91d05999ed629c95fd5d6973fee5/src/router.js#L54-L56).

Simply make `match = 1` and if there is no matches in that `.replace()` function match would be falsey value, otherwise will be 1.

```js
path.replace(new RegExp(re.re, "g"), function () {
  for (var i = 1; i < arguments.length - 2; i++) {
    params[re.keys.shift()] = arguments[i]
  }

  match = 1
})

if (match) {
  return function (model, actions) {
    return routes[route](model, actions, params)
  }
}
```

src/index.js
tunnckoCore: agree with @dodekeract

jbucaran: Still not convinced about that. Router looks better.

src/app.js
jbucaran: Seems more verbose. Specially if more hooks are added in the future. I'll see if there are any performance gains! 👍

src/index.js
jbucaran: One solution could be to not ship the router with HyperApp anymore. That way you can import the router like

```jsx
import router from "hyperapp-router"

// or

import Router from "hyperapp-router"

// or

import RoUTer from "hyperapp-router"
```

src/app.js
jbucaran: Maybe I'm missing something, but I still think this is not going to work when sharing onError with plugins. This is a minor thing, so please send me a PR after we land this one 👍.

src/router.js
jbucaran: I prefer:
```
params.id
// or
(model, actions, { id }) -> id
```

over

```
model.router.params.id
```

jbucaran: But, I was able to extract the match functionality outside onRender.

dodekeract: @jbucaran Your syntax proposal is compatible with mine. You can just call the render function with `model, actions, model.router`.

src/app.js
dodekeract: There will be performance gains. https://jsperf.com/for-in-or-for

src/index.js
dodekeract: @jbucaran Generally I'm in favor of stuff like this. Especially now that the router is basically just an app/plugin I think we can go for this.

I don't think it's necessary yet though. Basically the decision is quite opinionated.

This won't really solve the examples problem though. People will still see that it's `import Router from 'hyperapp-router'`.

src/app.js
jbucaran: Are you talking about for-in when used in arrrays?

src/index.js
jbucaran: Exactly. I think Plugins written in title case would look better, just like Components.

```jsx
app({ plugins: [Router, Logger, Debugger, LevelDB, URLParser, Portals] })
```

src/router.js
jbucaran: Hehe, done, locally. I haven't pushed anything 😄.

src/app.js
dodekeract: Right, yes. Hard to keep track of 10 conversations at the same time.

src/index.js
lukejacksonn: Router could be a plugin but that doesn't mean it has to be. If it is then it negates the `R` vs `r` issue. The awkward thing about the router is that it is not needed for all cases, but in reality it is needed for _a lot_ of cases and if it is not needed (they only have one view/route) then hyperapp already has that covered with very little redundant bytes. Good frontend routers are hard to come by. I think this..
```
model.router = {
	matched: '/some/route/:id',
	url: '/some/route/34593458',
	params: {
		id: 34593458
	}
}
```
with an `onrender` hook, feels like a very robust approach. I'd say for the sakes of argument.. abstract the router out, use it as an example of a really useful plugin (`npm i hyperapp-router -S` is a good practice for developers to get into anyway if we are encouraging a plugin ecosystem) and let people call it what they want.

jbucaran: Thanks @lukejacksonn. Notice the right name is `onRender`. Hooks are application level events so we use camelCase. DOM events are lowercase so we use oncreate, onupdate, onremove for lifecycle methods. Just in case someone sees this in the future.

SkaterDad: For what it's worth, the convention in Hapi is to capitalize anything you 'require'.  It makes it more clear which variables are from an external file, and I agree with @jbucaran that it looks nicer.

src/app.js
```diff
@@ -14,11 +14,12 @@ export default function (app) {
 	var root
 	var batch = []
 
-	var onError = function (err) {
-		if (!hook("onError", function (fn) {
-			fn(err)
-		})) {
-			throw err
+	var onError = function (error) {
+		for (var hookKey in hooks.onError) {
```
dodekeract: `for in` 😢

src/app.js
```diff
@@ -52,17 +53,17 @@ export default function (app) {
 
 			if (typeof action === "function") {
 				container[key] = function (data) {
-					hook("onAction", function (fn) {
-						fn(name, data)
-					})
+					for (var hookKey in hooks.onAction) {
```
dodekeract: 😢

src/app.js
```diff
@@ -52,17 +53,17 @@ export default function (app) {
...
+					}
 
 					if (shouldRender) {
 						var oldModel = model
 						model = merge(model, action(model, data))
 
-						hook("onUpdate", function (fn) {
-							fn(oldModel, model, data)
-						})
+						for (var hookKey in hooks.onUpdate) {
```
dodekeract: 😢

src/app.js
```diff
@@ -97,25 +98,21 @@ export default function (app) {
...
 		}
-		return hookfn !== undefined
 	}
-
+	
 	function render(model, view) {
-		hook("onRender", function (fn) {
-			view = fn(model, view)
-		})
+		for (var i in hooks.onRender) {
```
dodekeract: 😢

jbucaran: @dodekeract Using for in here is the easiest way to avoid checking if onRender is undefined.

src/app.js
lukejacksonn: Just out of curiosity how come `for ... in` is being preferred over `forEach` or `map`

jbucaran: @lukejacksonn for in is used to loop through props of an object. This can be used to traverse an array too, but it's a "hack". The best practice to traverse arrays is using a regular for loop or forEach.

I've changed them all to use a regular for loop, so say bye to for in! :)

lukejacksonn: 🎉

jbucaran:
@dodekeract Thanks for the comments. I removed several instances of for..in in favor of a for loop, updated the hooks system as suggested and revamped the router.

jbucaran:
Hi @tzellman @danigb! Could you take a loot at the new tests? I rewrote the entire spec, but even with 100% coverage maybe there are still features untested or missing cases?

src/app.js
jbucaran: With the new hook system this problem has gone away!

src/index.js
jbucaran: Good to know! Also, components are also capitalized, but they are not classes.

.editorconfig
```diff
@@ -5,11 +5,6 @@ end_of_line = lf
...
 trim_trailing_whitespace = true
 insert_final_newline = true
-indent_style = tab
-
-[*.{diff,md}]
-trim_trailing_whitespace = true
-
-[*.json]
 indent_style = space
 indent_size = 2
```
lukejacksonn: <3

Issue #126, opened 2017-03-04
What if it was only actions?


jbucaran:
I am rewriting parts of the documentation and currently revisiting the eternal reducers Vs effects.

Why do we need to force users to think about reducers and effects instead of just **actions**? 

If an action returns a promise, we know it's an effect, otherwise it's a reducer.

```jsx
app({
  model: {
    counter: 0,
    waiting: false
  },
  actions: {
    add: model => ({ counter: model.counter + 1 }),
    toggle: model => ({ waiting: !model.waiting }),
    waitThenAdd: (model, actions) => {
      actions.toggle()
      return new Promise(resolve => setTimeout(_ => resolve(), 1000))
        .then(actions.add)
        .then(actions.toggle)
    }
  },
  view: (model, actions) =>
    <button
      onclick={actions.waitThenAdd}
      disabled={model.waiting}
    >
      {model.counter}
    </button>
})
```

The only drawback is that Promises are not supported in IE11, so obviously we can't change this right now.

leeoniya:
Promise shims come pretty small: https://github.com/taylorhakes/promise-polyfill/blob/master/promise.min.js

jbucaran:
@leeoniya Thanks! There must be an even smaller implementation that targets only IE10 or IE11, which is what HyperApp supports.

jbucaran:
Alternatively, we could provide this out of the box like this:

```jsx
app({
  model: {
    counter: 0,
    waiting: false
  },
  actions: {
    add: model => ({ counter: model.counter + 1 }),
    toggle: model => ({ waiting: !model.waiting }),
    waitThenAdd: (model, actions) => {
      actions.toggle()
      return resolve => setTimeout(_ => resolve(actions.add).then(actions.toggle), 1000)
    }
  },
  view: (model, actions) =>
    <button
      onclick={actions.waitThenAdd}
      disabled={model.waiting}
    >
      {model.counter}
    </button>
})
```

selfup:
The problem with setTimeout is that we aren't sure how long the request is going to take

jbucaran:
@selfup See my edit. HyperApp receives the function you return from the effect and calls it immediately with our own home-brewed `then` function. 

setTimeout is called and eventually our then is called.

selfup:
Ohh I see! Hmm using Events?

selfup:
I was going to mention Proxy but no IE support. It seems the polyfill is small though:

[...]

Issue #69, opened 2017-02-09
Should HyperApp Wait For The DOM Before Rendering?


dodekeract:
If you only want to make your vote count please just :+1: or :-1: this comment. I added both so it's easy to click.

## Things that need to be done

- [ ] Decide if `window.addEventListener('load')` should be implicit
- [ ] Add documentation for agreed upon behavior

---

## The Problem

When including `app()` in a script in `<head>` hyperapp will crash because it tries to append itself onto the not-yet-existing DOM.

*Note: Some might suggest to just include the script at the bottom of the page, however - in some use-cases (like mine) that's not possible. It also feels like a workaround IMHO.*

This can be solved in (at least) 2 ways:

### Solution 1

Force users to add `window.addEventListener('load', () => app(...))` to their scripts.

```javascript
import {app, html} from 'hyperapp'

window.addEventListener('load', () => app({
	view: html`<div>Example</div>`
}))
```

#### Pro

- less implicit behavior, more control/information for the dev

#### Contra

> It just crashes for no reason, even though I copied the whole example!
\- Frustrated first-time users.

### Solution 2

Implicitly do the same as in solution 1.

```javascript
import {app, html} from 'hyperapp'

app({
	view: html`<div>Example</div>`
})
```

#### Pro

- more straight-forward code
- less stuff to consider/know for the dev

> It just works!

#### Contra

Might be barely considered "Magic". Should be properly documented. However I don't see any use-case where hyperapp is supposed to crash. :smile:

---

## Meta:

[Follow-up issue](https://github.com/hyperapp/hyperapp/issues/62)
[Initial issue: `subs` didn't work when ran by `onload`](https://github.com/hyperapp/hyperapp/issues/61)

jbucaran:
Also related #64.

terkelg:
I'd prefer to keep it out for now. It's good to be explicit, and the majority have no problems with the way it is now. However, that being said I see both sides and I don't really mind if it get included.

dodekeract:
@terkelg Yes, since the `subscriptions` issue is fixed this doesn't bug me anymore either. I'm used to DOM frameworks breaking if not attached to an event listener.

It might also be worth it to investigate if `window.addEventListener('load', () => app(...))` would not **do anything** if the `load` is implicit.

tzellman:
[...]

PR #127, opened 2017-03-04
Collapse reducers and effects into actions. Implement #126.

jbucaran:
**Remove** reducers/effects for a familiar concept: _actions_.

* Actions can be used to update the model/render the view. How? Return the new model, just like with reducers before:

```jsx
app({
    model: 0,
    actions: {
        add: model => model + 1,
        sub: model => model - 1
    }
})
```

* In addition, actions can be used to cause side effects or update the view asynchronously. How? You have two options:

1. Same as before. Run your async tasks inside the function and when you are done call further actions to update the view if necessary.
```jsx
app({
  actions: {
    add: model => model + 1,
    sub: model => model - 1,
    waitThenAdd: (model, data, actions) => {
      setTimeout(actions.add, 1000)
  }
})
```
2. Return a Promise. In this fashion, you can have multiple actions that return promises and chain them together using the [.then](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then) or catch errors via [.catch](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch). 
```jsx
app({
  actions: {
    add: model => model + 1,
    sub: model => model - 1,
    waitThenAdd: (model, data, actions) =>
      new Promise(resolve => setTimeout(resolve, 1000)).then(actions.add)
  }
})
```
If you don't know Promises work or don't want to use them, that's okay. Promises **are not necessary** in order to use hyperapp, but it's nice you could use them if you wanted and hyperapp does the right thing (which is returning them to the caller instead of updating the model/view).

### Caveats

* Before, reducers and effects had different function signatures. Now all actions have the same signature: (model, data, actions, error).
* To know whether an action updates the model/view or not, Hyperapp checks its return value. If you return a primitive/model then it's logical to assume you want to update the model.
* If an action returns undefined, it doesn't make much sense to update the model or render the view.

**Related**:
* https://github.com/hyperapp/hyperapp/issues/126

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -3,14 +3,45 @@ export default function (app) {
     return ""
   }
 
-  var model, actions = {}, reducers, effects
-
+  var model
+  var actions = {}
   var subscriptions = []
   var hooks = {
     onError: [],
     onAction: [],
     onUpdate: [],
     onRender: []
+  }
+
+  var plugins = [app].concat((app.plugins || []).map(function (plugin) {
+    return plugin(app)
+  }))
+
+  var node
+  var root
+  var batch = []
+
+  for (var i = 0; i < plugins.length; i++) {
+    var plugin = plugins[i]
+
+    if (plugin.model !== undefined) {
+      model = merge(model, plugin.model)
+    }
+
+    if (plugin.actions) {
+      init(actions, plugin.actions)
+    }
+
+    if (plugin.subscriptions) {
+      subscriptions = subscriptions.concat(plugin.subscriptions)
+    }
+
+    var _hooks = plugin.hooks
+    if (_hooks) {
+      Object.keys(_hooks).forEach(function (key) {
+        hooks[key].push(_hooks[key])
+      })
+    }
   }
 
   function onError(error) {
@@ -23,52 +54,7 @@ export default function (app) {
     }
   }
 
-  var plugins = app.plugins || []
-
-  var node
-  var root
-  var batch = []
-
-  use(app)
-
-  for (var i = 0; i < plugins.length; i++) {
-    use(plugins[i](app))
-  }
-
-  init(actions, effects, false)
-  init(actions, reducers, true)
-
-  load(function () {
-    root = app.root || document.body.appendChild(document.createElement("div"))
-
-    render(model, view)
-
-    for (var i = 0; i < subscriptions.length; i++) {
-      subscriptions[i](model, actions, onError)
-    }
-  })
-
-  function use(app) {
-    if (app.model !== undefined) {
-      model = merge(model, app.model)
-    }
-
-    reducers = merge(reducers, app.reducers)
-    effects = merge(effects, app.effects)
-
-    if (app.subscriptions) {
-      subscriptions = subscriptions.concat(app.subscriptions)
-    }
-
-    var _hooks = app.hooks
-    if (_hooks) {
-      Object.keys(_hooks).forEach(function (key) {
-        hooks[key].push(_hooks[key])
-      })
-    }
-  }
-
-  function init(container, group, shouldRender, lastName) {
+  function init(container, group, lastName) {
     Object.keys(group).forEach(function (key) {
       if (!container[key]) {
         container[key] = {}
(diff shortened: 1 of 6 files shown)


jbucaran:
@dodekeract Review please 🙏.

lukejacksonn:
Have you seen my proposal [here](https://github.com/hyperapp/hyperapp/issues/126#issuecomment-284192686). I have only just skimmed this PR but I think we are trying to achieve the same thing.

src/app.js
```diff
@@ -84,27 +70,35 @@ export default function (app) {
...
+    root = app.root || document.body.appendChild(document.createElement("div"))
+
+    render(model, view)
+
+    for (var i = 0; i < subscriptions.length; i++) {
+      subscriptions[i](model, actions, onError)
+    }
+  })
+
   function load(fn) {
```
jbucaran: @liadbiz Thanks! :) 

See #69 thread for the original reason we introduced this code.

jbucaran:
I'd like to proceed with this PR 🔥.

Please if you have any further comments, post them soon 🙏 🙏.

lukejacksonn:
I _think_ I am up to date with updates haha!

> Correct me if I am wrong, but your proposal requires actions to always return a Promise?

I'm assuming you are saying this because of `Promise.resolve(action(m,d,a))` if so then no this is not true as `Promise.resolve(42) /=> 42`. If it is not passed a promise then it will resolve immediately with the  primitive value (a way overlooked quality imo).

Seeing as action used to return nothing (or more recently actions which could be done here but there would be no point). The only issue I see with it is what @selfup mentioned which is..

> The problem with that is that Promises are not very performant even if they resolve immediately

I have no solid evidence but I have read that resolving a primitive value is evaluated as `If !IsPromise(x) return x;`, would be interesting to see because this is nice..

```javascript
function (data) {
  for (i = 0; i < hooks.onAction.length; i++) {
    hooks.onAction[i](name, data)
  }
  return Promise.resolve(app.actions[key](model, data, actions))
  .then((result) => {
    if(result) { // Check it wasn't just an effect
      for (i = 0; i < hooks.onUpdate.length; i++) {
        hooks.onUpdate[i](model, result, data)
      }
      model = merge(model, result)
      render(model, view)
    }
  })
}
```

jbucaran:
@lukejacksonn Ah, okay, right. In that case, I don't see what prevents you from doing this already then.

jbucaran:
@lukejacksonn It seems you were suggesting to use Promise directly inside hyperapp?

lukejacksonn:
Yes here https://github.com/hyperapp/hyperapp/blob/master/src/app.js#L82

jbucaran:
@lukejacksonn Even if promises were super fast, which they are not, the problem with this is forcing users to polyfill `Promise`.

lukejacksonn:
I wasn't aware that was a limitation of the project. Most developers are very accustomed to importing [babel-polyfill](https://babeljs.io/docs/usage/polyfill) these days especially if using a bundler, it would mean having to serve an IE friendly version on CDN though perhaps.

Performance is a real issue but I would be intrigued to see how it compares. I [don't think there is going to be much in it]((https://jsperf.com/promise-vs-callback)) given the simplicity of `Promise.resolve` and how much browsers are trying to optimize it.

Let's get this PR merged and I can play around 👍

jbucaran:
@liadbiz I've added a few comments to explain what's going on in certain places. If there are other places you think need further explanation let me know.

jbucaran:
@lukejacksonn Thanks, for looking into perf.

See this: https://github.com/lukeed/polkadot/blob/master/benchmarks/readme.md by @lukeed.

> I wasn't aware that was a limitation of the project. Most developers are very accustomed to importing babel-polyfill these days especially if using a bundler, it would mean having to serve an IE friendly version on CDN though perhaps.

The project has a determined goal of not straying too far away from the vicinity of 1kb. If we introduce `Promise` we break this promise! 

Well, not true, if you are using an ultra modern browser, you wouldn't need a polyfill, so technically we would be 1kb, but right now, we can support IE10 and still be 1kb, which is nice.

The entire code base is a mix of mostly ES3 and some ES5 (forEach, Object.keys) and I've been really trying to keep it that way. 

Not everyone is using promises. Whatever their reasons are, if we use Promise out of the box that's a reason for a Promise-denier to not use hyperapp.

PR #130, opened 2017-03-05
Internal routing hook for <a href=/*>

lukejacksonn:
This behavior was lost during the router refactor. It was quite a nice feature of the router, offering an intuitive means of navigating between views.
```html
<a href='/'>Home</a> <!-- Prevents page reload because /* indicates internal route
```
Is this the kind of thing you'd expect to see as plugin now?
```es6
export const Interlink = app => ({
  subscriptions: [
    (m,a) => {
      window.addEventListener("click", function (e) {
        if (e.metaKey || e.shiftKey || e.ctrlKey || e.altKey) return
        var target = e.target
        while (target && target.localName !== "a") {
          target = target.parentNode
        }
        if (target && target.host === location.host && !target.hasAttribute("data-no-routing")) {
          var element = document.querySelector(target.hash === "" ? element : target.hash)
          if (element) {
            element.scrollIntoView(true)
          } else {
            e.preventDefault()
            a.router.go(target.pathname)
          }
        }
      })
    },
  ]
})
```
Or is it suggested that the developer use `a` tag but with `onclick` instead of `href`
```jsx
<a onclick={e => actions.router.go('/')}>Home</a>
```
Or make a component out of it like
```jsx
export const Link = href =>
  <a href={href} onclick={e => e.preventDefault() || a.router.go(href)}>Internal Link</a>
```
I'm not entirely sure which option to choose as a convention (for myself at least) going forward. Regardless of personal preference, might it be worth promoting a _recommended_ way of navigating between views? Seeing as it is a common common action in many applications.

This probably should have been posted as an issue for discussion but the changed code here is functional and mergable in case it is decided that this is something that should come out the box.

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -26,7 +26,25 @@ export default function (options) {
         addEventListener("popstate", function () {
           actions.router.match(location.pathname)
         })
-      }
+      },
+      function (_, actions) {
+        addEventListener("click", function (e) {
+          if (e.metaKey || e.shiftKey || e.ctrlKey || e.altKey) return
+          var target = e.target
+          while (target && target.localName !== "a") {
+            target = target.parentNode
+          }
+          if (target && target.host === location.host && !target.hasAttribute("data-no-routing")) {
+            var element = document.querySelector(target.hash === "" ? element : target.hash)
+            if (element) {
+              element.scrollIntoView(true)
+            } else {
+              e.preventDefault()
+              actions.router.go(target.pathname)
+            }
+          }
+        })
+      },
     ]
   }
 }


dodekeract:
This could actually work very well as a plugin. Not sure what the best thing to do is though.

jbucaran:
> It won't be needed in all apps and it might be unwanted behavior in some cases.

That's one reason I removed it. Another reason was the implementation is weak, the router never had support for querystrings or hash-based routing. I think a super incredible 5kb router as a plugin that someone can built would be nice to have, but HyperApp's default router should be as simple as possible.

What do you guys think? @dodekeract @lukejacksonn

dodekeract:
@jbucaran I think it depends on the file size difference. With gzip I'd guess this is ~100-200 Bytes. If it's that low I'd say we could add it as an option.

`plugins: [router({inject: true})]`

jbucaran:
That's an interesting syntax.

dodekeract:
@jbucaran It's basically just the normal way to use middleware in frameworks like express and koa.

jbucaran:
@dodekeract Cool. I meant that in the context of hyperapp, obviously. But it doesn't matter, plugins can work any way they want as long as they return a function that has a signature `(options)`.

jbucaran:
@lukejacksonn What do you think? would you like to see Router take care of this out of the box?

lukejacksonn:
> To simplify the examples etc. it'd be best to just ship this with the router

This would in turn also simplify everyone else's code going forward. Imagine everywhere that used to be `<a href='/demo'>Demo</a>` now becomes `<a onclick={e => a.router.go('/demo')}>Demo</a>` or better semantically but worse syntactically `<a href='/demo' onclick={e => e.preventDefault() || a.router.go('/demo')}>Demo</a>`. There Standardizing the intent of this action. 

> however it won't be needed in all apps

True, a project that doesn't require a router certainly would not need this, but the router is shipped as part of the bundle anyway so an extra 100 bytes is going to have minimal impact. If the router is employed then, generally this is a behavior that you'd expect to come with it (given that as soon as an accidental page reloads happens that SPA feeling is shattered and state is lost).

>  it might be unwanted behavior in some cases.

For those cases, `data-no-routing` is available or as you suggest some option like `Router({ internalLinks: true })` might suffice.

### In conclusion..

*It is a tough call..* as a developer I much much prefer the `<a href='/demo'>Demo</a>` syntax to describe links to internal views and want to go to as little effort as possible to take advantage of that SPA behavior (not have the page reload and lose state).

Arrghhhh! 😅 can anyone else offer any insight, preference?

jbucaran:
> True, a project that doesn't require a router certainly would not need this, but the router is shipped as part of the bundle anyway so an extra 100 bytes is going to have minimal impact.

The bundle is included in the hyperapp distribution for the CDN, but remember that will be stripped from your production bundle when you use webpack/rollup tree-shaking. Incidentally, **that's** why we're using ES6 `import`/`export` module syntax.

> Arrghhhh! 😅 can anyone else offer any insight, preference?

Yeah, I know and you're right. I'll put it back in and try to add some tests.

dodekeract:
@jbucaran If we could add it as opt-in or opt-out it'd be really nice. It's kind of a hack anyway.

Other routing libraries solve it by providing a Link component e.g. `import Link from 'react-router/Link'.

jbucaran:
> Other routing libraries solve it by providing a Link component e.g. `import Link from 'react-router/Link'.

That'd work too.

jbucaran:
Sorry, I'm doing keys and #141. Will take care of this soon tho.

lukejacksonn:
No worries @jbucaran I will _try_ write some tests for it also

jbucaran:
@lukejacksonn Can we move this to https://github.com/hyperapp/router ? 💥

PR #131, opened 2017-03-06
Updated examples in Readme

vishnun:
The code said `actions: {}` but the codepen says `reducers: {}`
So, updating the readme to be in sync with the actual working code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ app({
 ```jsx
 app({
   model: 0,
-  actions: {
+  reducers: { // actions
     add: model => model + 1,
     sub: model => model - 1
   },
@@ -76,7 +76,7 @@ app({
 ```jsx
 app({
   model: "",
-  actions: {
+  reducers: { // actions
     text: (_, value) => value
   },
   view: (model, actions) =>
@@ -205,7 +205,7 @@ app({
         <button onclick={actions.add}>add</button>
       </p>
     </div>,
-  actions: {
+  reducers: { // actions
     add: model => ({
       input: "",
       todos: model.todos.concat({


jbucaran:
Thanks @vishnun, after #127, reducers/effects became _actions_, and I'm still updating everything. Everything should be in synced soon.

jamen:
So I'm just using hyperapp today.  I think I remember things in the wiki need revisions too.

jbucaran:
Hi @jamen! Yep. Going through those too as we speak. The CodePen examples are out of date because I haven't released 0.7.0, but I intend to do that very soon too.

dodekeract:
@jamen That's actually a good point against using the wiki. If we move the docs to master they will always fit the revision you checked out in git.

jbucaran:
@dodekeract I don't think the most common use case for hyperapp is to clone the repo. 

I prefer a wiki (and will use a wiki from now on in future projects too). This is a decision I reached after more than 200 commits to the now-deprecated user manual and weighing cons and pros in my head. The main reasons I favor public wikis are:

1. Makes editing docs easy for maintainers. I'd like to commit something and see how it looks and change my mind based on gut and feeling a lot.
2. Makes editing docs easy for anyone else since they don't have to create a PR. They don't even have to let maintainers know. I like this openness.
3. It's a GitHub feature and I am using GitHub. If I provide docs in the repo or as a separate document, then I'd have to use the wiki for something else or disable it. Disabling the wiki is also acceptable if docs go on a website, but websites cost incur a cost, whereas using the wiki is "free".



Number 2. above is

dodekeract:
@jbucaran The use-case isn't to "clone the repo" the use case is to be able to work with older versions of hyperapp **without** knowing every single detail.

There'll be issues when we switch from v1 to v2 since we have to destroy the old wiki.

jbucaran:
> There'll be issues when we switch from v1 to v2 since we have to destroy the old wiki.

No idea what that means.

jbucaran:
Okay, I see what you mean.

Hyperapp v1, then v2 changes everything and there would be no v2 documentation.

dodekeract:
@jbucaran `v2` will by definition have breaking changes compared to `v1`. We only have one wiki. Having the docs in master means they're always the important ones, no matter which revision you look at.

jbucaran:
🤔 I didn't consider that possibility. That definitely poses an interesting problem. 

I think this particular project has the potential to avoid that scenario, since I hope to freeze the API with 1.0. If there's a 2.0 there _may_ be new features, but no removal of older ones. 

Well, that's not something I'm worried about at this moment, but thanks for opening my eyes.

liadbiz:
> If there's a 2.0 there may be new features, but no removal of older ones.
not really, IMO, vue and angular actually have some big changes between v1 and v2, you can not always control what software is going on...however, wiki is a good place to place some tutorial that explains concepts and architecture, which mostly will not change between versions, to help users and contributors to undersatnd repo, but api need another place like standalone website.

PR #133, opened 2017-03-06
Support camelCase lifecycle/DOM event keys

farism:
Backstory: 

When trying to adhere to Typescripts DOM library typings (and potentially other vdom implementations), it is expecting camelCased names for event handlers, e.g. `onInput` rather than `oninput`. When this property key is parsed by hyperapp and the `on` prefix is removed, we are potentially left with capitalized event names.

Changes:

This PR forces lowercased event names. According to https://developer.mozilla.org/en-US/docs/Web/Events event names are all lowercased, so this should be a safe change.

I did not see any tests surrounding event handling, so I didn't alter anything there. If this is something that should be tested I'd be happy to add it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 .vscode/
 
 # Logs
-.log*
+*.log*
 yarn.lock
 
 # Misc
--- a/src/app.js
+++ b/src/app.js
@@ -171,7 +171,7 @@ export default function (app) {
         : document.createElement(node.tag)
 
       for (var name in node.data) {
-        if (name === "oncreate") {
+        if (name === "onCreate") {
           defer(node.data[name], element)
         } else {
           setElementData(element, name, node.data[name])
@@ -233,7 +233,7 @@ export default function (app) {
       if (value === undefined) {
         removeElementData(element, name, oldValue)
 
-      } else if (name === "onupdate") {
+      } else if (name === "onUpdate") {
         defer(value, element)
 
       } else if (
@@ -263,8 +263,8 @@ export default function (app) {
 
       batch.push(parent.removeChild.bind(parent, element))
 
-      if (oldNode && oldNode.data && oldNode.data.onremove) {
-        defer(oldNode.data.onremove, element)
+      if (oldNode && oldNode.data && oldNode.data.onRemove) {
+        defer(oldNode.data.onRemove, element)
       }
 
     } else if (shouldUpdate(node, oldNode)) {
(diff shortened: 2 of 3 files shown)


dodekeract:
@farism While I personally am in favor of allowing `onClick` `onInput` etc. hyperapp's current policy is that this is invalid syntax. The given reasoning was that `toLowerCase()` has a small cost.

So as long as @jbucaran doesn't change that policy this PR isn't necessary.

farism:
@dodekeract Understood, there is a cost both in file size (extra characters) and in performance. Although I believe that unless you are emitting thousands of events per second the performance loss is probably negligible. It might be a case of premature optimization. So the filesize issue may be the biggest cost? In the vein of keeping things 1kb that is.

I think it could be important for hyperapp to support camel cased event names if we want to support/have good interoperability with other (potentially typed) JSX implementations.

dodekeract:
@farism I completely agree.

jbucaran:
@farism I have no issue with this change, but it will create a consistency problem, because this does not account for lifecycle methods: oncreate, onupdate and onremove.

farism:
I see @jbucaran, good point. Do you think this is something that should be addressed at all, and if so in this PR? This PR is targeting compatibility more than anything, there wouldn't be anything  preventing someone from continuing to use lowercased key names. Allowing the _option_ for inconsistent casing could be a bad from a style point of view, so it comes down to whether you are willing to make that compromise.

Perhaps it isn't such a big deal, as I think most people would understand the difference between hyperapp lifecycle events and native DOM events.

jbucaran:
Another option would be to use camelCase for everything. What do you think?

dodekeract:
@jbucaran I'm always in favor of that idea.

jbucaran:
Cool! In that case, we can close here and simply make change the lifecycle methods to be onCreate, onUpdate and onRemove. .toLowerCase() will not be necessary either!

farism:
I think camelCase would be more inline with what other vdom implementations are doing, and probably more natural for developers coming from React et al.

Sounds good!

jbucaran:
@farism Do you want to take care of this? I can update the docs.

farism:
Yes I can take care of updating the lifecycle events, but I am confused by getting rid of `.toLowerCase()`. It seems it is still needed if we want to support native DOM events, which are expected to be all lowercase? Or am I missing something.

jbucaran:
@farism Yes, you are right. I screwed up hehe. 👍

farism:
@jbucaran I think it should be good now

jbucaran:
LGTM

jbucaran:
Available from 0.7.1.

PR #134, opened 2017-03-06
Enable component nesting with own model and actions

j3ski:
Hyperapp is heavily inspired by Elm, however it lacks one of its best features.
Elm enables to compose applications by nesting components, which themselves could work as separate apps.
In Hyperapp a nested component must know the app's model tree structure to work correctly.
This PR enables actions to take 2 additional arguments: the model they are working on & the actions they can trigger, hence they can be made fully independent of the whole app tree & reusable. The only limitation here is that their state must be an object to easily update it when necessary.
This came with a cost of making the internal model mutable (doesn't matter as long as actions/subscriptions don't modify it as a side effect - it would break anyway).

An analog of Elm's App.map would come in handy here, not included to keep the footprint small.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -25,7 +25,7 @@ export default function (app) {
     var plugin = plugins[i]
 
     if (plugin.model !== undefined) {
-      model = merge(model, plugin.model)
+      updateModel(model, plugin.model)
     }
 
     if (plugin.actions) {
@@ -65,12 +65,14 @@ export default function (app) {
       var i
 
       if (typeof action === "function") {
-        container[key] = function (data) {
+        container[key] = function (data, modelFragment, localActions) {
+          modelFragment = modelFragment || model
+          localActions = localActions || actions
           for (i = 0; i < hooks.onAction.length; i++) {
             hooks.onAction[i](name, data)
           }
 
-          var result = action(model, data, actions, onError)
+          var result = action(modelFragment, data, localActions, onError)
 
           if (result === undefined || typeof result.then === "function") {
             return result
@@ -80,7 +82,7 @@ export default function (app) {
               hooks.onUpdate[i](model, result, data)
             }
 
-            model = merge(model, result)
+            updateModel(modelFragment, result)
             render(model, view)
           }
         }
@@ -122,8 +124,8 @@ export default function (app) {
     batch = []
   }
 
-  function merge(a, b) {
-    var obj = {}
+  function merge(a, b, into) {
+    var obj = into? a: {}
     var key
 
     if (isPrimitive(b) || Array.isArray(b)) {
@@ -138,6 +140,12 @@ export default function (app) {
     }
 
     return obj
+  }
+
+  function updateModel(modelFragment, newModelFragment) {
+    var isRoot = modelFragment === model;
+    var merged = merge(modelFragment, newModelFragment, !isRoot);
+    model = isRoot? merged: model;
   }
 
   function isPrimitive(type) {
--- a/src/app.js
+++ b/src/app.js
@@ -124,28 +124,33 @@ export default function (app) {
     batch = []
   }
 
-  function merge(a, b, into) {
-    var obj = into? a: {}
-    var key
+  function merge(a, b) {
+    var obj = {}
 
     if (isPrimitive(b) || Array.isArray(b)) {
       return b
     }
 
-    for (key in a) {
-      obj[key] = a[key]
-    }
-    for (key in b) {
-      obj[key] = b[key]
-    }
+    assign(obj, a)
+    assign(obj, b)
 
     return obj
   }
 
+  function assign(obj, src) {
+    var key
+    for(key in src) {
+      obj[key] = src[key]
+    }
+  }
+
   function updateModel(modelFragment, newModelFragment) {
     var isRoot = modelFragment === model;
-    var merged = merge(modelFragment, newModelFragment, !isRoot);
-    model = isRoot? merged: model;
+    var merged = isRoot
+      ? merge(modelFragment, newModelFragment)
+      : assign(modelFragment, newModelFragment)
+
+    model = isRoot? merged: model
   }
 
   function isPrimitive(type) {
@@ -206,9 +211,7 @@ export default function (app) {
 
   function setElementData(element, name, value, oldValue) {
     if (name === "style") {
-      for (var i in value) {
-        element.style[i] = value[i]
-      }
+      assign(element.style, value)
 
     } else if (name[0] === "o" && name[1] === "n") {
       var event = name.substr(2)
--- a/src/app.js
+++ b/src/app.js
@@ -138,8 +138,7 @@ export default function (app) {
   }
 
   function assign(obj, src) {
-    var key
-    for(key in src) {
+    for(var key in src) {
       obj[key] = src[key]
     }
   }
--- a/test/app.test.js
+++ b/test/app.test.js
@@ -329,6 +329,47 @@ describe("app", () => {
         view: _ => h("div", {}, "")
       })
     })
+
+    it("allows to pass a model fragment into an action", () => {
+      app({
+        model: {
+          todos: [{id: 1}]
+        },
+        actions: {
+          updateTodo: (todo, text) => ({text})
+        },
+        view: model => h("div", {}, `${model.todos[0].text}`),
+        subscriptions: [
+          (model, actions) => actions.updateTodo('foo', model.todos[0])
+        ]
+      })
+
+      expectHTMLToBe(`
+					<div>
+						<div>foo</div>
+					</div>
+				`)
+    })
+
+    it("passes scoped actions when they are passed", () => {
+      app({
+        model: {
+          todos: [{id: 1}]
+        },
+        actions: {
+          todos: {
+            update: (_, d, actions) => {
+              expect(Object.keys(actions)).toEqual(['update'])
+            }
+          },
+          other: _ => _
+        },
+        view: _ => h("div", {}, ''),
+        subscriptions: [
+          (model, actions) => actions.todos.update('foo', model, actions.todos)
+        ]
+      })
+    })
   })
 
   describe("hooks", () => {


jbucaran:
Hi @j3ski! This sounds interesting, but I don't really understand what's going on here.

Can you provide a simple example that illustrates what's this?

jbucaran:
`merge` is also heavily used throughout the engine, so I suspect some perf cost here. I'll pull your changes and run them through the js-framework-benchmarks to see.

jbucaran:
I ran the test twice, and this was the best result.

screen shot 2017-03-06 at 22 25 46

We'd go from 1.26-ish to 1.33-ish.

j3ski:
Hi @jbucaran! Express feedback :) I'll try my best (and also provide a sample repo).
If you recall the Elm tutorial, there's the Counter example, where you firs build the Counter and then you can use it building a Counter List, not changing anything in the Counter code. I don't see this possible in Hyperapp.

Let me explain on a ToDo List app:
First the todo component, it doesn't know anything about the rest of the app, it's only concerned with its model/view and updates.
```javascript
//todo.js

const view = (model, actions) => (
<li onclick={actions.toggle} 
      style={{textDecoration: model.completed? 'line-through': 'none'}}>
      {todo.text}
  </li>
)
const init = text => ({text, completed: false})
const actions = {
   toggle: model => ({...model, completed: !model.completed})
}

export {view, init, actions}
```

Now the todo is used by the todolist, which only can add (and probably delete) todos

```javascript
//todolist.js
import {init as todoInit, view as todoView, actions as todoActions} from 'todo'
import appMap from 'appMap' //binds the model and/or actions to the additional params of actions.

const view = (model, actions) => (<div>
  <form onsubmit={actions.add}>...</form>
  <ul>
    {model.todos.map(todo => todoView(todo, appMap(actions.todo, {model: todo})))}
  </ul>
</div>)

const init = () > ({
  todos: [],
  inputText: ''
})

const actions = {
  todo: todoActions,
  addTodo: model => ({...model, inputText: '', [...model.todos, initTodo(model.inputText)]})
  //probably some more actions
}

export { init, actions, view}
```

And now call app on it.
```javascript
//index.js
import {app} from 'hyperapp'
import {init, actions, view} from 'todolist'

app({model: init(), actions, view})
```

I see the slowdown. If we would restrict the app model to accept only objects, maybe we could swap merge to Object.assign for model updates?

jbucaran:
@j3ski Thanks. While I go through this code, can you check the dynamic counters [CodePen](http://codepen.io/jbucaran/pen/RKYNeM) and let me know if we're talking about the same thing?

jbucaran:
> @j3ski ...there's the Counter example, where you first build the Counter and then you can use it building a Counter List, not changing anything in the Counter code.

Sounds good.

> Let me explain on a ToDo List app:

I'd rather like to see an example using a Counter / Counter List first. We can move on to a Todo example later.

tunnckoCore:
> maybe we could swap merge to Object.assign for model updates?

HyperApp is oriented to IE10, that's why we don't use Object.assign.

dodekeract:
@tunnckoCore Hmm. Maybe we could rely on a polyfill for that though?

jbucaran:
> If we would restrict the app model to accept only objects, 

🤔 Don't know.

> maybe we could swap merge to Object.assign for model updates

Object.assign is not available in IE10/11, so no. But maybe someday :)

j3ski:
@tunnckoCore, @dodekeract @jbucaran 
probably we could do what merge does, only one way (merge props of the new state to the old one, will work on it tomorrow)

@jbucaran I'll come up with some more code tomorrow. Until then, can I somehow run that perf. benchmark? It'd be useful when tweaking.

As for the counter example:
```javascript
// counter.js

const view = (model, actions) => (
  <div>
    <button onclick={actions.increment}>+</button>
    <span>{model.value}</span>
    <button onclick={actions.decrement}>-</button>
  </div>
)

const init = (value = 0) => ({
  value
})

const actions = {
  increment: model => ({value: model.value + 1}),
  decrement: model => ({value: model.value - 1})
}

export {init, actions, view}

// now you could just run app({model: init(), view, actions}) and get a working counter
```

```javascript
//counter-list.js
import Counter from 'counter'
import appMap from 'app-map'

const view = (model, actions) => (
  <div>
    <div>
      <button onclick={actions.add}>Add counter</button>
      <button onclick={actions.remove}>Remove counter</button>
    </div>

    <ul>
      {model.counters.map(counter=> (
        <li>
          {Counter.view(counter, appMap(actions.counter, {model: counter}))}
        </li>
      ))}
    </ul>
  </div>
)

const init = () => ({
  counters: []
})

const actions = {
  add: model => ({ counters: [...model.counters, Counter.init()] }),
  remove: model => ({ counters: model.counters.slice(0, -1)}),
  counter: Counter.actions
}

//now we can run app({model: init(), actions, view}) and get a list of counters
```

appMap would be something like:
```javascript
appMap(actions, options) {
    const mappedActions = {};
    Object.keys(actions).forEach(key => {
       if(typeof actions[key] === 'function') {
           mappedActions[key] = function(data, model, localActions) {
               return actions[key](data, model || options.model, options.actions || localActions)
           }
       } else {
           mappedActions[key] = appMap(actions[key], options);
       }
    });

    return mappedActions
}
```

Thanks for all the comments.

dodekeract:
I don't understand how I'd call `actions.router.go` from any other namespace with this.

j3ski:
@dodekeract it's up to the dev to pass actions&model to the nested component's view. You can pass the global namespace or add the router actions to your new namespace (and bind them or not to any model)

j3ski:
I've pushed some changes & created a pen with a working counter list (hyperapp is loaded from a <script> tag in the html)

http://codepen.io/Jedrski/pen/zZKQbp

jbucaran:
@j3ski What is appMap?

j3ski:
@jbucaran 
An analog of http://package.elm-lang.org/packages/elm-lang/html/2.0.0/Html#map
I'm pretty sure it was previously exposed in Html.App hence I followed that convention.

The role of appMap is to bind actions to the local model/actions. If it wasn't there, all event handlers inside the Counter would have to be something like `{e => actions.action(data, model, actions)}`. There's a simple version of this function in the codepen I provided [here](https://github.com/hyperapp/hyperapp/pull/134#issuecomment-284641167)

jbucaran:
@j3ski Gotcha. And how do you suggest we inform users about this? It seems that if this feature is accepted hyperapp should provide a map function out of the box.

j3ski:
@jbucaran 
I don't think it's necessary to include the map function in the core library if we want to keep it small? If that's not an issue than I can provide the mapping utility.

This change should come with updated docs (a nesting/component architecture part?) & an example to communicate it.

jbucaran:
@j3ski I think you've proved your point, this is doable! Now comes the hardest part: make it accessible to the masses.

We need to come up with a more polished UX for this. What do you think?

src/app.js
```diff
@@ -124,28 +124,33 @@ export default function (app) {
...
-      obj[key] = b[key]
-    }
+    assign(obj, a)
+    assign(obj, b)
 
     return obj
   }
 
+  function assign(obj, src) {
+    var key
```
dodekeract: Should be moved into the `for`.

j3ski: Done

j3ski:
@jbucaran absolutely agree. Incorporating an appMap (I'll call it map from now) would add more than a 100 bytes, but maybe there's a better way to get this working.

 Definitely needs good documentation when shipped

jbucaran:
@j3ski Thanks! I'm more worried about how to explain this in layman's terms. What are the advantages of this feature in plain English? Incidentally, this is always the problem with FL.

At "first glance" it looks rather as if it will make things more complicated. Don't worry so much about the implementation, but coming up with a simpler example to get started.

AutoSponge:
I made an example of how someone could achieve nested components where children don't know the entire model structure.

[gomix example](https://gomix.com/#!/project/hypure-nested)

The ideas are borrowed from redux using purestate. By passing a dispatch function and the appropriate model, each sub-component can separately mutate its state and call for a re-render.

rbiggs:
For me, this is the biggest drawback with Hyperapp. I have been building UI libraries since '96, the ole DHTML days. It's all about creating reusable widgets/components that other people can drop into their project and use. Angular, Aurelia, Ember, Inferno, Preact, React, Riot, Vue - all have component architectures. Elm has one too which is used by Elm UI. I'd like to see either a baked-in component solution like this or similar.

jbucaran:
@rbiggs I really want hyperapp to encourage users to think simply. 

You can totally create staless components and reuse them in totally different projects/

A stateless component doesn't know its host's _model_ or _actions_. It only needs to know how to render itself given some data. These data can be props, actions or a mixture of both.

farism:
@jbucaran i think what @rbiggs is saying (and the point of this PR and @AutoSponge example) is that while render can be isolated, the actions currently are not. If I define an action in a deep component and then call that action, I am passed the full application model from the root. Therefore to update the correct section of the model just for that component, I would need to know about the whole model structure in between. This is what I've been referring to when I say fractal.

jbucaran:
@farism Can we justify this feature by posing a problem hyperapp can't solve easily? 

Then we can say feature X was added because without it you can't create components with own model/actions. See example [?].

AutoSponge:
Because there are so many ways to solve state (streams, redux store, cursors, event emitters), I don't want to solve for or encourage one over another. I'd rather figure out how to enable all of them.

Things that cause boilerplate so far:
1. Only returning `model` from an `action` causes render.
1. Subcomponents can not trigger a render of the app without being passed an `action`.
1. Subcomponents can not trigger a render of themselves.

If there were a separate library for folks wanting to do this called `hyperapp-component`, I could see handling the interface so that these problems are abstracted away. And if the implementation was tied to a specific model, like `hyperapp-redux-component`, you can make even more affordances for convenience. 

But I'm not sure `hyperapp` needs to solve this by enforcing a particular pattern in core.

jbucaran:
> @j3ski Elm enables to compose applications by nesting components, which themselves could work as separate apps.

Do you mean completely isolated / stateful components? If so, how do they communicate with each other?

jbucaran:
I was reading this:
* https://guide.elm-lang.org/reuse/
* https://guide.elm-lang.org/reuse/checkboxes.html

j3ski:
@jbucaran It looks like I was stuck with the old Elm approach.
If you read the [1.1.0 docs](): http://package.elm-lang.org/packages/elm-lang/html/1.1.0/Html-App for the 'map' function it states:
> This function is useful when nesting components with the Elm Architecture

While for [2.0.0](http://package.elm-lang.org/packages/elm-lang/html/2.0.0/Html) it reads:
> This should not come in handy too often. Definitely read [this](https://guide.elm-lang.org/reuse/) before deciding if this is what you want.

However it's still useful for more complex solutions like an [autocomplete](https://github.com/thebritican/elm-autocomplete) component.

The main difference between Hyperapp and Elm in this term is that Elm is event-based, while Hyperapp passes state-mutation functions into the view and hence, as @farism said, it'd be good to have a way to decouple actions & model.

The main point is - if you want to develop some *reusable* piece of code i Hyperapp, that not only *displays* something, but actually *does* something, you'll gonna have a hard time.

This may not be required by regular devs (in NoRedInk they needed this [twice in 30k LOC](https://guide.elm-lang.org/reuse/more.html)), but probably is desireable by people who want to create libraries for Hyperapp.

jbucaran:
@j3ski Worth noting that _this_ is this link: https://guide.elm-lang.org/reuse/.

Quoting the Elm Wizards:
> If you are coming from JavaScript, you are probably wondering “where are my reusable components?” and “how do I do parent-child communication between them?” A great deal of time and effort is spent on these questions in JavaScript, but it just works different in Elm. We do not think in terms of reusable components. Instead, we focus on reusable functions. It is a functional language after all!
> ...
>In the end, I think we end up with something far more flexible and reliable than “reusable components” and there is no real trick. We will just be using the fundamental tools of functional programming!

jbucaran:
@j3ski Can you take a moment and go through my [multicounter example](http://codepen.io/jbucaran/pen/RKYNeM) and tell me what you think? 🙏

j3ski:
@jbucaran I was just thinking of creating something similar. I see your point :+1: 
Looks as if this feature isn't needed after all?
The approach with designing views with an API should definitely be documented in the guide.

jbucaran:
> The approach with designing views with an API should definitely be documented in the guide.

Everything should be documented! You know it works if you create a new page in the Wiki, it doesn't have to perfect or anything, we'll get there.

jbucaran:
https://github.com/hyperapp/hyperapp/wiki/_new

j3ski:
@jbucaran on it!

farism:
I have looked at the example before, the problem I have with this approach is with `counterReduce`, specifically how you are doing `model.map()`. This is a simplistic case because your model is simple, but it does not scale because it is tying this reducer to the structure of the model.

Imagine a had a model with two pages, each with their own sets of counters, e.g. 

```js
{
  pageA: [],
  pageB: [],
}
```

I would have to bind additional structure information to the `add` and `sub` (more than just my index) when calling. This is not considered fractal design. At any layer of your application you should only care about your direct parent or your direct children. Anything outside of that can be handle with an effect/cmd.

RE the elm guide, I think that it is open to interpretation. From my view, it is not saying "don't do nested architecture". I think the message to be had is "don't make everything into a 'component' for the sake of making it into a component", i.e. if it's a one time use just keep that view's state at the top level of the model and create a view helper. But if I have N of these views, and they aren't static and I want them to manage updating themselves, I may need some form of encapsulation. It's a common pattern in libraries/reusable components to follow a nested/fractal architecture. 

Here is an example of simple nested architecture in elm, I apologize if not everyone is familiar with the syntax

https://github.com/farism/mmo-client/blob/elm/src/Main.elm
https://github.com/farism/mmo-client/blob/elm/src/UI.elm
https://github.com/farism/mmo-client/blob/elm/src/Chat.elm

The interesting parts are the `Msg` union types in each file and the `update`. 

Notice how in `Main` I am created a `UI` msg constructor that accepts a `UI.Msg` (imported from UI module). I handle that message in `Main`, but all I am doing is calling `UI.update` and passing the piece of state that I own. 

This pattern is continued with `UI > Chat`. I handle the `Chat.Msg` and call `Chat.Update`, passing in the `chat` slice of my model. 

But when I get down to chat, my `update` method has no idea about the outside world. I am simply passing a model from my parent, and the original message that I sent, and I handle it and return the result, and my parent is responsible for setting it into the state. At each level I am doing a `Cmd.map` and `Html.map` on the `Msg`'s that the children are emitting.

This is what I was referring to when I said a few days ago about "nested" actions. I wasn't talking about the structure of the `actions` object, but rather the way that messages are composed up the hierarchy. 

If something like this is easily obtainable without too much boilerplate in hyperapp's current form, we should talk about how to document it. My issue is primarily all the boilerplate required, it's easy to get lost in it through many layers.

jbucaran:
I thank you all for your time and effort on this issue, but I think this conversation lacks the formulation of a problem, something that @farism or @j3ski (or anyone) think current hyperapp won't make easy, hence the need for this feature.

If you produce a problem, I'll try to find a way to fix it using current hyperapp. While I do that, anyone can try to come up with a "pseudo" implementation using hyperapp + requested feature.

I think we'll learn a lot more from this exercise than any conversation.

Please, don't say that this problem exists when building a very large application, as that is not going to get this feature landed.

farism:
A common pattern for solving this in `Redux` is to distribute a reducer + HOC pair with your library/shared component that reserves a top level namespace in your state. This is how [`redux-form`](https://github.com/erikras/redux-form) works, for example. 

Then, they often provide a higher-order component that accepts some form of 'key', and does the Redux `connect` under the hood, and then they manage the state on a component-by-component basis using these keys. But then you run into issues of wanting the keys to be dynamic, and a whole other host of problems. Not to mention if the parent of one of these components wants to access the child date, is not stored under that model; it's stored under some other location in the state tree.

It's not the best pattern, and a little clunky, but it works. This is how we have currently architected a few of our shared components being used in production now. And this is in a fairly large project with at least 50k loc. We are finding it hard to maintain and add cross-component features.

farism:
@jbucaran I think the issue at hand, or the actionable that could come out of this, is to keep hyperapps _capabilities_ aligned with what elm is capable of (when it comes to architecture, not feature set). You may not reach for nested architecture frequently, but in my eyes it is an essential first-class paradigm of the Elm language. It may not be how things should always be written, but if you are writing something reusable it is very elegant.

jbucaran:
> @farism is to keep hyperapps capabilities aligned with what elm is capable of

Sure, but this is not a requirement either. HyperApp is far from an Elm clone, merely Elm-like, You could also say it's Redux-like.

farism:
@jbucaran Personally, when it comes to this particular PR and how it is implemented and what the API is, I am not too opinionated. And it's getting somewhat lengthy so perhaps we should open an issue for this discussion and have an ongoing forum about architecture/DX?

jbucaran:
@farism SGTM! Please make one 👍 and let's take it from there.

@j3ski Can we close here for now? 🙏

j3ski:
@jbucaran Sure thing. Once again thanks for your time and efford :+1:

foxdonut:
I am late to this discussion but am wondering the same. Repeating here what I wrote in the slack channel: I coded the following example in an attempt to follow the advice given in the Elm links that were posted: write view functions instead of "reusable components" and essentially pass in callbacks for those views to be able to trigger model updates.

http://codepen.io/foxdonut/pen/gmgOOJ?editors=0010

The only thing that doesn't quite sit right with me is the way to wire up communication between the `randomGif` part "signaling" that `newGifSuccess` has occurred, and the `counter` part that needs to "listen" to that signal -- while having access to the whole model -- in order to produce a model update of its own, i.e. increasing the counter. Does anyone have any ideas about that?

I've explored and experimented with a few variants, but the core ideas/principles remain the same:

- a model as the single source of truth;
- views as functions of the model;
- views need a way to trigger model updates, which automatically re-renders the view;
- views have their own  model, which is then included within the top-level model. views should not know "where" their model is within the top-level model;
- views should be reusable multiple times within the same top-level model.

I really appreciate the simplicity of `hyperapp`. But we need some guidance to how to organize code beyond simple examples.

jbucaran:
> I really appreciate the simplicity of hyperapp. But we need some guidance to how to organize code beyond simple examples.

I hope I can help with this part. There's a lot of stuff on my todo, and this is one is at the top for sure.

Issue #117, opened 2017-02-27
prevent images and iframes reloading on patch?


zaceno:
Hi!

I have a view which renders two iframes (one on screen, and one translated off-screen). I slide the visible one off-screen, and the hidden one slides in, using CSS transitions.

 There are three states to my transition:
1 when we are just showing the current slide (just one iframe),
2 when the next slide is loading off screen while the current slide is still on screen,
3 when the current slide is sliding out,  and the new one is sliding in

(and it goes in a loop, so after 3, we go back to 1 for Xms, then 2 for Yms and so on)

I have it all working correctly using reducers to set the states, and a "transition" effect that calls the reducers after setTimeouts. My view sets classes on the iframe-tags to make it use a CSS transition and slide

The only problem is, that the view we render at stage 1 causes the iframe to reload (making a blinking effect, which defeats the whole idea of this "load and slide in")

My only guess as to why this is, is that the patch-function is somehow determining that the whole iframe tag needs to be updated (thereby changing the src attribute as well), when I remove the "entering" class. But it doesnt seem to blink when I *add* the "entering" class.

I tried to read the patch-logic to find a way to add and remove classes without touching the src attribute but couldn't figure it out. Is there a way?

Or perhaps there is a way to use the lifecycle methods to capture and prevent the iframe from being updated, and it could let me "manually" modify the class property?

jbucaran:
> I tried to read the patch-logic to find a way to add and remove classes without touching the src attribute but couldn't figure it out. Is there a way?

What does the src attribute has to do with adding/removing classes?

zaceno:
Ok, to explain my issue more simply (probably should have started here): I have a view that renders an iframe basically according to this:

```
function (model, actions) {
    var cls;
    if (model.state === 'loading') cls = 'loading'
    if (model.state === 'transitioning') cls = 'entering'
    if (model.state === 'showing') cls = 'showing'
    h('iframe', {src: model.url, class: cls)})
```

I have noticed that when `model.state` goes from 'loading' to 'transitioning', there is no visible reload of the iframe. Indicating to me that hyperapp did not touch the actual dom-node, but only updated the class attribute of the domNode (as I intended)

From 'transitioning' to 'showing' there IS a visible reload of the entire iframe. Indicating either that the entire DOMNode was replaced (or at the very least, the src attribute was fiddled with.) -- which is what I would like to avoid.

[...]

PR #135, opened 2017-03-06
Allow specifying nodes to insert

zaceno:
In order prevent unnecessary rebuilding of the dom, it is some-
times necessary to specify that an element *must* be inserted
as a new element, and not replace or update an existing element.

This commit allows you to specify this via:

    h(tag, {_insert: true, ...}, [...])

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -252,8 +252,11 @@ export default function (app) {
 
   function patch(parent, oldNode, node, index) {
     if (oldNode === undefined) {
-      parent.appendChild(createElementFrom(node))
-
+      if (parent.children.length > index) {
+        parent.insertBefore(createElementFrom(node), parent.children[index])
+      } else {
+        parent.appendChild(createElementFrom(node))
+      }
     } else if (node === undefined) {
       var element = parent.childNodes[index]
 
@@ -277,10 +280,14 @@ export default function (app) {
 
       var len = node.children.length, oldLen = oldNode.children.length
 
-      for (var i = 0; i < len || i < oldLen; i++) {
+      for (var i = 0, j = 0; i < len || j < oldLen; i++) {
         var child = node.children[i]
-
-        patch(element, oldNode.children[i], child, i)
+        if (child && child.data && child.data._insert) {
+          patch(element, undefined, child, i)
+        } else {
+          patch(element, oldNode.children[j], child, i)
+          j++
+        }
       }
     }
   }


zaceno:
Fixes https://github.com/hyperapp/hyperapp/issues/117

jbucaran:
@zaceno I see! Why don't we just use insertBefore for everything instead?

zaceno:
@jbucaran Not sure I follow. We can't use insertBefore in a dom parent with no children, because there's nothing before which to insert. But perhaps I misunderstood your question?

jbucaran:
@zaceno So, instead of using this flag, we need to check if it's possible to use insertBefore and use it, otherwise fallback to appendChild.

Does that make sense?

zaceno:
Still not sure I understand. It is only thanks to this new flag that we have reason to add nodes anywhere else but the end. When _insert != true, we always should append (because if we're creating a new node, it means we're out of oldnode children to update, so we'll always be creating new nodes at the end of the parent). When _insert is truthy, however we *might* need to insert new nodes higher up than the end.

So what I don't get is when you say "instead of using this flag"  If we don't have the flag, we should always append new nodes (and update existing nodes)

The flag is not meant to force insertBefore over appendChild, but to force us to create a new Element instead of updating an existing one. Does that clarify?

zaceno:
... sorry for being so thick :P maybe it would help me if you explained the benefit you see in (or what problem you want to avoid by) "using insertBefore for everything"

leeoniya:
`insertBefore` works with a `null` reference element, which functions exactly as `appendChild`, and probably gzips better to use a single method.

zaceno:
Ah, cool. Well then we should totally use that instead. Will update the PR. (But my home laptop is out for repair, so it will have to wait until tomorrow)

zaceno:
@jbucaran, I think I understand what you mean now (from our discussion in the issue-thread) that the problem you want to address is the need to specify a flag in the h() call. And that is ugly.

I have commented further on that in the issue thread. Just mentioning it here too, in case anyone is having a hard time following ;)

jbucaran:
@zaceno Thank you for your persistence and work on this patch. Before we proceed, I'm going to ask you (and please take your time, there's no rush) to come up with a new simpler example that illustrates the problem. I find the iframe/img example with the transitions extremely confusing.

zaceno:
@jbucaran, You're right -- the transitions example _is_ confusing. There's a lot there that isn't relevant to the actual problem. Now that I've figured out what the problem *is*, it shouldnt be too hard to come up with something simpler. I have a few ideas to try out...

zaceno:
Here is a much simpler example demonstrating the problem (and how this PR solves it) 

http://codepen.io/zaceno/pen/bqwPVe

In the demo, you see an image gallery, and an "update" button below it. Click the button and the "updating..." message appears above the gallery (nothing is actually updating of course, it's just a demo)

But notice that as the "updating..." message appears, all the images in the gallery reload (you see them flicker)

Then go to line 11 in the pen and uncomment the `_insert:true` line. Now when you click the button, the "updating.." message appears *without* causing the images to flicker.

The reason this works is that without, `insert: true`, the patch function would *replace* the image gallery node with the message node, and then make a new image gallery node.

*With* `insert: true` the patch function is forced to insert the message in the dom, and then compare the old gallery with the new one. And since there is no diff there are no further operations to the dom.

zaceno:
By the way, I could write a test for this to get full coverage of the diff -- but the test itself would look a little redundant in the suite, because the resulting html should be the same wether you use _insert: true or not.

jbucaran:
@zaceno Something I've been meaning to ask for a while, why do those images take so long to load?

leeoniya:
this issue is something that would/could be solved by implementing keyed diffing into the vdom engine. if that more general-case solution is planned, this _insert feature would not be necessary. i would probably be wary of adding it and encouraging its use since the writing is already on the wall for its eventual deprecation.

zaceno:
They take a while to load because they are quite large. The problem is most obvious for large images (or really long lists). I just display them scaled down because code-pen's viewing area is too small for a demo with the images in their natural size.

jbucaran:
@leeoniya Is there really no other way to solve this problem without introducing a keyed implementation?

Someone who has already written a vDOM engine may disregard my complain, but while I understand the problem better now, I still find this concept difficult to grasp and explain to others. 

Like the first time I saw: _Warning: Each child in an array or iterator should have a unique "key" prop._

leeoniya:
keyed diffing is likely to be the only solution that is both intuitive for users and doesnt suffer from multiple edge cases. i think this example is too contrived and only solves a very specific case.

for example, this dom reuse bug can also be caused when removing and re-ordering. try this same example with 5 child nodes, eavh having a different number of images. you'll run into many other cases where even the bug also manifests and even the _insert terminology stops making sense.

i will stop short of saying there is no other solution, but i think keyed diffing is the most obvious and simple one that covers all mutation cases.

zaceno:
@leeoniya After giving this some thought, I believe this PR obviates the need for a full keyed vdom diffing solution. At least for all cases I can come up with.  And I agree with @jbucaran that explaining a keyed vdom *with its dangers* is quite difficult.

So I would hate to see this PR conflated with the idea of keyed dom diffing, and disregarded for the same reasons.

Even this solution feels a little wonky and inelegant somehow. But in terms of documenting / explaining it, it has the benefits that:

a) It doesn't need to be explained or understood by beginners. It can be a power-user secret. Because wether or not you use _insert:true, the end result will be the same. It would only be used in certain situations to optimize performance and prevent reloading of images and iframes. (I guess the same could be said for keyed vdoms, but...)

b)  There are no dangers to using `_insert:true` the "wrong way". There is no need to issue a warning statement like the one @jbucaran quoted above. Worst case, the patch function performs as "poorly" as it already does. `_insert:true` can only help, when used the right way.

zaceno:
Actually, I want to correct myself. Disregard statement "b" in the above. The warning @jbucaran quotes is basically warning that you will fail to get performance benefits for using keys the wrong way -- *not* that you can make matters worse than if you don't use keys at all. So in that respect, it is the same as my _insert:true solution.

So, with that said, my opposition to a keyed vdom solution is only that I think it would make the patch-function a lot more complicated.

I disagree that `_insert: true` is less intuitive to explain than keyed vdom, but that is a subjective matter.

> you'll run into many other cases where the bug also manifests and even the _insert terminology stops making sense.

I'm open to changing my opinion, but I'd like to see some examples where _insert:true doesn't make sense and a keyed vdom does.

zaceno:
Sorry for rambling, I'll clean up & consolidate my previous posts here:

I believe that:

- both `_insert:true` as well as keyed diffing are wonky inelegant solutions and I wish there was a better way.
- neither `_insert:true` nor keyed diffing needs to be understood by beginners. It can be considered a power-user secret. Used correctly, they can give you performance benefits and prevent reloading iframes/images. Used incorrectly they will not make performance any worse than if you hadn't used them at all.
- the `_insert:true` solution covers all cases a keyed vdom solution does
- the `_insert:true` solution is *no less* intuitive than keyed diffing in any case I can think of
- keyed diffing will make the patch function much more complex than `_insert:true` does. Hence I lean toward `_insert:true` (besides my obvious bias of being the PR-author)

I am open to being proven wrong of course :)

zaceno:
@leeoniya 

> for example, this dom reuse bug can also be caused when removing and re-ordering. try this same example with 5 child nodes, each having a different number of images. you'll run into many other cases where the bug also manifests and even the _insert terminology stops making sense.

The example I gave above works the exact same wether there are 5 child nodes with different numbers of images or just one. As demonstrated by this fork of the previous pen: http://codepen.io/zaceno/pen/ryWBbG

Not sure I get the problem you're talking about.

zaceno:
Well now I've thought even more about it, and realize there is one case that keyed vdom diffing might support, that `_insert:true` does not: reordering.

The only case I can come up with where this would make a visible difference were if you wanted to allow drag & drop reordering of large images in a list / grid.

zaceno:
I submitted a PR for a keyed dom-patching implementation here:

https://github.com/hyperapp/hyperapp/pull/141

...turns out it wasn't *too* crazy complicated to implement. Marginally more than my `_insert: true` solution.  It obviates the need for this solution to my problem, besides solving others this PR doesn't solve. 

Hence I am closing this PR.

PR #137, opened 2017-03-06
Retain selection position on element value update

mrkstwrt:
If setElementData updates the value of an element that has selectionEnd
and selectionStart properties it keeps the old values. This fixes a
"bug" in hyperapp which stops the user from editing text in inputs
properly.

Added tests to cover this, open to opinions on other ways of doing this.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -208,7 +208,7 @@ export default function (app) {
       element.removeEventListener(event, oldValue)
       element.addEventListener(event, value)
 
-    } else if (name === "value" && "selectionEnd" in element && "selectionStart" in element) {
+    } else if (name === "value" && "selectionEnd" in element) {
       var selectionEnd = element.selectionEnd
       var selectionStart = element.selectionStart
 
--- a/src/app.js
+++ b/src/app.js
@@ -209,11 +209,8 @@ export default function (app) {
       element.addEventListener(event, value)
 
     } else if (name === "value" && "selectionEnd" in element) {
-      var selectionEnd = element.selectionEnd
-      var selectionStart = element.selectionStart
-
       element.setAttribute(name, value)
-      element.setSelectionRange(selectionStart, selectionEnd)
+      element.setSelectionRange(element.selectionStart, element.selectionEnd)
 
     } else {
       if (value === "false" || value === false) {
--- a/src/app.js
+++ b/src/app.js
@@ -213,8 +213,7 @@ export default function (app) {
       var selectionStart = element.selectionStart
 
       element.setAttribute(name, value)
-      element.setAttribute("selectionEnd", selectionEnd)
-      element.setAttribute("selectionStart", selectionStart)
+      element.setSelectionRange(selectionStart, selectionEnd)
 
     } else {
       if (value === "false" || value === false) {
--- a/src/app.js
+++ b/src/app.js
@@ -208,6 +208,14 @@ export default function (app) {
       element.removeEventListener(event, oldValue)
       element.addEventListener(event, value)
 
+    } else if (name === "value" && "selectionEnd" in element && "selectionStart" in element) {
+      var selectionEnd = element.selectionEnd
+      var selectionStart = element.selectionStart
+
+      element.setAttribute(name, value)
+      element.setAttribute("selectionEnd", selectionEnd)
+      element.setAttribute("selectionStart", selectionStart)
+
     } else {
       if (value === "false" || value === false) {
         element.removeAttribute(name)
--- a/src/app.js
+++ b/src/app.js
@@ -208,10 +208,6 @@ export default function (app) {
       element.removeEventListener(event, oldValue)
       element.addEventListener(event, value)
 
-    } else if (name === "value" && "selectionEnd" in element) {
-      element.setAttribute(name, value)
-      element.setSelectionRange(element.selectionStart, element.selectionEnd)
-
     } else {
       if (value === "false" || value === false) {
         element.removeAttribute(name)
@@ -222,7 +218,19 @@ export default function (app) {
         if (element.namespaceURI !== "http://www.w3.org/2000/svg") {
           // SVG element's props are read only in strict mode.
 
+          var oldSelStart, oldSelEnd
+          var type = element.type
+
+          if (type && type.substr(0, 4) === "text") {
+            oldSelStart = element.selectionStart
+            oldSelEnd = element.selectionEnd
+          }
+
           element[name] = value
+
+          if (oldSelStart >= 0) {
+            element.setSelectionRange(oldSelStart, oldSelEnd)
+          }
         }
       }
     }
--- a/test/app.test.js
+++ b/test/app.test.js
@@ -211,7 +211,7 @@ describe("app", () => {
       })
     })
 
-    it("keeps selectionStart and selectionEnd properties intact on update", () => {
+    it("keeps selectionStart and selectionEnd properties on text inputs intact on update", () => {
       app({
         model: "foo",
         actions: {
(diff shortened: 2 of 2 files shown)


jbucaran:
@mrkstwrt Thanks! I know it's more work, but can you provide a demo with the bug? 🙏

mrkstwrt:
@jbucaran No problem, what kind of demo would you like me to provide? You can see the bug in question in the todo list example that you provide in the readme (https://hyperapp.gomix.me/todo). Try typing in the input box, placing your cursor somewhere else inside the input (say the middle) and type again and you'll see the cursor will jump back to the end of the input.

I've seen this behaviour on the latest Chrome and Safari, however Firefox does not seem to be effected.  If you'd like me to put something else together just let me know.

src/app.js
```diff
@@ -208,6 +208,13 @@ export default function (app) {
       element.removeEventListener(event, oldValue)
       element.addEventListener(event, value)
 
+    } else if (name === "value" && "selectionEnd" in element && "selectionStart" in element) {
```
dodekeract: Is it necessary to check for both?

src/app.js
```diff
@@ -208,6 +208,13 @@ export default function (app) {
...
+      var selectionEnd = element.selectionEnd
```
dodekeract: It's good to be verbose, but does the minifier pick this up and optimize it? Otherwise I'd prefer this to be inline.

mrkstwrt: Which are you referring to here sorry?

src/app.js
mrkstwrt: Quite right, checking both is redundant. I'll update 👍

jbucaran:
Hey @mrkstwrt! I can't believe I didn't see this bug before. 🤔 I wonder if this is something that would just go away by improving the patch function. Instead of replaceChild, insertChild or similar strategy that does not create a new element.

mrkstwrt:
@jbucaran I'm not sure, when I was stepping through the code in the debugger to work out why it was happening I could see the `selectionStart/End` staying the same until `value`  gets changed. I think when `value` is changed on an element certain browsers reset the selection properties along with it, which makes sense most of the time.

mrkstwrt:
@jbucaran Check out this little pen to see what I mean https://codepen.io/mrkstwrt/pen/KWNwQJ

Simply setting `value` on the input sets the `selectionStart` and `selectionEnd` to the end of the input (in Chrome at least)

dodekeract:
@mrkstwrt also happens in Safari and Firefox Developer Edition.

jbucaran:
@mrkstwrt Ah, yeah, I can totally repro this. What I am saying is that there may be a more general solution to this problem, given that I can't find any reference to selectionRange in other vDOM engines.

jbucaran:
> @mrkstwrt I think when value is changed on an element certain browsers reset the selection properties along with it, which makes sense most of the time.

`value` is deliberately changed in the TodoList example since we want to clear the input box after a new item is added.

mrkstwrt:
The problem occurs whenever you're setting `value` on an input to be a string from the model, which is probably quite often.

jbucaran:
@mrkstwrt What does React or any other framework do in this case?

mrkstwrt:
@jbucaran I'm not that knowledgeable on other frameworks to be honest however a quick Google seems to show that this is a problem in React, they have an open issue here https://github.com/facebook/react/issues/955 with some different approaches.

Some use the same approach I have and others seem to involve not updating while focused, however in my case at least I don't think that would work as my views change as the user types.

jbucaran:
> @mrkstwrt  I don't think that would work as my views change as the user types.

Yes, and also when doing something like in the todolist example.

jbucaran:
It seems the consensus in https://github.com/facebook/react/issues/955 is that this is "impossible". 🍭

jbucaran:
@mrkstwrt Further testing indicates this patch introduces another problem, making it impossible to clear the input text.

mrkstwrt:
@jbucaran What do you mean by clear? I'm not sure I can replicate that in my setup.

jbucaran:
@mrkstwrt I was able to fix the bug (as can be see in the TodoList example), using a bit less code.

Do you have another example of your own that you could share so that I can check all is good?

mrkstwrt:
@jbucaran That falls over for me on other elements, I have a checkbox in my view and get the following:

`Uncaught DOMException: Failed to read the 'selectionStart' property from 'HTMLInputElement': The input element's type ('checkbox') does not support selection.`

I think we need to maintain a check for `selectionStart` or `selectionEnd`

dodekeract:
@mrkstwrt `undefined >= 0` should be false. Why does that error happen?

jbucaran:
@mrkstwrt Thanks, I'll fix it.

jbucaran:
This is the problem.
* https://codereview.chromium.org/2706033004/patch/20001/30008

mrkstwrt:
So there's no way around checking for the existence of selectionStart or selectionEnd at the moment?

This works for me (and lets me clear the field)
```
} else if (name === "value" && "selectionEnd" in element && value.length) {
      element.setAttribute(name, value)
      element.setSelectionRange(element.selectionStart, element.selectionEnd)
}
```

but it's probably a bit more code than you're looking for. I'm sure there's a more elegant solution.

jbucaran:
@mrkstwrt Yes, there is!

```jsx
          var oldSelStart, oldSelEnd
          var type = element.type

          if (type && type.substr(0,4) === "text") {
            oldSelStart = element.selectionStart
            oldSelEnd = element.selectionEnd
          }

          element[name] = value

          if (oldSelStart >= 0) {
            element.setSelectionRange(oldSelStart, oldSelEnd)
          }
```

jbucaran:
Mmm, I didn't try "selectionEnd" in element, let me see.

mrkstwrt:
I think checking for selectionEnd is the better option than type being text as this effects a range of input types (number for example) and stuff like textareas as well

mrkstwrt:
@jbucaran The solution is my last comment above seems to cover those bases in my tests

jbucaran:
@maraisr `"selectionEnd" in element` fails if the element is a type="number" and you are forgetting to element[prop] = value too.

jbucaran:
> Failed to read the 'selectionStart' property from 'HTMLInputElement': The input element's type ('number') does not support selection.

mrkstwrt:
Yeah just spotted that too, my bad. Seems that number inputs aren't effected by the bug though which helps things.

jbucaran:
@mrkstwrt Cool, let's go with this then:

```
          var oldSelStart, oldSelEnd
          var type = element.type

          if (type && type.substr(0, 4) === "text") {
            oldSelStart = element.selectionStart
            oldSelEnd = element.selectionEnd
          }

          element[name] = value

          if (oldSelStart >= 0) {
            element.setSelectionRange(oldSelStart, oldSelEnd)
          }
```

jbucaran:
BTW this code covers input text and also textarea, since a textarea's type is text (that's why I'm using substr there).

mrkstwrt:
That works great for me, no more errors. I just have one question, would it not be slightly better on performance to only do this when the `name === "value"` instead of every property? If that's not going to be an issue worth thinking about I'll push it up.

jbucaran:
@maraisr You're right, let me run this through the benchmarks.

mrkstwrt:
@jbucaran I've pushed up the changes, let me know if you find that it'll be worth the extra conditional to check that it's `value` being changed

jbucaran:
@mrkstwrt Thank you Mark! I may refactor this a bit more and create a new release later today! 🎉

jbucaran:
@mrkstwrt Turns out my logic was wrong, it was more similar to what you had originally (in order to support input types like url, date, etc), but we also needed to set the property, so it had to be done within the last else clause.

Can you please check that everything is still working for you after https://github.com/hyperapp/hyperapp/commit/d926ba803d92791c8e4ddf8e387d4b54195f50ba?

mrkstwrt:
@jbucaran Yeah that's what my thinking was before, this way should cover all the edge cases and any element with a value and selection properties will be handled. Works great for me, fixes the issue in my apps, great work.

jbucaran:
@mrkstwrt Thanks for the feedback! This was a tricky one.

PR #139, opened 2017-03-07
Allow dashes in dynamic path segments

### LINKED ISSUE #138: Router does not accept dashed in dynamic path segments
For example: with the path `/:slug` and the url `/this-slug` the router will fail to match the path. The url `this_slug` will match.

judas-christ:
Fixes #138

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -45,7 +45,7 @@ function match(routes, path) {
       .replace(/\//g, "\\/")
       .replace(/:([A-Za-z0-9_]+)/g, function (_, key) {
         keys.push(key)
-        return "([A-Za-z0-9_]+)"
+        return "([-A-Za-z0-9_]+)"
       }) + "/?$", "g"), function () {
 
         for (var i = 1; i < arguments.length - 2; i++) {
--- a/test/router.test.js
+++ b/test/router.test.js
@@ -80,7 +80,7 @@ describe("Router", () => {
   })
 
   it("collects matched route keys", () => {
-    window.location.pathname = "/FOO/BAR/BAZ"
+    window.location.pathname = "/FOO/BAR/BAZ-BAZ"
 
     app({
       view: {
@@ -96,7 +96,7 @@ describe("Router", () => {
 				<ul>
 					<li>FOO</li>
 					<li>BAR</li>
-					<li>BAZ</li>
+					<li>BAZ-BAZ</li>
 				</ul>
 			</div>
 		`)
--- a/test/router.test.js
+++ b/test/router.test.js
@@ -80,7 +80,7 @@ describe("Router", () => {
   })
 
   it("collects matched route keys", () => {
-    window.location.pathname = "/FOO/BAR/BAZ-BAZ"
+    window.location.pathname = "/FOO/BAR/BAZ"
 
     app({
       view: {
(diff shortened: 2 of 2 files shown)


jbucaran:
Thanks @judas-christ!

LGTM. @dodekeract

judas-christ:
I'm actually thinking maybe that part of the regex could be made even more lenient (and smaller), something like `([^\/?]+)` instead of `([-A-Za-z0-9_]+)` but I'm not sure about all the edge cases so it would take some more thorough testing.

farism:
`([\w-]+)` would also work.

dodekeract:
Code itself is good, but there is a reason why dashes weren't allowed in the first place.

> `varchar = ALPHA / DIGIT / "_" / pct-encoded`

[RFC 6570](https://tools.ietf.org/html/rfc6570#section-2.3)

**TL;DR*: RFC and Express don't allow this.

judas-christ:
@dodekeract isn't that for the variable names, though? I.e not allowed in the template `/:field`, but allowed in matched URLs?

dodekeract:
@judas-christ No, that's `varname`. As far as I understand `varchar` is for the values.

judas-christ:
Hmm I think they're talking about variable names until the next section about expansion. I might be wrong.

jbucaran:
@judas-christ I'm confused. What do we want to achieve again? The goal of this patch is so that the router can match routes (slugs) with dashes in it right?

judas-christ:
@jbucaran Yes, exactly.

jbucaran:
Thank! That's what I thought. Now, what's this discussion about varname, RFC, etc.?

dodekeract:
@jbucaran It's from https://github.com/pillarjs/path-to-regexp/issues/63 and the reason why they don't support it. (path-to-regexp is used by express)

jbucaran:
> Basically if you need - in your variable values you should apply urlencoding to that section.

🤔 But why? What's wrong with /:name-id?

I assume that would match slugs like `/foo-id` and `/bar-id`.

dodekeract:
@jbucaran currently it would, yes. As far as I understand this PR is proposing that `/:name` matches `/some-name-with-dashes`.

jbucaran:
Gotcha. @judas-christ So, what do you want to do?

I personally wouldn't use dashes in URLs, or does this work:

```
/:foo-:bar
```

?

dodekeract:
@jbucaran It should work. At least it does in `path-to-regexp`.

The thing is that `-` should just be urlencoded if used as a variable. Otherwise it's more like a delimiter. e.g. `en-US` -> `:language-:country`.

judas-christ:
@jbucaran @dodekeract I just tested in express and it allows both routes like `/:var` and `/:var1-:var2`. The URL `/one-two` will match both. In the first case `var` will be `one-two` and in the second you get `one` as `var1` and `two` as `var2` as expected.

I don't know what's more common but personally I use dash-delimited slugs as URL segments all the time. So if it were up to me, I'd go with that, but if everyone else does it differently it makes more sense to keep it like it is and I can just make a variant of the router that works like I want it to.

dodekeract:
@judas-christ Maybe we can make the regex an option? How does express do that? Supporting both would be nice.

jbucaran:
Out of the box we should support most use cases. @judas-christ Can you fix the PR so that /:var matches also `/foo-bar`?

dodekeract:
@jbucaran If this supports both I'm 100% fine with it. Didn't know express managed to do that.

judas-christ:
@jbucaran this PR works like that already. It does not support `/:var1-:var2` though.

dodekeract:
@judas-christ does the current hyperapp version support `/:var1-:var2`?

judas-christ:
@dodekeract just tested the current router, and yes it does support `/:var1-:var2`

jbucaran:
@judas-christ So now, both styles work?

judas-christ:
@jbucaran Yes, they're all good. I added a couple of tests to verify both use cases.

jbucaran:
@judas-christ Awesome! I'll be merging this a bit later today! 🔥

jbucaran:
@judas-christ Thanks!! 👍

judas-christ:
@jbucaran No worries, I'm just happy to help. Love what you're doing with this.

dodekeract:
@jbucaran Is this published? I think I'm experiencing this issue.

judas-christ:
@dodekeract it doesn't seem to be published on npm (it's not in the 0.7.1 release)

Issue #128, opened 2017-03-05
Children appended to script element when using root: document.body


ngryman:
Hi,

When I specify `root: document.body` I get children of my component appended to the `script` tag:

![image](https://cloud.githubusercontent.com/assets/892048/23589884/2b4f44ee-01d6-11e7-90c8-7cb5e422d65c.png)

<details>
<summary>Code</summary>
<br>

**View**
```js
const { h } = require('hyperapp')
const hyperx = require('hyperx')

const html = hyperx(h)

const ScheduleEvent = (event) => html`
<li class="schedule-event">
  <span class="event">${event.title}</span>
  <span class="hours">
    <time>${event.start}</time> - 
    <time>${event.end}</time>
  </span>
</li>
`

const ScheduleDay = (day) => html`
<li class="schedule-day">
  <span class="day"></span>
  <ul class="events">
    ${day.events.map(event => ScheduleEvent(event))}
  </ul>
</li>
`

const ScheduleWidget = (schedule) => html`
<ul class="schedule-widget">
  ${schedule.days.map(day => ScheduleDay(day))}
</ul>
`

module.exports = ScheduleWidget
```

**Index**

```js
const { app } = require('hyperapp')

const model = require('./lib/model')
const view = require('./lib/view')
const reducers = require('./lib/reducers')
const subscriptions = require('./lib/subscriptions')

app({ model, view, reducers, subscriptions, root: document.body })
```

</details>

jbucaran:
@ngryman Yup, this is not in the docs, but hyperapp does not render itself in the body as that would create edge complexity. You can either use a regular element inside the body or provide none and hyperapp will add one for you.

ngryman:
@jbucaran Ok thanks! It would be worth mentioning in the docs. I can provide a PR if you want, but my english is not ... ✨

jbucaran:
@ngryman That's okay, the docs are not in the repo but in the wiki, which is open / public so you can change them anytime, no need to ask for permission, just do it! :)

Btw, even better would be to fix `patch` so that hyperapp can render a tree in document.body. I remember I tried once and failed, but maybe it's time to try again.

ngryman:
Well I added this test case which passes:
```js
```

ngryman:
@jbucaran If I add other nodes to `document.body` before calling `app`, it always insert the children in the **first** element. It must be related somehow to the fact that you expect the `root` to only have one child (`ul` here) and so append other children in the first root child that end up being `script` here.
[...]

PR #140, opened 2017-03-07
Support roots with existing children (#128)

ngryman:
Here it is :)

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -15,7 +15,7 @@ export default function (app) {
     onRender: []
   }
 
-  var node
+  var node, nodeElement
   var root
   var batch = []
   var plugins = app.plugins || []
@@ -111,7 +111,7 @@ export default function (app) {
       view = hooks.onRender[i](model, view)
     }
 
-    patch(root, node, node = view(model, actions), 0)
+    nodeElement = patch(root, nodeElement, node, node = view(model, actions))
 
     for (var i = 0; i < batch.length; i++) {
       batch[i]()
@@ -224,11 +224,9 @@ export default function (app) {
     }
   }
 
-  function patch(parent, oldNode, node, index) {
-    var element = parent.childNodes[index]
-
+  function patch(parent, element, oldNode, node) {
     if (oldNode === undefined) {
-      parent.appendChild(createElementFrom(node))
+      element = parent.appendChild(createElementFrom(node))
 
     } else if (node === undefined) {
       batch.push(parent.removeChild.bind(parent, element))
@@ -245,7 +243,9 @@ export default function (app) {
       if (typeof node === "string") {
         element.textContent = node
       } else {
-        parent.replaceChild(createElementFrom(node), element)
+        var newElement = createElementFrom(node)
+        parent.replaceChild(newElement, element)
+        element = newElement
       }
     } else if (node.tag) {
       updateElementData(element, node.data, oldNode.data)
@@ -254,8 +254,10 @@ export default function (app) {
       var oldLen = oldNode.children.length
 
       for (var i = 0; i < len || i < oldLen; i++) {
-        patch(element, oldNode.children[i], node.children[i], i)
+        patch(element, element.childNodes[i], oldNode.children[i], node.children[i])
       }
     }
+
+    return element
   }
 }
--- a/test/app.test.js
+++ b/test/app.test.js
@@ -779,6 +779,62 @@ describe("app", () => {
 				</main>
 			`)
     })
+
+    it("appends view to given non-empty root", () => {
+      var main = document.createElement("main")
+      main.appendChild(document.createElement('div'))
+
+      app({
+        root: document.body.appendChild(main),
+        view: _ => h("div", {}, "foo")
+      })
+
+      expectHTMLToBe(`
+        <main>
+          <div></div>
+          <div>
+            foo
+          </div>
+        </main>
+      `)
+    })
+
+    it("updates view in a mutated root", () => {
+      var main = document.createElement("main")
+
+      app({
+        root: document.body.appendChild(main),
+        model: 'foo',
+        actions: {
+          bar: model => 'bar'
+        },
+        subscriptions: [
+          (_, actions) => {
+            expectHTMLToBe(`
+              <main>
+                <div>
+                  foo
+                </div>
+              </main>
+            `)
+
+            main.insertBefore(document.createElement('header'), main.firstElementChild)
+            main.appendChild(document.createElement('footer'))
+            actions.bar()
+            expectHTMLToBe(`
+              <main>
+                <header></header>
+                <div>
+                  bar
+                </div>
+                <footer></footer>
+              </main>
+            `)
+          }
+        ],
+        view: model => h("div", {}, model)
+      })
+    })
   })
 
   describe("lifecycle methods", () => {


dodekeract:
Does this also work when the third-party element vanishes or another one gets added at runtime?

ngryman:
@dodekeract I don't think so, my bad...
We should track the node element instead of the node index and compute the index dynamically. I've tried this approach but couldn't make it work with the router which basically changes the node element. I have to go now, but I'll figure out this in the afternoon.

ngryman:
Also it would be nice to expose somehow internal `app` functions in order to unit test them.

jbucaran:
@ngryman Awesome! I still need to run this thru the benchmarks however. Patch is heavily used and we don't want worse performance just to have the ability to insert an app in document.body right? :)

dodekeract:
@jbucaran I don't think this issue is related to `document.body`.

leeoniya:
accessing childNodes and using indexOf in a hot function is gonna be a perf killer

jbucaran:
@dodekeract No? Do you have an failing example to back that up?

dodekeract:
@jbucaran https://github.com/hyperapp/hyperapp/issues/128#issuecomment-284685968

jbucaran:
@dodekeract I see what you mean. The issue is embedding the app in an element that is not empty.

Why would anyone do that anyway?

dodekeract:
@jbucaran Only relevant for `document.body` since browser plugins may add elements there.

ngryman:
@leeoniya @jbucaran A cleaner solution would be perhaps to invoke a patch *outer* function that executes some specific code for the root and then call recursively a patch *inner* function.

jbucaran:
@dodekeract What do other libraries do btw?

@ngryman Do we have to run this code on every patch or just once?

dodekeract:
@jbucaran React just decided to not support `<body>` because of this. I don't know what happens if the root has other elements, but I'd assume that react clears them.

ngryman:
@dodekeract If I remember well React warns the user about using `document.body` but supports it.
@jbucaran Basically, the patched code would be executed only the first time the render function is invoked, and each time a route is matched when using the `router` plugin.

jbucaran:
@ngryman Then why is the code inside patch? Maybe outside render, and change render to take an index argument too?

The router...

ngryman:
@jbucaran Because I need the actual DOM element which is created inside `patch`. But there is probably a cleaner way to do so, I'm trying to understand the code at the same time :)

dodekeract:
@ngryman Is that new? I remember that this was one of the few things I disliked when starting with react. Could have changed "recently".

jbucaran:
JSX as well as Hyperx don't allow you to do stuff like this:

```
<h1>Hi</h1>
<p>Hello.</p>
```

Why should the vDOM engine allow you to embed an app in a container with children already in it?

ngryman:
First of all some terminology, let's define the `host` as the existing dom element that receives the `root` node of the application.

Here is an *unelegant* solution, but it highlights what could be done: treat `host` operations separately from the `root` and all of its children.

ngryman:
@jbucaran That's a different problem. Any `jsx` markup needs **one** `root` element to be inserted into any `host` element. I think we lack some terminology where we differentiate those 2 elements.

ngryman:
@jbucaran If you agree with this I could factorize this and add associated tests.

jbucaran:
@ngryman Go for it and let's see! 🙏

ngryman:
@jbucaran Here is the idea.

After refactorization, we end up with just 1 more `if` for `create`, `replace` and `update` operations which should be perfectly fine. You can still bench it to be 100% sure. But I'm pretty sure there will be no difference. If there is, encapsulating each operation in its own `function` would let `v8` and brothers optimize each one.

I just need to add one more test to check if it works well when mutating the root node outside `hyperapp`. If you have an example on how to test this it would be awesome. I'll finish this tomorrow :)

farism:
I'm on the fence about this feature. First gut instinct is that it is unnecessary as there are current easy workarounds (e.g. creating a placeholder element inside of the targeted root element and targeting that instead).

And I think for the most part, people coming from other view libraries are expecting mounting a DOM tree to completely replace the contents.

jbucaran:
> And I think for the most part, people coming from other view libraries are expecting mounting a DOM tree to completely replace the contents.

Can you mention at least one library that works like this?

leeoniya:
domvm *can* replace and/or clear a target placeholder via `.mount(placeHolder, true)`. but regular `.mount(container)` will append the view's root node to the container. [1]

[1] https://github.com/leeoniya/domvm#views

farism:
@jbucaran Yes: React. Which is arguably the most familiar to masses.

Here's my opinion: hyperapp is not a full fledged vdom implementation. After trying other vdom implementations it was decided that it would be best to keep the footprint small and implement a custom vdom. Should this vdom have all of the capabilities of another implementation? If so, we should probably just be using another library, which has smart people maintaining it and interested in solving those domain problems.

In my eyes this project is more about the elm architecture than it is about giving people options. If anything, being restrictive on these types of edges cases is probably better than being open, as it coerces people into doing one thing consistently.

If this feature request is to easier enable people to mount components into a DOM tree without blowing away the existing DOM, then my suggestion is to just make a placeholder element at end (or beginning, or wherever you want, this technique is actually more flexible) and render into that. It seems to be a perfectly acceptable solution.

ngryman:
@farism **tldr** I understand your opinion and would totally agree with you if the cost of this feature was actually high. But it's not.

`hyperapp` should be and stay *KISS*. But it doesn't mean that every edge case should be thrown away and not implemented. IMHO judging a feature implementation based on what others do and *hypothetical complexity* is not sufficient enough.

In fact many factors are involved:

- [x] **Legitimity**: Is is something useful?
- [ ] **Usage**: How many users would it impact?
- [x] **Stability**: Does it resolve a bug?
- [x] **Weight**: Would it add weight to the library?
- [x] **Performance**: Will it have a negative impact on the performance?
- [x] **Maintenance**: Will it increase code complexity?

I've take the liberty to check what I *think* this feature is OK with. The only remaining item would potentially be the impact of this feature, which I agree with you is pretty low if we consider the current usage.
But I'm pretty sure I'm not the only one annoyed by this additional placeholder.
IMHO I would even say that asking the user to *workaround* because we don't want to handle this is not a solution, it's laziness. If for a given use case everybody has to work around, well there is a conception problem somewhere...

`hyperapp` exists to be an opinionated alternative to `react`. So why wouldn't it also be opinionated on this too?

Still, you have 5/6 items checked meaning that you have a pretty good score on this feature. And it's actually 4 loc... I see a huge benefit for a tiny cost here.

---

> And I think for the most part, people coming from other view libraries are expecting mounting a DOM tree to completely replace the contents.

Well I don't expect this at all tbh. In fact, what is the benefit for the user to do so? I don't see it, I only see a benefit for the library author that don't want to mess with this.
Also, it's not because `react` *sempai* does it *this way* that it's **the way**. If people come from other libraries there is a reason: they are unsatisfied with some key points and expect different (opinionated) things.

Sorry if it was veeeery long, but it's an interesting discussion because it raises the question: is `hyperapp ~= react + elm` or `hyperapp = vdom + elm` ?

jbucaran:
@ngryman 4 loc?

jbucaran:
> react sempai

🤣

ngryman:
@jbucaran If we discard the refactorization (which is basically wrapping DOM operations in functions), the patch itself is 4 lines of code 🎉 : [L274](https://github.com/ngryman/hyperapp/blob/f4d9d4cb57f4f2252b7862f0496985607afc0d0b/src/app.js#L274) [L278](https://github.com/ngryman/hyperapp/blob/f4d9d4cb57f4f2252b7862f0496985607afc0d0b/src/app.js#L278) [L281](https://github.com/ngryman/hyperapp/blob/f4d9d4cb57f4f2252b7862f0496985607afc0d0b/src/app.js#L281) [L296](https://github.com/ngryman/hyperapp/blob/f4d9d4cb57f4f2252b7862f0496985607afc0d0b/src/app.js#L296)

jbucaran:
@ngryman Cool. So, is this PR ready for review?

ngryman:
@jbucaran Yes, I still have to reach the 100% coverage.

rbiggs:
Hmmm... To me it seem natural to want to have structure in the body tag before rendering your app. That's how PWAs work - they have a server-rendered app shell into which you inject your dynamic content. I do this all the time. Was the first big WTF when I started playing with JSX and Hyperx that I couldn't insert siblings into my app. I wound up creating my own tagged template function to allow that.
I really hate being forced to Russian-doll my components, which winds up breaking the code for layout that was expecting a more semantic structure. For example, I'd expect to be able to build a semantic layout like this:

```html
<body>
  <nav>
    <h1>The Title</h1>
    <menu></menu>
  </nav>
  <article id='foo1'>
    <section>
      <!-- Content -->
    </section>
  </article>
  <article id='foo2'>
    <section>
      <!-- Content -->
    </section>
  </article>
  <article id='foo3'>
    <section>
      <!-- Content -->
    </section>
  </article>
  <footer>
    <!-- Links, etc. -->
    <menu></menu>
  </footer>
</body>
```
The Russian doll "feature" brakes this really quickly with endlessly nest components and child components :-|

jbucaran:
@ngryman Are you familiar with [createDocumentFragment](https://developer.mozilla.org/en-US/docs/Web/API/Document/createDocumentFragment)? I wonder if we could use that instead to get this working.

dodekeract:
@jbucaran sounds promising. I think it's worth a shot. Might even make things easier or faster since we can trust the fragment.

jbucaran:
@dodekeract I don't have a lot of experience (basically none) with fragments, so I'm experimenting with this for the first time. It seems browser support is stellar, but I wonder if maybe be any other implications I'm missing.

rbiggs:
Using `createDocumentFragment` lets you create sibling elements and then append them to a container without the need for a wrap, unlike JSX and Hyperx. 

```js
const createElements = (template) => {
  const temp = document.createElement('div')
  temp.innerHTML = template
  const frag = document.createDocumentFragment()
  // Use childNodes to allow creating element nodes or text nodes:
  const children = Array.prototype.slice.apply(temp.childNodes)
  children.map(el => frag.appendChild(el))
  return frag
}

var myTemplate = createElements(`
<nav>
  <h1>The Title</h1>
</nav>
<article id='post1'>
  <p>The stuff here</p>
</article>
`)
document.body.appendChild(myFragment)
```
When you append the fragment, it goes bye-bye and the siblings get inserted without problem into the target element.

jbucaran:
Thanks @rbiggs. Do you know we could use it to solve this issue?

dodekeract:
@rbiggs Wait a minute

> it goes bye-bye

So we can't access it anymore later? That sucks. I hoped that it'd stay and we can just manipulate it.

rbiggs:
Document fragments exist to create temporary containers for element creation, as in the example. Docs from MDN:

> DocumentFragments are DOM Nodes. They are never part of the main DOM tree. The usual use case is to create the document fragment, append elements to the document fragment and then append the document fragment to the DOM tree. In the DOM tree, the document fragment is replaced by all its children.

> Since the document fragment is in memory and not part of the main DOM tree, appending children to it does not cause page reflow (computation of element's position and geometry). Consequently, using document fragments often results in better performance.

rbiggs:
Document fragments have all DOM methods you would expect: appendChild, etc. But they do not support `innerHTML`. Got use methods to append, etc.

leeoniya:
fragments get cleared after insertion so cannot be used as transparent holders in the live dom. they used to provide a speed boost, but this is no longer the case.

any vdom fragment implementation has to be manually implemented in code, including fragment boundaries, nested fragments, etc. i'll tell you that the impl was not a fun one for me. discovering how useless createDocumentFragment was was a bummer.

leeoniya:
i didnt end up using this approach [1] due to some limitations it has, but it has some merit.

[1] https://github.com/trueadm/fragment-node/issues/1

dodekeract:
@leeoniya That's a shame, they sounded perfect. 😢

jbucaran:
Everyone please note I only want to see if fragments can be used to support inserting trees in document.body (easily) and since this PR has a large diff.

rbiggs:
Document fragments are only useful when you're dealing with sibling element insertion, like list items into a list, etc. If you inserting a tree, then no need for document fragment.

jbucaran:
Got it "working", but I wasn't happy w/ the implementation, plus some tests failed. We can try this again after keyed nodes are implemented, since this tweaks `patch` a lot and so does keyed nodes.

ngryman:
@jbucaran I know fragments but I don't understand how they resolve the issue? You are talking about first patch?

jbucaran:
@ngryman Yes, exactly.

ngryman:
@jbucaran It think it should probably sit in a different issue then.

I use fragments when I manipulate the `dom` myself and I need some control over live `dom` operations (basically batch insertion). As far as I know they provide a speed boost at insertion into the live `dom`, compared to a traditional detached node [[1]].
But as @leeoniya points out, it may not be the case anymore. Browser internals evolve crazy fast and might have optimized the naive way of doing so...

[1]: http://ejohn.org/blog/dom-documentfragments/

ngryman:
> this PR has a large diff / is more complicated than what I was expecting.

I factorized some basic operations into their own function hence the diff. The *real* patch is 4 LOC.
@jbucaran I can revert the additional refactorization for the review to be simpler if you want.

lukejacksonn:
I haven't dug too deeply into the code around this issue so I am not entirely sure this is relevant _but_ all the talk of document fragments got me thinking about [`range.createContextualFragment`](https://developer.mozilla.org/en-US/docs/Web/API/Range/createContextualFragment) which I have always found more useful than [`createDocumentFragment `](https://developer.mozilla.org/en-US/docs/Web/API/Document/createDocumentFragment). I'm sure there are some downsides (seeing as it is rarely ever employed) but some things I like about it are..

- It works great with template literals
- No need for wrapper elements
- Browser support back to IE11 (not 10 I know)


```es6
const frag = template => {
  const range = document.createRange();
  range.selectNode(document.body);
  return range.createContextualFragment(template.trim());
};

const model = { name: 'Joe Bloggs' }
const view = (model) =>`
  <img src='https://unsplash.it/200'
  <h1>Welcome ${model.name}</h1>
  <button>Click Me</button>
`

const node = frag(view(model))
document.body.appendChild(node)
```

jbucaran:
@ngryman I found a simpler way (maybe).

One way to do it:

1. Have patch return the appended element (when oldNode === undefined) this guarantees we get the application root node inside render.
2. Store it in another global variable; ignore all future returned nodes 🎈.
3. In render, before we call patch, calculate the correct index based in the stored root element.

To calculate the index I am using:

```jsx
  function getElementIndex(elm) {
    for (var i = 0; elm = elm.previousSibling; i++);
    return i
  }
```

Notice I'm using [previousSibling](https://developer.mozilla.org/en-US/docs/Web/API/Node/previousSibling) and not [previousElementSibling](https://developer.mozilla.org/en-US/docs/Web/API/NonDocumentTypeChildNode/previousElementSibling), since we want to account for text nodes too. 

Now, I'm trying to see how to cache the calculated index, so that we don't have to run getElementIndex on every render.

ngryman:
@lukejacksonn I didn't know `createContextualFragment`, that's some interesting stuff!

@jbucaran Yes right, that would remove a bunch of `if`. So basically you would store the new `nodeElement` outside `patch`.
Another idea to explore (and bench) would be to avoid the initial index lookup you proposed by accepting `index` to also be an element. So instead of searching in `parent.childNodes` you could simply patch the element itself. The only cost for this would be a `var element = 'object' === typeof index ? index : parent.childNodes[index]`

jbucaran:
@ngryman I think I ended up with something similar to that. So, I was able to get rid of the index. Now, I'm just passing the parent and the first child of the parent to patch. I cache the first child inside `load` (not render) so it only happens once when the app starts (which makes more sense).

The only changes are that both render and patch return an element (the one we cache inside `load`) and that we calculate the index of this element relative to app.root once in load like this:

```jsx
for (var i = 0; element = element.previousSibling; i++);
```

Also, the default app.root is now `document.body`. 😄

ngryman:
@jbucaran Sounds promising 👍

From what you describe, I just want ot make sure you cover:
1. `document.body` mutations outside `hyperapp`. Those mutations would invalidate the index that you only computed once in `load`. Handling those mutations would involve an index lookup each time `render` is called, to ensure it's up-to-date.
2. The router use case that basically replace your cached element with a new one. From what I understand it would be updated in `render` form the returned element of `patch`.

I think we would need your code to discuss this 🤓

jbucaran:
@ngryman There's one use case that won't work for sure.

```jsx
app({
  model: 0,
  actions: {
    add: model => model + 1
  },
  view: (model, actions) => {
    var el
    if (model === 0) {
      el = <h1 onClick={actions.add}>Small</h1>
    } else if (model === 1) {
      el = <h2 onClick={actions.add}>Smaller</h2>
    } else {
      el = <h3>Tiny</h3>
    }
    // return el // BAD
    return <code>{el}</code> // OK
  }
})

**Comment by dodekeract** [comment]
@jbucaran I think if there is no *relatively* easy solution that works in **all** cases we shouldn't consider it. Having apps that sometimes break for users with browser addons vs having to use a `<div>` is clearly in favor of the `<div>` IMO.

**Comment by jbucaran** [comment]
@dodekeract Ah, I see what you are saying. Bummer. In that case we need a different solution because my code doesn't work with elements added dynamically, unless we make render slow and we don't want that right.

**Comment by jbucaran** [comment]
@dodekeract @ngryman Wait, wait my bad. We are caching an element, not an index. We calculate the index only once, inside load, and then get `app.root`'s first child.

Even app.root changes (has nodes appended or inserted at the top) we always start from the cached first child. So, it's okay.

**Comment by jbucaran** [comment]
@ngryman Like I said above, it works, but I found it fails in a strange way when the root node is an SVG element and the body is modified.

Would you like to continue working on this feature?

**Comment by ngryman** [comment]
@jbucaran Sure! I won't be able before this WE thought. Can you provide your code on some branch so I can start from here?

**Comment by jbucaran** [comment]
@ngryman Sure thing! 

Would you mind joining the Slack room? 🙏

**Comment by zaceno** [comment]
I've been trying to catch up with what's going on here... Am I correct in understanding the following:

-  that the original ("4 LOC") implementation of this feature is no longer relevant?
- we are exploring a different solution to the problem (based on caching the app's top element)?
- ... there is, as yet no code to play with (in a PR or codepen or whatever) for this new solution-in-progress?

**Comment by jbucaran** [comment]
@zaceno The code to play with is on this gist.

* https://gist.github.com/jbucaran/fc3651391b45e417f4dc04e771a88b00

**Comment by ngryman** [comment]
@zaceno Yes this implementation works, or at least worked (`hyperapp` is evolving fast 🤓 ). And it's effectively 4 LOCs. We basically cache the topmost node element in order to patch it directly and thus avoiding conflicts with other external elements.

@jbucaran wanted to explore another way to do so in the provided gist.

I'm going to patch this PR given the idea of @jbucaran to return the element from patch.

**Comment by ngryman** [comment]
And here it is 😄

**Comment by jbucaran** [comment]
@ngryman Is it necessary to return the element at the end of `patch`? I think we only care about it when items are appended to the DOM, no?

BTW, how did you fix the issue I posted as a gif on Slack? 🤔

**Comment by ngryman** [comment]
@jbucaran Well it's only necessary when we `append` and `replace` the element. My guts tell me it's better to have one return point for the whole function in order to kiss and have a clear contract established.

I don't see your codepen link in Slack history, you told me about a svg clock if I remember. That said, I added 2 additional tests that covers all additional use cases I can see. But I would gladly validate if it works with your svg clock if you want.

**Comment by jbucaran** [comment]
@ngryman Cool, that would be this one http://codepen.io/jbucaran/pen/PWMBLp

But really, the problem has to do with what I wrote on Slack. 

This may happen after keyed dom, so maybe we can close this PR and create a new one. I don't know if it's my browser or GitHub, but it takes a while to load this thread.

**Comment by ngryman** [comment]
Here is the working version with my PR: http://codepen.io/ngryman/pen/mWxoRo.

**Comment by ngryman** [comment]
> This may happen after keyed dom, so maybe we can close this PR and create a new one.

**Comment by jbucaran** [comment]
@ngryman The code works, but it breaks when you try to edit the HTML on the dev console, which is how I am testing an imaginary browser extension that can modify document.body.

The svg was where I first found the problem, but even the simplest example can repro the issue.

**Comment by ngryman** [comment]
@jbucaran Can you reproduce it with programmatic calls? I'm not sure how devtools handle live edits exactly. IMHO this is a super edge case. If we consider the current behavior and the good things that this PR brings, we should merge it, and then address your specific issue. I have the feeling we want to resolve too many things at the same time. What do you think?

**Comment by jbucaran** [comment]
Thank you @ngryman! 🤘 🤘

**Comment by rbiggs** [comment]
Thanx a ton @ngryman. I was playing with your pull request with the svg clock. I noticed I could insert the clock into this structure and update the `p` without affecting the clock :-)

```html
<h1 id='testTitle'>
    <p>Hello</p>
</h1>
```
However, inserting the clock into a tag with only text and trying to update the textContent replaced the clock. Which is what I would expect:
```html
<h1 id='testTitle'>Hello</h1>
```
👍

zaceno:
Good work @ngryman (and @jbucaran) -- great to see this finally merged! (...now I gotta go fix the conflict in my PR though... ;) )

Issue #132, opened 2017-03-06
does hyperapp's virtual dom support keyed children?


yelouafi:
Looked at the source (https://github.com/hyperapp/hyperapp/blob/master/src/app.js#L280-L284) and it seems like element's children are only patched by position.

liadbiz:
@yelouafi  yeah, not supported yet...

jbucaran:
@yelouafi Correct. I am not sure I can keep the 1kb promise if we implement keyed children too.

On the other hand, hyperapp 0.7.0 already performs better than React, this is a local run of js-framework-benchmarks up to date for 0.7.0.

<img width="1249" alt="screen shot 2017-03-06 at 21 32 21" src="https://cloud.githubusercontent.com/assets/56996/23610070/79744000-02b4-11e7-8bea-b21ee69c5676.png">

zaceno:
This PR is not to support keyed children, but goes a bit of the way to solve the same problem:

https://github.com/hyperapp/hyperapp/pull/135

Specifically, it frees you a little from the positional-only updating of dom nodes, by letting you specify some elements that should be inserted, rather than replacing an existing dom node (in a potentially long chain of unnecessary dom operations)

zaceno:
@jbucaran: gotta ask about the benchmark table: how did you make that (a tool I could download, or did you put it together yourself?) I thought it would be interesting to check the performance effects of using the _insert flag of PR 135.

leeoniya:
@zaceno https://github.com/krausest/js-framework-benchmark

zaceno:
Thanks @leeoniya!

yelouafi:
@jbucaran 
It's more about preserving identity. This can be an issue when you have state in the real DOM that's not in sync with virtual DOM (eg user input, focus, scroll position, animation ...). When you replace that node with a new one all that state goes away (which can lead to undesirable UI effects). Keys prevent this by letting the lib track when old stateful nodes has moved elsewhere.

jbucaran:
Closing in favor of https://github.com/hyperapp/hyperapp/pull/141 🔥

PR #141, opened 2017-03-07
Keyed VDom

zaceno:
Allow instructing the patch-function what dom nodes to reuse in update, via 

```
h(tag, {_key: '...', ...}, [])
```

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -1,6 +1,7 @@
 var SVG_NS = "http://www.w3.org/2000/svg"
 
 export default function (app) {
+
   var view = app.view || function () {
     return ""
   }
@@ -107,6 +108,7 @@ export default function (app) {
     }
   }
 
+
   function render(model, view) {
     for (var i = 0; i < hooks.onRender.length; i++) {
       view = hooks.onRender[i](model, view)
@@ -138,12 +140,6 @@ export default function (app) {
     return obj
   }
 
-  function defer(fn, data) {
-    setTimeout(function () {
-      fn(data)
-    }, 0)
-  }
-
   function createElementFrom(node, isSVG) {
     if (typeof node === "string") {
       var element = document.createTextNode(node)
@@ -154,9 +150,7 @@ export default function (app) {
         : document.createElement(node.tag)
 
       for (var name in node.data) {
-        if (name === "onCreate") {
-          defer(node.data[name], element)
-        } else {
+        if (name !== "onCreate") {
           setElementData(element, name, node.data[name])
         }
       }
@@ -164,6 +158,10 @@ export default function (app) {
       for (var i = 0; i < node.children.length; i++) {
         element.appendChild(createElementFrom(node.children[i], isSVG))
       }
+
+      // call onCreate handler *after* child nodes are added,
+      // so they'll be available to the handler for manipulation
+      if (node.data.onCreate) node.data.onCreate(element)
     }
 
     return element
@@ -171,8 +169,8 @@ export default function (app) {
 
   function setElementData(element, name, value, oldValue) {
     name = name.toLowerCase()
-
-    if (!value) {
+    if (name === "key") {}
+    else if (!value) {
       element[name] = value
       element.removeAttribute(name)
 
@@ -211,7 +209,7 @@ export default function (app) {
       var realValue = element[name]
 
       if (name === "onUpdate") {
-        defer(value, element)
+        value(element)
 
       } else if (value !== oldValue || realValue !== value) {
         setElementData(element, name, value, oldValue)
@@ -219,32 +217,102 @@ export default function (app) {
     }
   }
 
-  function patch(parent, element, oldNode, node) {
+  function batchRemove (parent, element, node) {
+    batch.push(function () {
+      if (node.data && node.data.onRemove) node.data.onRemove(element)
+      parent.removeChild(element)
+    })
+  }
+
+  function patch (parent, element, oldNode, newNode) {
     if (oldNode == null) {
-      element = parent.appendChild(createElementFrom(node))
+      var n = createElementFrom(newNode)
+      parent.insertBefore(n, element)
+      element = n
 
-    } else if (node == null) {
-      batch.push(function () {
-        if (oldNode.data && oldNode.data.onRemove) {
-          oldNode.data.onRemove(element)
-        }
-        parent.removeChild(element)
-      })
-    } else if (node.tag && node.tag === oldNode.tag) {
-      updateElementData(element, node.data, oldNode.data)
+    } else if (
+      newNode.tag !== oldNode.tag ||
+      typeof newNode !== typeof oldNode ||
+      typeof newNode === "string" && newNode !== oldNode
+    ) {
+      var n = createElementFrom(newNode)
+      parent.replaceChild(n, element)
+      element = n
 
-      var len = node.children.length
-      var oldLen = oldNode.children.length
-
-      for (var i = 0; i < len || i < oldLen; i++) {
-        patch(element, element.childNodes[i], oldNode.children[i], node.children[i])
-      }
-
-    } else if (node !== oldNode) {
-      var i = element
-      parent.replaceChild(element = createElementFrom(node), i)
+    } else if (oldNode.tag){
+      updateElementData(element, newNode.data, oldNode.data)
+      patchChildren(element, oldNode, newNode)
     }
 
     return element
+  }
+
+  function patchChildren(element, oldNode, newNode) {
+    // We need to index all keyed nodes in oldChildren in order to
+    // distinguish new keyed nodes from nodes to reuse in newChildren
+    var oldLen = oldNode.children.length
+    var newLen = newNode.children.length
+    var oldChildren = []
+    var oldKeys = {}
+
+    for (var i = 0; i < oldLen; i++) {
+      if (oldNode.children[i].data && oldNode.children[i].data.key) {
+        oldKeys[oldNode.children[i].data.key] = true;
+      }
+      oldChildren[i] = [element.childNodes[i], oldNode.children[i]]
+    }
+
+    var newIndex = 0
+    var oldIndex = 0
+    var cache = {}
+
+    while (newIndex < newLen || oldIndex < oldLen) {
+      var newChild = newNode.children[newIndex]
+      var oldChildElem = oldChildren[oldIndex][0]
+      var oldChild = oldChildren[oldIndex][1]
+      var newKey = newChild && (newChild.data != null) && oldKeys[newChild.data.key] ? newChild.data.key : null
+      var oldKey = (oldChild && (oldChild.data != null) && oldChild.data.key) || null
+
+
+      if (
+        newChild != null &&
+        newKey != null &&
+        cache[newKey]
+      ) {
+        var i = cache[newKey]
+        element.insertBefore(i[0], oldChildElem)
+        patch(element, i[0], i[1], newChild)
+        delete cache[newKey]
+        newIndex++
+
+      } else if (
+        oldChild == null ||
+        (oldKey != null && newKey == null)
+      ) {
+        patch(element, oldChildElem, null, newChild)
+        newIndex++
+
+      } else if (
+        newChild == null ||
+        (oldKey == null && newKey != null)
+      ) {
+        batchRemove(element, oldChildElem, oldChild)
+        oldIndex++
+
+      } else if (oldKey == newKey) {
+        patch(element, oldChildElem, oldChild, newChild)
+        oldIndex++
+        newIndex++
+
+      } else {
+        cache[oldKey] = oldChildren[oldIndex]
+        oldIndex++
+
+      }
+    }
+
+    for (var key in cache) {
+      batchRemove(element, cache[key][0], cache[key][1])
+    }
   }
 }


zaceno:
Solves https://github.com/hyperapp/hyperapp/issues/117 (among many other important use-cases)

Here is the code pen from that issue, demonstrating how it solves the problem. 
(wierdly though, something makes it break in safari, although it works in chrome. Need to look more at that before mergeing)

zaceno:
Here is a nother, simpler problem this PR solves.

http://codepen.io/zaceno/pen/ryWddK

To demonstrate the problem, click the update button and notice how the images reload.
Now uncomment line 17 (_key: 'somekey') and try again. Notice that because the photo-gallery is rendered with the same _key property, the patch function knows to reuse it's dom-node, rather than replace it with the "updating..." message (and then create a new image-gallery node from scratch)

leeoniya:
keyed nodes solve many problems, broadly summarized here:

https://github.com/hyperapp/hyperapp/issues/132#issuecomment-284704949

farism:
It is definitely something that will be needed in some form or another for long term performance reasons.

I haven't looked at the implementation, but the first thing that jumps out at me is why the leading `_` in `_key`?

leeoniya:
keys rarely help performance if implemented inefficiently can reduce performance. the point of keys is to prevent unintended node reuse and busting any intermittent dom state that's not explicitly managed by the vdom lib.

the `_` prefix is to discern vars used by the vdom lib itself from real dom attributes applied to the elements.

farism:
Great, thank you for the explanation @leeoniya

jbucaran:
@zaceno In your example what is `model.current`? Is it a number, a string? There is no model schema to infer this so one needs to understand all the code to figure it out.

jbucaran:
@leeoniya Could you review this PR? 🙏

jbucaran:
I'm going to need everyone's help to review this PR. @farism @dodekeract @danigb @selfup @jamen @liadbiz 🙏 🙏

leeoniya:
i can sanity check it tomorrow. keep in mind though, an optimal keyed algo is thoroughly described here:

https://github.com/localvoid/kivi/blob/master/lib/vnode.ts#L1284

zaceno:
@jbucaran: in my example model.current and model.next are strings with URLs

zaceno:
Amen to some review on this! At this stage it's mostly a proof of concept that it's doable. But certainly needs some community-scrutiny

src/app.js
```diff
@@ -275,12 +276,50 @@ export default function (app) {
...
+      var len = node.children.length, oldLen = oldChildren.length
+      for (var i = 0, j = 0; i < len || j < oldLen; i++) {
         var child = node.children[i]
+        //in case the next new node is keyed (and a previous node is known
+        //we want the element inserted at this index, and then patched
+        //normally, with the corresponding oldNode (which may be moved)
+
+        if (child && child.data && child.data._key && keyedChildren[child.data._key]) {
+          var keyed = keyedChildren[child.data._key]
+          element.appendChild(keyed.element)
```
zaceno: Probably more effincient in most cases, if this line were: `element.insertBefore(keyed.element, element.childNodes[i])`

zaceno:
There is an issue with https://github.com/hyperapp/hyperapp/issues/141 and css transitions. It seems that the  way we remove and reinsert keyed elements in the dom somehow breaks css-transitions in Chrome and Safari. 

In order to make my sliding images demo work, I have to add classes to the element *after* the patch function is run, i e in oncreate and onupdate lifecycle methods.

http://codepen.io/zaceno/pen/dvOmNr

... not sure if there is something we should do about that in the patch-function, or just call it a browser-bug and accept the workaround.

leeoniya:
quick update, because i've been somewhat busy. i want to be clear about what i plan check for this PR, and if anyone gets to this first, go ahead and do it.

for [obvious reasons](https://github.com/leeoniya/domvm), i don't have much interest in dedicating much time to hyperapp. i i spot anything obvious that i've learned to avoid, i will certainly point it out, but i will not be digging into the source here to validate that all things are coded optimally and without errors - that job is painstakingly huge, and i've already gone through it all in domvm.

that being said, here's an easy way to figure out whether this PR does what it's supposed to from external testing alone:

1. create a list of 5 sibling nodes via hyperapp, each should have a `textContent` of a different color like "red". each should also be `_key`ed by the same value.
2. use vanilla dom to add a `.style.color = "red"` to match the created nodes' textContent.
3. return various templates in hyperapp that represent the following mutations:
  a. remove first node
  b. remove last node
  c. remove a node from middle
  d. swap 2 adjacent middle nodes
  e. swap first and last nodes
  f. insert a node in front
  g. insert a node at end
  h. insert a node in middle
  i. insert 2 disjoint nodes in middle.
  j. remove 2 disjoint nodes from middle.
4. after each "redraw", use iterate the DOM and ensure that the textContent matches `.style.color`.
5. validate which DOM operations were performed to achieve the needed dom state. you can use mutationObservers for this or you can use a lib i made for testing this in domvm [1]. Youc an see how it's used in domvm's regression tests [2].

I don't have a timeline for when i can get around to this, life is quite hectic these days, sorry.

[1] https://github.com/leeoniya/domvm/blob/2.x-dev/test/dominstr.js
[2] https://github.com/leeoniya/domvm/blob/2.x-dev/test/index.html

zaceno:
@leeoniya thanks for that excellent suite of tests. I couldn't help myself but started checking my implementation against the suite you recommended (even though I know self-verification is not real verification) -- and *it did* uncover bugs in the implementation that I will start looking at now.

So there's no rush for anyone to verify this just yet (unless you're curious what the bugs were, and interested in fixing them, of course ;) )

leeoniya:
> (even though I know self-verification is not real verification)

that would imply that developers cannot be trusted to write unit tests for their own code 😆

zaceno:
Ok, so the problem was the `element.appendChild...` line I commented on yesterday. It wasn't inefficcient -- it was *wrong* ;) After fixing that, @leeoniya 's tests all pass.

To "prove" it, here's a code-pen that cycles through the tests. One each second so that you can follow them visually.

If at any point the element.style.color doesn't match the element.textContent, the console should shout at you. Fortunately it doesn't :)

... Again, I prove nothing by verifying my own code of course. But consider it a green light to proceed with more verification & reviewing.

zaceno:
@leeoniya ... hehe, you make a good point. But then I at least hope someone looks at my tests ;)

leeoniya:
> Again, I prove nothing by verifying my own code of course. But consider it a green light to proceed with more verification & reviewing.

no implementation is bug free. but code that passes a non-trivial set of smell tests *and* has automated regression tests can be merged. if all existing tests pass and no new tests fail, that's the best anyone can do. the only enemy is not enough existing tests - there are never enough.

zaceno:
I can't explain how a test failed in circleci from my most recent push. Locally they're all green!

Besides all I did was remove a commented-out line I'd left in by mistake. (But you wouldn't know, because I did a `git commit --amend` and then a `git push --force`)

ngryman:
IMHO `_key` should be `key` (https://github.com/Matt-Esch/virtual-dom/tree/master/virtual-hyperscript#special-properties-in-h).

zaceno:
@ngryman -- fine with me. Makes sense considering onremove et c also have no special prefix.

jbucaran:
→ onRemove :)

AutoSponge:
So, I put in https://github.com/hyperapp/hyperapp/pull/151. Based on my work with morphdom solving a very similar issue, my code should solve this if the "key" element implements `isSameNode` as an expando property.

Here's the code that changed:
```js
function shouldUpdate(a, b) {
    return a.tag !== b.tag || typeof a !== typeof b || isPrimitive(a) && a !== b || isElement(a) && !a.isSameNode(b)
}
```

By declaring a virtual node as the same as another (existing DOM node) with `isSameNode` you can make the renderer implementation "skip" adding/mutating the in-place node.

jbucaran:
@AutoSponge Do you mean #151 is an alternative to adding key (this PR) or something else? I'm afraid I don't understand what you are saying.

AutoSponge:
@jbucaran it made more sense in my head...

To allow for keys to work and allowing real elements to be held by reference in the model (my PR), the `setElementData` process would need to set `isSameNode` to the element (since this implementation only knows how to use keys based on `node.data.key` which would not exist on an element).  The `isSameNode` fn would return true when passed a vnode having the expected key.

jbucaran:
@AutoSponge Haha okay.

I am currently working on my own keys implementation based in #141. Like you say I am not sure if it will be compatible with #151, but on the other hand, I still haven't quite understood what's the purpose of that PR. Don't take that as criticism please, on the contrary, I just need more time or perhaps if you have time you can try ELI5 to me again.

AutoSponge:
@jbucaran no problem! I'll add explanation to #151

jbucaran:
@zaceno 👋 Is this done or you planning tweaking something more?

zaceno:
@jbucaran, Hi sorry for the silence :) 

No, it's not complete. I am working on a rewrite where I do removeChild(...) only when absolutely necessary (to fix the css-transition issue, among other things). It's going to make the diff bigger, but it will be a more complete solution.  

I didn't get as much time as I had hoped this weekend, but I am working away at it little by little, and hope to have an update to this PR soon. By the end of the week at the latest.

zaceno:
@jbucaran So *now* I believe this solution is complete. It ended up bigger than I hoped, and perhaps touching more of the code than you would like. But it *works*.

(And because I don't needlessly remove & reinsert dom-nodes any longer, it does not break css transitions.)

Here are updated versions of all the code pens I've mentioned before:

@leeoniya s suite of smoke tests visualized here: http://codepen.io/zaceno/pen/wJqmjY
...notice that the console is silent. Proof that the keyed vdom solution does what it should

My css transitions slideshow with images: http://codepen.io/zaceno/pen/WpEJQV
... no flickering reloading images. See it break when you comment out the `key: ...` lines.

My css transitions slideshow with iframes: http://codepen.io/zaceno/pen/oZeqRG
... no reloading iframes! See it break when you comment out the `key: ...`lines.

the image gallery example: http://codepen.io/zaceno/pen/bqrvQO
... notice how when you press the update button, the gallery images *don't* reload (as intended). but if you comment out the `key: ...` line, then they *do* reload when the update button is pressed (a problem keys solve)

AutoSponge:
I solved a similar problem in my PR #151. The problem comes when replacing an element with a sibling element instead of just removing the preceding sibling. 

I simply perform the remove then, remove the reference from the `oldNode.children` array then update the array's length, `oldLen`. If `oldNode.children` held a reference to the element it created the first time it went through render, you could perform the same check I'm doing (since I'm looking at real dom nodes and using `.compareDocumentPosition`. 

If you want, I can try to add that to my PR and test your _original_ demos against it.

zaceno:
@AutoSponge, if I get what you're saying, I think a keyed vdom covers more cases than just the one you're describing, but the case you're describing is the problem I initially had with my iframe/image slidedshow demos.

I know my case can be solved with something simpler than keyed vdoms (I had a PR out before this one, that solved it in a more minimal way) -- but I'm absolutely curious to see your solution (I think I'd understand better what you're saying if I saw it anyway), so go for it! :)

That said, because keyed vdoms cover more cases (specifically: when you need to reuse but *reorder* dom elements), I'm hoping for that feature in hyperapp rather than a less complete solution.

Hope I understood you correctly! If I didn't, it wasnt intentional :)

AutoSponge:
@zaceno I started this work but I'm not quite done. As I worked, I thought `id` should be used as the key instead of `key` since (a) it should already be unique and (b) it's easy to get the list of "keyed" elements with `parent.querySelectorAll("[id]")`. 

I could continue to work on this implementation but don't let me hold up acceptance. My changes in #151 are likely to fold in easily since it's not that big.

zaceno:
@AutoSponge Cool, thanks for sharing that idea! The thing about keys vs ids is, though, is that while id's must be unique within the entire document, keys only need to be unique among sibling-nodes. And there are cases where you might want to have the same key in multiple places in a document. Like if you have multiple of the same components on a page.

By all means keep working, and don't hold your breath on this PR! I hope its not long before we can have keyed patching in hyperapp, but this may not be the implementation that makes it.

@jbucaran expressed hesitance on the slack-channel, against both the performance-degradation and increased bundle size this brings with it. So to make this happen I'll need to both optimize and condense it. But frankly, right now I'm not sure I can do any better, so I'm not feeling that hopeful :(

Thankfully @jbucaran is working on an implementation of his own, and I hope he fares better :)

zaceno:
Given the state of this PR, I want to express my purely personal and subjective opinion that: if no better solution (more performant, smaller) can be found in the near term, this PR should be merged anyway. 

There are several use cases where this functionality (not necessarily my implementation) is absolutely critical. Hyperapp *needs* this if it wants to be a "general purpose" frontend framework. 

Despite the shortcomings of my implementation in terms of size and performance, it *does work*. Once it's in, we can and should continue looking for optimizations. 

... that is of course *only if* no better implementation can be found in the short term. I'm still willing to wait for @jbucaran 's solution (or anyone else currently having a go at it)

jbucaran:
@zaceno Thanks for your work on this PR! I agree with you this is an important feature we need to ship ASAP. I hope I can dedicate some good time to this on the weekend.

zaceno:
So I made some adjustments and brought back the batch-delete. (thanks @jbucaran for brilliantly reminding me that dom nodes don't have to be detached from the document to be inserted!!) I also tried to refactor it into a hopefully more compressible way.

 I wasn't able to get the benchmark thing running, so I havent been able to check performance. But hopefully this should be much better in performance, and a little smaller as well.

src/app.js
```diff
@@ -215,48 +217,134 @@ export default function (app) {
...
+          element: element.childNodes[oldIndex + indexOffset]
+        }
+        oldIndex++
       }
     }
 
-    return element
+    for (var key in cache) {
+      var unused = cache[key]
+      batchRemove(element, unused.element, unused.node)
```
zaceno: I'm pretty sure that can't happen. Unkeyed nodes get scheduled for removal as we iterate over them. Keyed nodes are put in the cache, and scheduled for removal after we see we haven't used them.

test/app.test.js
```diff
@@ -288,6 +288,209 @@ describe("app", () => {
...
+                } else {
+                  return h("ul", {onUpdate: checkColors}, test.children)
+                }
+              }
+            })
+          })
+        })
+      })
+
+      it("fires onremove for ONLY the not-reused keyed nodes", done => {
```
zaceno: Ok, I'll remove them. But can we wait until it's definitely time to merge? Because they're pretty useful for knowing I don't break stuff when I fix merge-conflicts.

jbucaran: Of course, of course.

src/app.js
```diff
@@ -215,48 +217,134 @@ export default function (app) {
...
+    ) {
+      newChild = newNode.children[newIndex]
+      oldChild = oldNode.children[oldIndex]
+
+      if (
+        newChild !== undefined &&
+        isKeyed(newChild) &&
+        cache[newChild.data.key]
+      ) {
+        var c = cache[newChild.data.key]
```
zaceno: You mean `var i` compresses better than `var c`? Is that because `var i` is already used in a lot if places? 

Either way, I'll fix that, no problem.

src/app.js
```diff
@@ -215,48 +217,134 @@ export default function (app) {
...
       }
     }
   }
 
-  function patch(parent, element, oldNode, node) {
-    if (oldNode === undefined) {
-      element = parent.appendChild(createElementFrom(node))
-
-    } else if (node === undefined) {
+  function batchRemove (parent, element, node) {
```
zaceno: I don't se any problem batching the onRemove-handler, BUT even before this PR, the onRemove handler was deferred, meaning it gets was called *after* the node is removed. So... do you want me to fix that behavior as well, or leave it as it is?

src/app.js
```diff
@@ -215,48 +217,134 @@ export default function (app) {
...
-
-      if (oldNode && oldNode.data && oldNode.data.onRemove) {
-        defer(oldNode.data.onRemove, element)
+      if (node.data && node.data.onRemove) {
+        defer(node.data.onRemove, element)
       }
+  }
+
+  function patch (parent, element, oldNode, newNode) {
+    if (oldNode === undefined) {
```
zaceno: Yes, that is correct

src/app.js
```diff
@@ -215,48 +217,134 @@ export default function (app) {
...
+      var n = createElementFrom(newNode)
+      parent.insertBefore(n, element)
+      element = n
 
     } else if (
```
zaceno: If you're referring to the "insert app directly into document.body" merge, then yes. This code has been updated specifically to adapt to that (to return the new element of a patch)

src/app.js
```diff
@@ -215,48 +217,134 @@ export default function (app) {
...
+    }
+
+    return element
+  }
+
+  function patchChildren(element, oldNode, newNode) {
+    // We need to index all keyed nodes in oldChildren in order to
+    // distinguish new keyed nodes from nodes to reuse in newChildren
+    var keyedChildren = {}
+    for (var i = 0; i < oldNode.children.length; i++) {
```
zaceno: I'll try it :)

src/app.js
```diff
@@ -215,48 +217,134 @@ export default function (app) {
...
+    var newChild
+
+    var patchChild = function () {
+      patch(element, element.childNodes[oldIndex + indexOffset], oldChild, newChild)
+      oldIndex++
+      newIndex++
+    }
+
+    var insertChild = function () {
+      patch(element, element.childNodes[oldIndex + indexOffset], undefined, newChild)
```
zaceno: Ok. But it means the variable declaration for `node` at the top of the file will have to be instantiated to null as well. Trying that.

zaceno: Actually no. can't use null here. The patcher must be prepared to accept undefined values for oldChild parameter. And we can't just check `if(!oldChild)` because it could be an empty string. So we'd have to check `if (oldChild === null || oldChild === undefined)` ... which defeats the idea of *removing* bytes :)

src/app.js
```diff
@@ -215,48 +217,134 @@ export default function (app) {
...
+      newIndex++
+    }
+
+    var removeChild = function () {
+      batchRemove(element, element.childNodes[oldIndex + indexOffset], oldChild)
+      oldIndex++
+    }
+
+    while (
+      newNode.children[newIndex] !== undefined ||
```
zaceno: Totally. Don't know why I didn't think of that. Probably left over from some earlier attempt where the child-lists were mutated (so you couldn't rely on lengths). I'll fix this!

src/app.js
zaceno: Actually *YES*... I just realized that `'' == null` is false and `undefined==null` is true. So we can use null here with no problems. :)

src/app.js
```diff
@@ -215,48 +217,134 @@ export default function (app) {
...
+        removeChild()
+
+      } else if (!isKeyed(newChild)) {
+        insertChild()
+
+      } else if (oldChild.data.key === newChild.data.key) {
+        patchChild()
+
+      } else {
+        cache[oldChild.data.key] = {
```
zaceno: Hmm... not sure I follow. I'm happy to try it, but could you give me a code-snippet to explain more. What I'm envisioning is not smaller so I'm sure I'm misunderstanding :)

src/app.js
jbucaran: I pushed it to master a while ago, it's like 10 bytes smaller so...

jbucaran: 😆

src/app.js
```diff
@@ -215,48 +217,134 @@ export default function (app) {
...
+    var keyedChildren = {}
+    for (var i = 0; i < oldNode.children.length; i++) {
+      if (!oldNode.children[i].data) continue
+      if (!oldNode.children[i].data.key) continue
+      keyedChildren[oldNode.children[i].data.key] = true
+    }
 
-      var len = node.children.length
-      var oldLen = oldNode.children.length
+    var isKeyed = function (node) {
```
zaceno: Mostly for readability. Because they were repeated code snippets. But now that I reduced my 8-branch if-else down to its minimum 5-branches, I have got rid of those extra functions, and just inlined the code in the if-else.

zaceno: the `!!` was a dumb mistake. Fixed

I'll look more into how to avoid isKeyed!

zaceno:
Latest adjustments brings the total compressed bundle size down to 1954 bytes.

src/app.js
```diff
@@ -254,93 +254,66 @@ export default function (app) {
   function patchChildren(element, oldNode, newNode) {
     // We need to index all keyed nodes in oldChildren in order to
     // distinguish new keyed nodes from nodes to reuse in newChildren
-    var keyedChildren = {}
+    var oldChildren = []
```
jbucaran: @zaceno Nice!

src/app.js
jbucaran: I think there should be no `var x = function(){..}`. This pattern only adds more bytes.

src/app.js
```diff
@@ -211,40 +213,107 @@ export default function (app) {
...
 
-      for (var i = 0; i < len || i < oldLen; i++) {
-        patch(element, element.childNodes[i], oldNode.children[i], node.children[i])
-      }
+    var newIndex = 0
+    var oldIndex = 0
+    var cache = {}
+
+    while (
+      newIndex < newNode.children.length ||
```
jbucaran: @zaceno We should try to access stuff only once. `len`, `oldLen`, `oldNodeData`, etc.

src/app.js
```diff
@@ -211,40 +213,107 @@ export default function (app) {
...
+        element.insertBefore(i[0], oldChildren[oldIndex][0])
+        patch(element, i[0], i[1], newChild)
+        delete cache[newChild.data.key]
+        newIndex++
+
+      } else if (oldChild == null || (isKeyed(oldChild) && !isKeyed(newChild))) {
+        patch(element, oldChildren[oldIndex][0], null, newChild)
+        newIndex++
+
+      } else if (newChild == null || (!isKeyed(oldChild) && isKeyed(newChild)) ) {
```
jbucaran: @zaceno Perhaps you can store the oldKey and newKey in a variable, then just check `oldKey != null` to see if a node is keyed or not.

src/app.js
```diff
@@ -211,40 +213,107 @@ export default function (app) {
...
+      } else {
+        cache[oldChild.data.key] = oldChildren[oldIndex]
+        oldIndex++
+
+      }
     }
 
-    return element
+    for (var key in cache) {
+      batchRemove(element, cache[key][0], cache[key][1])
```
jbucaran: Noooice!

src/app.js
```diff
@@ -211,40 +213,107 @@ export default function (app) {
...
+    return element
+  }
+
+  function patchChildren(element, oldNode, newNode) {
+    // We need to index all keyed nodes in oldChildren in order to
+    // distinguish new keyed nodes from nodes to reuse in newChildren
+    var oldChildren = []
+    var oldKeys = {}
+    for (var i = 0; i < oldNode.children.length; i++) {
+      oldNode.children[i].data &&
```
jbucaran: @zaceno Let's if instead of && here. It adds no value since uglifyjs will do and undo with this code anyway.

src/app.js
```diff
@@ -211,40 +213,107 @@ export default function (app) {
...
+      oldNode.children[i].data.key &&
+      (oldKeys[oldNode.children[i].data.key] = true);
+      oldChildren[i] = [element.childNodes[i], oldNode.children[i]]
+    }
 
-      var len = node.children.length
-      var oldLen = oldNode.children.length
+    var isKeyed = function (node) {
+      return node.data && node.data.key && oldKeys[node.data.key]
```
jbucaran: @zaceno This means keys can't be numbers, specifically the number 0. I don't have an issue with that, but is that what we want?

leeoniya: 👎

src/app.js
zaceno: I tried that, but that change actually *added* 11 bytes. Either way I'll leave the change in the upcoming commit, so that you can see the result and perhaps help me find a shorter way to write it.

src/app.js
zaceno: Actually I think we can have numbers. The number 0 as well. Not sure exactly how various browsers handle this, but in v8 at least (just popped open my node console and tested this)

    > var a = {}
    undefined
    > a[0] = 'foo'
    'foo'
    > a 
    {'0': 'foo'}
    > a[0]
    'foo'

Sure, it will *cast* numbers to strings for the representation. But as long as you are not using a mix of numbers and stringed numbers to key your children (and *that* is a really unlikely scenario), it should work without conflict

src/app.js
jbucaran: Interesting! Thanks for letting me know that :)

src/app.js
jbucaran: @leeoniya Which one?

jbucaran: @zaceno If node.data.key is the number 0, then the above condition will result in iskeyed returning false.

What I meant was, if we are going to support keys as `Number`s, then the smallest check that works is using `null`.

leeoniya: right the condition will fail for all falsey values,which is wrong; all strings and numbers are valid keys. it's not a hash issue, which wull stringify all object keys.

src/app.js
zaceno: I managed to compensate for those 11 bytes with some other changes (in the upcoming commit)

src/app.js
zaceno: @leeoniya & @jbucaran: good points. Fixed in upcoming

zaceno:
Current bundle size: 1935 bytes

jbucaran:
@zaceno Now that #172 is out there, perhaps you can help there? We've been through this over Slack arleady. I am super thankful for all the work that went into this PR and all your feedback since you created #117 and I look forward more from you after this and from now on too! 👋

zaceno:
@jbucaran Of course! I've already left a couple of comments and am continuing to review it. As far as code style goes, I haven't got a lot to say since it's your project :). But I am going to pull it down and run it through some tests later tonight.

Anyhow until 172 is merged, I'll try to keep 141 in parity. If only for the sake of personal pride (Since you mentioned it would be pulled into the history with 172 on top of it, I don't want my name attached to something half-assed ;) ) Removing the deferred is one such way. It was a very small change really -- took all of 5 minutes, so...

zaceno:
Cleaned up the PR. Removed tests (as requested) and squashed to a single commit.

jbucaran:
@zaceno Cool! I benchmarked my implementation and it's not as bad as I thought, 1.72, but I am sure we can do better.

Cache the old keys and old element lists might make a difference. Cache everything, we can even memoize toLowerCase. Also, going to try a different approach at removing stuff.

It would be great if you could review my PR and suggest more changes or how to use some of the features from your PR, since they are both pretty similar after all.

zaceno:
@jbucaran Actually I suspect that our attempts to make it smaller have led to poorer performance. I've been hearing a lot of chatter lately about how specific declarative checks like `if (typeof foo === undefined)` run much faster than `if (!foo)` or `if (foo != null)` -- because of the casting that isn't necessary in the first case. Here's an example of such chatter I've been hearing 

https://github.com/developit/preact/pull/610

(I'll duplicate this post over on your PR so the discussion can continue there)

jbucaran:
@zaceno 🎉

PR #142, opened 2017-03-08
Change iterators to array methods

jwulf:
I changed the iterators to array methods, and used const and let.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -1,28 +1,27 @@
 export default function (app) {
-  var view = app.view || function () {
+  const view = app.view || function () {
     return ""
   }
 
   var model
-  var actions = {}
-  var subscriptions = []
-  var hooks = {
+  const actions = {}
+  let subscriptions = []
+  const hooks = {
     onError: [],
     onAction: [],
     onUpdate: [],
     onRender: []
   }
 
-  var plugins = [app].concat((app.plugins || []).map(function (plugin) {
+  const plugins = [app].concat((app.plugins || []).map(function (plugin) {
     return plugin(app)
   }))
 
-  var node
-  var root
-  var batch = []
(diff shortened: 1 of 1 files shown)


selfup:
Thanks for the PR 🎉  

However, the lib source is ES3-ES5 compliant to remove the need of a transpiler and be under 1kb.

@jbucaran 

I wonder about Array methods. forEach is nice to write but from a perf perspective I wonder what the impact will be. I know forEach has to check all Array properties prior to recursive iteration (no tail call optimization) so I would reckon performance will slow down and stack overflows would happen on large DOMs

jwulf:
Is the test failure due to the ES6 `let` and `const`?

selfup:
Good question! Could be arrow functions since uglify is more than likely like wutttttt

selfup:
It does seem to have blown up on the first instance of let

jwulf:
According to jsperf, forEach is 96% slower than for: http://jsperf.com/fast-array-foreach

selfup:
I didn't realize it was that drastic!

jwulf:
Yeah, forget this PR then. It needs to be fast.

selfup:
Thanks for creating meaningful discussion anyways 🎉

jbucaran:
Thanks @jwulf! We went thru this, just not in a specific PR, but when the Plugin architecture was introduced. Yes, forEach is slower than for, and also for consistency.

The code uses ES6 module export/import since that's what works best with Rollup/Webpack and enables tree-shaking, in our case that means stripping the router from the build bundle if you are not using it, etc.

dodekeract:
Here is a more fair comparison between for and foreach
https://jsperf.com/fast-array-foreach/3

PR #150, opened 2017-03-09
onRoute router

lukejacksonn:
Apologies, I have created a PR instead of an issue already, but the code here works as intended so good for discussion!

This implementation allows for `onRoute` style subscriptions using the router syntax:
```es6
view: {
  '/': [Default, (m,a) => a.counter.increment()], // <-- called on a.router.go('/')
  '*': [Fallback], // <-- no actions to call for this route
},
```

This could be easily altered to allow for the old router syntax too, just needs a few more checks as to what form the matching route takes (array or view function).


lukejacksonn:
hummm.. this doesn't work with `popstate` navigation because it is attached to the `router.go` action 🤔 I did have something similar working in a plugin with an`onUpdate` hook that looked like this:

```
var lastMatch = ''
export default (options) => ({
  hooks: {
    onUpdate: function (oldModel, newModel, data, actions) {
      if(newModel.router && newModel.router.match !== lastMatch) {
        lastMatch = newModel.router.match
        if (typeof options.view[newModel.router.match][1] === 'function')
          options.view[newModel.router.match][1](newModel, actions)
      }
    },
  },
})
``` 
..but it requires the router to pass `actions` to `onUpdate` hooks. Is there any reason why this would not be desirable?

lukejacksonn:
That did not work either 😅 because it is checking for difference of `router.match` which would return `/:id` any route of the form `/*`; so it would not trigger on navigation from `/1` to `/2`. A more reliable way to check if the route is new could be to compare `oldModel.router` and `newModel.router` for equality but this is awkward in JS. So maybe this would be better..
```
onUpdate: function (oldModel, newModel, data, actions) {
  if(newModel.router && oldModel.router.path !== newModel.router.path &&
      typeof options.view[newModel.router.match][1] === 'function') {
    options.view[newModel.router.match][1](newModel, actions)
  }
},
```
This requires that the `match` function returns an object with a `path` attribute from which the router derived the `match` and `params` attribute:
```
{
  match: '/:id',
  params: { id: 1 },
  path: '/1'
}
```
Now we know if the route has changed if there is an update where `newModel.router` exists and `oldModel.router.path !== newModel.router.path`.

jbucaran:
@lukejacksonn Hey man, how's this PR coming along? @dodekeract Do you have any comments?

.gitignore
```diff
@@ -11,6 +11,7 @@ sandbox
 node_modules
 coverage
 *.xml
+.DS_Store
```
dodekeract: :+1:

src/app.js
```diff
@@ -77,7 +77,7 @@ export default function (app) {
 
           } else {
             for (i = 0; i < hooks.onUpdate.length; i++) {
-              hooks.onUpdate[i](model, result, data)
+              hooks.onUpdate[i](model, result, data, actions)
```
dodekeract: Maybe we should consider changing this to `model, data, actions, result`? That's similar to the action signature.

src/router.js
```diff
@@ -17,8 +17,14 @@ export default function (options) {
       }
     },
     hooks: {
+      onUpdate: function (oldModel, newModel, data, actions) {
+        if(newModel.router && oldModel.router.path !== newModel.router.path &&
```
dodekeract: This seems to rely on `newModel.router` being a different object than `oldModel.router`. As far as I know we don't guarantee that.

dodekeract:
What if we change this:

`if (actions.router.onUpdate) actions.router.onUpdate()`

People would then be able to just specify a `router.onUpdate` action in their actions and the router would notice it and pick it up.

src/app.js
lukejacksonn: Ideally yes 👍 for the sakes of not introducing breaking changes that could be avoided, perhaps no? Or in a separate PR?

src/router.js
lukejacksonn: It relies on `oldModel.router.path !== newModel.router.path` where `path` is  https://github.com/hyperapp/hyperapp/pull/150/files#diff-15ee858ae5d5aa9b3e2846cfd1c4c466R71

The `newModel.router` check is there to prevent an error being thrown when trying to access `newModel.router.path` when `newModel.router` does not exist. Also quite a cheap way of checking if any action needs to be considered at all.

dodekeract: Yes but if it's the same object it will have the same path. We use a really simple merge/cloning algorithm and I'm not sure if this algorithm always guarantees that the compared paths will be different.

lukejacksonn: I'm not sure if I follow.. surely at some point there will be a difference in `location.pathname` from `/one` to `/two`. At the moment this code triggers only when the location is actually changed. Not on page load like a subscription and not if you navigate from `/one` to `/one`

lukejacksonn:
So you would only have one of these `router.onUpdate(/*...*/)` actions per application? What does it get passed and how does it detect if the route has actually changed?

dodekeract:
@lukejacksonn The detection part would still be how your code worked. But the way to trigger `onUpdate` wouldn't rely on changing the view anymore. It'd actually get passed nothing, because the model already has the current route & parameters.

lukejacksonn:
I like that better! Having to change the view didn't feel quite right. So you expect something like:

```
const actions = {
  router: {
    onUpdate: (m,d,a) => switch(m.router.path) { ... }
  }
}
```
to initiate actions for certain routes?

dodekeract:
@lukejacksonn Yes. Quite sure that you need `{}` around `switch` though.

lukejacksonn:
I have managed to abstract all the route change detection logic out to a plugin with an `onUpdate` hook. Thus decoupling the view from the _onRoute_ style subscriptions which was my big issue with my last approach. I tried implementing what you suggested @dodekeract but couldn't get it to play ball for one reason or another. Ended up with this, which I think is much easier to digest!

```es6
// Fetcher Plugin
export default () => ({
  hooks: {
    onUpdate: (om,m,d,a) => {
      if(m.router && om.router.path !== m.router.path) {
        switch (m.router.match) {
          case '/': return a.component.fetch(0)
          case '/:id': return a.component.fetch(m.router.params.id)
        }
      }
    },
  },
})
```

**This means that** the only fundamental changes are:

- To pass `actions` as the last parameter to `onUpdate` functions

This changes the function signature to `(prev, model, data, actions)` which is sensible given the action signature `(model, data, actions, error)`. Not a breaking change.

- To add a `path` attribute to the `router` model

The value of the `location.pathname` that got passed into `router.match` to produce a new model. Which I'm sure was proposed but never made it in. Not a breaking change either.

src/app.js
lukejacksonn: Thinking about this now.. I think `(oldModel, model, data, actions)` actually makes sense, see my last commit and comment

lukejacksonn:
This can probably be closed now due to the introduction of the `route` event. I am yet to try it out on one of the projects I have that requires this feature (but is still using a workaround). Will reopen an issue if I come across any limitations!

jbucaran:
Sounds good, thanks Luke.

Issue #148, opened 2017-03-09
Make app composable


AutoSponge:
Here are the changes I think need to be made to `app` and `h` to make sure that apps can be composed over each other. These features will also allow hyperapps to play nicely with independent, non-hyperapp components.

# 1
`app#render()` should not try to `createElement` when `node` is an element:

```js
const isElement = (node) => node instanceof window.HTMLElement
```

# 2
`h()` should skip elements. If `isElement`, return "`tag`"

# 3
`app()` should return `root`.

# 4
~~`render` should check for `root` existence in DOM before trying to `render` (making un-mounted apps less of a burden). When an app.root is added to the DOM by a parent app, a hook should fire `onAppend`. When removed from the DOM, a hook should fire, `onRemove`.~~
`root` should be upgraded to a node but should not be mounted to the DOM automatically.

Let's discuss :)

jbucaran:
@AutoSponge What you are saying is that you want to pass app into app?

```jsx
const MyApp = app({...})

app({ view: model => <div><MyApp /><OtherStuff/></div>
```

jbucaran:
> app() should return root.

This is easy to do. Does this imply that app will not mount itself automatically? I kind of like that.

AutoSponge:
@jbucaran I think it would look more like 

```js
app({
  view: (model) => <div>{ app({...}) }</div>
})
```

jbucaran:
1 and 2 also look easy, although I'd like to see what are the benefits of this, `h` is heavily optimized and one little more if you add there only slows everything down.

These concerns are probably "worthless". I am basically using js-framework-benchmark as a base. 0.7.1 currently scores ~1.30, which is pretty good.

AutoSponge:
>> app() should return root.

> This is easy to do. Does this imply that app will not mount itself automatically? I kind of like that.

Currently, you can provide a `root` that's not on the DOM. I think that's how you can make an app "lazy". Only when the parent appends the returned element would the subcomponent try to render (although actions, subscriptions, etc. will still fire depending on how they're coded).

jbucaran:
> Currently, you can provide a root that's not on the DOM.  I think that's how you can make an app "lazy".

That sounds interesting and worthy of its own separate issue. We could totally have that before any of the other stuff.

AutoSponge:
> That sounds interesting and worthy of its own separate issue. We could totally have that before any of the other stuff.

```js
const component = app({
  view: (model) => My Component,
[...]

PR #151, opened 2017-03-09
Composable app

AutoSponge:
addresses https://github.com/hyperapp/hyperapp/issues/148
technique for avoiding problems found in https://github.com/hyperapp/hyperapp/issues/117
may also solve https://github.com/hyperapp/hyperapp/pull/141 if `isSameNode` is used by virtual node, requires test

# Explanation
#151 is about trying to make apps smaller and with single-purpose. Then, using composition and communication (not enforced by the framework), construct your goal.

In my [example](https://gomix.com/#!/project/picayune-postbox), I created a self-contained `app` called `slide`. The `slide` component exports a create method. 

Because `app()` now returns its root and `h()` roots are upgraded to DOM nodes, the `main` app that composes over the `slide`s only has to worry about its `model`--the list of slides. Some additional state was added to "control" when to add/remove/transition slides, but that's gravy to demonstrate the extensibility of the pattern.

In this example, I used shared knowledge of an event namespace to communicate from `main` to the `slide`s. if 2-way communication was needed, the `slide` app could trigger a `register` event that passes an event emitter as data to the `main` and any apps that are composing over `main`.

When passing events to parents, use bubbling (capture=false) so that propagation can be stopped by a parent. When passing events "down" the stack, make sure the composed component uses capturing to get the event. These will process before any "upstream" apps receive the event.


AutoSponge:
I'm working on the example using this version.

jbucaran:
@AutoSponge Thanks! I've left a few notes, let's discuss :)

AutoSponge:
I made a thing based on the previous "problem" example in #117. I will need to add some tests, but please evaluate these changes to see if they fit the overall need.

[example](https://gomix.com/#!/project/picayune-postbox)

Things learned:
- If the `app` is lazy (passed a `root`) we can return that root. But if using dynamic `root`, we can't return because the root isn't created until the DOM is ready. This might be able to change with either a promise or by creating the root element before the DOM is ready and mounting it on DOMready. This wasn't changed because it wasn't needed for this example.
- There seems to be a bug in `patch` when `parent.childNodes[index]`  doesn't exist (needs test).
- I learned how to make this slider thing!

dodekeract:
> There seems to be a bug in patch when parent.childNodes[index] doesn't exist (needs test).

You could add a watch expression for that in the developer tools/debugger.

AutoSponge:
This is done but if it comes after the "keys" PR, it may need to be reworked.

AutoSponge:
This was a small change that supported several use cases. But master is changing constantly and I don't have time to keep rebasing this. Let me know when you want it and I'll fork again.

PR #172, opened 2017-03-28
Introduce keys. 

jbucaran:
Keys are a simple way to tell if two virtual nodes represent the same element.

If you are familiar with [keys in React](https://facebook.github.io/react/docs/reconciliation.html), this is basically the same idea.

Keys can be anything, but they are usually of type Number or String. Keys uniquely identify a group of elements among its siblings. Keys are assigned to nodes in the same way as attributes are.

```jsx
<div key="a"></div>
```

Consider the following example:

* Old tree.

```jsx
view: model =>
  <main>
    <div>A</div>
  </main>
```

* New tree.

```jsx
view: model =>
  <main>
    <div>B</div>
    <div>A</div>
  </main>
```

Usually, hyperapp will patch the old A div with the new B div and create a new A div.

Using keys, hyperapp can create a new div for B and insert it on top of A, which may be preferred depending on your program requirements. See #117.

```jsx
view: model =>
  <main>
    <div key="a">A</div>
  </main>
```

* New tree.

```jsx
view: model =>
  <main>
    <div key="b">B</div>
    <div key="a">A</div>
  </main>
```

### Reviewers 🙏
* @zaceno
* @dodekeract 
 
### Related 
* #141 
* #117 
* #135

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -7,18 +7,17 @@ export default function (app) {
 
   var model
   var actions = {}
-  var subscriptions = []
   var hooks = {
     onError: [],
     onAction: [],
     onUpdate: [],
     onRender: []
   }
+  var subscriptions = []
 
   var node
-  var element
   var root
-  var batch = []
+  var element
   var plugins = app.plugins || []
 
   for (var i = -1; i < plugins.length; i++) {
@@ -38,8 +37,8 @@ export default function (app) {
     }
 
     if (obj = plugin.hooks) {
-      Object.keys(obj).forEach(function (key) {
-        hooks[key].push(obj[key])
+      Object.keys(obj).map(function (key) {
+        (hooks[key] ? hooks[key] : hooks[key] = []).push(obj[key])
       })
     }
   }
@@ -49,76 +48,68 @@ export default function (app) {
 
     render(model, view)
 
-    for (var i = 0; i < subscriptions.length; i++) {
-      subscriptions[i](model, actions, onError)
-    }
+    subscriptions.map(function (cb) {
+      cb(model, actions, error)
+    })
   })
 
-  function onError(error) {
-    for (var i = 0; i < hooks.onError.length; i++) {
-      hooks.onError[i](error)
-    }
-
-    if (i <= 0) {
+  function error(error) {
+    if (hooks.onError.length === 0) {
       throw error
     }
+
+    hooks.onError.map(function (cb) {
+      cb(error)
+    })
   }
 
   function init(container, group, lastName) {
-    Object.keys(group).forEach(function (key) {
-      if (!container[key]) {
-        container[key] = {}
-      }
-
+    Object.keys(group).map(function (key) {
       var name = lastName ? lastName + "." + key : key
       var action = group[key]
 
       if (typeof action === "function") {
         container[key] = function (data) {
-          for (var i = 0; i < hooks.onAction.length; i++) {
-            hooks.onAction[i](name, data)
-          }
+          hooks.onAction.map(function (cb) {
+            cb(name, data)
+          })
 
-          var result = action(model, data, actions, onError)
+          var result = action(model, data, actions, error)
 
           if (result == null || typeof result.then === "function") {
             return result
 
           } else {
-            for (var i = 0; i < hooks.onUpdate.length; i++) {
-              hooks.onUpdate[i](model, result, data)
-            }
+            hooks.onUpdate.map(function (cb) {
+              cb(model, result, data)
+            })
 
             model = merge(model, result)
             render(model, view)
           }
         }
       } else {
-        init(container[key], action, name)
+        init(container[key]
+          ? container[key]
+          : container[key] = {}, action, name)
       }
     })
   }
 
-  function load(fn) {
+  function load(cb) {
     if (document.readyState[0] !== "l") {
-      fn()
+      cb()
     } else {
-      document.addEventListener("DOMContentLoaded", fn)
+      document.addEventListener("DOMContentLoaded", cb)
     }
   }
 
   function render(model, view) {
-    for (var i = 0; i < hooks.onRender.length; i++) {
-      view = hooks.onRender[i](model, view)
-    }
+    hooks.onRender.map(function (cb) {
+      view = cb(model, view)
+    })
 
     element = patch(root, element, node, node = view(model, actions))
-
-    for (var i = 0; i < batch.length; i++) {
-      batch[i]()
-    }
-
-    batch = []
   }
 
   function merge(a, b) {
@@ -128,20 +119,14 @@ export default function (app) {
       return b
     }
 
-    for (var key in a) {
-      obj[key] = a[key]
+    for (var i in a) {
+      obj[i] = a[i]
     }
-    for (var key in b) {
-      obj[key] = b[key]
+    for (var i in b) {
+      obj[i] = b[i]
     }
 
     return obj
-  }
-
-  function defer(fn, data) {
-    setTimeout(function () {
-      fn(data)
-    }, 0)
   }
 
   function createElementFrom(node, isSVG) {
@@ -155,7 +140,7 @@ export default function (app) {
 
       for (var name in node.data) {
         if (name === "onCreate") {
-          defer(node.data[name], element)
+          node.data[name](element)
         } else {
           setElementData(element, name, node.data[name])
         }
@@ -172,7 +157,8 @@ export default function (app) {
   function setElementData(element, name, value, oldValue) {
     name = name.toLowerCase()
 
-    if (!value) {
+    if (name === "key") {
+    } else if (!value) {
       element[name] = value
       element.removeAttribute(name)
 
@@ -204,14 +190,14 @@ export default function (app) {
     }
   }
 
-  function updateElementData(element, data, oldData) {
+  function updateElementData(element, oldData, data) {
     for (var name in merge(oldData, data)) {
       var value = data[name]
       var oldValue = oldData[name]
       var realValue = element[name]
 
       if (name === "onUpdate") {
-        defer(value, element)
+        value(element)
 
       } else if (value !== oldValue || realValue !== value) {
         setElementData(element, name, value, oldValue)
@@ -219,27 +205,122 @@ export default function (app) {
     }
   }
 
+  function getKeyFrom(node) {
+    if (node && (node = node.data)) {
+      return node.key
+    }
+  }
+
+  function removeElement(parent, element, node) {
+    if (node.data.onRemove) {
+      node.data.onRemove(element)
+    }
+    parent.removeChild(element)
+  }
+
   function patch(parent, element, oldNode, node) {
     if (oldNode == null) {
-      element = parent.appendChild(createElementFrom(node))
+      element = parent.insertBefore(createElementFrom(node), element)
 
-    } else if (node == null) {
-      batch.push(function () {
-        if (oldNode.data && oldNode.data.onRemove) {
-          oldNode.data.onRemove(element)
-        }
-        parent.removeChild(element)
-      })
     } else if (node.tag && node.tag === oldNode.tag) {
-      updateElementData(element, node.data, oldNode.data)
+      updateElementData(element, oldNode.data, node.data)
 
       var len = node.children.length
       var oldLen = oldNode.children.length
+      var reusableChildren = {}
+      var oldElements = []
+      var newKeys = {}
 
-      for (var i = 0; i < len || i < oldLen; i++) {
-        patch(element, element.childNodes[i], oldNode.children[i], node.children[i])
+      for (var i = 0; i < oldLen; i++) {
+        var oldElement = element.childNodes[i]
+        oldElements[i] = oldElement
+
+        var oldChild = oldNode.children[i]
+        var oldKey = getKeyFrom(oldChild)
+
+        if (null != oldKey) {
+          reusableChildren[oldKey] = [oldElement, oldChild]
+        }
       }
 
+      var i = 0
+      var j = 0
+
+      while (j < len) {
+        var oldElement = oldElements[i]
+        var oldChild = oldNode.children[i]
+        var newChild = node.children[j]
+
+        var oldKey = getKeyFrom(oldChild)
+        if (newKeys[oldKey]) {
+          i++
+          continue
+        }
+
+        var newKey = getKeyFrom(newChild)
+
+        var reusableChild = reusableChildren[newKey]
+        var reusableElement = 0
+        var reusableNode = 0
+
+        if (reusableChild) {
+          reusableElement = reusableChild[0]
+          reusableNode = reusableChild[1]
+        }
+
+        if (null == oldKey && null == newKey) {
+          patch(element, oldElement, oldChild, newChild)
+          j++
+          i++
+
+        } else if (null == oldKey && null != newKey) {
+          if (reusableElement) {
+            element.insertBefore(reusableElement, oldElement)
+            patch(element, reusableElement, reusableNode, newChild)
+          } else {
+            patch(element, oldElement, null, newChild)
+          }
+
+          j++
+          newKeys[newKey] = newChild
+
+        } else if (null != oldKey && null == newKey) {
+          i++
+
+        } else {
+          if (oldKey === newKey) {
+            patch(element, reusableElement, reusableNode, newChild)
+            i++
+
+          } else if (reusableElement) {
+            element.insertBefore(reusableElement, oldElement)
+            patch(element, reusableElement, reusableNode, newChild)
+
+          } else {
+            patch(element, oldElement, null, newChild)
+          }
+
+          j++
+          newKeys[newKey] = newChild
+        }
+      }
+
+      while (i < oldLen) {
+        var oldChild = oldNode.children[i]
+        var oldKey = getKeyFrom(oldChild)
+        if (null == oldKey) {
+          removeElement(element, oldElements[i], oldChild)
+        }
+        i++
+      }
+
+      for (var i in reusableChildren) {
+        var reusableChild = reusableChildren[i]
+        var reusableNode = reusableChild[1]
+        if (!newKeys[reusableNode.data.key]) {
+          removeElement(element, reusableChild[0], reusableNode)
+        }
+      }
     } else if (node !== oldNode) {
       var i = element
       parent.replaceChild(element = createElementFrom(node), i)
--- a/src/app.js
+++ b/src/app.js
@@ -173,7 +173,9 @@ export default function (app) {
         element.style[i] = value[i]
       }
     } else {
-      element.setAttribute(name, value)
+      if (typeof value !== "function") {
+        element.setAttribute(name, value)
+      }
 
       if (element.namespaceURI !== SVG_NS) {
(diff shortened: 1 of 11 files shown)


src/app.js
```diff
@@ -247,4 +229,114 @@ export default function (app) {
...
+        if (reusableElement) {
+          element.insertBefore(reusableElement, oldElement)
+          patch(element, reusableElement, reusableNode, newChild)
+        } else {
+          patch(element, oldElement, null, newChild)
+        }
+        j++
+        newKeys[newKey] = true
+
+      } else if (oldKey && !newKey) {
```
zaceno: just checking `!newKey` might result in false positives for keys such as `0` or `''` (not sure we should allow empty string as a key. But perhaps 0 should be ok.

selfup: Well I typically use `.map((element, index) =>` to create keys when I have very simple objects, so we either make sure that passes, or ask people to: `.map((element, index) => <h1 key={index + 1}>Hi</h1>`

jbucaran: @zaceno Yup, the same thing I brought up in #141. Not sure what's the best thing to do. Save bytes and force keys to be strings, always convert keys to strings, allow both? 🤔

jbucaran: @zaceno @selfup Okay, now it uses null to see if a key exists or not. So now, keys can be anything, except: `undefined` or `null`.

selfup: @jbucaran great! That was a good workaround 🎉

src/app.js
```diff
@@ -221,30 +216,128 @@ export default function (app) {
...
     } else if (node !== oldNode) {
       var i = element
       parent.replaceChild(element = createElementFrom(node), i)
     }
 
     return element
   }
+
+  function getKeyFrom(node) {
+    return node && (node = node.data) ? node.key : node
```
zaceno: If node is `''` then `getKeyFrom('')` will return `''`

jbucaran: @zaceno Good point, I'll fix that.

src/app.js
```diff
@@ -172,7 +166,8 @@ export default function (app) {
   function setElementData(element, name, value, oldValue) {
     name = name.toLowerCase()
 
-    if (!value) {
+    if (name === "key") {
+    } else if (!value) {
```
dodekeract: Any reason why this isn't `if (name !== 'key' && !value)`?

dodekeract: Any reason why this isn't `if (name !== 'key' && !value)`?

src/app.js
```diff
@@ -221,30 +216,128 @@ export default function (app) {
...
     return element
   }
+
+  function getKeyFrom(node) {
+    if (node && (node = node.data)) {
+      return node.key
+    }
+  }
+
+  function batchRemove(parent, element, node) {
```
dodekeract: `batchRemove` sounds like this would be the function that removes multiple items. It looks more like an `enqueue` function though.

src/app.js
```diff
@@ -221,30 +216,128 @@ export default function (app) {
...
     } else if (node !== oldNode) {
       var i = element
       parent.replaceChild(element = createElementFrom(node), i)
     }
 
     return element
   }
+
+  function getKeyFrom(node) {
+    if (node && (node = node.data)) {
```
dodekeract: Why does this overwrite `node`?

src/app.js
```diff
@@ -221,30 +216,128 @@ export default function (app) {
...
-        patch(element, element.childNodes[i], oldNode.children[i], node.children[i])
+        var oldChild = oldNode.children[i]
+        var oldKey = getKeyFrom(oldChild)
+
+        if (null != oldKey) {
+          reusableChildren[oldKey] = [oldElement, oldChild]
+        }
       }
 
+      var i = 0
```
dodekeract: Could there be better names for `i` and `j`? We minify so variable name length doesn't matter.

src/app.js
```diff
@@ -221,30 +216,128 @@ export default function (app) {
...
+
+        var reusableChild = reusableChildren[newKey]
+        var reusableElement = 0
+        var reusableNode = 0
+
+        if (reusableChild) {
+          reusableElement = reusableChild[0]
+          reusableNode = reusableChild[1]
+        }
+
```
dodekeract: Something like this will probably save some bytes:

```javascript
// evil bitwise hacking to add 0b1 if oldKey is null and 0b10 if newKey is null
// leads to both null => 3, newKey null => 2, oldKey null => 1, nothing null => 0

var select = null == oldKey | 2 * (null == newKey)

if (select == 3) {
// if (null == oldKey && null == newKey)
}
if (select == 1) {
// if (null == oldKey &&  null != newKey)
}
if (select == 2) {
// if (null != oldKey && null == newKey)
}
if (select == 0) {
// else
}
```

src/app.js
jbucaran: That would be more bytes after build.

src/app.js
jbucaran: Would yield a larger file size after build.

src/app.js
dodekeract: Why?

src/app.js
jbucaran: That's pretty evil, hehe. What does `2*` mean there?

src/app.js
jbucaran: Bytes.

src/app.js
jbucaran: I like `enqueue` better! 👍

src/app.js
dodekeract: @jbucaran Think of it as `0b10`.

It's basically just the binary representation of

| isNull? | oldKey | newKey |
|-------:|---------|--------|
| yes      | `0b01`     | `0b10` |
| no        | `0b00`   | `0b00` |

If you add these two up you'll get all 4 cases `0b00`, `0b01`, `0b10`, `0b11`.

src/app.js
dodekeract: @jbucaran But we use a minifier. That changes the variable names to `a`, `b`, etc. anyway.

src/app.js
jbucaran: Interesting. This may come later. Thanks for the summary @dodekeract 👍.

src/app.js
jbucaran: No clue.

src/app.js
jbucaran: No idea man. If you move the position of functions, like put some at the top, swapping their positions just to see it sometimes yield a larger file size too. Haha, it's weird.

leeoniya:
i'll give you one idiot's opinion here (none).

you guys are prematurely hyper-optimizing for size (pardon the pun). don't waste your time until you have proven the features/no-bugs and performance of your architecture. it often happens than many things need to be re-organized, edge cases and conditional branches added, etc.

your size is already great, even if it was 4k, it would still be equally great (no one will notice). but if your performance is 50% worse and there are shitloads of bugs, no one will give a hit about your 2k. don't waste cycles on this stuff, it can all be done later. be mindful of size, but leave size optimization passes for later since they seem to bog down features and bug fixes.

/$0.02

dodekeract:
@leeoniya It's the only thing I can suggest. I have no idea how the virtual DOM works besides the basics. 🙁

leeoniya:
> It's the only thing I can suggest. I have no idea how the VDOM magic works besides the basics. 

that's fair, still the suggestions may be premature (see below).

> Did I miss something?

I'm just saying in theory. For instance, you guys have no "Keyed" implementation on the board [1]. This keyed pull request could perform poorly and may need to be entirely rewritten, yet there is a lot of time dedicated in the initial implementation optimizing it for size, which could all be wasted effort if it all needs to be tossed. perhaps @jbucaran already benched it and found it to be acceptable, i dunno.

[1] https://rawgit.com/krausest/js-framework-benchmark/master/webdriver-ts/table.html

src/app.js
```diff
@@ -155,7 +149,7 @@ export default function (app) {
 
       for (var name in node.data) {
         if (name === "onCreate") {
-          defer(node.data[name], element)
+          node.data[name](element)
```
zaceno: calling the onCreate handler before we've created the child elements means that the element is empty when the oncreate handler is called.  It decreases the utility of the lifecycle method. I suggest moving the eventual call to onCreate handlers until after the child-node creation loop.

As a motivation to why I think it matters:

I noticed this because one of my tests depends on the children being available to the element (so I can loop over them). If left as is, I would have to wrap my onCreate in a setTimeout function which feels funky, and like something I shouldn't have to do.

jbucaran: @zaceno Ah, that explains why I originally was deferring these stuff! Well observed 👍

dodekeract: @jbucaran Pro tip: add a comment to make this obvious. 🙂

src/app.js
```diff
@@ -221,30 +216,128 @@ export default function (app) {
...
+          j++
+          i++
+
+        } else if (null == oldKey && null != newKey) {
+          if (reusableElement) {
+            element.insertBefore(reusableElement, oldElement)
+            patch(element, reusableElement, reusableNode, newChild)
+          } else {
+            patch(element, oldElement, null, newChild)
+          }
```
zaceno: When I ran my tests for #141 on this PR, lines 278-279 were the only un-covered lines. That means, either:

a) I have missed a case in my tests - totally possible, but I can't for the life of me figure out what that would be, or
b) this branch is redundant

Could you, @jbucaran, give an example of a case where the code path would go through line 278?

jbucaran: When the new node can't be reused, that is, it was not already present in the DOM, so it is a new _new_ node. Make sense? Or is this covered somewhere else already?

zaceno: Yes makes sense, except it's a little more specific (I figured out after looking a little closer at the algo): It's when we encounter a child in the new tree *with a key that was not used in the old tree*

It makes sense those lines wouldn't be covered then because I have no tests where the new tree has keys not present in the old.

So all is well... nothing to do here :)

zaceno: Added that case to my tests, and now coverage was full (and all passing, of course). So no redundant code as far as I can tell :)

src/app.js
zaceno: I could be wrong about this, but a possible explanation why it makes a difference is that if you already have one-letter variables, the minifier doesn't change their names. And since `i` and `j` are commonly used perhaps the compression algorithm will give them a shorter representation than other one-letter vars. Maybe. Just a suspicion

zaceno:
@leeoniya 
> I'm just saying in theory. For instance, you guys have no "Keyed" implementation on the board [1]. This keyed pull request could perform poorly and may need to be entirely rewritten

Yes, I'm really curious to see some numbers how this performs, both when using keys for the rows in the benchmark and not.

(Not as relevant but out of personall academic curiosity I'd be interested in those numbers for my implementation in PR 141 as well)

src/app.js
dodekeract: @zaceno That's what I suspected too, but IMO a minifier shouldn't even look at the original variable names. Not sure if it does.

Also @jbucaran made it sound like there is a significant difference in size.

zaceno: I agree that if that is indeed how minifiers behave it's pretty weird

leeoniya: Weird? You haven't seen weird. I just finished banging my head against a performance regression caused by Google Closure Compiler inclining a function into a parent function because it became smaller - "this will be faster", i thought.

It did this to reduce size, okay. But when it came to benchmarks, the Chrome's JIT wasn't fastpathing the parent now because the entire parent became larger as a result of the inlining. I had to hack in a thing to prevent it from inlining it to get the perf back (at the expense of 20 bytes.)

[1] https://github.com/leeoniya/domvm/commit/e3a02fc9e301e0c038f8ac4505a93d5846a8fb64

src/app.js
jbucaran: @zaceno BTW, I didn't try to figure out why, but your implementation fails to run the benchmarks at all. It crashes. So, I couldn't benchmark it to compare. 🤔

zaceno: 🙀 Oh dear... that's curious. Thanks for trying though!

zaceno:
@jbucaran Actually I suspect that our attempts to make it smaller have led to poorer performance. I've been hearing a lot of chatter lately about how specific declarative checks like `if (typeof foo === undefined)` run much faster than `if (!foo)` or `if (foo != null)` -- because of the casting that isn't necessary in the first case. Here's an example of such chatter I've been hearing 

https://github.com/developit/preact/pull/610

jbucaran:
Good to know! Thanks.

jbucaran:
> My vote is on making it fast. Because to me, the benefit of smallness is that I can read and understand the code. But that benefit goes away the less declarative, more implicit and "magical" the code becomes. If hyperapp were 4k it wouldn't stop me from using it. If it were 0,5k and completely unintelligble -- that would stop me from using it.

Let's try to keep the discussion focus on reviewing this PR and offer suggestions on how to improve it, whether that be size-wise or speed-wise. 

The argument that

> If it were 0,5k and completely unintelligble -- that would stop me from using it.

was already kind of made by @leeoniya and I agree and I think we all agree. 👍

src/app.js
```diff
@@ -221,30 +216,128 @@ export default function (app) {
 
   function patch(parent, element, oldNode, node) {
     if (oldNode == null) {
```
zaceno: Perhaps try with `if(oldNode === undefined)` (and in the following clauses) to see if that improves performance

src/app.js
```diff
@@ -221,30 +216,128 @@ export default function (app) {
...
+      element = parent.insertBefore(createElementFrom(node), element)
 
-    } else if (node == null) {
-      batch.push(function () {
-        if (oldNode.data && oldNode.data.onRemove) {
-          oldNode.data.onRemove(element)
-        }
-        parent.removeChild(element)
-      })
     } else if (node.tag && node.tag === oldNode.tag) {
```
zaceno: See if `else if (node.tag !== undefined && ...` improves performance

leeoniya:
https://jsperf.com/undefined-vs-null-check

src/app.js
zaceno: ... yet another conditional where we could try a more declarative form in order to get more speed. Perhaps like `if (node !== undefined && node.data !== undefined) return node.data.key`

jbucaran: @zaceno So, in summary, use === or !==. Got it. Let's see.

zaceno: Yes let's :) I'm not at all sure it will make a big difference, but I'm hearing this sort of recommendation a lot of late

dodekeract: @zaceno I think it's a good practice if there is no reason to use `==`. `===` prevents a lot of unexpected issues, is faster and more predictable.

dodekeract: @zaceno I also doubt that it makes a size difference when using gzip.

src/app.js
```diff
@@ -39,7 +38,7 @@ export default function (app) {
 
     if (obj = plugin.hooks) {
       Object.keys(obj).forEach(function (key) {
-        hooks[key].push(obj[key])
+        (hooks[key] ? hooks[key] : hooks[key] = []).push(obj[key])
```
jbucaran: @lukejacksonn With this, a plugin can now export unknown hooks and they will be initialized to an empty array here. 

Then, `Router` can check if app.hooks.onRoute exists, and if so, call it.

src/router.js
```diff
@@ -29,38 +23,47 @@ export default function (options) {
...
+      }
     }
-  }
 
-  return {
-    match: match || "*",
-    params: params
+    if (app.hooks && app.hooks.onRoute) {
+      app.hooks.onRoute(model)
+    }
```
jbucaran: @lukejacksonn @traducer @dodekeract This is how I implemented onRoute. Is this what you folks were asking for and do you think the method signature should be any different?

The model already has the matched route and params, so I am not sure what else we could pass to onRoute.

Note it's not possible to pass `actions` to onRoute.

jbucaran:
Hi everyone, sorry for the huge diff, it's mostly the refactor of tests. So, instead of a huge app.test.js, now there are test files for each area that should be tests, actions.test.js, hooks.test.js, etc.

src/app.js
lukejacksonn: This suggests to me that hooks from plugins get aggregated into an array of _callbacks_ for that type of hook.. this is not specific to the `onRoute` hook but all hooks right?
 
 If that assumption is true then the router would need to do something like..

```
app.hooks.onRoute.forEach(hook => hook(model))
```
Right?

src/router.js
lukejacksonn: This is essentially what @dodekeract was suggesting but I'm not sure how much use this is without passing `actions` to the  `onRoute` function. For example.. what if we have `actions.user.fetch` that updates the model with the most recent user data from the server. A developer might want this action to be called every time `model.router.path` changes or every time `model.router.path === '/profile'` for instance.

src/app.js
jbucaran: Yes. Good catch! Fixed and pushed.

src/router.js
jbucaran: Thought I had figured it out, but no. 🤔 Tough without rewiring a lot of stuff. Let me see.

lukejacksonn: I just read your discussion with @dodekeract on slack RE function signatures for hooks.. I agree with the notion of trying to somewhat standardize them. Is the only thing stopping us passing actions to `onRoute` the fear of invoking an infinite loop?

jbucaran: The only solution I can think of is to initialize the user's options, and only then initialize the plugin.

src/app.js
```diff
@@ -38,8 +37,8 @@ export default function (app) {
     }
 
     if (obj = plugin.hooks) {
-      Object.keys(obj).forEach(function (key) {
-        hooks[key].push(obj[key])
+      Object.keys(obj).map(function (key) {
```
jbucaran: In case anyone is wondering: Using `map` here makes no important difference as opposed to using forEach and we can save more bytes.

leeoniya: this obsessive bytes-first mindset is gonna bite you.

https://jsperf.com/object-keys-vs-for-in-with-closure/114
https://jsperf.com/object-keys-vs-for-in-with-closure/137

lukejacksonn: I generally prefer to use `map` over `forEach` whenever possible.. it feels much more pure as it has a return value, which also allows for chaining and AFAIK it is [actually faster](https://ryanpcmcquen.org/javascript/2015/10/25/map-vs-foreach-vs-for.html) than `forEach` in most cases so 👍 💯

lukejacksonn: Agrees with @leeoniya sentiment.. there are more important factors to consider choosing `map` over `forEach` and vice versa besides the saving of 4 bytes

lukejacksonn:
Yeh this feels like it has turned into a release branch :boom: If so it is probably better to abstract out various features into their own branches, squash them into single commits and then cherry-pick them onto a release branch. Not sure if there is a better way but that is they way I have always seen it done!

src/app.js
jbucaran: @leeoniya I am awful at taking notes, so if I don't push this code, I'll forget about it later. But you're right, obsessions don't lead anywhere good. 👍

jbucaran:
@zaceno I know, sorry about that! I've removed the onRoute, but there are some minimal router changes.

Issue #174, opened 2017-03-30
Merge hooks and subscriptions into "events".


jbucaran:
I'd like to simplify the API a bit more without losing functionality / power or compromising speed. I think this is possible by removing subscriptions and hooks and introducing a new concept: **events**.

```jsx
app({
  events: {
/* replaces subscriptions*/
    onLoad: [ (state, actions), ... ], 
/* below are the hooks you already know */
    onAction: [ (state, action, data), ... ], 
    onUpdate: [ (oldState, newState, data), ... ],
    onRender: [ (state, actions, view), ... ],
    onError:  [ (state, error), ... ]
  }
})
```

* Newcomers have a concept less to learn and a single signature to remember when using events. 
* Easier for plugins to add new events, for example the Router and onRoute.

### Before

```jsx
app({
  subscriptions: [
    (model, actions) => actions.doSomething()
  ]
})
```

### After
```jsx
app({
  events: {
    onLoad: (state, actions) => actions.doSomething()
  }
})
```

### Before

```jsx
app({
  hooks: {
    onAction: (action, data) => console.log(action, data)
  }
})
```

### After
```jsx
app({
  events: {
    onAction: [
      (state, action, data) => console.log(action, data)
    ]
  }
})
```

zaceno:
How does the plugin concept play into this?

zaceno:
And also the talk of the onRoute hook? (Maybe a dumb question - haven't been following that very closely)

jbucaran:
@zaceno Plugins would continue to work as usual, they should be able to add new events, just like they can attach multiple hooks. This is just another name overhaul, trying to make the API smaller and neater.

If this goes through we'd only have:

* model
* actions
* view
* events
* root

dodekeract:
@jbucaran In theory `events` could be predefined actions (`actions.on.load`). Only issue would be that actions would need to be able to handle arrays then.

arturi:
[...]

PR #177, opened 2017-04-03
Introduce events. Close #173, #174.

### LINKED ISSUE #173: Rename model to state.
Please 👍 or 👎.

jbucaran:
#174


jbucaran:
-927!

src/app.js
```diff
@@ -1,115 +1,78 @@
...
 
-  load(function () {
-    root = app.root || document.body
+    init(actions, plugin.actions)
 
-    render(model, view)
-
-    subscriptions.map(function (cb) {
-      cb(model, actions, error)
+    Object.keys(plugin.events || []).map(function (key) {
```
selfup: Maybe consider using an object 😄 

```javascript
Object.keys(plugin.events || {})
```

Is using `[]` more performant than `{}`?

src/app.js
```diff
@@ -1,115 +1,78 @@
...
-  function error(error) {
-    if (hooks.onError.length === 0) {
-      throw error
-    }
+  }
 
-    hooks.onError.map(function (cb) {
-      cb(error)
-    })
+  if (document.readyState[0] !== "l") {
```
selfup: can use ternary here:

```javascript
(document.readyState[0] !== "l")
  ? load()
  : addEventListener("DOMContentLoaded", load)
```

Otherwise 👍

src/app.js
```diff
@@ -1,115 +1,78 @@
...
+    load()
+  } else {
+    addEventListener("DOMContentLoaded", load)
   }
 
-  function init(container, group, lastName) {
-    Object.keys(group).map(function (key) {
+  function init(namespace, children, lastName) {
+    Object.keys(children || []).map(function (key) {
```
selfup: `children || {}` here as well maybe?

src/app.js
```diff
@@ -1,115 +1,78 @@
...
-          })
-
-          var result = action(model, data, actions, error)
+        namespace[key] = function (data) {
+          var result = action(state, emit("action", {
+            name: name,
+            data: data
+          }).data, actions, emit)
 
           if (result == null || typeof result.then === "function") {
```
selfup: Is `==` just checking for a read only primitive?

src/app.js
```diff
@@ -1,115 +1,78 @@
...
+    emit("loaded", emit)
   }
 
-  function render(model, view) {
-    hooks.onRender.map(function (cb) {
-      view = cb(model, view)
+  function emit(name, data) {
+    (events[name] || []).map(function (cb) {
+      var result = cb(state, actions, data, emit)
+      if (result != null) {
```
selfup: Repeating my question again but is `!=` sufficient here vs. `!==`?

src/app.js
```diff
@@ -173,20 +130,15 @@ export default function (app) {
...
-        }
-
+      try {
         element[name] = value
+      } catch (_) { }
 
-        if (oldSelStart >= 0) {
-          element.setSelectionRange(oldSelStart, oldSelEnd)
+      if (typeof value !== "function") {
+        if (value) {
```
selfup: Consider ternary in nested logic gate 😄

src/router.js
```diff
@@ -1,37 +1,40 @@
 export default function (app) {
   return {
-    model: match(app.model, location.pathname),
+    state: {},
```
selfup: Is this why you were hesitant on setting a default value for `var state //= app.state || []`? 

On line **2** of `app.js`

package.json
```diff
@@ -27,9 +27,18 @@
...
+      [
+        "transform-react-jsx",
+        {
+          "pragma": "h"
+        }
+      ]
+    ]
   },
   "devDependencies": {
+    "babel-plugin-transform-react-jsx": "^6.23.0",
```
selfup: This is to test jsx? 👍

src/app.js
```diff
@@ -1,115 +1,78 @@
-var SVG_NS = "http://www.w3.org/2000/svg"
-
 export default function (app) {
-  var view = app.view || function () {
-    return ""
-  }
-
-  var model
+  var state //= app.state || {}
```
selfup: Are `state` and `view` now required I am assuming? Just wondering why the hesitation 😂

src/app.js
selfup:
```javascript
value
  ? element.setAttribute(name, value)
  : element.removeAttribute(name)
```

src/app.js
```diff
@@ -155,14 +118,8 @@ export default function (app) {
   }
 
   function setElementData(element, name, value, oldValue) {
-    name = name.toLowerCase()
-
     if (name === "key") {
```
selfup: since you are comparing strings here, can a switch/case be used here (more performant)

```javascript
switch(name.toLowerCase()) {
  // case stuff here
}
```

src/app.js
```diff
@@ -1,115 +1,78 @@
...
-      subscriptions = subscriptions.concat(obj)
-    }
+  for (var i = -1, plugins = app.plugins || []; i < plugins.length; i++) {
+    var plugin = plugins[i] ? plugins[i](app) : app
 
-    if (obj = plugin.hooks) {
-      Object.keys(obj).map(function (key) {
-        hooks[key].push(obj[key])
-      })
+    if (plugin.state != null) {
```
selfup: with single line `if`s you can remove curly braces. Unclear if how it minifies though 🤔 

```javascript
if (plugin.state != nulll) state = merge(state, plugin.state)
```

package.json
jbucaran: Yeah, kinda. This is so that I can use JSX in tests, not testing JSX per se.

src/app.js
jbucaran: Kinda. There's no default `view` anymore; it is required now. So, perhaps it follows that `state` should be required too?

src/app.js
jbucaran: Let me see if I can rewrite it using switch and it makes any difference.

src/app.js
jbucaran: Fortunately, that is taken care for us during minification.

src/router.js
jbucaran: Hehe maybe!

src/app.js
jbucaran: Yields the same amount of bytes because it gets converted to that during minification.

```
n?e.setAttribute(t,n):e.removeAttribute(t)
```

src/app.js
jbucaran: Using == null and != null later in order to save bytes. That is equivalent to

```result === null || result === undefined```

File: src/app.js (hunk 5 above)
Comment (jbucaran) [inline src/app.js]: Answered above.

File: src/app.js (hunk 1 above)
Comment (jbucaran) [inline src/app.js]: These || [] are just so that Object.keys or .map don't break. I actually just tried using {} and it yielded 2 extra bytes, perhaps it has to do with compression?

File: src/app.js (hunk 6 above)
Comment (selfup) [inline src/app.js]: Oh interesting! Welp I tried 😄

File: package.json (hunk 8 above)
Comment (selfup) [inline package.json]: Right that is what I meant. Makes sense 👍

File: src/app.js (hunk 4 above)
Comment (selfup) [inline src/app.js]: ahh ok makes sense!

File: src/app.js (hunk 1 above)
Comment (selfup) [inline src/app.js]: Yea odd. Keep what saves the bytes!

File: src/app.js (hunk 9 above)
Comment (selfup) [inline src/app.js]: Yes this makes sense.

File: src/app.js [hunk 12]
```diff
@@ -1,115 +1,78 @@
-var SVG_NS = "http://www.w3.org/2000/svg"
-
 export default function (app) {
-  var view = app.view || function () {
-    return ""
-  }
-
-  var model
+  var state = {}
```
selfup: should we do: `var state = app.state || {}` or is initializing with `{}` ok from the get go?

src/app.js
selfup: Yep I thought so

selfup:
Thanks for taking the time to answer each comment! ❤️

src/app.js
selfup: Let me know how it goes. Theoretically it *should* be more performant

src/app.js
jbucaran: In this case it's fine because the first plugin from which we'll try merge the local state variable with is actually the options object passed into app. We'll have an annotated hyperapp someday which is better than comments!

src/app.js
jbucaran: Haha 👍

src/app.js
jbucaran: Yeah, sure, it's a matter of style. Since uglify converts it to that anyway, I'd rather keep the `if` which is more familiar (readable?) than using the ternary op.

jbucaran:
Hi hypernauts! @dodekeract @zaceno @ngryman @pedroborges @selfup 🙋‍♂️

I think this PR is basically done. I'd like to refactor/add more tests to patch.test.js and vdom.test.js, but other than that, it's pretty much this. If you have any comments or questions, please go over the code changes and leave your feedback 🙏.

Special thanks to Pedro for suggesting to use an emitter to implement ~hooks~ **events** and all his work on #178. 👏

src/app.js
selfup: Since this has heavy traffic in terms of ops, I would be really interested in seeing performance benchmarks with `switch` vs.`if` here 🎉

Issue #144, opened 2017-03-08
Add hook to router


pcariel:
hi, guys. Is it possible to add some type of validation in the route before activating the view?

OnActivate similar to angular2.

dodekeract:
Could be done with a plugin that hooks into the `onRender` hook.

jbucaran:
@pcariel https://github.com/hyperapp/hyperapp/wiki/api#onrender :)

dodekeract:
@jbucaran https://github.com/hyperapp/hyperapp/wiki/API#onrender

lukejacksonn:
This is quite relevant https://github.com/hyperapp/hyperapp/pull/150.. just looking now at how it could be abstracted out to a plugin or how it can be integrated without having to alter the existing router 👍

jbucaran:
@pcariel Epa Pablo! The router now exposes an event, `route` that may be exactly what you need. I am going to close here, but feel free to reopen or ask other questions! 👋

PR #178, opened 2017-04-04
Implement event/hook emitter

pedroborges:
This PR relates to #177, #174, #150, #144.

After an interesting discussing on Slack, I've modified the emitter implementation. Below is a detailed explanation of what have changed since my [last comment](https://github.com/hyperapp/hyperapp/issues/174#issuecomment-291198564) on issue #174.

From the Slack discussion it was clear that `subscriptions` is the best name to group both `events` and `hooks` since @jbucaran wants to avoid having both. The following is a list of all events and hooks emitted by the core of Hyperapp:

```js
subscriptions: {
    '*': (type) => {},                   // Listens to all events/hooks
    'actions': (name, data) => {},       // When any action is called
    'action:save': (data) => {},         // When a specific action is called
    'update': (name) => {},              // When model updates
    'load': (model, actions) => {        // When the core loads right before the first render
        if (model.contacts.length < 1) {
            actions.contacts.addRandom()
        }
    },
    'render': (model, view) => {         // When rendering
        // do something and
        // return a modified view
        return view
    }
}
```

I have integrated [Mitt](https://github.com/developit/mitt) in Hyperapp storing all handlers on the `subscription` object. Mitt methods `on` and `off` have been removed. The only way to register event/hook handlers now is via the `subscriptions` object when initialising the app/plugin.

Also, Mitt is no longer exposed to the app. Instead only actions will receive an `emit` object allowing them to trigger an `event` and/or `hook`:

- **`emit.event` -** is a way to notify that something happened **without** expecting any value to be returned.

```js
actions: {
  router: {
    match: match,
    go: function (_, data, actions, emit) {
      history.pushState({}, "", data)
      actions.router.match(data)
      emit.event('route:update', data)
    }
  }
}
```

- **`emit.hook` -** **expects** a value to be returned allowing the app/plugin to act based on it. Under the hood this method uses `Array.reduce` to store the value returned since more than one handler may exist for a given hook. The value that will be returned from the hook is stored on the **last argument** allowing the app/plugin to change or replace this value.

```js
function render(model, view) {
  view = emit.hook("render", model, view)

  element = patch(root, element, node, node = view(model, actions))
}
```

In this example from the core, any handler that subscribes to the `render` hook will be able to change or completely replace the `view`, which will be returned since it's the last argument. The router subscribes to the `render` hook:

```js
subscriptions: {
  "render": function (model, view) {
    return view[model.router.match]
  }
}
```

`onLoad` is now `load` and must defined inside `subscribers`. It also no longer receives the `emitter` as I first proposed.

`onRender` is now `render` for short.

These changes maintain the flexibility brought by an event emitter while limiting its use to avoid undesired side effects.

Prefixing core events/hooks and encouraging plugin authors to do the same is a good idea, keep that mind.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -5,9 +5,10 @@ export default function (app) {
     return ""
   }
 
-  var emitter = emitter()
+  var emit = {event: event, hook: hook}
   var model
   var actions = {}
+  var subscriptions = {}
 
   var node
   var root
@@ -26,17 +27,17 @@ export default function (app) {
       init(actions, obj)
     }
 
-    emitter.on('render', plugin.beforeRender)
-
-    emitter.on('load', plugin.onLoad)
+    Object.keys(plugin.subscriptions || {}).forEach(function (type) {
+      (subscriptions[type] || (subscriptions[type] = [])).push(plugin.subscriptions[type])
+    })
   }
 
   load(function () {
     root = app.root || document.body
 
-    render(model, view)
+    emit.event("load", model, actions)
 
-    emitter.emit('load', model, actions, emitter)
+    render(model, view)
   })
 
   function init(container, group, lastName) {
@@ -46,16 +47,16 @@ export default function (app) {
 
       if (typeof action === "function") {
         container[key] = function (data) {
-          emitter.emit('action:'+name, data)
-          emitter.emit('actions', name, data)
+          emit.event("action:"+name, data)
+          emit.event("action", name, data)
 
-          var result = action(model, data, actions, emitter)
+          var result = action(model, data, actions, emit)
 
           if (result == null || typeof result.then === "function") {
             return result
 
           } else {
-            emitter.emit('update', model, result, data)
+            emit.event("update", model, result, data)
 
             model = merge(model, result)
             render(model, view)
@@ -78,7 +79,7 @@ export default function (app) {
   }
 
   function render(model, view) {
-    view = emitter.stack('render', model, view)
+    view = emit.hook("render", model, view)
 
     element = patch(root, element, node, node = view(model, actions))
   }
@@ -302,33 +303,17 @@ export default function (app) {
     return element
   }
 
-  function emitter(all) {
-    all = all || Object.create(null)
+  function event(type) {
+    var args = Array.prototype.slice.call(arguments, 1);
+    (subscriptions[type] || []).forEach(function(cb) { cb.apply(null, args) });
+    (subscriptions["*"] || []).forEach(function(cb) { cb.apply(null, type) });
+  }
 
-    return {
-      on: function(type, handler) {
-        (all[type] || (all[type] = [])).push(handler)
-      },
-
-      off: function(type, handler) {
-        var e = all[type] || (all[type] = [])
-        e.splice(e.indexOf(handler) >>> 0, 1)
-      },
-
-      emit: function(type) {
-        var args = Array.prototype.slice.call(arguments, 1)
-        var handler = function(cb) { cb.apply(null, args) }
-        ;(all[type] || []).map(handler)
-        ;(all['*'] || []).map(handler)
-      },
-
-      stack: function(type) {
-        var args = Array.prototype.slice.call(arguments, 1);
-        return (all[type] || []).reduce(function(result, handler) {
-          return handler.apply(null, args.concat(result))
-        }, args.pop());
-      }
-    }
+  function hook(type) {
+    var args = Array.prototype.slice.call(arguments, 1)
+    return (subscriptions[type] || []).reduce(function(result, handler) {
+      return handler.apply(null, args.concat(result))
+    }, args.pop())
   }
 }
 
--- a/src/app.js
+++ b/src/app.js
@@ -26,13 +26,9 @@ export default function (app) {
       init(actions, obj)
     }
 
-    if (obj = plugin.beforeRender) {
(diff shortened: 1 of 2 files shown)


danigb:
Wow, that's a good explanation 👍

src/app.js
```diff
@@ -330,5 +302,18 @@ export default function (app) {
 
     return element
   }
+
+  function event(type) {
+    var args = Array.prototype.slice.call(arguments, 1);
+    (subscriptions[type] || []).forEach(function(cb) { cb.apply(null, args) });
+    (subscriptions["*"] || []).forEach(function(cb) { cb.apply(null, type) });
```
jbucaran: @pedroborges Can we get rid of "*"?

src/router.js
```diff
@@ -10,18 +10,16 @@ export default function (app) {
...
     },
-    hooks: {
-      onRender: function (model) {
-        return app.view[model.router.match]
-      }
-    },
-    subscriptions: [
-      function (_, actions) {
+    subscriptions: {
+      "render": function (model, view) {
```
jbucaran: @pedroborges Why is this a string?

src/app.js
pedroborges: At one point I removed this, then I thought about the Chrome Dev tools [some](https://chrome.google.com/webstore/detail/vuejs-devtools/nhdogjmejiglipccpnnnanhbledajbpd) [frameworks](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi) have to inspect apps written with them. If we get an tool like that this will be useful. Can always add it later anyways.

src/router.js
pedroborges: Most for consistency with the others. I suggest that we namespace this names, like: `core:render`, `route:enter`, and so on.

lukejacksonn: I presume this is just for consistency, given that a common syntax for name spacing events is `namespace:event` which is an invalid key when not a string. Could be wrong though..

lukejacksonn:
Really great explanation indeed.. I will review the actual code as soon as I get a free moment. Much of what you are saying _a lot_ of sense! 🤓

src/app.js
pedroborges: To be frankly even the whole `event` function could be removed. I just left it there because the `hook` function uses `Array.reduce` which I bet is less performant than `Array.forEach`.

jbucaran: 🤔 Sorry, I still don't understand what "*" is for. @pedroborges @lukejacksonn

src/app.js
```diff
@@ -5,15 +5,10 @@ export default function (app) {
     return ""
   }
 
+  var emit = {event: event, hook: hook}
```
jbucaran: Can we make this a function?

src/app.js
pedroborges: That's a catch all for events. If you register a listener to `*` it will be called whenever an event or hook is called.

jbucaran: Thanks, is this necessary?

pedroborges: It was part of Mitt, as I said above. If we get a Chrome Dev tool for Hyperapp in the future it will be useful. Since that's a maybe, we can remove it for now and add when needed.

jbucaran: Ah, gotcha.

jbucaran: Actually, I lost you again 😆. Can't `onAction` be used for that?

pedroborges: `*` would catch everything: `action`, `update`, `render`, `load`, and whatever plugins emit too.

src/app.js
pedroborges: If we can find a way to consolidate functions that returns a value (for hooks) and function that don't return a value (event) the yes.

I'll can give it a try, right now I can't think of how to solve this.

jbucaran:
@pedroborges I don't think it's a good idea to pass the emitter to actions. It will confuse users or open the door to interact with the application with something other than actions.

src/app.js
pedroborges: This works, but it longer. You may help me get it shorter…

```js
function hook(type) {
  var args = Array.prototype.slice.call(arguments, 1);
  var result = args.pop();

  (subscriptions[type] || []).forEach(function(handler) {
    var val = handler.apply(null, args.concat(result))
    result = val !== undefined ? val : result
  })

  return result
}
```

Since hooks depend on the returned value and events don't, we need to check if the handler returns a value.

pedroborges: If we can fit in one function, a few name ideas for that one function:

- emit
- fire
- stream
- broadcast

jbucaran: What about just `on`?

pedroborges: `on` would be for registering events, we are no longer exposing an API for registering them. The current implementation registers everything in the `subscriptions` method.

pedroborges:
@jbucaran `error` is already the forth argument on the action. It does basically the same thing, it send a message to an event handler, except it's limited for just errors.

I thought of adding it to actions as a way for plugins to talk to the app. The router examples I showed on #174 explains what the benefits are.

Do you have any other idea on how to solve cases like that?

jbucaran:
@pedroborges Plugins can talk to the app by adding new actions, e.g. Router.

jbucaran:
@pedroborges Thanks for all the good work man! The emitter made it to the finals and won after all.

PR #181, opened 2017-04-14
Add size badge

ngryman:
Hey!

A small auto-promo here: https://github.com/ngryman/badge-size 🎉

I'm proposing to add a size badge to advertise (and keep track) of the gzip minified size of `hyperapp`: [Size](https://unpkg.com/hyperapp/dist/hyperapp.js)

I fetches the compiled file via `unpkg` and computes its gziped size.


jbucaran:
Hey @ngryman! Thanks, but I'll pass on this one. You'd think that with all the byte saving taking place in this neck of the woods this is the badge I'd be looking for; and it is, but I feel the README would look rather bloated with it, so nope. 🙏

ngryman:
@jbucaran No problem, I get it.

lukejacksonn:
Great plugin though @ngryman!

PR #184, opened 2017-04-14
Not mandatory to concat positional Number/String nodes.

jbucaran:
@ngryman @selfup @dodekeract 

The idea here is to remove the code that concatenates contiguous Number/String nodes. 

```jsx
<div>
    {"string"}
    {"or"}
    {1}
    {"another string"}
</div>
```

Before, h would generate a node with a single child  `"stringor1another string"`, now it generates a node with 4 children, one for each string or number. This is not the same as generating 4 nodes, that doesn't happen.

It may seem like this was a good thing to have, but:

* Such scenario, like in the markup posted above is unlikely.
* Benchmark results were similar (slightly faster without the code actually).
* We can shave more bytes off the bundle for no cost.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -105,11 +105,11 @@ export default function (app) {
         element.appendChild(createElementFrom(node.children[i++], isSVG))
       }
 
-      for (var name in node.data) {
-        if (name === "onCreate") {
-          node.data[name](element)
+      for (var i in node.data) {
+        if (i === "onCreate") {
+          node.data[i](element)
         } else {
-          setElementData(element, name, node.data[name])
+          setElementData(element, i, node.data[i])
         }
       }
     }
@@ -277,4 +277,3 @@ export default function (app) {
     return element
   }
 }
-
--- a/src/h.js
+++ b/src/h.js
@@ -1,37 +1,22 @@
 export default function (tag, data) {
   var node
-  var canConcat
-  var oldCanConcat
-
   var stack = []
   var children = []
 
   for (var i = arguments.length; i-- > 2;) {
-    stack.push(arguments[i])
+    stack[stack.length] = arguments[i]
   }
 
   while (stack.length) {
     if (Array.isArray(node = stack.pop())) {
-      i = node.length
-
-      while (i--) {
-        stack.push(node[i])
+      for (var i = node.length; i--;) {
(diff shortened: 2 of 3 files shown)


ngryman:
The PR seems fine for me. But I still don't know the consequences it has when diffing and patching with this.

ngryman:
In my report I wrongly believed that concatenation needed a stack, which is the true perf killer here.
But you explained me that it's because of `hyperx`, so I let you decide if this PR is worth or not 👑

jbucaran:
So far it looks and works good. I'll do some more tests tomorrow and triple check we're not breaking anything.

jbucaran:
Okay I was able to get the bundle down to 1677, best was 1641 but perf would decrease. With this perf actually increased a bit.

src/app.js
```diff
@@ -105,11 +105,11 @@ export default function (app) {
         element.appendChild(createElementFrom(node.children[i++], isSVG))
       }
 
-      for (var name in node.data) {
-        if (name === "onCreate") {
-          node.data[name](element)
+      for (var i in node.data) {
```
jbucaran: Reuse `i`.

src/h.js
```diff
@@ -1,37 +1,22 @@
...
   var node
-  var canConcat
-  var oldCanConcat
-
   var stack = []
   var children = []
 
   for (var i = arguments.length; i-- > 2;) {
-    stack.push(arguments[i])
+    stack[stack.length] = arguments[i]
```
jbucaran: `array[array.length] = value` beats `array.push(value)`, but it's more bytes. This time I choose speed.

test/h.test.js
```diff
@@ -28,29 +28,21 @@ test("vnode with a single child", () => {
...
+    children: ["foo", "bar", "baz"]
   })
 
   expect(
     h("div", {}, 1, "foo", 2, "baz", 3)
   ).toEqual({
     tag: "div",
     data: {},
-    children: ["1foo2baz3"]
+    children: ["1", "foo", "2", "baz", "3"]
```
jbucaran: `h` transforms Numbers into String types.

selfup:
Looks great! Glad it was merged 🎉

PR #186, opened 2017-04-15
Add bench and trace scripts for h

ngryman:
I'm adding a quick way to evaluate how `h` is doing, so we have some material for our discussions.

- `yarn bench` will start some benchmarks (you are welcomed to add more cases).
- `yarn trace` will trace the code and generate artifacts that can be used with http://mrale.ph/irhydra/2.

I'm going to switch to the official repo but I started this branch on my fork.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 # IDE
 .idea/
 .vscode/
+code.asm
+hydrogen.cfg
 
 # Logs
 *.log*
--- a/package.json
+++ b/package.json
@@ -23,6 +23,8 @@
   "scripts": {
     "test": "jest --coverage --no-cache",
     "build": "rollup -cm -n hyperapp -f umd -i src/index.js -o dist/hyperapp.js",
+    "bench": "rollup perf/h.js -f cjs | node",
+    "trace": "rollup perf/h.js -f cjs | node-irhydra",
     "prepublish": "npm run build",
     "release": "npm run build && npm test && git commit -am $npm_package_version && git tag $npm_package_version && git push && git push --tags && npm publish"
   },
@@ -32,6 +34,8 @@
   "devDependencies": {
     "babel-preset-es2015": "^6.22.0",
     "jest": "^19.0.2",
+    "nanobench": "^2.1.0",
+    "node-irhydra": "^1.0.0",
     "rollup": "^0.41.4",
     "rollup-plugin-uglify": "^1.0.1"
   }
--- a/perf/h.js
+++ b/perf/h.js
@@ -1,0 +1,18 @@
+import bench from 'nanobench'
+import h from '../src/h'
+
+bench("vnode with no child 1M times", b => {
+  b.start()
+  for (let i = 0; i < 10000000; i++) {
+    h("div", {})
+  }
+  b.end()
+})
+
+bench("vnode with a single child 1M times", b => {
+  b.start()
+  for (let i = 0; i < 10000000; i++) {
+    h("div", {}, ["foo"])
+  }
+  b.end()
+})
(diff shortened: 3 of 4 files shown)


jbucaran:
Looks interesting. I'm already using [js-framework-benchmark](https://github.com/krausest/js-framework-benchmark) to benchmark h/app, but I'll keep an open mind to this too. 

I'll like to take some time to study these tools, their philosophies and how they work before I merge this. In the meantime, let's keep this PR open and hopefully get more feedback from other people.

@leeoniya What do you think?

ngryman:
@jbucaran Alright, it's a first shot tho. The aim here is to be able to bench specific parts of `hyperapp` in order to focus on important optimizations.

Three tools would be included:
* **benchmarks**: which put digits on use cases and help to keep track of the overall performance.
* **profiling** (todo): which allow us to spot where exactly we are slow and fix it.
* **tracing**: which allow us to know if hot parts of the code are well optimized by the `vm`.

leeoniya:
irhydra2 should tell you everything you need to know about perf. i'm not familiar with it and the artifacts output could vary significantly between commits simply because they're diffetent but not necessarily faster or slower. so the trace output may need to be hand analyzed anyhow. it may be possible to count the number of deoptimizations as a rough regresssion indicator but it may not be very useful. for sure irhydra2 will help with investigating slowness.

as for benchmarks, i think it makes sense to do benches without dom mutation just to track framework overhead. i do this manually by tracking ms taken to do a 0% (no change) dbmonster redraw, which does a full vtree rebuild/normalization and diff/patch/reconcile pass.

https://github.com/leeoniya/domvm/blob/2.x-dev/demos/bench/dbmonster/app.js

ngryman:
@leeoniya Yup i'm using `irhydra2` and the awesome resources provided by @mraleph on the subject 😍

I'm still in a learning phase on the subject, but my workflow to hunt slowness is starting to get a shape:
 - Create a bench (like a unit test) and have reference numbers.
 - Profile the code to see where we lose time. Then optimize things, change of data structure, etc... Do back and forth between profiling and benching. Try to get numbers low.
 - And finally, when satisfied, I enter the nerdy stuff and try to hunt for deopts. After some time, you know what principles to follow: stay monomorphic, take advantage of hidden classes, and avoid things like out-of-array bounds accesses, plus of course everything mentioned here:
   - https://github.com/petkaantonov/bluebird/wiki/Optimization-killers
   - http://mrale.ph/blog/2011/12/18/v8-optimization-checklist.html
   - https://www.html5rocks.com/en/tutorials/speed/v8/

leeoniya:
FWIW in domvm the major perf boosts didnt come from micro-optimizations. some stuff i had to do:

- dont mutate vnodes across function boundaries
- reduce number of passes over the vtrees, do as much as possible in 1 pass
- minimize GC pressure by reducing intentional and accidental object/array/string allocations whenever possible
- sometimes DRY is good for bytes but bad for perf since the JIT cannot optimize one function for a diverse set of args/signatures.
- also, sometimes minifiers that inline functions (like Closure) can reduce perf by bloating their callers and deopting them in the process. banged my head against this recently during what should have been an optimization commit but regresseed after running through min.

ngryman:
@lukejacksonn @jbucaran What is the priority on this one?
I'd like to wait early august, so that I free up so time to release SR 0.3.0 which will allow to do pretty much everything we want and run some regression tests on CI.

lukejacksonn:
Nothing urgent @ngryman just a nice to have.. I look forward to SR 0.3.0 🏎 🎉

lukejacksonn:
@ngryman perhaps we close this PR until then?

ngryman:
Sure. I'll refer to this one when shipping fresh code.

Issue #185, opened 2017-04-14
Don't process nested children


ngryman:
Let's continue on this topic here.

So, from what I see in `hyperxify` [examples](https://github.com/substack/hyperxify/blob/master/example/vdom.js), they are using [virtual-dom](https://github.com/Matt-Esch/virtual-dom). `virtual-dom` does not seem to handle nested children: https://github.com/Matt-Esch/virtual-dom/blob/master/virtual-hyperscript/index.js.

My guess is that those nested children are only text nodes, but when a real node is met, a `h` call is made. Given the fact that `virtual-dom` does not handle this, I'm pretty confident we could get rid of this.

@jbucaran How could we validate that it's not mandatory?

jbucaran:
> @ngryman How could we validate that it's not mandatory?

You mean other than changing the code and run a few hyperx examples and see if everything still works?

jbucaran:
Hyperx generates nested children which are arrays and sometimes may contain further nested arrays, it's why h is a mess.

ngryman:
@jbucaran Ok, do you have an example of this code? Thanks!

ngryman:
Linking the potential implementation: https://github.com/hyperapp/hyperapp/issues/183#issuecomment-294284671

ngryman:
Ok, I've generated an output from `hyperxify` taking their example.
From what I see we don't have to manage nested children.

```js
function render(state) {
  return html`<div>
    <h1>clicked ${state.times} times</h1>
    <button onclick=${onclick}>click me!</button>
  </div>`
}
```

```js
function render(state) {
  return h("div", {}, [
    "\n",
    h("h1", {}, [
      "clicked ",
      state.times,
      " times"
    ]),
    "\n",
    h("button", { "onclick": onclick }, [
      "click me!"
    ]),
    "\n"
  ])
}
```

ngryman:
Next potential implementation:
```js
```

jbucaran:
@ngryman That didn't pass the tests.

jbucaran:
Closing in favor of your other issue #183. 

> Ok, I've generated an output from hyperxify taking their example.
> From what I see we don't have to manage nested children.

Try with this:

```js
html`
  <div class="someClass">
    <ul>
      ${[1, 2, 3, 4].map(i => html`
      <li>
        ${i}:
        <ul>${[9, 8, 7, 6].map(i => html`
          <li>
            ${i}
          </li>`)}
        </ul>
      </li>
[...]

PR #187, opened 2017-04-15
Optimize h

ngryman:
Let's start from this implementation which passes the current tests.

Just to have an idea, here are the benchs:

**Before**
```sh
NANOBENCH version 2
> node

# vnode with no child 1M times
ok ~315 ms (0 s + 315373353 ns)

# vnode with a single child 1M times
ok ~1.35 s (1 s + 353013098 ns)

# vnode with a single array child 1M times
ok ~1.66 s (1 s + 657042054 ns)

all benchmarks completed
ok ~3.33 s (3 s + 325428505 ns)
```

**After**
```sh
NANOBENCH version 2
> node

# vnode with no child 1M times
ok ~287 ms (0 s + 287287753 ns)

# vnode with a single child 1M times
ok ~692 ms (0 s + 692102600 ns)

# vnode with a single array child 1M times
ok ~822 ms (0 s + 822467710 ns)

all benchmarks completed
ok ~1.8 s (1 s + 801858063 ns)
```

### DIFF
--- a/src/h.js
+++ b/src/h.js
@@ -4,7 +4,7 @@ function addChild(children, val) {
       children.push(val[j])
     }
   }
-  else if (val && val !== true) {
+  else if (val != null && val !== true && val !== false) {
     if (typeof val === 'number') {
       val = val + ''
     }
--- a/src/h.js
+++ b/src/h.js
@@ -5,10 +5,7 @@ function addChild(children, val) {
     }
   }
   else if (val != null && val !== true && val !== false) {
-    if (typeof val === 'number') {
-      val = val + ''
-    }
-    children.push(val)
+    children.push(typeof val === 'number' ? val + '' : val)
   }
 }
 
@@ -21,11 +18,9 @@ export default function(tag, data, values) {
     }
   }
   else {
-    addChild(children, values)
-  }
-
-  for (var i = 3; i < arguments.length; i++) {
-    addChild(children, arguments[i])
+    for (var i = 2; i < arguments.length; i++) {
+      addChild(children, arguments[i])
+    }
   }
 
   return typeof tag === "string"
--- a/src/h.js
+++ b/src/h.js
@@ -1,10 +1,12 @@
 function addChild(children, val) {
+  if (val == null || typeof val === 'boolean') return
+
   if (Array.isArray(val)) {
     for (var j = 0; j < val.length; j++) {
       children.push(val[j])
     }
   }
-  else if (val != null && val !== true && val !== false) {
+  else {
     children.push(typeof val === 'number' ? val + '' : val)
   }
 }
@@ -12,14 +14,16 @@ function addChild(children, val) {
 export default function(tag, data, values) {
   var children = []
 
-  if (Array.isArray(values)) {
-    for (var i = 0; i < values.length; i++) {
-      addChild(children, values[i])
+  if (arguments.length > 2) {
+    if (Array.isArray(values)) {
+      for (var i = 0; i < values.length; i++) {
+        addChild(children, values[i])
+      }
     }
-  }
-  else {
-    for (var i = 2; i < arguments.length; i++) {
-      addChild(children, arguments[i])
+    else {
+      for (var i = 2; i < arguments.length; i++) {
+        addChild(children, arguments[i])
+      }
     }
   }
 
--- a/src/h.js
+++ b/src/h.js
@@ -1,19 +1,31 @@
-export default function(tag, data) {
+function addChild(children, val) {
+  if (Array.isArray(val)) {
+    for (var j = 0; j < val.length; j++) {
+      children.push(val[j])
+    }
+  }
+  else if (val && val !== true) {
+    if (typeof val === 'number') {
+      val = val + ''
+    }
+    children.push(val)
+  }
+}
+
+export default function(tag, data, values) {
   var children = []
 
-  for (var i = 2; i < arguments.length; i++) {
-    var val = arguments[i]
-    if (Array.isArray(val)) {
-      for (var j = 0; j < val.length; j++) {
-        children.push(val[j])
-      }
+  if (Array.isArray(values)) {
+    for (var i = 0; i < values.length; i++) {
+      addChild(children, values[i])
     }
-    else if (val && val !== true) {
-      if (typeof val === 'number') {
-        val = val + ''
-      }
-      children.push(val)
-    }
+  }
+  else {
+    addChild(children, values)
+  }
+
+  for (var i = 3; i < arguments.length; i++) {
+    addChild(children, arguments[i])
   }
 
   return typeof tag === "string"
--- a/src/h.js
+++ b/src/h.js
@@ -1,22 +1,18 @@
 export default function(tag, data) {
-  var node
-  var stack = []
   var children = []
 
-  for (var i = arguments.length; i-- > 2; ) {
-    stack[stack.length] = arguments[i]
-  }
-
-  while (stack.length) {
-    if (Array.isArray((node = stack.pop()))) {
-      for (var i = node.length; i--; ) {
-        stack[stack.length] = node[i]
+  for (var i = 2; i < arguments.length; i++) {
+    var val = arguments[i]
+    if (Array.isArray(val)) {
+      for (var j = 0; j < val.length; j++) {
+        children.push(val[j])
       }
-    } else if (node != null && node !== true && node !== false) {
-      if (typeof node === "number") {
-        node = node + ""
+    }
+    else if (val && val !== true) {
+      if (typeof val === 'number') {
+        val = val + ''
       }
-      children[children.length] = node
+      children.push(val)
     }
   }
 
--- a/test/h.test.js
+++ b/test/h.test.js
@@ -15,6 +15,12 @@ test("vnode with a single child", () => {
     children: ["foo"]
   })
 
+  expect(h("div", {}, [["foo"]])).toEqual({
+    tag: "div",
+    data: {},
+    children: ["foo"]
+  })
+
   expect(h("div", {}, "foo")).toEqual({
     tag: "div",
     data: {},


src/h.js
```diff
@@ -1,22 +1,18 @@
...
-
-  while (stack.length) {
-    if (Array.isArray((node = stack.pop()))) {
-      for (var i = node.length; i--; ) {
-        stack[stack.length] = node[i]
+  for (var i = 2; i < arguments.length; i++) {
+    var val = arguments[i]
+    if (Array.isArray(val)) {
+      for (var j = 0; j < val.length; j++) {
+        children.push(val[j])
```
jbucaran: @ngryman `val[j]` can be an array, as we previously discussed on #185.

jbucaran:
@ngryman Good start! Remember that when using hyperx, children can be arrays that contain nested arrays too.

jbucaran:
Here is the implementation that I was able to find that works with both hyperx and jsx and drops the stack:

```js
export default function (tag, data) {
  var children = []

  for (var i = 2; i < arguments.length;) {
    push(arguments[i++], children)
  }

  return typeof tag === "string"
    ? {
      tag: tag,
      data: data || {},
      children: children
    }
    : tag(data, children)

  function push(node) {
    if (Array.isArray(node)) {
      for (var i = 0; i < node.length; i++) {
        push(node[i], children)
      }
    } else if (node != null && node !== true && node !== false) {
      children.push(node)
    }
  }
}
```

It's small, but performance was bad, and I couldn't optimize it. I hope it gives you some ideas.

src/h.js
```diff
@@ -1,22 +1,18 @@
...
       }
-    } else if (node != null && node !== true && node !== false) {
-      if (typeof node === "number") {
-        node = node + ""
+    }
+    else if (val && val !== true) {
```
jbucaran: @ngryman You want:

```js
} else if (val != null && val !== true && val !== false) {
```

Otherwise we will not be able to process text nodes with the number 0.

ngryman:
@jbucaran Coverage drops because no tests expose the way `hyperx` pass children as 3rd parameter and not as splats.

ngryman:
New benchs, code is being deopt, working on it.

```sh
```

ngryman:
New benchs, I can't really do better I think:
```js
NANOBENCH version 2
> node

# vnode with no child 1M times
ok ~277 ms (0 s + 277363055 ns)

# vnode with a single child 1M times
ok ~905 ms (0 s + 905257714 ns)

# vnode with a single array child 1M times
ok ~934 ms (0 s + 934230712 ns)

all benchmarks completed
ok ~2.12 s (2 s + 116851481 ns)
```

But I'm gonna tease you @jbucaran, if we enforce every node to be the exact same object (even for text nodes), here is what we get:
```sh
NANOBENCH version 2
> node

# vnode with no child 1M times
ok ~329 ms (0 s + 328797485 ns)

# vnode with a single child 1M times
ok ~547 ms (0 s + 546626621 ns)

# vnode with a single array child 1M times
ok ~552 ms (0 s + 552139589 ns)

all benchmarks completed
ok ~1.43 s (1 s + 427563695 ns)
```

I could explain further why such a big boost, but basically the current vdom implementation mixes different types of nodes, it can be `string`s or `object`. If we *statically type* the node object by creating a constructor for it and we make sure that the tree only manipulates instances of this constructor, then `v8` and friends can optimize this heavily because it always manipulates data of the same shape. Hence the boost.

src/h.js
```diff
@@ -1,20 +1,32 @@
-export default function(tag, data) {
+function addChild(children, val) {
+  if (Array.isArray(val)) {
+    for (var j = 0; j < val.length; j++) {
+      children.push(val[j])
+    }
+  }
+  else if (val && val !== true) {
```
jbucaran: I left you a comment on the other diff about this.

Basically, you are forgetting about the number 0.

```js
} else if (val != null && val !== true && val !== false) {
```

jbucaran: @ngryman See above.

jbucaran:
> @ngryman Coverage drops because no tests expose the way hyperx pass children as 3rd parameter and not as splats.

Then we need to add those tests :)

src/h.js
```diff
@@ -4,7 +4,7 @@ function addChild(children, val) {
       children.push(val[j])
     }
   }
-  else if (val && val !== true) {
+  else if (val != null && val !== true && val !== false) {
```
jbucaran: 👍

jbucaran:
@ngryman Your final implementation is very similar to the recursive one I created, but bigger. Performance is better than that one though! 👍 

However, this version is not faster than the current one on master either.

jbucaran:
> @ngryman if we enforce every node to be the exact same object (even for text nodes), here is what we get...

Sounds good, but that does not seem to be what this PR is addressing.

ngryman:
> Then we need to add those tests :)

They have been added, 100% is back :)

> @ngryman Your final implementation is very similar to the recursive one I created, but bigger.

Because it treats the 2 functions signatures in an optimal way:
 - `h(tag, data, children)`
 - `h(tag, data, ...children)`

`arguments` object must be used very carefully if you don't want to kill perfs. I also added some `if`s to avoid to call `Array.isArray` with `null` values which deoptimize the code.

> Now, this version is not really faster than the current one on master (mine wasn't either which is why I didn't merge it).

 - `master` version: `ok ~3.34 s (3 s + 338336241 ns)`
 - `h-optimized` version: `ok ~2.12 s (2 s + 116851481 ns)`

It depends how you bench it. What is your bench I'm curious?

> Sounds good, but that does not seem to be what this PR is addressing no? Sorry, I am terrible at discussing two different things at the same time.

Yes it's just the next step. I also give too much information lately instead of giving you a concise report. It's because it's a complex subject also.

src/h.js
ngryman: We're reviewing an old commit. It has been addressed since then.

jbucaran:
@ngryman I use https://github.com/krausest/js-framework-benchmark.

src/h.js
```diff
@@ -1,22 +1,29 @@
...
-      for (var i = node.length; i--; ) {
-        stack[stack.length] = node[i]
+export default function(tag, data, values) {
+  var children = []
+
+  if (arguments.length > 2) {
+    if (Array.isArray(values)) {
+      for (var i = 0; i < values.length; i++) {
+        addChild(children, values[i])
       }
```
jbucaran: @ngryman What's the point of this loop here? I replaced it with just `addChild(children, values)` and everything works too.

jbucaran:
> @ngryman Because it treats the 2 functions signatures in an optimal way:

`tag` can be either a String or a Function, no? Doesn't that defeat what V8?

jbucaran:
Is it necessary to repeat the `for` loop 3 times? What about this implementation that does not change bundle size?

```js
export default function(tag, data, values) {
  var children = []

  if (arguments.length > 2) {
    if (Array.isArray(values)) {
      loop(children, values, 0)
    } else {
      loop(children, arguments, 2)
    }
  }

  return typeof tag === "string"
    ? {
        tag: tag,
        data: data || {},
        children: children
      }
    : tag(data, children)
}

function loop(children, child, i) {
  for (; i < child.length; ) {
    addChild(children, child[i++])
  }
}

function addChild(children, child) {
  if (child != null && child !== true && child !== false) {
    if (Array.isArray(child)) {
      loop(children, child, 0)
    } else {
      children.push(typeof child === "number" ? child + "" : child)
    }
  }
}
```

ngryman:
@jbucaran Yes, like @leeoniya says, sometimes we have to trade bytes for performance.

Your proposed solution gives this:
```js
```

ngryman:
I'm not saying that there is not a better implementation than mine though, not at all 😆

ngryman:
@jbucaran btw, do you test your bundle size gziped? `gzip` logically handles redundancies, multiple for loops should be *grasped* by `gzip` somehow.

ngryman:
Your version give a gziped bundle of `1702B`, mine `1710B`.

jbucaran:
Bummer. I just ran it through the benchmark suite and it was definitely slower. 👍 

I'm still not getting any perf gains from yours against the one on master though, so I am not compelled to merge this one, at least not until other priorities are sorted out like improving the patch algorithm.

ngryman:
> Bummer. I just ran it through the benchmark suite and it was definitely slower. 👍

I'm curious how this version could be slower than the one on `master`, it does not make sense.
Well, let's define what are the priorities then 😆

jbucaran:
@ngryman Read my comment again.

jbucaran:
Your version is certainly not slower than the current one on master, but it's not faster either. I'm using js-framework-benchmark. And they both score 1.35.

leeoniya:
js-framework-benchmark won't always show improvements from micro-optimizations since timings in all its metrics are dominated by dom ops & relayouts/reflows/repaints. that's why it's good to have an additional bench that stresses framework internals without any effect on the dom.

jbucaran:
> @ngryman Well, let's define what are the priorities then 😆

Which is why I think we can leave this PR open for when the time is right. It's just weird to merge this now, that's what I mean.

jbucaran:
@developit Can you add your feedback here? 
Your https://github.com/developit/preact/blob/master/src/h.js is very similar and perhaps could benefit from this discussion too.

ngryman:
To make things short:
 - The PR version is faster but when tested in a global benchmark there is no difference.
 - The true perf gain will come when VNodes will all share the same constructor (hidden class: https://gist.github.com/twokul/9501770).
 - It's the start of several passes of optimizations, imho we have to start somewhere :smile:

jbucaran:
@ngryman Thanks for the summary and the short article. Just what I needed for my low attention span issues 😄.

And yes, we have to start somewhere, and that somewhere is possibly patch, who's doing all the heavy lifting around here. 💪

developit:
That little function has been through probably 100 different implementations.  Regarding changing JSX: even if you were to get VNodes extended to support new nodeTypes like Text and Comment, `children` is still a passthrough - people provide all sorts of things in that Array, like functions.  Your best bet is to design the function to perform well without trying to force it into being monomorphic.

I've experimented with making `h()` a small normalization function and moving the heavy lifting into another function (such as the VNode constructor) called with consistent types/shapes:
https://esbench.com/bench/587567b899634800a0347555

You can see some other contenders for `h()` benchmarked here:
https://esbench.com/bench/57ee8f8e330ab09900a1a1a0

jbucaran:
@ngryman I'd like to revisit this PR and if possible, with your help, merge it.

Can we first double check that these changes don't break any existing code? JSX-users must not suffer from this.

👋😄🙏

ngryman:
@jbucaran Sure, I'll do it in the week.

ngryman:
I'm going to create a new PR for this one if it's ok for you. This one was from my fork and has diverged quite a lot form `master`.

jbucaran:
SGTM

PR #188, opened 2017-04-18
Fix case-sensitive attributes

madebyherzblut:
Attributes were always converted to lowercase which breaks case-sensitive attributes like `viewBox` for most browsers.

Example: http://codepen.io/anon/pen/VbvGwZ?editors=0010 (the circle should fill the entire viewport)

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -130,7 +130,7 @@ export default function(app) {
       }
     } else {
       try {
-        element[name] = value
+        element[attr] = value
       } catch (_) {}
 
       if (typeof value !== "function") {
--- a/src/app.js
+++ b/src/app.js
@@ -122,8 +122,9 @@ export default function(app) {
   }
 
   function setElementData(element, name, value, oldValue) {
-    if (name === "key") {
-    } else if ((name = name.toLowerCase()) === "style") {
+    var attr = name.toLowerCase();
+    if (attr === "key") {
+    } else if (attr === "style") {
       for (var i in merge(oldValue, (value = value || {}))) {
         element.style[i] = value[i] || ""
       }
--- a/src/app.js
+++ b/src/app.js
@@ -122,15 +122,14 @@ export default function(app) {
   }
 
   function setElementData(element, name, value, oldValue) {
-    var attr = name.toLowerCase();
-    if (attr === "key") {
-    } else if (attr === "style") {
+    if (name === "key") {
+    } else if (name === "style") {
       for (var i in merge(oldValue, (value = value || {}))) {
         element.style[i] = value[i] || ""
       }
     } else {
       try {
-        element[attr] = value
+        element[name.toLowerCase()] = value
       } catch (_) {}
 
       if (typeof value !== "function") {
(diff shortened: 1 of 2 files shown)


jbucaran:
IIRC the only reason we're running .toLowerCase() is to allow users to write events like onclick as onClick since camelCase plays nicer with TypeScript or so I heard (TypeScript users please correct me).

So, instead we could patch line 133 like this:

```js
element[name.toLowerCase()] = value
```

@madebyherzblut Tell me if that makes sense.

dodekeract:
@jbucaran There are a lot of multi-word attributes in HTML. It's not just about `onClick`. It's also about `frameBorder` and similar attributes.

jbucaran:
@dodekeract I think you've misunderstood my comment.

> IIRC the only reason we're calling .toLowerCase() is to allow users to write events like onclick as onClick...

So, we don't need to call this for attributes at all and `element[name.toLowerCase()] = value` should suffice.

jbucaran:
@farism Ping.

Ref:
* https://github.com/hyperapp/hyperapp/pull/133

src/app.js
```diff
@@ -122,14 +122,15 @@ export default function(app) {
   }
 
   function setElementData(element, name, value, oldValue) {
-    if (name === "key") {
-    } else if ((name = name.toLowerCase()) === "style") {
+    var attr = name.toLowerCase();
```
jbucaran: I think we can skip this check altogether, since we only want to call toLowerCase once in line 133.

test/vdom.test.js
```diff
@@ -395,6 +395,7 @@ test("svg", () => {
 
   const svg = document.getElementById("bar")
   expect(svg.namespaceURI).toBe(SVG_NS)
+  expect(svg.getAttribute('viewBox')).toBe('0 0 10 10')
```
jbucaran: 👍

madebyherzblut:
Updated the pull request with your suggestion @jbucaran. It works but I have an uneasy feeling about the whole case-(in)sensitivity to be honest. Does anyone know why React maintains a list of valid attributes?

src/app.js
```diff
@@ -123,13 +123,13 @@ export default function(app) {
...
     if (name === "key") {
-    } else if ((name = name.toLowerCase()) === "style") {
+    } else if (name === "style") {
       for (var i in merge(oldValue, (value = value || {}))) {
         element.style[i] = value[i] || ""
       }
     } else {
       try {
-        element[name] = value
+        element[name.toLowerCase()] = value
```
jbucaran: @madebyherzblut This only sets props, so attrs are not even touched now, which is what we wanted of course! 💥

jbucaran:
@madebyherzblut After your PR is merged we won't even be touching attributes, which is good. The only issue left is props. I'd really prefer to not use toLowerCase at all TBH, I don't like onClick so much, but this was introduced in #133 for a reason, so I'll let it rest and hope to get some feedback from Faris.

jbucaran:
@madebyherzblut Also, don't worry about SVG attributes, since some of their prop equivalents are readonly, it doesn't matter if element[name] fails. That's why it's inside a `try, catch`. It's our little wart.

jbucaran:
I think this is a good opportunity to go back to not camel case for anything. oncreate/onupdate/onremove. We save toLowerCase(), avoid any potential errors from case sensitive attrs, it's less code and one less thing to worry about.

Issue #190, opened 2017-04-19
Normalize actions/events signatures.


jbucaran:
# tl;dr 

Make all functions in the API follow the same signature:

### (state, actions, data, emit)

<hr>

### Now

* view(state, actions)
* actions.myAction(state, data, actions, emit)
* events.loaded(state, actions, emit)
* events.action(state, actions, data, emit)
* events.update(state, actions, data, emit)
* events.render(state, actions, data, emit)

## After

* view(state, actions)
* actions.myAction(state, actions, data, emit)
* events.loaded(state, actions, emit)
* events.action(state, actions, data, emit)
* events.update(state, actions, data, emit)
* events.render(state, actions, data, emit)

Tentative: Pass the `options` (the same from app(options)) object to `loaded` as data, so it can match the proposed signature too. 🤔

dodekeract:
> events.loaded(state, actions, emit) — @jbucaran

Maybe it's better to just use `state, actions, _, emit` here, where `_` is (right now) always `undefined`? It streamlines everything and if we ever decide that we need `data` we don't have to break the API.

jbucaran:
@dodekeract Good point!

jbucaran:
This is a breaking change, as you'll need to modify all your actions signature from (state, data, actions) to **(state, actions, data)**. 

@selfup @pedroborges @cdeutmeyer @lukejacksonn Are we okay with this?

lukejacksonn:
The difference between `(s,a,d)` and `(s,d,a)` has always niggled me and it would certainly mean one less thing for users to remember.. for these reasons I am all for it.

I also support @dodekeract's statement of using `state, actions, _, emit` for `loaded`. Just to keep things super consistent. Your proposal @jbucaran of passing options as the data for that function is also not a bad idea at all 👍

jbucaran:
https://github.com/hyperapp/hyperapp/commit/c29970c1bfa6741463646a7edc50ca1b0feb9fb4 🎉

PR #191, opened 2017-04-19
0.9.0

### LINKED ISSUE #189: Rename plugins
The mechanism for plugins has proven a generally useful way to separate the concerns of your app into separate modules. 

If for example, you are writing an admin interface for a blogging platform, you might want to separate your code into `auth`, `posts`, `comments`, et c modules, containing the state, actions, events relevant to each respective topic.

By exposing each module's state, actions, events as a *plugin*, it becomes very convenient to bring them together in a single app using `app({ ... plugins: [auth, posts, comments], ...})`

However, the word "plugins" implies it is only meant for more general, reusable things. Not as a way to structure your app. Therefore, I move that the name "plugins" be changed to something else. The question is *what*?

Either, we find a suitable noun, such as "modules" (except "modules" might cause some confusion), or, perhaps we don't hard-code a name into hyperapp, and instead declare the plugins with a verb, such as: `app({ ... compose: [auth, posts, comments] ...})`. Other options, besides "compose", might be "include" or "merge".

If we go the "verb" route, we'll still need a noun to use instead of "plugins", for docs and discussions. But we're more free to use a longer, descriptive name, like "hyperapp-modules" for example.

At the moment, I think my vote is on: `app({ ... include: [foo, bar] ...})` and calling the includees "hyperapp-modules" in the docs

jbucaran:
Close #189 and #190.


README.md
```diff
@@ -16,7 +16,7 @@ HyperApp is a JavaScript library for building frontend applications.
 * **Stateless components**: Build complex user interfaces from micro-components. Stateless components are framework agnostic, reusable and easier to debug.
 * **Batteries-included**: Out of the box, HyperApp has Elm-like state management and a virtual DOM engine; it still weighs `1kb` and has no dependencies.
 
-[Get started with HyperApp](https://github.com/hyperapp/hyperapp/wiki/Getting-Started).
+[Get started with HyperApp](...)
```
dodekeract: Shouldn't this be `/docs/getting-started.md`?

jbucaran: Fixed!

PR #193, opened 2017-04-24
[Fix] Strange bug where `node.data` isn't set within the `removeElement` method.

maraisr:
I have a working example where I'm using jsx through babel and doing some odd filtering on that data. Without investigating why the `node.data` isn't set, I've just added that equality in there to make it work.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -162,7 +162,7 @@ export default function(app) {
   }
 
   function removeElement(parent, element, node) {
-    if (node.data.onRemove) {
+    if (node.data && node.data.onRemove) {
       node.data.onRemove(element)
     }
     parent.removeChild(element)


jbucaran:
@maraisr Thanks! I'll investigate why the data isn't set.

zaceno:
He, I was about to send in a PR to fix this also. Node data isn't set when removeElement is called for text nodes. In the PR I was going to send in, I included a test which a) fails when the fix is not present and b) passes when the fix is present.

The fix, of course is to change `if (node.data.onRemove) ...` to `if (node.data && node.data.onRemove)...`

And the test I was going to send in looks like this:

~~~~
test("remove text node", () => {
  TreeTest([
    {
      tree: h("main", {}, [h("div", {}, ["foo"]), "bar"]),
      html: `<main><div>foo</div>bar</main>`
    },
    {
      tree: h("main", {}, [h("div", {}, ["foo"])]),
      html: `<main><div>foo</div></main>`
    }
  ])
})
~~~~

jbucaran:
@zaceno That's it! I remember noticing this at some point when implementing keys, but forgot about it. I'll still hold this up a bit. I think we can improve how we handle text nodes (and by doing so not needing to check anything in onremove since it doesn't make sense to fire for strings). 

In patch, we need to detect if we're dealing with a text node and skip the loop.

jbucaran:
@zaceno The test is okay, send it.

zaceno:
@jbucaran Ok I'll send in the test as a separate PR then. 

BTW: handling strings differently than normal vnodes earlier in the patch does sound like a good idea. But if I were you, I would bring in @maraisr s fix in the mean time, because right now, hyperapp@0.8.1 is broken for the app I'm working on (and I guess @maraisr too). I guess it depends on how long you expect to wait before fixing the patch function wrt string-nodes.

jbucaran:
@zaceno You are right.

PR #195, opened 2017-04-24
Add test for when plain-string vnodes are removed

zaceno:
This test breaks the suite because there is currently a bug in hyperapp, that occurs when textnodes are removed, where we check for `node.data.onRemove` without knowing for sure that `node.data` exists (which it doesn't, if node is just a plain string). Causes a TypeError to be thrown.

### DIFF
--- a/test/vdom.test.js
+++ b/test/vdom.test.js
@@ -111,6 +111,19 @@ test("insert children on top", () => {
   ])
 })
 
+test("remove text node", () => {
+  TreeTest([
+    {
+      tree: h("main", {}, [h("div", {}, ["foo"]), "bar"]),
+      html: `<main><div>foo</div>bar</main>`
+    },
+    {
+      tree: h("main", {}, [h("div", {}, ["foo"])]),
+      html: `<main><div>foo</div></main>`
+    }
+  ])
+})
+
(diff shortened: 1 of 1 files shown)


zaceno:
Submitted as verification of:

https://github.com/hyperapp/hyperapp/pull/193

... and to prevent regressions in the future :)

maraisr:
@zaceno might just have to merge master back into your branch, and I believe the tests should pass again.

maraisr:
Looking like this guy can be merged? @jbucaran @zaceno?

jbucaran:
Whoops, forgot about this. Thanks @zaceno and @maraisr.

maraisr:
haha @jbucaran all good man! 

Are we going to see a patch version bump?

lukejacksonn:
+1 on the patch version bump. Would be great if we could.. I'm waiting on [this](https://github.com/hyperapp/hyperapp/commit/451b0c22699305d91d8159c668c2619da17df25d) to fix svgs which could fall under a patch (although maybe its minor 😕)

dodekeract:
I see no reason why we should ever delay a non-breaking bugfix.

PR #197, opened 2017-04-30
Add prettier format script & run

tiaanduplessis:


src/h.js
```diff
@@ -28,4 +28,3 @@ export default function(tag, data) {
       }
     : tag(data, children)
 }
-
```
zigomir: This tiny change somehow doesn't conform to https://github.com/hyperapp/hyperapp/blob/master/.editorconfig#L7 imo 🐰

selfup:
going to pull down and take a look!

src/h.js
selfup: It added a newline for me when I ran `yarn run format` after deleting the line on purpose.

But yes agreed, somehow this got missed :P

dodekeract: There still is a newline. It removed a duplicate newline.

selfup: Ahh yes the way github diffs. Makes sense 👍

jbucaran:
Sorry for the delay! Thanks @tiaanduplessis! 👏

PR #200, opened 2017-05-02
Refactor patch for less bytes

lazarljubenovic:
Didn't check it on any live project, I put my faith in Boolean algebra and 100% coverage :sparkles:

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -217,21 +217,11 @@ export default function(app) {
           reusableNode = reusableChild[1]
         }
 
-        if (null == oldKey && null == newKey) {
-          patch(element, oldElement, oldChild, newChild)
-          j++
-          i++
-        } else if (null == oldKey && null != newKey) {
-          if (reusableElement) {
-            element.insertBefore(reusableElement, oldElement)
-            patch(element, reusableElement, reusableNode, newChild)
-          } else {
-            patch(element, oldElement, null, newChild)
+        if (null == newKey) {
+          if (null == oldKey) {
+            patch(element, oldElement, oldChild, newChild)
+            j++
           }
-
-          j++
-          newKeys[newKey] = newChild
-        } else if (null != oldKey && null == newKey) {
           i++
         } else {
           if (oldKey === newKey) {


jbucaran:
@zaceno Can you have a look? 🙏

jbucaran:
@lazarljubenovic I can confirm that all of my "sandbox" examples work with this change. I still need to study this more carefully.

zaceno:
@jbucaran Sure! I've looked at it and it looks fine. At first I suspected it would fail when you move a keyed node to above an unkeyed node, but I tested that and it works fine. So 👍 from me :)

src/app.js
```diff
@@ -217,21 +217,11 @@ export default function(app) {
...
-          patch(element, oldElement, oldChild, newChild)
-          j++
-          i++
-        } else if (null == oldKey && null != newKey) {
-          if (reusableElement) {
-            element.insertBefore(reusableElement, oldElement)
-            patch(element, reusableElement, reusableNode, newChild)
-          } else {
-            patch(element, oldElement, null, newChild)
+        if (null == newKey) {
```
selfup: This looks great! Super short and concise 👍

src/app.js
```diff
@@ -217,21 +217,11 @@ export default function(app) {
...
         }
 
-        if (null == oldKey && null == newKey) {
-          patch(element, oldElement, oldChild, newChild)
-          j++
-          i++
-        } else if (null == oldKey && null != newKey) {
-          if (reusableElement) {
-            element.insertBefore(reusableElement, oldElement)
-            patch(element, reusableElement, reusableNode, newChild)
```
dodekeract: Why can reusable be removed here?

jbucaran: @lazarljubenovic Ping.

ngryman:
> At first I suspected it would fail when you move a keyed node to above an unkeyed node, but I tested that and it works fine.

I think we should add a test case for that, what do you think?

lazarljubenovic:
The git's diff accidentally hides the logic behind it and makes you make a wrong assumption of how the optimization was made. It's all actually just Boolean algebra, without even thinking about this specific algorithm. Here's the breakdown to help.

We won't write whole code here, instead we notice repeating patterns in code and write them here as letters A, B, C and D.

```
if (null == oldKey && null == newKey) {
  A // first two lines
  B // last line
} else if (null == oldKey && null != newKey) {
  C // everything
} else if (null != oldKey && null == newKey) {
  B
} else {
  if (oldKey === newKey) {
    D
  } else {
    C
  }
}
```

Since all top-level ifs are disjoint, we can reorder them. Swapping second and third gives us this.

```
if (null == oldKey && null == newKey) {
  A
  B
} else if (null != oldKey && null == newKey) {
  B
} else if (null == oldKey && null != newKey) {
  C
} else {
  if (oldKey === newKey) {
    D
  } else {
    C
  }
}
```

Now we group first two and second two under an umbrella if. Basically instead of having separate stuff like

```
if(X and Y) {}
else if (X and !Y) {}
else if (!X and Y) {}
else /* if (!X and !Y) */ {}
```

we have 

```
if X {
  if Y {}
  else {}
} else /* if !X */ {
  if Y {}
  else {}
}
```

When we apply this to our example, we get this.

```
if (null == newKey) {
  if (null == oldKey) {
    A ; B
  } else {
    B
  }
} else {
  if (null == oldKey) {
    C
  } else {
    if (oldKey === newKey) {
      D
    } else {
      C
    }
  }
}
```

Let's format for easier reading.

```
if (null == newKey) { // (1)
  if (null == oldKey) A B
  else B
} else { // (2)
  if (null == oldKey) C
  else { if (oldKey === newKey) D else C }
}
```

Now let's take a look at first branch. (1) In one case we're doing `{A ; B}`, in the other caase we're doing just `B`. So we're doing `B` either way and doing `A` only if the condition holds. (1) is equivalent to:

```
if (null == newKey) { // (1)
  if (null == oldKey) A // do A only if condition holds
  B // always do B
}
```

The second is a _little_ bit tricky. It's much easier to see how this can be further optimized if we flip the innermost if, like this:

```
// ... (1)
} else { // (2)
  if (null == oldKey) C
  else { if (oldKey !== newKey) C else D }
}
```

Let's also get rid of `else {if ...}` and just write this as `else if`, because it all chains nicely.

```
// ... (1)
} else { // (2)
  if (null == oldKey) C
  else if (oldKey !== newKey) C
  else D
}
```

Let's take a look at those two cases when we do `C`. Keep in mind that we're inside `else` branch which means that for the whole block we know that `null != newKey`.

It turns out that just the second check is enough here. In other words, the above snippet is equivalent with this one.

```
// ... (1)
} else { // (2)
  if (oldKey !== newKey) C
  else D
}
```

Let me prove it to you. Say that there _is_ a situation where the first check will catch something that the second one cannot catch. This would obviously be true only when `oldKey` is `null`. So, what happens in `oldKey` is `null`?

Well, **since we're in the `else` branch, we already know that `newKey` is not `null`. We're assuming that `oldKey` is `null`. This means that if `oldKey != newKey` is true, then `oldKey == null` must also be true, because we already know `newKey != null`.

Now we can just flip it back.

```
// ... (1)
} else { // (2)
  if (oldKey === newKey) D
  else C
}
```

The whole block is now

```
if (null == newKey) {
  if (null == oldKey) A
  B
} else {
  if (oldKey === newKey) D
  else C
}
```

Hope it's clear now.

As I said, there are no changes in the algorithm itself. It's pure "mechanic" optimisation (though it takes some persistance and luck to notice it).

src/app.js
lazarljubenovic: @dodekeract The diff is a bit misleading as it gives the wrong impressing that I "removed" it. It was just squashed together with another equvalent combination of `oldKey` and `newKey` values. Check out https://github.com/hyperapp/hyperapp/pull/200#issuecomment-298674240

jbucaran:
> @zaceno At first I suspected it would fail when you move a keyed node to above an unkeyed node, but I tested that and it works fine.

How would this test look like?

jbucaran:
@lazarljubenovic Great explanation. 

Merged! 🎉

ngryman:
@lazarljubenovic Awesome write up, thanks for clarification.

I would just add some tests around this particular area to avoid sever things in the future:
 - long write ups (even if I enjoyed it, time is the enemy).
 - avoid future regressions.
 - validate by *experience* your algebra skills 👍

zaceno:
@ngryman @jbucaran Here's the test I ran.

~~~

test("move keyed node above unkeyed", () => {
  TreeTest([
    {
      tree: h("main", {}, [
        h("div", {}, "A"),
        h("div", {}, "B"),
        h("div", {}, "C"),
        h("div", {oncreate: el => (el.id = "d"), key: "d"}, "D"),
        h("div", {}, "E")
      ]),
      html: `
        <main>
          <div>A</div>
          <div>B</div>
          <div>C</div>
          <div id="d">D</div>
          <div>E</div>
        </main>
      `
    },
    {
      tree: h("main", {}, [
        h("div", {}, "A"),
        h("div", {key: "d"}, "D"),
        h("div", {}, "B"),
        h("div", {}, "C"),
        h("div", {}, "E")
      ]),
      html: `
        <main>
          <div>A</div>
          <div id="d">D</div>
          <div>B</div>
          <div>C</div>
          <div>E</div>
        </main>
      `
    }
  ])
})

~~~~

It cold probably be simplified to just:

~~~

test("move keyed node above unkeyed", () => {
  TreeTest([
    {
      tree: h("main", {}, [
        h("div", {}, "A"),
        h("div", {key: "b"}, "B"),
      ]),
      html: `
        <main>
          <div>A</div>
          <div>B</div>
        </main>
      `
    },
    {
      tree: h("main", {}, [
        h("div", {key: "b"}, "B"),
        h("div", {}, "A")
      ]),
      html: `
        <main>
          <div>B</div>
          <div>A</div>
        </main>
      `
    }
  ])
})

~~~

PR #201, opened 2017-05-02
Remove unneeded strict equality checks

lazarljubenovic:


src/app.js
```diff
@@ -98,10 +98,10 @@ export default function(app) {
   }
 
   function createElementFrom(node, isSVG) {
-    if (typeof node === "string") {
+    if (typeof node == "string") {
       var element = document.createTextNode(node)
     } else {
-      var element = (isSVG = isSVG || node.tag === "svg")
+      var element = (isSVG = isSVG || node.tag == "svg")
```
lazarljubenovic: This can technically fail because `tag` does not have to be a string (or it has to, up at this point? it can be a function at some point, but not sure if here we can guarantee that that's been resolved).

Anyway, here's the weird thing with `==` in case it's a function.

```
const foo = () => {}
foo.toString() // "() => {}"
foo == foo.toString() // true
```

However, `"svg"` is not a valid function, so I guess nothing bad can happen here.

src/app.js
```diff
@@ -110,7 +110,7 @@ export default function(app) {
       }
 
       for (var i in node.data) {
-        if (i === "oncreate") {
+        if (i == "oncreate") {
```
lazarljubenovic: Again, this _can_ fail if someone really wants to make it fail.

```
const a = {}
a.toString = () => 'oncreate'
a == 'oncreate' // true
```

Although this might actally be expected behaviour if you think about i.

src/app.js
```diff
@@ -147,9 +147,9 @@ export default function(app) {
       var value = data[name]
       var oldValue = oldData[name]
 
-      if (name === "onupdate") {
+      if (name == "onupdate") {
         value(element)
-      } else if (value !== oldValue || value !== element[name]) {
+      } else if (value != oldValue || value != element[name]) {
```
lazarljubenovic: This one is what I'm worried about the most. No yet 100% familiar with the source or intended behvaiour so can't say for sure what can happen here. We should add a test for it anyway.

src/app.js
```diff
@@ -234,7 +234,7 @@ export default function(app) {
         } else if (null != oldKey && null == newKey) {
           i++
         } else {
-          if (oldKey === newKey) {
+          if (oldKey == newKey) {
```
lazarljubenovic: Keys are always strings, right? Otherwise this must remain `===`.

src/app.js
```diff
@@ -265,7 +265,7 @@ export default function(app) {
           removeElement(element, reusableChild[0], reusableNode)
         }
       }
-    } else if (node !== oldNode) {
+    } else if (node != oldNode) {
```
lazarljubenovic: If it's come to this point, we're talking about two exact same objects (with rfs unchanged), or two same text nodes (strings). Either way, I guess it _can_ fail if we have something like

```
<div>[object Object]</div>
```

turning into

```
<div><div></div></div>
```

So I guess this should be remain `!==` actually.

src/app.js
jbucaran: @lazarljubenovic Keys could be numbers actually.

src/app.js
jbucaran: A node can be a text node, so this is really testing "Hello" !== "Hi".

jbucaran:
We're using == and != in a few places, but that's mostly to handle null/undefined while saving some considerable bytes at the same time. I am however skeptical of using == and !== anywhere else.

== and !== type coerce which is slower than === or !==, so there's also that.

lazarljubenovic:
I understand the concerns about a few instances of `===`/`==`, however, performance is _not_ an issue.

Abstract equality checks (`==` and `!=`) [are **not** slower](https://github.com/getify/You-Dont-Know-JS/blob/master/types%20%26%20grammar/ch4.md#equality-performance) than strict equality checks (`===` and `!==`) if operands of `==` are the same type, because it actually does `===` internally in that case.

Here's the spec: http://www.ecma-international.org/ecma-262/7.0/index.html#sec-abstract-equality-comparison

dodekeract:
I'm against this change because the minifier should already do it for everything that doesn't change meaning (e.g. String comparison).

Making it `==` and `!=` implies that it's necessary IMHO and therefore leads to more complex to understand code.

jbucaran:
It's possible to have a "modern front-end framework", based in functional principles in a 1kb (or so), but not at the expense of readability!

I admit there may be a bit of golfing, here and there, but nothing too extreme IMO.

jbucaran:
@lazarljubenovic 

> I understand the concerns about a few instances of ===/==, however, performance is not an issue.

I read the provided link and was convinced (more or less) that it's not really an issue after all, so you are right.

> Abstract equality checks (== and !=) are not slower than strict equality checks (=== and !==) if operands of == are the same type, because it actually does === internally in that case.

Yep, I think you are right about this too.

@lazarljubenovic My first reaction to the PR is not the reason I don't want to proceed with this change, but basically what @dodekeract commented above. I think this makes the code a bit harder to read since now the user needs to think about the difference between == and === everywhere.

lazarljubenovic:
I aggree that in certain cases it sacrifices too much readability for a single character.

Personally I'd still use `==` in `typeof` comparisons. There's 100% no confusion here. It's the same argument as with semi-colons: most people use semis everywhere because they think it protects them against something, when it actually doesn't. It's the same here: `typeof foo === 'string'` to my eyes looks like you're actually expecting `typeof` to be anything other than string, so you're forcing the type check, too. `===` makes it _less_ readable for me because it makes me think "wait, so this other thing can not be a string so we're making sure that it _is_ a string?". With `==`, I see a comparison with string and know that we expect string on the other side, too.

But as @dodekeract mentions, if it's optimized by a minifier, that's nice. Haven't thought about those at all.

I'm closing this. We can discuss this further if you want, but in the end, it's more a matter of style and it doesn't really matter that much at all. :beers:

PR #203, opened 2017-05-05
reduce router regex length

lazarljubenovic:
The tests are passing but I'm not 100% sure this is equivalent, although by definition it should be. Tests aren't really targeting this area much.

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -41,9 +41,9 @@ export default function(app) {
             "^" +
               route
                 .replace(/\//g, "\\/")
-                .replace(/:([A-Za-z0-9_]+)/g, function(_, key) {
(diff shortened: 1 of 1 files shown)


jbucaran:
@lazarljubenovic Can we add a test for this? 🙏

lazarljubenovic:
I think that whoever's written the original regex should add tests. Whatever I write will be biased towards `\w` (and will work with both versions) because that's how I "see" the original one.

lazarljubenovic:
Got merged here https://github.com/hyperapp/hyperapp/pull/207

jbucaran:
Thanks @lazarljubenovic! 🙏

PR #204, opened 2017-05-05
remove array extracting

lazarljubenovic:
This gets rid of 5 bytes in the bundle (I think). I'm not sure how expensive accessing `[0]` and `[1]` a few times is, compared to additional initialization to `0` and an `if` statement. 

Note that previously `if` was executed all the time, and these accesses are happening only when the conditions are satisfied.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -208,14 +208,7 @@ export default function(app) {
 
         var newKey = getKeyFrom(newChild)
 
-        var reusableChild = reusableChildren[newKey]
-        var reusableElement = 0
-        var reusableNode = 0
-
-        if (reusableChild) {
-          reusableElement = reusableChild[0]
-          reusableNode = reusableChild[1]
-        }
+        var reusableChild = reusableChildren[newKey] || []
 
         if (null == newKey) {
           if (null == oldKey) {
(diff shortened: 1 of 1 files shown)


jbucaran:
> This gets rid of 5 bytes in the bundle (I think).

I run

```
yarn run build; zopfli --i2000 dist/hyperapp.js > dist/hyperapp.js.gz; size dist/hyperapp.js.gz
```

when golfing.

jbucaran:
@lazarljubenovic

lazarljubenovic:
That makes it from 3605 to 3600, raw; and **1636 to 1625**, gzipped. I used `stat` as `size` complains that "file format is not recognized" for the `.js.gz` file. :confused: 

What do you think about this change, performance-wise?

jbucaran:
@lazarljubenovic Not sure, (probably/hope there's no difference) but I'll definitely run the [benchmarks](https://github.com/krausest/js-framework-benchmark) for this one and see.

ngryman:
@lazarljubenovic Performance-wize there should be no difference as a `gvn` pass will simply create a temporary variable under the hood.

GVN: https://www.wikiwand.com/en/Global_value_numbering

jbucaran:
Thanks @lazarljubenovic!

PR #205, opened 2017-05-06
make use of inline assignment

lazarljubenovic:
I learned this trick from this repo and it's awesome. Here's 10 bytes less.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -162,8 +162,8 @@ export default function(app) {
   }
 
   function removeElement(parent, element, node) {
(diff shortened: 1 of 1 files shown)


jbucaran:
LGTM

lazarljubenovic:
Ha, this actually drops bytes from minified version (3605 to 3595), but increases gzipped (1636 to 1640). :sob:

jbucaran:
@lazarljubenovic Ha. Well, no worries, this will eventually go away after patch is optimized in how it deals with texst nodes.

PR #206, opened 2017-05-06
save bytes by re-using variable names

lazarljubenovic:
It's all function scope with `var` anyway.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -180,11 +180,11 @@ export default function(app) {
       var oldElements = []
       var newKeys = {}
 
-      for (var i = 0; i < oldLen;) {
+      for (var i = 0; i < oldLen; i++) {
         var oldElement = element.childNodes[i]
         oldElements[i] = oldElement
 
-        var oldChild = oldNode.children[i++]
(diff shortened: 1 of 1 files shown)


jbucaran:
@lazarljubenovic I think this doesn't actually end up saving any bytes in the compressed bundle. If you are developing hyperapp locally, you can run the `build` script and compare the bundle size. Then post here the before and after.

lazarljubenovic:
Minified: 3605 to 3595
Gzipped: 1636 to 1637.

Gzip is a bitch. :unamused:  

A problem with gzip is that it's not really easily predictable how it wil behave. Maybe two separate changes both increase gzip filesize, but together they decrease it (if there are repeating parts in them).

lazarljubenovic:
Now gzip stays the same, but minified version gets from 3605 down to 3596. Do we care for this or only gzip? It doesn't make code any less easier to read imo.

jbucaran:
> Do we care for this or only gzip?

I'd say gzip.

lazarljubenovic:
No need for this since gzipped version doesn't improve.

PR #207, opened 2017-05-08
Doc tweaks, router fixes.

jbucaran:
Fixes a bug in the router originally reported by @lukejacksonn on Slack.

The bug consists of a call stack overflow generated when calling a reducer-like action from the router's `route` event. The reason for the overflow is because the `route` event is fired inside the `render` event, so any action that changes the state would also trigger another render which would again trigger `route` and so on.


jbucaran:
@lukejacksonn Nothing in particular, I think we're good to merge.

lukejacksonn:
:shipit:

src/router.js
```diff
@@ -15,20 +18,19 @@ export default function(app) {
     },
     events: {
       loaded: function(state, actions) {
-        addEventListener("popstate", function() {
+        match()
```
SkaterDad: Sorry to comment on such an old commit, but I'm curious why `match()` is being called here?  It's already called when the app starts up and the `state` object is instantiated.

Basically I was playing around with Hyperapp on Webpackbin.com, and logging various things.  Doing so, I noticed `match` gets called twice on app startup.  I commented out that line, and everything still seemed to work, but I'm sure the embedded browser window isn't perfectly representative of real use.

lukejacksonn: humm..  I see what you are saying, have you tried testing it locally to see if you get the same results?

jbucaran: @SkaterDad By calling the local function match here, we trigger actions.router.match which calls emit("route"). So, the reason there's this match() here is so that events.route is triggered when the app loads.

SkaterDad:
Thanks for explaining, that makes total sense now!

Issue #209, opened 2017-05-11
Support for xlink:href


ngryman:
Hey guys,

SVG support is currently incomplete, `use` does not work because `xlink:href` is a namespaced attribute that needs the couple `setAttributeNS`/`removeAttributeNS`.

I'm not opening a PR directly for this, but it's a rather common use case and I think we should support it.

Some references:
 - preact issue on the subject: https://github.com/developit/preact-svg/issues/15
 - preact-way: https://github.com/developit/preact/blob/33fc697ac11762a1cb6e71e9847670d047af7ce5/src/dom/index.js#L81

jbucaran:
I was looking for an example to test xref:link, and found this on MDN.
> xlink:href
> Deprecated since SVG 2
> This feature has been removed from the Web standards...

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href

Am I looking at the wrong page?

@lukejacksonn @ngryman @leeoniya

ngryman:
No, but then there is browser support: https://www.chromestatus.com/feature/5760616295825408.
In real life, everybody still uses `svg 1.1`.

ngryman:
On a side note, every frontend frameworks that I know and claims to support `svg` does support namespaced attribute.

jbucaran:
@ngryman Sure, I just wanted to make sure I was not looking at the wrong page because I've never used xlink.

lukejacksonn:
> What is the cost? ~10B minified gziped? What is the gain? It just vastly increase svg support for modern apps that use svg sprites.

I'm really in agreement here.. my hyperapp projects are the only projects that don't employ `<use>` tags, not because I don't prefer them anymore but because it isn't supported.

How much cleaner is:

```
const DeleteButton = () =>
  <button>
    <svg><use xlink:href="#icon-delete" /></svg>
  </button>
```

than the alternative:

```
const DeleteButton = () =>
  <button>
    <svg fill="#000000" height="24" viewBox="0 0 24 24" width="24" xmlns="http://www.w3.org/2000/svg">
      <path d="M0 0h24v24H0V0z" fill="none"/>
[...]

Issue #210, opened 2017-05-13
Managing Complex State


zaceno:
Hi!

Here and there there has been some discussion on some problems and potential solutions around managing state when it gets complex. Possibly related topic: how to handle state for custom components (since hyperapp doesn't allow local component state like React/Vue)

I just wanted to open this issue as a central place to have such discussions.

zaceno:
Here are a couple of scenarios where I think that the initialization, modification or reading of state could use some improvement somehow (feel free to add to this list!!)

I'm hoping we can distill down from this either some best practices, or some more specific key things that make these situations difficult, so that some simple yet broadly helpful solutions can be designed.

- When you have a list (array) of "things", where each thing has some state of it's own. Say you're building an email client. Your state would likely be `emails: [ {...}, {...} ...]` where each object in the list has a bunch of properties like:  `{body: ..., attachments: [], .... recipients: [] ...., etc...}` 

- When you have some complex conditions for rendering a thing. A lengthy logical formula obfuscates the intent and makes things less readable. Especially if the same logic formula needs to be used in many places of your view.

- When state gets deep, your actions need to be defined in ways like `(state, actions, {index, jndex, value}) => { state.foo.bar[index].baz[jndex] = value; return state; }. You don't have much choice but to modify the current state (rather than return a new modified copy). It would be nice to have a helper to abstractly so you could just write: `(state, actions, {index, jndex}) => stateMutateHelper('foo.bar.index.baz.jindex', value)`.

zaceno:
Come to think of it -- I think an email client would be a good example project to:

- Answer the question: how difficult is it *really* to build an app with complex state in hyperapp?
- Demonstrate the particular difficulties
- Demonstrate possible "best practices" to alleviate difficulties
- Be a proving ground for some new solutions.

It would be a pretty big undertaking though-- one that I'm not up for, personally (not solo anyway). So if anyone has any ideas for a simpler app that still makes use of complex state, please suggest it!

zaceno:
The most requested state management feature (on Slack anyway) has been the notion of scoping actions to work on subsets of the state (aka "scoped actions")

This gist: https://gist.github.com/zaceno/239e384dd914f1cb83a4d4b36af25ea2
shows an example of how I see that working (several examples in successive refinement -- according to myself anyway)

zaceno:
Actually, arter thinking about it some more, I no longer feel like "scoped actions" are a good idea (not definitely anyway)

[...]

Issue #147, opened 2017-03-09
local state for a select box


queckezz:
There was a lot of discussion regarding sub-components in #2 and #73 already.

I was wondering how should one go about creating **a reusable select box**. There is a lot of state involved that does not necessary need to be stored in the root model:

* `filteredSuggestions` - Filtered suggestion displayed in the dropdown menu
* `open` - Is the dropdown menu open?
* `value` - Current value of the select input box

In React, I created two components `StatefulSelect` and `Select`, the first maintaining it's local state and the latter a stateless functional component representing it's visual appearance.

```js
const Select = () => { /* ... */ }

class StatefulSelect extends Component {
  constructor (props) {
    super(props)

    this.state = {
      value: '',
      open: false,
      filteredSuggestions: props.suggestions,
      selected: null
    }
  }

  // ... state update methods ...

  render () {
    return <Select open={this.state.open} />
  }
}
```

So in `hyperapp` there is no concept of *local state*, right? I need to attach the `<Select />` state directly to the root model. Wouldn't that be a use case for nested `app()`'s? And if not, I would love to see an implementation of a select box in `hyperapp`

dodekeract:
I'd really love to see a solution for this problem. Right now I'm always avoiding it.

jbucaran:
> @queckezz So in hyperapp there is no concept of local state, right?

That's right. Just like in Elm.

> I would love to see an implementation of a select box in hyperapp

Me too. This sounds like a good first non-trivial challenge to tackle the idea of reusable functions as opposed to reusable components.

From [Scaling The Elm Architecture](https://guide.elm-lang.org/reuse/):

> If you are coming from JavaScript, you are probably wondering “where are my reusable components?” and “how do I do parent-child communication between them?” A great deal of time and effort is spent on these questions in JavaScript, but it just works different in Elm. We do not think in terms of reusable components. Instead, we focus on reusable functions. It is a functional language after all!
>
> ...
>
> In the end, I think we end up with something far more flexible and reliable than “reusable components” and there is no real trick. We will just be using the fundamental tools of functional programming!



**Related**:
* https://guide.elm-lang.org/reuse/

zaceno:
I agree this is a problem.

[...]

Issue #146, opened 2017-03-09
Website? ⚡️


jbucaran:
There is of course <https://hyperapp.gomix.me>, but that's just a bunch of examples I put together for people to get started with HyperApp more quickly.

1. Should we have a website?
2. What information should be on the website?
3. What should be the goal of the website?
4. Thoughts?

dodekeract:
If we have a website I can host it easily. Got a dedicated root server with lots of free resources.

We should get a domain then though.

jbucaran:
<https://github.com/dodekeract/deploy> looks nice with yellow, magenta and lime, but what is it?

> If we have a website I can host it easily. Got a dedicated root server with lots of free resources.

Awesome! I'd suggest we use something more substainable like plain ole GitHub Pages tho. The scope of the website is mostly promotional and HyperApp is not a product, but why not, I'm basically cool with that, just want to know what people think and whether this is a good idea or what.

dodekeract:
@jbucaran It's a small service that runs on my server to automatically pull GitHub repos when a web hook arrives. That way a push to master automatically updates the associated website. Right now it's only used by https://dinc.hacker.digital, which I host for a friend.

cdeutmeyer:
Somewhat related to this, it's too bad the /r/HyperApp/ is taken but neglected. Reddit is a pretty good platform for discussion and questions.

lukejacksonn:
I think this is a good idea (mainly because I can see it looking awesome) but content is a great question. What should be covered that isn't in the repo/wiki?

farism:
I think it's a great idea. Having a slick website never hurts.

grantcarthew:
Again @jbucaran as I stated on #165, have a look what I have done here: https://github.com/grantcarthew/site-auto-tech

Very easy to do and hosted by GitHub.

jhwohlgemuth:
@jbucaran @dodekeract
Regarding the desire for a domain, you might consider [js.org](https://dns.js.org/). I think **hyperapp.js.org** would be pretty cool. Getting a domain from js.org is free, easy to setup, and [seems to be available](https://github.com/js-org/dns.js.org/blob/master/cnames_active.js#L349).

Swizz:
Hyperapp do well to "animate" a small part of a webpage. Thanks to the root param.

So, it could be great to have an interactive documentation.

Idered:
Here's the design. Please leave some feedback. If everything is ok, I can start codding it.

[...]

PR #211, opened 2017-05-14
Support xlink:ref attribute

ngryman:
Following #209.

@jbucaran Is this worth to add the `removeAttributeNS`?

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -134,7 +134,12 @@ export default function(app) {
 
       if (typeof value !== "function") {
         if (value) {
-          element.setAttribute(name, value)
+          if (name === 'xlink:href') {
+            element.setAttributeNS("http://www.w3.org/1999/xlink", "href", value)
+          }
+          else {
+            element.setAttribute(name, value)
+          }
         } else {
           element.removeAttribute(name)
         }
--- a/src/app.js
+++ b/src/app.js
@@ -134,7 +134,7 @@ export default function(app) {
 
       if (typeof value !== "function") {
         if (value) {
-          if (name === 'xlink:href') {
+          if (name === "xlink:href") {
             element.setAttributeNS("http://www.w3.org/1999/xlink", "href", value)
           }
           else {
--- a/test/vdom.test.js
+++ b/test/vdom.test.js
@@ -427,6 +427,20 @@ test("svg", () => {
   }
 })
 
+test("xlink", () => {
+  const XLINK_NS = "http://www.w3.org/1999/xlink"
+
+  app({
+    view: _ =>
+      h("svg", { id: "foo", viewBox: "0 0 10 10" }, [
+        h("use", { id: "bar", "xlink:href": "about:blank" })
+      ])
+  })
+
+  const use = document.getElementById("bar")
+  expect(use.getAttributeNS(XLINK_NS, "href")).toBe("about:blank")
+})
+
 test("style", () => {
   TreeTest([
     {


jbucaran:
@ngryman Thanks! Are you using this on a project?

ngryman:
@jbucaran Yes, making an app :)

jbucaran:
@ngryman Sounds nice, but can you share more and how are you using this? I've never used xlink:ref.

ngryman:
`xlink:href` is used to reference and instantiate a SVG symbol from a SVG store. It basically allows you to reuse an inline SVG multiple times while keeping your markup DRY. The advantage of doing so is that you sill can style each instance differently.

You create a SVG store once:
```html
<svg>
  <symbol id="svg-logo" viewBox="0 0 128 128">
    <!-- logo content -->
  </symbol>
</svg>
```

And then you instantiate it wherever you want:
```html
<svg>
  <use xlink:href="#svg-logo" />
</svg>
```

`xlink:href` is needed to reference the symbol you want to instantiate.

Comprehensive lectures on the subject:
 - https://css-tricks.com/svg-sprites-use-better-icon-fonts/
 - https://tympanus.net/codrops/2015/07/16/styling-svg-use-content-css/.

jbucaran:
@ngryman Thanks for the explanation. So, you are using this on a project? Nice.

> @jbucaran Is this worth to add the removeAttributeNS?

Sorry, how do you mean? Like, when calling removeElement?

ngryman:
There is a `*AttributeNS` family of functions:

 DOM Level 1 | DOM Level 2
 ------------ | -------------
`setAttribute` | `setAttributeNS`
`getAttribute` | `getAttributeNS`
`hasAttribute` | `hasAttributeNS`
`removeAttribute` | `removeAttributeNS`

So when setting a NS attribute, you have to use all the `*NS` functions, including `removeAttributeNS`. I don't know if `removeAttribute` would work and I suspect it won't.

That said, I've never been in a use case where I had to remove the `xlink:ref` attribute, it does not really make sense.

For example here is how `preact` handles this topic: https://github.com/developit/preact/blob/33fc697ac11762a1cb6e71e9847670d047af7ce5/src/dom/index.js#L81

jbucaran:
@ngryman Thanks again for all the info!

lukejacksonn:
The `<use xlink:href='#id'>` tag is my preferred method of using SVGs, so this is great! 💯 

I also see little point in adding `removeAttributeNS` but perhaps for the sakes of completeness it is worth the extra few bytes?

jbucaran:
Not trying to downplay this issue, but the following, albeit unrelated here, deserve more attention:

* Documentation. Blog posts. 
* Examples
* Plugins.
* Presentation: https://github.com/hyperapp/hyperapp/issues/146, #41.
* Experimenting and finding common patterns: https://github.com/hyperapp/hyperapp/issues/147, #210, etc.

jbucaran:
Hi @ngryman! I was wondering, would it be possible to use one of the lifecycle events to setElementNS?

ngryman:
@jbucaran What do you mean exactly? I don't understand the relationship between lifecycle events and vdom.

jbucaran:
@ngryman Sure. I mean, it should be possible to use `onupdate` (or `oncreate`, depending on whether you need to set this only once, which seems likely to be the case) and call `setAttributeNS` yourself.

```js
const view = state => (
    <svg>
      <use
        oncreate={e =>
          element.setAttributeNS(
            "http://www.w3.org/1999/xlink",
            "href",
            "/svg/svg-sprite#my-icon"
          )}
      />
    </svg>
  )
```

ngryman:
Well then let's drop `vdom` totally and do this for every attribute and element 🤣 

Let's be honest, I know your are pretty conservative to code additions, and I totally get that. I think it's a good practice. But, and correct me if I'm wrong, I'm under the impression you over emphasis on bundle size, which seems to become the only reason you don't merge stuff.

Either `hyperapp`'s goal is to be the smallest library on earth, but then it has to assume it lacks several things or it applies a good balance between features / weight.

Speaking of this feature, we increase `svg` support (as advertised), which is just good 👍. What is the cost? ~10B minified gziped? What is the gain? It just vastly increase `svg` support for modern apps that use `svg` sprites.

jbucaran:
> @ngryman Well then let's drop vdom totally and do this for every attribute and element 🤣

I think it's okay to accept some things, and reject others. For example, we removed builtin support for controlled input elements at some point. It was convenient to have it, but it can be easily solved in userland and people are going to be creating custom tags often anyway.

**I'm playing the devil's advocate** here, but how often do you use xref:link? Actually, how often do you even use svg in _hyperapp_-managed views? Perhaps we can also remove svg and provide it as a separate module? /cc @lukejacksonn 

> Either hyperapp's goal is to be the smallest library on earth, but then it has to assume it lacks several things or it applies a good balance between features / weight.

😆 Not really, but you are right I've definitely been hesitant about a lot of things and that's exactly why I've just started a little new project, still WIP, that for now just exposes hyperapp's vdom. 

Turbodom will be more tolerant of size changes and hopefully can evolve to outgrow our built-in engine and be used by other libraries that want to provide their own custom state management. /cc @zaceno @foxdonut @terkelg.

* https://github.com/turbodom/turbodom

ngryman:
@jbucaran Thanks for the explanation. To answer 👿's advocate, I use it pretty much everywhere.
But if we had a poll, I would assume that 10% of developers use it. So I can understand your point.

Then I only see 2 solutions to features like this in the future:
 - recipes: we could have a reference on things that are not supported builtin and provide recipes to users.
 - vdom plugins: some kind of plugins system that would leverage existing vdom implementation.

What do you think?

On a side note, nice initiative on Trubodom (nice name too haha) 👍

jbucaran:
@ngryman We might merge this though. I'd just like to hear more feedback from other folks. Turbodom will definitely need this (a more complete PR using removeAttributeNS perhaps). 

I'm also trying to figure out what would be a more efficient way to implement this PR.

jbucaran:
@ngryman What are vdom plugins?

ngryman:
> @ngryman What are vdom plugins?

Something like this: https://github.com/snabbdom/snabbdom#hooks. With such hooks, the community could provide additional features around `hyperapp` core. That's just an idea.

jbucaran:
@ngryman Maybe. Definitely something I want to have in some form on Turbodom.

lukejacksonn:
> What is the cost? ~10B minified gziped? What is the gain? It just vastly increase svg support for modern apps that use svg sprites.

I'm really in agreement here.. my hyperapp projects are the only projects that don't employ `<use>` tags, not because I don't prefer them anymore but because it isn't supported.

How much cleaner is:

```
const DeleteButton = () =>
  <button>
    <svg><use xlink:href="#icon-delete" /></svg>
  </button>
```

than the alternative:

```
const DeleteButton = () =>
  <button>
    <svg fill="#000000" height="24" viewBox="0 0 24 24" width="24" xmlns="http://www.w3.org/2000/svg">
      <path d="M0 0h24v24H0V0z" fill="none"/>
      <path d="M6 19c0 1.1.9 2 2 2h8c1.1 0 2-.9 2-2V7H6v12zm2.46-7.12l1.41-1.41L12 12.59l2.12-2.12 1.41 1.41L13.41 14l2.12 2.12-1.41 1.41L12 15.41l-2.12 2.12-1.41-1.41L10.59 14l-2.13-2.12zM15.5 4l-1-1h-5l-1 1H5v2h14V4z"/>
      <path d="M0 0h24v24H0z" fill="none"/>
    </svg>
  </button>
```

jbucaran:
@lukejacksonn Would you be able to use `oncreate` for this?

lukejacksonn:
Sure you can.. but that isn't how `html` is intended to be used imo. At the end of the day, the `use` element is just a tag (a tag that hyperapp doesn't support). Not supporting the `audio` tag for example, would not be acceptable. Why is not support this acceptable?

A bit like the whole accessible link discussion:

```
<a onclick={e => actions.router.go('/')}>Home</a>
```
VERSUS
```
<a href='/'>Home</a>
```

One is much cleaner and more inline with web standards.

ngryman:
I'm gonna make the devil's advocate too 😈 

## Recipes

If we choose to propose recipes to users, then the `oncreate` solution could be refactored with something like this (I'm not using jsx so the syntax may be wrong, but you get the idea):

```jsx
const Use = ({ href }) =>
  <use
    oncreate={e =>
      e.setAttributeNS(
        "http://www.w3.org/1999/xlink",
        "href",
        ${href}
      )}
  />

const DeleteButton = () =>
  <button>
    <svg>
      <Use href="#icon-delete" />
    </svg>
  </button>
```

## Plugin

If we choose to propose plugins (which is my prefered solution), we could do something like this (hypothetical code):

**index.js**
```jsx
import { app } from 'hyperapp'
import { Use } from 'hyperapp-plugin-svg'

app({
  // ...
  plugins: [Use]
})
```

**use.js**
```jsx
const Use = () => ({
  events: {
    create: (e) => {
      if (!e.hasAttribute('xlink:ref') return

      const href = e.getAttribute('xlink:href')
      e.removeAttribute('xlink:href')
      e.setAttributeNS(
        "http://www.w3.org/1999/xlink",
        "href",
        href
      )
    }
  }
})
```

Then you could simply declare the `use` tag like this:

**view.js**

```jsx
const DeleteButton = () =>
  <button>
    <svg>
      <use xlink:href="#icon-delete" />
    </svg>
  </button>
```

jbucaran:
@ngryman I think the following is bearable.

```
const Use = ({ href }) =>
  <use
    oncreate={e =>
      e.setAttributeNS(
        "http://www.w3.org/1999/xlink",
        "href",
        href
      )}
  />

const DeleteButton = () =>
  <button>
    <svg>
      <Use href="#icon-delete" />
    </svg>
  </button>
```

It's really that bad?

ngryman:
@jbucaran Yes, but we need to document it for users in order to avoid them lose tons of time. So were are going for recipes?

jbucaran:
@ngryman Yes, but we call them [custom tags](https://github.com/hyperapp/hyperapp/blob/master/docs/custom-tags.md).

I think the custom tags is the way to go. It's how you would go about creating a controlled input text, a link that calls `actions.router.go` on your behalf, etc.

jbucaran:
@zaceno What are your thoughts on this?

zaceno:
@jbucaran I'm not really sure. I think it kind of depends on the cost of implementing the *NS support. On the one hand, you can get pretty far without them, on the other hand, we're not supporting a standard attribute... but we can't support everything and remain a tiny simple vdom solution. 

I think generally, I'm pretty ok with having to use lifecycle methods for some things, but like @ngryman said, it should be clear in the docs what is not supported out of the box, and how to work around it.

ngryman:
@jbucaran Yes, by recipes I just mean a collection of proposed custom tags to know use cases. So the user just has to copy/paste it instead of posting an issue, etc...

jbucaran:
@ngryman Okay, let's start with that. Perhaps we can start adding a new recipes section to the docs? What would be other names instead of _recipes_ that you can think of? 🤔

lukejacksonn:
@ngryman I like that idea! @jbucaran could we perhaps create a repo `hyperapp/parts|recipes|components` or something along those lines that has all of these useful _custom-tags_ in it.. or is that going too far?

jbucaran:
@lukejacksonn Maybe we can just add "recipes" to the docs for now and think about creating a repo later? 🙏

lukejacksonn:
Sounds reasonable to me!

lukejacksonn:
I've changed my mind on this one (and would like to see this PR closed)! Can we reconsider adding this now we have more bytes to play with?

The `use` tag is fast becoming the de facto way of using svgs.. it allows for sprite sheeting and external sprite/symbol sheets.

An example from Google (MD) docs:
https://github.com/google/material-design-icons/tree/master/sprites

This is something I use in **every** project, it is a standard issue tag and makes is super duper easy to newcomers to add sprites to their projects.

What are the downsides?

jbucaran:
@ngryman @lukejacksonn So let's close here then! :)

lukejacksonn:
Merge and close?

jbucaran:
I am not sure about this implementation. Can we improve it?

lukejacksonn:
Here is how preact handles it https://github.com/developit/preact/blob/33fc697ac11762a1cb6e71e9847670d047af7ce5/src/dom/index.js#L81

What concerns you about this implementation.. bytes, perf, redundancy?

ngryman:
@jbucaran An easy improvement would be to handle `removeAttributeNS`. But as I mentioned before, nobody is likely to set a `xlink:href` on a `use` tag, and then remove this specific attribute, it does not really make sense.

`preact` supports all `xlink:*` attributes which is good. In the real world, only `xlink:ref` is widely used, so I think it's ok not to include any regexp check (if it saves some bytes).

jbucaran:
@ngryman Thanks, that makes sense. The only concern I have left is, quoting @leeoniya

> i'm 👍 on this from a practical standpoint. my current concern with adding it to domvm is the perf overhead of testing each attr name for namespace presence in a hot function. since it cannot be cheaply prefix-tested, running a regex of indexOf on every attr would greatly impact perf. there are some options tho.

> in domvm svg elems use their own factory, so perhaps plain vnodes can be saved from the overhead.

> a non-standard extra prefix would be cheap to test more generally: :xlink:href but is surprising
rather than supporting any namespaced, just explicitly test for xlink:href, which is cheap.
for what it's worth, i plan to add support for this in domvm, but have to figure out which mix of the above ideas would result in least perf degredation. i think custom prefixing is DOA due to non-uniformity :(

lukejacksonn:
> just explicitly test for xlink:href, which is cheap

Isn't this what is going on in this PR? Have we tested the perf of this branch?

jbucaran:
Maybe @leeoniya can rephrase what he himself wrote, I myself might be misunderstanding what was said. I think some commas are missing?

leeoniya:
this PR tests for "xlink:href" directly rather than adding broad namespaced attr support, so it should be fine.

ngryman:
@leeoniya One a side note if you do want to support all namespaced attribute keeping good perfs, you could have a quick/then/slow test:

```js
if (name[5] === ":" && (name !== (name = name.replace(/^xlink\:?/, ''))) {
  // ...
}
```

leeoniya:
sure, if `xlink` is a universal prefix for all namespaces ;)

jbucaran:
@lukejacksonn With the new `oninsert` method and standarization of `oncreate` it is predictable to implement this outside of core, would you reconsider or do you still want this merged? Just a friendly question. 🤔

lukejacksonn:
It was predictable before too right (through setting the attribute in `oncreate`, there was no need to wait for the node to be attached to the DOM).. hence my initial acceptance of having to implement a stateless component for `use`. More recently as I have been showing/teaching people HyperApp, there has need to be explanation as to:

- Why don't my svgs render? (because we don't support the xlink:href attribute)
- Why don't you support the attribute? (to which I say I don't actually know..)

So my question is.. why? What is the good reason not to be able to add icons, from a sprite sheet (a very common workflow for designer/developers) into your apps without looking up, implementing or being dependant on a custom stateless component.

If perf is not an issue then I consider this byte diff acceptable given the convenience factor.

jbucaran:
@lukejacksonn Fair enough! I was just checking again. BTW is this single check good enough / is xlink:href is all you need (like love)?

lukejacksonn:
`xlink:href` is all that is needed to support svg `use` so good enough for the time being ❤️ 👍

>  I have mixed feelings about leaving out removeAttributeNS, but i agree that its absence actually causing problems in real apps is pretty slim.

I'm with @leeoniya and @ngryman on this one too. If it becomes an apparent issue then fixing it wouldn't be so hard either.

lukejacksonn:
Holddddd on a minute! 🐴 

> https://developer.mozilla.org/en/docs/Web/SVG/Element/use

image

> https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href

image

https://codepen.io/lukejacksonn/pen/dRjGMZ

leeoniya:
i ran it through IE10 & IE11 in Win10x64. works fine!

https://codepen.io/anon/pen/KqBVWK

lukejacksonn:
I was looking for what was/not supported and found this issue which suggests Safari 9 does not

https://github.com/jonathantneal/svg4everybody/issues/128

NBD

leeoniya:
hmm, we're still at 2.76% Safari 9 traffic last 30 days, accounting for 2.08% revenue. it's not much, but it's not gone just yet.

ngryman:
@lukejacksonn: @jbucaran mentioned this before and I answered [that](https://www.chromestatus.com/feature/5760616295825408), support seems fuzzy.
Then there is adoption rate, bunch of user will use `xlink:href` for years and will eventually open issues again and again for this.
My last statement on this issue : for 10B that's too much talk (oh those tomatoes again... 🍅)

lukejacksonn:
@leeoniya hummmm, that is some quite good insight 🤔 so will it stay in domvm then?
@ngryman  I know.. but I'd love to just close/forget this PR and start using `href` from now on!

I though this revelation was going to make this decision easier!

leeoniya:
i tagged v3 a couple days ago, so i'll leave it in for now, deprecate in a point release and rip it out in v4 when the metrics drop < 1%, hopefully by 2018. we don't use xlink:href so it doesnt matter too much and for my own projecys i'll just use href from now on.

ngryman:
Adaption is always late regarding to standards. We are not the only users of HyperApp. We invest time on seeking for new standards and so on because we are library **authors**. But that's not the case of library **consumers**.

As I mentioned before, it's very likely that consumers will continue to use `xlink:href` and open issues about it. Standard is not the key metric here, usage is.

So let me rephrase my point of view: this PR buys us **tranquility**. And when `xlink:href` disappears (usage-wize) I would be glad to simply remove this chunk of code and gain those 10B back :)

jbucaran:
@ngryman That is a double edged sword argument, because if we support then feature, then we encourage people to use it. By not supporting it we force some users to make the change. I am not saying that is what would actually happen, only that your argument also has a seemingly equally valid counter. 😏:wink:

leeoniya:
> @ngryman That is a double edged sword argument, because if we support then feature, then we encourage people to use it too. By not supporting it we force encourage other users to make the switch to href.

came here to say the same thing. you gotta start with deprecating expectations.

ngryman:
@jbucaran @leeoniya I get the feeling, really. But let's not make feeling biased decisions. Let me explain one more time because I think it would be a wrong decision:

## Philosophy

I truly think that supporting a living standard / current usage is not *encouraging* it. It's just offering support for it. That's an important distinction. Moreoever the main role of HyperApp is not advocating support of latest DOM features, it is:

> a JavaScript library for building frontend applications - /HyperApp tagline/

And "fronted applications" are using `xlink:href` in 2017. The vast majority of other frameworks support it, so we support it (as advertised). The point is, it's not our role to take the decision to deprecate things, people will do naturally thanks to many influencers in the community that do this job very well. Let's only focus on what we advertise and do it well.

## User Experience

HyperApp is a mindblowing framework for some new comers (Elm architecture, different paradigms, ...). Please let's not add more noise to their brain. They will expect this to work (as advertised).
Also the project is getting a good momentum, anything that may slow down adoption should be avoided.

## Maintenance

Inevitably this topic will come again and again. And everytime we will have to tell the guy : "use SVG2 or here is a complex recipe". I personally dislike to defer problems, "lack of foresight leads to grief" is my mantra.

## The end

Finally let's put emotional stuff aside, here is a factual costs vs benefits:

| Costs | Benefits
|------|---------
| Roughly 10B increase | Full SVG support (as advertised!)
| | No more related issues in the future
| | Conformance with other frameworks / ease of adoption

This message will self destruct in 5 seconds .... 💥

leeoniya:
> people will do naturally thanks to many influencers in the community that do this job very well

allow me to vehemently disagree. i consider myself to be pretty well informed for modern web tech, and i'm old enough to have lived through the dark ages of IE6. until this thread, i was completely unaware that non-namespaced `href` in SVG was already supported back in *IE9*.

whatever channels you feel should have been adequate for disseminating this info have clearly failed in a spectacular way, in my experience.

jbucaran:
@ngryman Whether we decide to merge this or not, I'd like to address the elephant in the room first.

>  non-namespaced href in SVG was already supported back in IE9.

If non-namespaced href is supported in +IE10, then why is anyone using it?

jbucaran:
Whoops, I forgot to clarify. I didn't know non-namespaced href **was supported** on IE9.

leeoniya:
ok, sorry. i might be wrong. i tested in IE11's IE9 compat mode on Win10x64. those emulation modes are pretty loose approximations.

https://cdn.rawgit.com/leeoniya/1bcab5a7299e9669a599c76af1a3f8f0/raw/e7cfbb9aefd89a1742738bf68648ee64fe0263a8/svg2-href.html

via https://netrenderer.com/index.php

seems to work only in ie11

ngryman:
@leeoniya Oh IE6... and 7-8... 😂

You haven't been informed until this thread because support is simply fuzzy, so nobody communicates about it. I would not make a bet on non existing stats and drop something. Either we have precise stats, either it's safer to support it.

Nowadays IEs are some IOS Safari / old Androids stock browser. There are still a bunch of devices stuck with those versions out there, with no SVG 2 support I'm pretty sure about that. Making the choice to not support them is more than informing the developer: it's making the choice to crash (which is more unpleasant than informing people).

If even you (well informed library author) are not aware of this feature, the lambda user sure won't be. We agree on that.

He will want to test HyperApp, read the docs really quickly, create a new project, hack around, and suddenly: "why this f** SVG is not displaying??" He will lose a bunch of minutes, hours to try to figure it out? "I don't get it I see the DOM element, but nothing is displaying, hum...". He will continue trying to figure out what's hell is happening? His daughter comes, and get yield at for no reason 😢 
Then he comes to the bug tracker, angry as hell, and report a misspelled nonsense issue in capslock explaining how he found a virus in HyperApp 🐛. To which even more pissed we will answer: YOU SON OF 🐒 RTFM, DUPLICATE! USE SVG2 💥

Ok a little bit satirical I admit 😂 Stil, it highlights one key point I'm desperately trying to explain: a really bad user experience. I know because I've experienced it and opened this issue (at the difference I have no daughter and I'm quite peaceful ☮️).

My point and philosophy is: you don't inform people by not implementing things but by documenting things, more precisely by advocating not enforcing. You support every browser while conveying a message that this will soon be deprecated. Every modern successful framework / tool does that.

SkaterDad:
> Ok a little bit satirical I admit 😂 Stil, it highlights one key point I'm desperately trying to explain: a really bad user experience. I know because I've experienced it and opened this issue (at the difference I have no daughter and I'm quite peaceful ☮️).

I could yell at one of mine for you if you need?  🤣

leeoniya:
> you don't inform people by not implementing things but by documenting things, more precisely by advocating not enforcing

of course not, that's why dev builds (and TypeScript/Flow) exist [1] - to inform new users they're doing things wrong or sub-optimally and offer recommendations. if you feel this could be a common wtf, just add a warning that's stripped from production.

[1] https://github.com/leeoniya/domvm#devmode

ngryman:
Well fuck it, let's raise a 444 exception to SVG1 users. We won't see $1 revenue drop anyway. Oh wait, we are doing that for free. Damn... 🙄

leeoniya:
certainly don't let *me* pry your IE9/FF3.6/telnet:80/js-off crowd from you! :D

ngryman:
> certainly don't let me pry your IE9/FF3.6/telnet:80/js-off crowd from you! :D

Android Stock Browser / iOS / Windows Phone / Opera Mobile (other than US/CA users)? 😂

ngryman:
I've got lots of arguments against your vision but as you said it would go off topic and we could continue on and on until we reach the Godwin point ⚔️ 😂 
Let's agree we disagree on some points. For the rest, I let @jbucaran and other contributors decide.

leeoniya:
> Android Stock Browser / iOS / Windows Phone / Opera Mobile

you're comparing users who may not be able to afford a new device with those who cannot be bothered to install any modern browser despite half the web already being significantly broken for them. there's a non-zero [very non-zero] cost to supporting legacy tech. either it's worth it to you or it isnt on a project-by-project basis. Google agrees, Github agrees, even Microsoft agrees. i don't have some blanket vendetta against all low% user-agents, heh.

that being said, xlink:href support in hyperapp *is* pretty much free, so sure why not.

ngryman:
> i sincerely hope we're better than resorting to adolecent ad-hominem arguments :)

I hope so 👍

ngryman:
Sooooo, ladies and gentlemen, I'm proud to announce that this PR gets 3 awards today 🏆:
 - The oldest of HyperApp 👴 
 - The most discussed 🗣 
 - The less bytes 🏋️ 

🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 

Ok, more seriously, let's decide its faith. I'm ok with both merging it or not, really. Let's do a popular vote, so please 👍 or 👎 my comment. And we'll act in consequence.

jbucaran:
@ngryman I think there is no need too rush this PR. I'd rather focus on improving other areas. Fix the router, make sure the new mixin API is final, introduce node recycling, etc.

We can sort this one out before 1.0. 🙏:smile:

lukejacksonn:
I was going to come back to this last night.. as I was finding issues with using the `oncreate` method (sounds like some kind of birth control).

I am pretty sure there is some unintentional side effect of setting `xlink:href` in the element lifecycle methods..

```
const $use = href =>
  h("use", {
    oncreate: e =>
      e.setAttributeNS("http://www.w3.org/1999/xlink", "href", href)
  })

h('svg', {}, $use(s.playing ? '#pause' : '#play'))
```

When the playing state is toggled the `href` that gets passed to the `$use` changes.. but the AttributeNS is not reset (because oncreate is only fired once and the element appears not to change shape so no diff?). To fix this for all devices I have had to do..

```
const $use = href =>
  h("use", {
    href,
    onupdate: e =>
      e.setAttributeNS("http://www.w3.org/1999/xlink", "href", href)
  })
```

So it works.. but the amount of time it took me to test/stabilise was a bit ridiculous for something as simple as a play/pause button. Also this highlights the importance of the `onupdate` fix.. with 100 icons on screen, it would have been running `e.setAttributeNS` 100 times _every_ state change not just every time one icon changes.

They are my findings thus far.. I would like to see this PR closed too 😅

jbucaran:
@lukejacksonn What about using [`oninsert`](https://github.com/hyperapp/hyperapp/blob/master/docs/lifecycle-events.md#lifecycle-events)?

lukejacksonn:
I suspect `oninsert` will work the same as `oncreate` given that the `use` element won't be removed and reinserted.. just the `href` attribute changes 🤔

Still feels a bit hacky atm.

jbucaran:
@lukejacksonn Sorry, I am not sure, I am just pointing out that oncreate and onisert are very different.

andyrj:
I don't have skin in the game on this, but if it's small change I say we support it even if it is on the way out.  Fewer surprises fewer questions.

tunnckoCore:
Damn long thread. Can't read it fully, but it not worth so much writing, when most simpler thing that can be done [is to always use `createElementNS`](https://github.com/hyperapp/hyperapp/blob/master/src/app.js#L133-L135) and use always use `*AttributeNS` everywhere later. Correct me if i'm wrong?

This probably may be even smaller on bytes perspective.

ngryman:
@tunnckoCore AFAIK default namespace for `html` is `http://www.w3.org/1999/xhtml` so if we want to use `createElementNS` we would have to use this namespace, not `http://www.w3.org/1999/xlink`. But I don't really see the point of doing so honestly, because our only namespaced element is `svg`.

Back to the `setAttributeNS`, again our only need is for `xlink:href`, namespacing every attribute of every node explicitly feels weird for an exception. And performance-wize it's likely that `setAttribute` does less internally than `setAttributeNS`. So again I don't really see the benefit here.

That said, I totally agree that's way too much talk for such a small addition. But the later discussion was more around SVG 2 dropping `xlink` namespace and if we should support SVG 1.1, than the proposed implementation.

tunnckoCore:
> we would have to use this namespace

oh yea, that's what i meant.

> because our only namespaced element is `svg`.

not very true. 

> every tag inside svg should be with svg ns.
> – @bfred-it 

Better/full svg support includes things that some tags like above. I'm not very dom person, but last months i'm digging very much things around that and seen very much implementations and threads.

ref: https://github.com/vadimdemedes/dom-chef/issues/5

ngryman:
> because our only namespaced element is svg

> every tag inside svg should be with svg ns

Yes wrongly said: implicitly`svg` and descendants.

We detect any `svg` element and namespace it and all its descendants. So in the code we're only referring to `svg` explicitly, hence my sentence.

jbucaran:
@lukejacksonn @ngryman This is how we're doing it on the [website](https://github.com/picodom/website/blob/856d689105209200d4eb1d1f15287ec50c09f20c/src/index.js#L5). Doesn't seem so bad to me.

lukejacksonn:
I ran into more issues with this the other day.. here is a demo of one problem with using the `oncreate` method as demonstrated in [website](https://github.com/picodom/website/blob/856d689105209200d4eb1d1f15287ec50c09f20c/src/index.js#L5).

**Dynamically changing the `href` causes no update:**
https://codepen.io/lukejacksonn/pen/eENeGv

My hypothesis being that because `href` is not set like a _normal_ prop then to patch it doesn't look like the element has changed so.. no diff no (re)rendered?

If you add `href` as an actual prop as well as `oncreate` (which is advised best practice for modern browsers anyway) then the icon re-renders as it should in chrome, no re-render in iOS.

If you just use `href` (as it _should_ be!) then, fine in chrome, you get nothing on iOS.

If you change `oncreate` to `onupdate` then it works in chrome, but prevents the icon being rendered at all on iOS at first, it only appears on state the first change. Not to mention this function will get called on every update (cause we can't diff props that are functions), not so good if you have hundreds of icons on your page and rapidly changing state.

Adding `href`, `oncreate` and `onupdate` works as expected in chrome and on iOS but the component looks like this:

```
const Icon = href =>
  h('svg', {},
    h('use', {
      href,
      oncreate: e =>
        e.setAttributeNS(
          "http://www.w3.org/1999/xlink",
          "href",
          href
        ),
      onupdate: e =>
        e.setAttributeNS(
          "http://www.w3.org/1999/xlink",
          "href",
          href
        )
     })
   )
```

Which is nearly as many bytes as HyperApp's `h` function.. and gross.

The amount of time I have spent debugging this 😅 I'd give anything to be able to be able to just be able to just write:

```
h('svg', h('use', { href: '#happy' }))
```

and it just _work_ everywhere!

jbucaran:
Did you use oninsert?

lukejacksonn:
@jbucaran I did try setTimeout with `oncreate` but it makes no difference (wouldn't make sense if it did really), see the demo pen.

jbucaran:
@lukejacksonn It doesn't work because you are using the version of hyperapp that calls onupdate based on our internal props diffing. It will work in 0.11.0. 😄

jbucaran:
@lukejacksonn By the way, what is the alternative "modern" way to do it without `xlink:ref`?

lukejacksonn:
@jbucaran _It will work in 0.11.0_ ...I don't believe you 😛

The alternative _modern_ way to do this is simply using an `href` prop.. so this works in chrome:
```
h('svg', h('use', { href: '#happy' }))
```
but will render nothing on iOS (I am not attempting to check other browsers until at least these two work consistently)

lukejacksonn:
Also, given the nature of `onupdate` don't want to attach `onupdate` to all my icons.. `oncreate` already feels a bit 🤢 and I _believe_ are the cause of this bug

jbucaran:
@lukejacksonn `oncreate` runs only once, since you are using a button to toggle the href:link thing, you need to use onupdate. 

Having said that, I will deal with this issue a bit later and figure out the best thing to do

This might be a good chance to draw a line on the browser support issue.

jbucaran:
@ngryman @lukejacksonn 

- https://codepen.io/hyperapp/pen/preYMW?editors=0010
- https://css-tricks.com/creating-svg-icon-system-react/

jbucaran:
As Luke already pointed out, xlink:href has been:

> [**Deprecated since SVG 2**](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href)
>
> This feature has been removed from the Web standards. Though some browsers may still support it, it is in the process of being dropped. Avoid using it and update existing code if possible; see the compatibility table at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

The reality is that support for standalone href is still not great and using `xlink:href` is probably the best way to go in the near future.

While, I understand the situation, I don't want to introduce this feature on the basis that `xlink:href` is the only oddball from the other `xlink:` bunch, this change adds a bunch of extra bytes, and introduce another `if` check for **every** attribute that will be set (except functions). This PR does not take into account `removeAttributeNS` either. While, we could probably get away without it, might as well have it for consistency. 

If we didn't have another choice, I'd merge this, but we do have two.

#### Options

**I**. Embed the full SVG in the JavaScript like shown in the example above.

**II**. Create a `Use` component that calls `setAttributeNS` / `removeAttributeNS`.
```jsx
const Use = (props, children) => {
  const href = props.href
  const XLINK_NS = "http://www.w3.org/1999/xlink"

  const setLink = (element, oldProps) => {
    if (href) {
      element.setAttributeNS(XLINK_NS, "href", href)
    } else if (!oldProps || href !== oldProps.href) {
      element.removeAttributeNS(XLINK_NS, "href")
    }
  }

  return (
    <use {...props} oncreate={setLink} onupdate={setLink}>
      {children}
    </use>
  )
}

const Icon = href =>
  <svg>
    <Use href={href} />
  </svg>
```

I am sorry this PR has been open for so long and would like to thank @ngryman for his patience. I also apologize to those that were expecting me to merge this after #317. 🙇 

/cc @ngryman @lukejacksonn

ngryman:
@jbucaran It's ok I got [3 rewards](https://github.com/hyperapp/hyperapp/pull/211#issuecomment-315024886) with this PR 😂

jbucaran:
For posterity. 😆 

> The oldest of HyperApp 👴
> The most discussed 🗣
> The less bytes 🏋️

Issue #14, opened 2017-02-03
Server Rendering?


dodekeract:
## The Problem

Server-side rendering mostly improves the following aspects of web-applications. They're arguably the most important ones:

1. Time to load. The client can already start rendering without even loading the `.js`
2. Search engine optimization. Most SEs (notable exception: Google) don't parse JavaScript-only websites or rank them down. Responding with a "real" HTML solves this perfectly.
3. Especially reduces initial load times on mobile.
4. Greatly improve the experience for people that disabled javascript. Especially given that `hyperapp`'s router uses plain links that work on the server-side too.

---

## Proposed Usage

1. Create an `app()` that works on the client-side.
2. Require said `app()` on the server-side
3. Render it somehow (maybe with a `hyperapp/server`, so that the client-version isn't polluted )
4. Get a string or buffer back
5. Return said buffer via HTTP/etc.

So *basically, the same as react*. :smile:

## Notable Challenges

- `Router` has to be able to run on the server-side. I suggest we add a `state.router.url` property to the `state` for both browser and server usage. If `state.router.url` is defined, the router should always make sure to set the browser URL to that URL too.

---

## Bonus feature?

Just thought about file-sizes. Wouldn't it be possible to somehow re-use the already rendered HTML version to make the javascript components even smaller? Maybe gzip already fixes that if you bundle the JS as well. Not sure.

---

Is this possible or planned? Would be awesome if this could completely replace React.

jbucaran:
@dodekeract Planned, if not already possible!

But how would that look like? Can you describe a simple scenario or usecase?

tunnckoCore:
This was my question previous evening (sun is rising up now here at Bulgaria) too.

SkaterDad:
`snabbdom/snabbdom-to-html` would seem to be a good starting point for implementation.  If you're open to it, I could try to hack something together.

It would also be a bonus if streaming SSR could be implemented, similar to what Vue 2.0 is doing.  Their implementation is a bit complex due to their component design, but it could serve as inspiration.  I guess it's possible that `hyperapp` is so lightweight that you won't benefit from streaming, though.

This is a really intriguing project!  95% of the reason I'm currently trying to implement my project in Vue was the promise of easy, high performance server rendering.  But... all the extra baggage is making my eyes wander back to small, more functional approaches like `hyperapp`.

jbucaran:
> If you're open to it, I could try to hack something together.

**Yes**,  absolutely!

[...]

PR #212, opened 2017-05-15
Render virtual DOM node to HTML string

benjaminj6:
#14 

Maybe a starting point rather than full implementation for server side rendering as it currently only spits out static html. 

The other thing I could see is to put it in a separate package so the client-side bundle size doesn't suffer. I'd love to hear what the preferred path forward is for that.

Let me know your thoughts!


AutoSponge:
I didn't see `key` being omitted from the string renderer. I also think dataset may need to be handled explicitly. There may be other edge cases I'm not thinking of.

jbucaran:
@AutoSponge Thanks!

@benjaminj6 Good job! With regards to:

> The other thing I could see is to put it in a separate package so the client-side bundle size doesn't suffer. I'd love to hear what the preferred path forward is for that.

I think definitely put it in a separate package is the way to go.

/cc @dodekeract @lukejacksonn @selfup

dodekeract:
@jbucaran React handles this by letting you import `react-dom/server`. I see no reason why we can't do the same here.

jbucaran:
@dodekeract Ah, that makes sense too. 

@benjaminj6 I think this means we don't need the changes in index.js.

benjaminj6:
@jbucaran @dodekeract That makes total sense, will take those changes out. While we're at it, I was thinking about file names this morning and I wonder if `hyperapp/server` would be preferable. Let me know and I'll go ahead and change that too.

jbucaran:
@benjaminj6 @dodekeract What do you think? I always thought `toString` was the best.

benjaminj6:
I'd be happy with `toString`

dodekeract:
@jbucaran You're forgetting that server rendering doesn't mean "offer a toString function". It means that the whole app can run on the server-side. So we need to **at least** accept state and fire `load`.

jbucaran:
@dodekeract Actually, I have no idea. 😅 

If you want to take care of this issue and sort it out, please do! 👋

dodekeract:
@jbucaran Well, we do need a `toString` function for sure. But we also need to take care of the rest.

Essentially, when server-rendering the only thing that should change is the `app()` function. It should render to a string instead of the DOM, but it should still work the same. E.g. contain actions, etc..

The only problem here is that we'll need to make sure to never leak memory. That's fatal on a server.

To begin working on this we basically only need `app({state, view})`. Then, we need to find a solution for the router. My suggestion regarding that still is to include the URL-state in `state.router.url`. I think now you're slowly noticing the point I talked to you about regarding "all state should be in `state`, URL gets rendered from `state`, not the other way around".

In the long term, we need a complete drop-in replacement for `app()`, which accepts `state`, `view`, `events` (load is probably important?), `actions`, `plugins`.

jbucaran:
@dodekeract I thought a toString was good enough. If we're going to go this route, might as well create a separate repo.

benjaminj6:
@dodekeract @jbucaran 

I think a `toString` function is still valuable/needed — especially with the rise of static site generators (the whole [JAMstack](https://jamstack.org/) thing comes to mind).

But I also agree that eventually we should implement true ssr, perhaps in a separate repo. Especially if we want people to start using `hyperapp` more 😎

jbucaran:
Thanks @benjaminj6! You are probably right. In that case, I think we should go ahead and create a new repo and take care of it proper. Would you like to take the lead?

I can send you an org invite so you can create it yourself.

benjaminj6:
@jbucaran I'd love to! Shoot me that invite and I'll get on it .

benjaminj6:
Cool! Hopefully gonna be able to get it done this weekend now that I have a little bit of free time 🎉

benjaminj6:
Ok so we've got a [repo](https://github.com/hyperapp/hyperapp-server) for ssr set up 🎉

Let's move this discussion over there. I just started by moving the `toString` into it for static rendering and opening an [issue](https://github.com/hyperapp/hyperapp-server/issues/2) regarding a server-side implementation of `app`. I still need to do a little more research/need a little input on _how_ to go about that though 😎

Issue #176, opened 2017-04-02
Remove Router from core?


Facel3ss1:
Hello!
Just looking through HyperApp and I am very impressed! 👍 I am just wondering why `Router` is in the core package that everyone downloads because it's just a plugin. Wouldn't it be better to separate Router from core and put Router and similar plugins elsewhere? It's just a question, as I was wondering about community plugins/extensibility. E.g. The core hyperapp repo is just `h` and `app` and `Router` and other similar plugins could be in other repos and the user could choose which one they install alongside hyperapp.
```
npm install hyperapp
npm install hyperapp-router
...
```

What do you think?

dodekeract:
@Facel3ss1 we thought about this a lot. Main reason is primarily that this doesn't have any major disadvantages, while it maintains the "one system" aspect of hyperapp.

Look at react, there are 20 different routers, 20 different state manangement libraries and they don't integrate nicely with each other. Hyperapp (while not forcing you to use our router) gets rid of most of these problems by providing it out-of-the-box.

lukejacksonn:
humm, I'm still torn on this one.. there are pros and cons to abstraction imho.

Reasons for abstraction:
- Makes it very clear that it is in fact just a plugin, not core (blurred lines atm)
- Makes it easier to focus on and develop in parallel with core (not blocked by core issues/PRs)
- Makes it easier for others to fork/copy/experiment/adapt to suit their own needs
- Makes for less of a locked in feel (preempting _why you ship me a router I don't want_ comments)
- Makes for a good plugin example (encouraging people to build/install/import plugins is good?)
- Makes for a smaller library overall (there is the tree shaking argument but not everyone does)

Reasons not to abstract:
- Makes for a more simple and out-of-the box solution for multipage apps
- Makes for one less `npm install` or `<script>` tag include
- Makes for one centralized place for hyperapp issues/PRs

Anyone got anything to add to either of these lists?

As I see it, nearly the best way to handle it would be to create `hyperapp/router` repo. That way it is made very clear that this is the recommended general purpose router _but_ we understand that frontend routing is non trivial and encourage adaptation and feedback. Although it pains me so to not ship frontend routing with core.. core does exactly as advertised without the router.

 ¯\_(ツ)_/¯

dodekeract:
@lukejacksonn You dropped something: \\

jbucaran:
Hi @Facel3ss1 👋

[...]

PR #214, opened 2017-05-18
Make router deterministic

igl:
Using route paths as keys is not a reliable way of describing routes if you later want determinstic server side rendering + client side rehydration or fallthroughs.

Therefore i changed it to a object[] with the objects describing the route:

```js
... view: [{ path: "/:foo", view: state => h("div", {}, state.router.params.foo) }]
```

I changed the tests and docs too. Maybe i missed something or you want to change the docs further. I am not a good writer ;) Review welcome.

FYI: Objects are an unordered collection of properties. The order is purely accidental by your javascript engine doing it's optimizations. If you create a similar object with additional keys you can see what happens.

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -25,23 +25,29 @@ export default function(app) {
         }
       },
       render: function(state, actions, view) {
-        return view[state.router.match]
+        for (var i = 0, len = view.length; i < len; i++) {
+          var route = view[i]
+          if (route.path === state.router.match) {
+            return route.view
+          }
+        }
       }
     }
   }
 
-  function match(data) {
+  function match(path) {
     var match
     var params = {}
 
-    for (var route in app.view) {
+    for (var i = 0, len = app.view.length; i < len; i++) {
+      var routePath = app.view[i].path
       var keys = []
 
-      if (!match && route !== "*") {
-        data.replace(
+      if (!match && routePath !== "*") {
+        path.replace(
           RegExp(
             "^" +
-              route
+              routePath
                 .replace(/\//g, "\\/")
                 .replace(/:([\w]+)/g, function(_, key) {
                   keys.push(key)
@@ -54,7 +60,7 @@ export default function(app) {
             for (var i = 1; i < arguments.length - 2; ) {
               params[keys.shift()] = arguments[i++]
             }
-            match = route
+            match = routePath
           }
         )
       }
--- a/test/router.test.js
+++ b/test/router.test.js
@@ -13,9 +13,9 @@ beforeEach(() => {
 
 test("/", () => {
   app({
-    view: {
-      "/": state => h("div", {}, "foo")
-    },
+    view: [
+      { path: "/", view: state => h("div", {}, "foo")}
+    ],
     plugins: [Router]
   })
 
@@ -27,9 +27,9 @@ test("/", () => {
 
 test("*", () => {
   app({
-    view: {
-      "*": state => h("div", {}, "foo")
-    },
+    view: [
+      { path: "*", view: state => h("div", {}, "foo") }
+    ],
     plugins: [Router],
     events: {
       loaded: (state, actions) => {
@@ -59,9 +59,9 @@ test("routes", () => {
   window.location.pathname = "/foo/bar/baz"
 
   app({
-    view: {
-      "/foo/bar/baz": state => h("div", {}, "foo", "bar", "baz")
-    },
+    view: [
+      { path: "/foo/bar/baz", view: state => h("div", {}, "foo", "bar", "baz") }
+    ],
     plugins: [Router]
   })
 
@@ -76,8 +76,9 @@ test("route params", () => {
   window.location.pathname = "/be_ep/bOp/b00p"
 
   app({
-    view: {
-      "/:foo/:bar/:baz": state =>
+    view: [{
+      path: "/:foo/:bar/:baz",
+      view: state =>
         h(
           "ul",
           {},
@@ -85,7 +86,7 @@ test("route params", () => {
             h("li", {}, `${key}:${state.router.params[key]}`)
           )
         )
-    },
+    }],
     plugins: [Router]
   })
 
@@ -102,8 +103,9 @@ test("route params separated by a dash", () => {
   window.location.pathname = "/beep-bop-boop"
 
   app({
-    view: {
-      "/:foo-:bar-:baz": state =>
+    view: [{
+      path: "/:foo-:bar-:baz",
+      view: state =>
         h(
           "ul",
           {},
@@ -111,7 +113,7 @@ test("route params separated by a dash", () => {
             h("li", {}, `${key}:${state.router.params[key]}`)
           )
         )
-    },
+    }],
     plugins: [Router]
   })
 
@@ -128,9 +130,10 @@ test("routes with dashes into a single param key", () => {
   window.location.pathname = "/beep-bop-boop"
 
   app({
-    view: {
-      "/:foo": state => h("div", {}, state.router.params.foo)
-    },
+    view: [{
+      path: "/:foo",
+      view: state => h("div", {}, state.router.params.foo)
+    }],
     plugins: [Router]
   })
 
@@ -143,10 +146,13 @@ test("routes with dashes into a single param key", () => {
 
 test("popstate", () => {
   app({
-    view: {
-      "/": state => "",
-      "/foo": state => h("div", {}, "foo")
-    },
+    view: [{
+      path: "/",
+      view: state => "",
+    }, {
+      path: "/foo",
+      view: state => h("div", {}, "foo")
+    }],
     plugins: [Router]
   })
 
@@ -168,12 +174,12 @@ test("go", () => {
     expect(url).toMatch(/^\/(foo|bar|baz)$/)
 
   app({
-    view: {
-      "/": state => "",
-      "/foo": state => h("div", {}, "foo"),
-      "/bar": state => h("div", {}, "bar"),
-      "/baz": state => h("div", {}, "baz")
-    },
+    view: [
+      { path: "/", view: state => "" },
+      { path: "/foo", view: state => h("div", {}, "foo") },
+      { path: "/bar", view: state => h("div", {}, "bar") },
+      { path: "/baz", view: state => h("div", {}, "baz") }
+    ],
     plugins: [Router],
     events: {
       loaded: (state, actions) => {


dodekeract:
LGTM. Definitely better than an object. Didn't check the source yet though.

jbucaran:
Thanks @igl, this looks interesting!

> if you later want deterministic server side rendering + client side rehydration or fallthroughs.

Can you explain or provide a link to a document/resource that explains the above? The current API design is at first glance simpler and less verbose than the one proposed here, but not by far.

While we're at it, @dodekeract we should revisit some of our earlier router designs. At this point, and with #176 on the pipeline (though not guaranteed to happen just yet), I am not opposed to a different router API.

igl:
> http://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order

I learned this the hard way: fixing flakey tests ;)

igl:
The router is very primitive. But if you want to stay below 1kb... What else can you do.

If you want a more elaborate router with params you could use something like [path-to-regexp](https://github.com/pillarjs/path-to-regexp). But its around 7kb (minified)

#176 looks interesting. You could do the currently popular "mono-repo" with multiple packages that have first class support but are still exchangeable

jbucaran:
@igl I am sure the router is primitive, otherwise everyone else's 10kb router is doing something deeply wrong. But one thing this router does have is params. 

> does-javascript-guarantee-object-property-order

🤔 Hmm, okay that's a start. What is client side rehydration? There are no fallthroughs, only "*", but that is treated in a special way so it works correctly.

igl:
> But one thing this router does have is params

But that means you could have fallthroughs like:

- `/documents`
- `/:document_id`

 right?

Rehydration is a term coming from doing server-side-rendering with react.

Once you do server rendering you can have the client pick up the existing html structure and create the shadow-dom without re-rendering all the nodes. It's very important that the rendering is deterministic in a cross browser/cross js-engine environment.

React will print a warning in dev mode if it finds server-rendered html that does not match with what it expects to create in the client.

lukejacksonn:
I'm swaying in favour of this, it is a breaking change but makes it easier to extend in the future. If https://github.com/hyperapp/hyperapp/issues/176 happened then size wouldn't be so much of issue either.

jbucaran:
@lukejacksonn My gut tells me this is a good thing, but I still don't understand why I want this.

lukejacksonn:
I think the fact your route is now an object not just a function makes things much more interesting. What else could be associated with a route other than path and view? Perhaps things like guards, beforeenter and onleave or fetch. Just thinking out loud here. Like you say, it feels right.. looks quite good too:

```
const routes = [{
  path: '/',
  view: (s,a,d) => 'Home',
}, {
  path: '*',
  view: (s,a,d) => 'Lost',
}]

app({
  view: routes,
  plugins: [Router],
})
```

jbucaran:
> What else could be associated with a route other than path and view?

Yes, but the current design is also simple or simpler. What's the problem with the current deisng again?

lukejacksonn:
The problem at the moment is that an objects order is not deterministic, whereas an array ensures order.

jbucaran:
@lukejacksonn Good. I can see that would be an issue when you have a list and want to ensure order. But why is that a problem here?

dodekeract:
@jbucaran Because we have a list and want to ensure a particular order.

> `/documents`
> `/:document_id`

Depending on the order, different routes get matched for `/documents`.

jbucaran:
@dodekeract That made sense. Now I understand the issue. Thanks! 🙇 

Now, perhaps we can try matching /: routes first and then proceed as normal?

The problem I see with this approach is that an app with a lot of routes will run a loop inside the critical `render` handler and using a dictionary (current approach) is much faster.

dodekeract:
@jbucaran Au contraire.

Currently, we [loop over an object using `for ... in ...`](https://github.com/hyperapp/hyperapp/blob/master/src/router.js#L37). The order that we loop in is random/implementation-dependant. It's also [**really slow**](https://jsperf.com/for-loop-vs-for-in-loop).

Since the order is deterministic in an array apps can actually micro-optimize their routes by adding the "most often used" ones in the lowest index position, granting even more performance.

Dictionaries/Objects **can** be faster, but that's only if you access a known key, not if you try to find one key that matches your route.

If you're worried about the loop inside render to find the appropriate render function, there is no reason why we can't add an "index" key to the state to make that access even faster than object-access.

src/router.js
```diff
@@ -25,23 +25,29 @@ export default function(app) {
         }
       },
       render: function(state, actions, view) {
-        return view[state.router.match]
+        for (var i = 0, len = view.length; i < len; i++) {
```
jbucaran: I was talking about this loop here @dodekeract.

dodekeract: So am I, in the last sentence of my last comment.

lukejacksonn: I thought that is what you meant.. like `view[state.router. matchIndex]`

igl:
> If you're worried about having a loop inside render to find the appropriate render function, there is no reason why we can't add an "index" key to the state to make that access even faster than object-access.

I agree the render loop is a unfortunate side effect of `state.match` only saving the matched path. The best would probably be to save the ref the whole matched route object instead so no lookup is required at all.

dodekeract:
@igl I'd save the index instead.

jbucaran:
> @igl The best would probably be to save the ref the whole matched route object instead so no lookup is required at all.

I'd prefer this approach. 

@dodekeract What would storing the index be superior to storing the object?

igl:
Right now, match() spits out `*` on miss even if there is no '*' route defined.
Should it throw an error or just silently fail and render nothing?

jbucaran:
> Right now, match() spits out * on miss even if there is no '*' route defined.

Correct.

> Should it throw an error or just silently fail and render nothing?

I think the current behavior, to fail, is okay.

dodekeract:
@jbucaran Storing the index is superior because it can actually be stringified. Storing the object would mean "storing h-functions", which means the state is — by design — not stringifiable.

igl:
@dodekeract I see,  stringifiable state is neat for SSR, localstore caching and the like.

I already started doing some changes and continue working on it this weekend. Ill keep you posted here.

jbucaran:
@dodekeract @igl I am basically (probably) okay about storing the index or the view, but I'd like to see both implementations before I know for sure. I may play with this a bit myself to see how either approach works in action.

jbucaran:
@igl Hey, hi! If you have any updates would you share them? 🙏

igl:
Sorry guys. I am swamped with work right now. Will come back to this ASAP - i have not forgotten this ;)

jbucaran:
Revisiting this issue, the problem is that the router treats the view as a key/value store.

```
localhost:8080/documents
```
```js
{
  "/documents": 0,
  "/:foo": 1
}
```

In the above example, the matched view may be 0 or 1 depending on which route key was checked first.

jbucaran:
Here's yet another option: instead of using an object of key/value pairs, we could express routes using a function and this pattern:

```jsx
app({
  state: 0,
  view: Router.default(() => {})
    .route("/", () => {})
    .route("/documents", () => {})
    .route("/:document_id", () => {})
    .route("/:foo", () => {})
})
```

@lukejacksonn @igl @dodekeract /cc

jbucaran:
Alas, there's a way like this:

```js
app({
  view: Router
    .default(() => {})
    .route("/", defaultView)
    .route("/documents", documentsView)
    .route("/:document_id", documentIDView),
  plugins: [Router] /* This would work */
})
```

So instead of asking users to write views like this:

```js
... view: [{ path: "/:foo", view: state => h("div", {}, state.router.params.foo) }]
```

We make Router's `.default`, `.route`, etc., return a similar object. 

```
[["/", defaultView], ["/:documents", documentsView], ["/:document_id", documentIDView]]
```

jbucaran:
I think we can update this PR to fix the non-determinism issue and improve the Router use experience in one shot.

Currently, we use the Router like this:

```
const { h, app, Router } from "hyperapp"
```

and to "activate it":

```
plugins: [Router]
```

But `Router` is not used anywhere else in the code, so it is a bit magical or too indirect. The proposal below involves a bit more boilerplate, but it makes the router role more obvious.

```js
app({
  ...
  plugins: [Router],
  view: Router
    .route("*", () => {})
    .route("/", defaultView)
    .route("/documents", documentsView)
    .route("/:document_id", documentIDView)
})
```

jamen:
@jbucaran There is [`wayfarer`](https://npmjs.com/wayfarer) and [`sheet-router`](https://npmjs.com/sheet-router) you might have heard of.  The usage you describe sounds similar so I thought I would mention them to take some ideas.

Wayfarer is a generic router that has the method syntax similar to what you describe.  Then `sheet-router` uses it for front-end routing, and exposes the array API you show in the end.

Are you saying to use `Router.route`/`Router.default` as sugar for the array routes? Because I prefer the arrays like with `sheet-router`. Using functions looks unnecessary to me.  I will just throw that opinion in there, feel free to take or not.

If you implementation ends up using methods like that, maybe also consider how `wayfarer` names theirs: `router.on(route)`, `router(route)`

jbucaran:
@jamen Thanks! Yes, I want to use a "monadic" / chainable pattern, .route(..).route(..) with functions. Basically, instead of @igl's array style:

```
...
view: [
  { 
    path: "*", 
    view: wildcardView
  },
  { 
    path: "/:foo", 
    view: fooView
  },
  { 
    path: "/:bar", 
    view: barView
  }
]
```

like this:

```js
view: Router
  .route("*", wildcardView)
  .route("/foo", fooView)
  .route("/bar", barView)
```

And behind the scenes, that will translate to the following array:

```
[["*", wildcardView], ["/foo", fooView], ["/bar", barView]]
```

So, technically, you could just pass the array yourself, but I'll promote the more verbose style as it more provides structure and readability IMO (I just think it looks better).

igl:
Interesting. This could open up support for middlewares that can transform state or simply parse url parameters when needed and not always.

```js
[["/foo", parseQueryString, fooView], ... ]
```

Using objects instead of arrays however has the advantage of being extendible for features like this:

match only when trailing slash exists:
```js
{ path: "/foo/", view: fooView, exact: true } 
```

redirect:
```js
{ path: "/foo/", redirect: "/bar" } // redirect
```

set status codes if rendering on the server:
```js
{ path: "/foo", view: fooView },
{ path: "*", view: error404View, status: 404 }
```

I am really just taking hints from react-router here, which also has support for exclusively and inclusive route matching: [Switch](https://reacttraining.com/react-router/web/api/Switch)

I think there is probably room for a "Simple Router" and a advanced router if you want to keep footprint of hyperapp as small as possible.

I would prefer declaring the routes as literals so:
- less api = easier.
- i can write a route factory myself without having to learn/know about intrinsic details of a chaining API.
- switching to a advanced router package you simply replace the dependency and not the definitions and gives you more features.

jbucaran:
> I think there is probably room for a "Simple Router" and an advanced router if you want to keep the footprint of hyperapp as small as possible.

I agree with this.

> I can write a route factory myself without having to learn/know about intrinsic details of a chaining API.

The API would be .on or .route, whichever we go with (most likely .on since it's short and cute), but looking at both code examples, the array doesn't look bad at all if it's just arrays.

```js

app({
  state: {
    count: 0
  },
  view: Router
    .on("/", defaultView)
    .on("/dashboard", dashboardView)
    .on("/billing", billingView),
  actions: {},
  plugins: [Router]
})
```
vs

```js
app({
  state: {
    count: 0
  },
  view: [
    ["/", defaultView],
    ["/dashboard", dashboardView],
    ["/billing", billingView]
  ]
})
```

@jamen BTW I'm not proposing something as sophisticated as sheet-router, i.e., no nested routes.

jbucaran:
@igl A router like react-router would be awesome to have as a separate plugin too. HyperSwitch, or something.

jbucaran:
@jamen @igl Thank you for the input! So, going with this:

```js
app({
  view: [
    ["/", defaultView],
    ["/dashboard", dashboardView],
    ["/billing", billingView],
    ["*", errorView]
  ],
  plugins: [Router]
})
```

SkaterDad:
@jbucaran Could the `router.on()` stuff be embedded inside other v-dom elements?  That's the most limiting part of the current implementation, in my mind.  It's easily solvable by wrapping all the route views in common function, of course, but I wonder how other plugins would interact with the router's view take-over.

You could also do something more `elm`-like and embrace switch statements in the user code?  Instead of the `state.router.match` being the vdom function itself, it would return the route path definition (ex: `"/home", "user/:id"`), and you could switch on it.

jbucaran:
> You could also do something more elm-like and embrace switch statements in the user code? Instead of the state.router.match being the vdom function itself, it would return the route path definition (ex: "/home", "user/:id"), and you could switch on it.

I am not sure I am following. `state.router.match` is a string with the matched route. There's `actions.router.match`, but it's for internal use only, so it's not documented.

Yes, we're going arrays now! I've updated the docs too. 🎉

SkaterDad:
> I am not sure I am following. state.router.match is a string with the matched route. There's actions.router.match, but it's for internal use only, so it's not documented.

Good point!  I've been playing around with a customized version of the router (with simplistic query param support that suits my backend), so my memory failed me there.  I think I'll hold off on any other suggestions until I have a fully formed thought :sweat_smile:

jbucaran:
@igl I will push my changes from #231 on top of this and merge so that you are credited for introducing this change! Thanks for your patience and making HyperApp better! 👍

PR #220, opened 2017-05-30
Fix setTimeout example

t8g:
setTimeout returns an integer and so addOneDelayed updates the state, it should'nt.


jbucaran:
@t8g You are right! Thanks.

PR #223, opened 2017-06-02
Ignore package-lock.json

benjaminj6:
Since `yarn.lock` is ignored already, and now npm 5.0 creates a `package-lock.json` that functions in a similar way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 # Logs
 *.log*
 yarn.lock
+package-lock.json
 
 # Misc
 sandbox


jbucaran:
LGTM.

lazarljubenovic:
Has ignoring yarn.lock been discussed? It shouldn't be. https://github.com/yarnpkg/yarn/issues/1583

jbucaran:
> Has ignoring yarn.lock been discussed? 

I can't recall it has been discussed, nope.

> It shouldn't be. yarnpkg/yarn#1583

Yep, I think you are right, even for libraries. I also found this:
- https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all

Luckily, this project is extremely simple (that's on purpose) and development deps are few (it's unlikely that one of them will break hyperapp suddenly), so I think we don't have to worry about it this time.

dodekeract:
@jbucaran Usually, I would add these files, but since we don't have any real dependencies I wouldn't bother with them. They make no sense in this case.

PR #224, opened 2017-06-03
Fix input type=text selection reset.

jbucaran:
ᕕ( ᐛ )ᕗ 🎉

Related #137.
Ping @mrkstwrt.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -145,11 +145,13 @@ export default function(app) {
   function updateElementData(element, oldData, data) {
     for (var name in merge(oldData, data)) {
       var value = data[name]
-      var oldValue = oldData[name]
+      var oldValue = name === "value" || name === "checked"
+        ? element[name]
+        : oldData[name]
 
       if (name === "onupdate") {
         value(element)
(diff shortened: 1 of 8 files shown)


src/app.js
```diff
@@ -145,11 +145,13 @@ export default function(app) {
   function updateElementData(element, oldData, data) {
     for (var name in merge(oldData, data)) {
       var value = data[name]
-      var oldValue = oldData[name]
+      var oldValue = name === "value" || name === "checked"
```
dodekeract: Why check for `checked` and `value`? Any explanation?

jbucaran: These are the two attributes whose actual value change when you interact with the component. So, when diffing an element's data, we want to look at the real DOM (not the stuff in oldData).

zaceno:
@jbucaran -- will this fix be coming to picodom as well (or is it not necessary?)

jbucaran:
@zaceno Yeah, already fixed in picodom. https://github.com/picodom/picodom/issues/6

PR #229, opened 2017-06-07
Docs:  Fix onremove lifecycle example

SkaterDad:
`element.parent` does not exist.

Copied example from: 
https://developer.mozilla.org/en-US/docs/Web/API/Node/removeChild

### DIFF
--- a/docs/lifecycle-events.md
+++ b/docs/lifecycle-events.md
@@ -25,7 +25,9 @@ Fired before the element is removed from the DOM.
 Note that when using this event you are responsible for removing the element yourself.
 
 ```js
-element.parent.removeChild(element)
+if (element.parentNode) {
+  element.parentNode.removeChild(element);
+}
 ```
 
 ## Example


docs/lifecycle-events.md
```diff
@@ -25,7 +25,9 @@ Fired before the element is removed from the DOM.
 Note that when using this event you are responsible for removing the element yourself.
 
 ```js
-element.parent.removeChild(element)
+if (element.parentNode) {
```
jbucaran: I think `if (element.parentNode) {` is not needed because the element is guaranteed to have a parent (the root property, which defaults to document.body if none is given).

jbucaran: @SkaterDad What do you think?

jbucaran:
I left you a comment. Other than that, LGTM!

docs/lifecycle-events.md
SkaterDad: We can certainly remove that guard statement, I just copied it from MDN.

I was wondering if it could still be useful, though, depending on what you're doing in the `onremove` hook.? Is there a chance that a node would be orphaned if you're doing something async in `onremove`, and patch removes it parent?  Or will the parent node stick around until the child gets removed?

For example, what would happen to the `div` if i delay the `h1` removal on a route change?
```
view: {
  '/': () => h('div', {}, [
    h('h1', {onremove: somethingAsyncThatTakesAwhile}, 'Test')
  ]),
  '/about': () => h('p', {}, 'About page')
}
```

jbucaran: @SkaterDad So, an element other than the root could have no parent?

SkaterDad: I did some more experimenting, and have a better idea of when `onremove` is called.

It seems like you could never have an orphaned child node calling `onremove` to begin with.  `onremove` only gets called when the parent node is staying the same, which makes perfect sense in hindsight.

For example, this will never call the h2's  `onremove` hook, since the `<p>` is changing to a `<div>`.    If both containers were `<div>`, the `onremove` will get called.
```javascript
import {app, h} from 'hyperapp'

function delayedOnRemove(element) {
  console.log('Immediate parent check')
  console.log(element.parentNode)
  setTimeout(() => {
  console.log('Delayed parent check')
    console.log(element.parentNode)
    element.parentNode.removeChild(element)
  }, 3000)
}

app({
  state: 0,
  view: (state, actions) => h('main', null, [
    h('div', {}, [
      h('button', {onclick: () => actions.add() }, 'Add'),
      h('button', {onclick: () => actions.sub() }, 'Sub'),
      h('span', {}, 'Current value is ' + state)
    ]),
    state > 1
    ? h('div', {}, [
      h('h1', { key: '1' }, 'State is greater than 1')
    ])
    : h('p', {}, [
      h('h2', { onremove: delayedOnRemove }, 'State is less than or equal to 1')
    ])
  ]),
  actions: {
    add: state => state + 1,
    sub: state => state - 1
  }
})

```

jbucaran: Right, onremove is not called when elements are replaced. This has been poorly discussed before and no agreement has been reached yet. 😅

jbucaran: @SkaterDad BTW, I wasn't sure how to document this feature, but the second argument to the onremove function is a _function_ that can remove the element, so we can actually do:

```js
onremove(el, removeElement) => removeElement()
```

SkaterDad: Ah, nice!  That's certainly convenient.

I suppose you could call it "cleanup" or "done", and most people will understand.

PR #230, opened 2017-06-07
Add `.` to the allowed characters in params

fvj:
I needed to capture dots in params, so I added them. 

Maybe we can talk about what hyperapp allows in params?

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -45,7 +45,7 @@ export default function(app) {
                 .replace(/\//g, "\\/")
                 .replace(/:([\w]+)/g, function(_, key) {
                   keys.push(key)
-                  return "([-\\.\\w]+)"
+                  return "([-\\w.~!$&'()*+,;=:@]+)"
                 }) +
               "/?$",
             "g"
--- a/src/router.js
+++ b/src/router.js
@@ -45,7 +45,7 @@ export default function(app) {
                 .replace(/\//g, "\\/")
                 .replace(/:([\w]+)/g, function(_, key) {
                   keys.push(key)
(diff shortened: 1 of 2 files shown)


jbucaran:
Hi @fvj! LGTM.

@lukejacksonn?

jbucaran:
> Can anyone see an issue arising in which any of these characters would cause a problem? If not then we might as well construct a regex that accommodates all possible permutations.

Maybe [YAGN](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) applies here?

fvj:
I've changed the regex to what RFC 3986 describes as valid characters. From what I understand YAGN would imply that we add each valid character separately, whenever someone needs it. I don't think that's worth the effort. Also, having the router conform to RFC 3986 has no disadvantage as far as I can tell.

jbucaran:
I was looking at this:

- https://stackoverflow.com/questions/7109143/what-characters-are-valid-in-a-url

jbucaran:
@lukejacksonn Aren't ? and & used for the query string?

jbucaran:
@fvj The reason I prefer to do this character by character is because it allows me to understand what, why and how those characters are actually used / their use case.

I can see the value for `.` in the pathname, e.g., http://test.com/some/path/version/1.2/thing, but I am still not sure about the others, so if you can revert the commit to only handle `.` which is what you needed:

> I needed to capture dots in params, so I added them.

then I'll merge right away.

fvj:
Done, thanks.

jbucaran:
@fvj Sorry for prolonging this! Merging 👍

PR #231, opened 2017-06-14
Make router deterministic.

jbucaran:
See the documentation for details. 

But in a nutshell, this PR changes the behavior and how you use the Router.

- Before
```jsx
app({
  view: {
    "/": defaultView,
    "/dashboard": dashboardView,
    "/billing": billingView
  },
  plugins:[Router]
})
```
- After
```jsx
app({
  view: [
    ["/", defaultView],
    ["/dashboard", dashboardView],
    ["/billing", billingView],
    ["*", errorView]
  ],
  plugins: [Router]
})
```

This is a bit more verbose, but it makes the router route resolution deterministic.

Bundle increase: 13 bytes. 💥 

@igl Ping.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -22,7 +22,7 @@ Type: ([tag](#h-tag), [data](#h-data), [children](#h-children)): [vnode]
 
 * <a name="h-tag"></a>tag: string | ([props](#h-data), [children](#h-children)): [vnode]
 * <a name="h-data"></a>data: {}
-* <a name="h-children"></a>children: string | [vnode]\[\]
+* <a name="h-children"></a>children: string | Array\<[vnode]\>
 
 ## app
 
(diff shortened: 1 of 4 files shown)


jbucaran:
To sugarcoat the API even further:

```jsx
app({
  view: Router
    .on("/", defaultView)
    .on("/dashboard", dashboardView)
    .on("/billing", billingView)
    .on("*", errorView),
  plugins: [Router]
})
```

We could do something like this:

```jsx
Router.on = function on(route, view, parent) {
  ;(parent ? parent : (parent = []))[parent.length] = [route, view]
  parent.on = function(route, view) {
    return on(route, view, parent)
  }
  return parent
}
```

dodekeract:
@jbucaran I'd still prefer something like React Router 4.

```javascript
{view :
	<div>
		<Route match="/" exact>Index</Route>
		<Route match="/:something">{state.router.params.something}</Route>
	</div>
}
```

SkaterDad:
@dodekeract This could basically already be done by removing the `render` piece of the router plugin, so its only job is mapping the url bar to a state object.

That `<Route match="/">` tag is just a simple function, then, unless I'm missing something?  I haven't really played with React Router since v2 or so.
```
{
  view:
        h('div', {}, [
          state.router.match === "/" ? "Index" : null,
          state.router.match === "/:something" ? state.router.params.something : null
        ])
}
```

jbucaran:
@dodekeract Maybe someday or maybe as a separate "advanced" router plugin. It would be awesome to have a react-router like router for hyperapp, but our default router is rather simplistic.

jbucaran:
Closing in favor of #214.

Issue #194, opened 2017-04-24
Rename plugins to mixins.


jbucaran:
What the title says.

Ref: https://github.com/hyperapp/hyperapp/issues/189

jamen:
@jbucaran Not sure if this is a relevant place to discuss, but...  Is there any value in adding support for plugins/mixins that are just objects:

```js
app({
  plugins: [ { ...partialApp } ]
})
```

I think this is a better mechanism for plugins that don't want the app options, but rather just some localized options.  This is good if you have one plugin who depends on another, and you don't want the dependency's options to bleed through `app(opts)`.

So, they could be designed something like this:

```js
app({
  plugins: [ plugin(opts) ]
})
```

Or using it inside another plugin,

```js
Object.assign({
  state: { foo: 123 },
  // ... other plugin stuff
}, depPlugin(opts))

**Comment by jamen** [comment]
Actually, thinking over what I posted there.  I don't like how it introduces some more complexity into the API...  What I describe is technically already possible, with a little bit of extra work.   E.g.:

```js
var bar = opts => app => Object.assign({
  state: { ... },
  // others
}, foo(opts)(app))
```

PR #233, opened 2017-06-15
Rename plugins to mixins

jamen:
- Renames `plugins` to `mixins`
 - Allows mixins to extend mixins
 - Updates documentation

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -76,7 +76,7 @@ Type: ([state](#state), [actions](#actions), [data](#update-data), [emit](#emit)
 
 Fired before the state is updated.
 
-#### rendero
+#### render
 
 Type: ([state](#state), [actions](#actions), [view](#view), [emit](#emit)): [view](#view) | Array\<[render](#render)\>
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -10,7 +10,7 @@
     * [action](#action)
     * [update](#update)
     * [render](#render)
-  * [props.plugins](#plugins)
+  * [props.mixins](#mixins)
   * [props.root](#root)
 * [emit](#emit)
 
@@ -33,7 +33,7 @@ Type: ([props](#app-props))
   * [view](#view)
   * [actions](#actions)
   * [events](#events)
-  * [plugins](#plugins)
+  * [mixins](#mixins)
   * [root](#root)
 
 ### state
@@ -82,15 +82,16 @@ Type: ([state](#state), [actions](#actions), [view](#view), [emit](#emit)): [vie
 
 Fired before the view is rendered.
 
-### plugins
+### mixins
 
-Type: [Plugin](#plugins-plugin)\[\]
+Type: [Mixin](#mixins-mixin)\[\]
 
-#### <a name="plugins-plugin"></a>Plugin
+#### <a name="mixins-mixin"></a>Mixin
 
-Type: ([props](#app-props)): [props](#plugin-props)
+Type: ([props](#app-props)): [props](#mixin-props)
 
-* <a name="plugin-props"></a>props
+* <a name="mixin-props"></a>props
+  * [mixins](#mixins)
   * [state](#state)
   * [actions](#actions)
   * [events](#events)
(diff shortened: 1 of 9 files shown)


docs/core.md
```diff
@@ -313,6 +313,19 @@ app({
   actions: {
     addOne: state => state + 1
   },
-  plugins: [Logger]
+  mixins: [Logger]
 })
 ```
+Mixins can also extend other mixins:
```
jbucaran: @jamen I know it's hard, very hard, to come up with good examples, but since you're introducing this feature, can you try to come up with a more practical example? 🙏

jbucaran:
@jamen LGTM. I'll merge this after #214. If you can come up with a better example to explain mixins with mixins, that would be a plus. 👍 👍 👍

jbucaran:
@jamen With #214 out of the way, let's rebase this, improve the example and merge.

One more thing, the syntax in the docs to indicate array types changed from:

`something[]`

`Array<something>`

docs/core.md
dodekeract: Wouldn't this cause an infinite recursion?

src/app.js
```diff
@@ -6,17 +6,21 @@ export default function(app) {
...
 
-  for (var i = -1, plugins = app.plugins || []; i < plugins.length; i++) {
-    var plugin = plugins[i] ? plugins[i](app) : app
+  for (var i = -1, mixins = app.mixins || []; i < mixins.length; i++) {
+    var mixin = mixins[i] ? mixins[i](app) : app
 
-    if (plugin.state != null) {
-      state = merge(state, plugin.state)
+    if (mixin.mixins != null && mixin !== app) {
+      mixins = mixins.concat(mixin.mixins)
```
dodekeract: That's an awful lot of `mixins`. 😉

jbucaran:
@dodekeract You'll recall https://github.com/hyperapp/hyperapp/issues/194 and https://github.com/hyperapp/hyperapp/issues/189.

I've been pretty much swaying towards mixins since the discussions begun. I think @zaceno's argument is sound too.

There's no need to rename plugins to mixins, but now it's the chance to do it, since 1.0.0 is just around the corner.

jamen:
Rebased with master and updated the example.  It was hard for me to think of one that fit in a good amount of space, so hopefully this is good enough.  It just relies on the previous `Logger` example.

jamen:
I think using a mixin on a mixin is better described as "compose" instead of "extend".  So I changed the description.

docs/core.md
jamen: @dodekeract Only way I could think of is if you intentionally cause a circular reference in the tree of mixins somewhere.  But I don't think that could happen naturally.

docs/api.md
```diff
@@ -76,21 +76,22 @@ Type: ([state](#state), [actions](#actions), [data](#update-data), [emit](#emit)
 
 Fired before the state is updated.
 
-#### render
+#### rendero
```
lukejacksonn: rendero 😅

jamen: Hah, whoops.  Fixed.

docs/core.md
```diff
@@ -326,15 +326,15 @@ const Counter = () => ({
     count: 0
   },
   actions: {
-    up: state => { count: state.count + 1 },
-    down: state => { count: state.count + 1 }
+    up: state => ({ count: state.count + 1 }),
+    down: state => ({ count: state.count + 1 })
```
jamen: Almost forgot these :sweat_smile:

jamen:
Just ran the example through the prettier too.  Was little errors.

jbucaran:
@jamen Awesome! Thanks.

jbucaran:
@jamen I invited you to join the org 🎉.

BTW I still want to make some tweaks, so I'll publish over the weekend.

jamen:
@jbucaran Awesome, thanks! :smile_cat:

Yeah feel free.  Happy to wait for a good v1 release.

terkelg:
Nice work @jamen! I really like this change.

PR #239, opened 2017-06-21
Refactor mixins for less code.

jbucaran:
Made it a bit less code.

@jamen Can you double check this still makes sense?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "hyperapp",
   "description": "1kb JavaScript library for building frontend applications.",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "hyperapp",
   "description": "1kb JavaScript library for building frontend applications.",
-  "version": "0.9.2",
+  "version": "0.9.3",
```
jbucaran: Whoops. My origin points to my fork of hyperapp so the last release actually updated my fork instead. 🙇

Issue #238, opened 2017-06-21
Components


jbucaran:
Let's discuss how components could look like in a future version of HyperApp.

### Submitted Proposals

- https://gist.github.com/zaceno/83286dfd2a18ebbd3b4f1d7cb5810a0c via @zaceno 
- https://gist.github.com/Swizz/8437df6ac5ec4985e72e648d242f6a9f via @Swizz 

---

### `counter.js`
```jsx
const Counter = (props, children) => ({
  state: {
    value: props.value
  },
  view: ({ value }, { inc, dec }) =>
    <div>
      <h1>{state.value}</h1>
      <button onclick={inc}>+</button>
      <button onclick={dec}>-</button>
    </div>,
  actions: {
    inc: state => state + 1,
    dec: state => state - 1
  }
})
```

### `index.js`
```jsx
app({
  state: {
    title: "Hello."
  },
  view: (state, actions) =>
    <main>
      <h1>{state.title}</h1>
      <Counter value={1} />
    </main>
})
```

## Credits
All credit goes to @MatejMazur:
- https://github.com/MatejMazur/hyperapp
- https://github.com/MatejMazur/hyperapp-component-experiment

## Related
- #147 
- #219

MatejMazur:
Yesterday I've created partially (doesn't work without IDs) working solution without `components: []`. Just import component and you can use it but I don't know how to track components without publicly exposed identification. I need some kind of tracking. 

Here is actual and working variant of writing apps with this version:

# Example

## app.js

```javascript
import { h } from 'hyperapp'
import Counter from './counter'
import Logger from './logger'

app({
  state: {
    who: 'world',
    counters: [
      { id: 0, initial: 0 }, // I don't know how to avoid ID
      { id: 1, initial: 1 },
      { id: 2, initial: 5 }
    ]
  },

  actions: {
    addCounter: (state) => {
      const counters = state.counters
      counters.push({ id: counters[counters.length - 1].id + 1 })
      return ({ counters })
    },
    removeCounter: (state, actions, index) => {
      state.counters.splice(index, 1)
      return {
        counters: state.counters
      }
    }
  },

  view: (state, actions) => (
    <div>
      <div>Hello {state.who}!</div>
      <button onclick={actions.addCounter}>Add counter</button>
      {state.counters.map((counter, index) => (
        <div>
          <Counter id={counter.id} initial={counter.initial} />
          <button onclick={() => actions.removeCounter(index)}>Delete</button>
        </div>
[...]

PR #241, opened 2017-06-21
Enable embedding self-contained apps in views

zaceno:
This allows an app to declare that it wants to be rendered at
a specific position among siblings within it's root, by
declaring the element at that position to the renderer.

The primary motivation for this, is that it is possible to embed
self contained apps inside views, by rendering a "dummy" virtual
node with an oncreate handler that instantiates the embedded app.

Without this change, however, apps embedded like this will be
rendered at the *end* of the list of siblings, because the
renderer does not know any better.

*With* this change, the instantiation of the embedded app can
inform the renderer to render the app in the position of the dummy
vnode

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -4,7 +4,7 @@ export default function(app) {
   var actions = {}
   var events = {}
   var node
-  var element
+  var element = app.element
 
   for (var i = -1, mixins = app.mixins || []; i < mixins.length; i++) {
     var mixin = mixins[i] ? mixins[i](app) : app


zaceno:
...yes this basically means enabling "components with local state" or "stateful components". But I think perhaps "embedded apps" is less controversial around here ;)

Nobody worry though! This *teensy* patch to hyperapp does not build in support for stateful components into hyperapp -- they must come from a user-land function. But that function can't work properly without this patch.

The userland function for "embedding an app" (= creating a stateful component) might look something like this:

```
function embed (opts) {
  return {
    tag: 'div',
    children: [],
    data: {
      key: opts.key,
      oncreate: el => setTimeout(_ => {
        opts.root = el.parentNode
        opts.element = el
        app(opts)
        el.parentNode.removeChild(el)
      })
    }
  }
}
```

And you might use it like this:

```
const MyStatefulInputComponent = (props, children) => embed({
  state: props.value,
  actions: {
    set: (_, __, newValue) => {
      props.notifyParent(props.fieldName, newValue)
      return newValue
    }
  },
  view: (val, {set}) => (<input type="text" value={val} oninput={ev => set(ev.currentTarget.value)} />)
})

app({
  state: ...
  actions: ...
  view: (state, actions) => (<form>
    {state.fields.map(({name, value}) => (
    <MyStatefulInputComponent
      fieldName={name}
      value={value}
      notifyParent={actions.notifyMe}
    />
    ))}
  </form>)
})

```

jbucaran:
@zaceno This would open the door to decentralized state, which is not a pattern I'd like to promote with HyperApp. This would be an interesting path and architecture to explore on a different project, perhaps even based on [picodom](https://github.com/picodom/picodom).

zaceno:
@jbucaran  Sure, and in fact I'm already using this approach (with my homemade state management + picodom) in my own projects. So I'm not heavily invested in this PR personally.

Still: I think it's wise to separate the ideas of what people *can* do, and what you would like (or not like) them to do. It's a one-line-diff after all.

> This would open the door to decentralized state

The door is already open, in the sense that you can call app('....') multiple times in a page. And that is as it should be, I think. If you want to close that door, it should be more clear from the design and/or docs that you're only meant to use  `app(...)` once per browser tab.

But I personally like the fact that there can be multiple apps. The only thing is that it is not very easy to set up interactions between them because there's no interface from the outside. Something equivalent to Elm's `ports` would be nice.

jbucaran:
> The door is already open, in the sense that you can call app('....') multiple times in a page.

Yes, but they can't communicate with each other unless you use some sort of global emitter. So, it's not an open door, but like a rope bridge. 

> I think it's wise to separate the ideas of what people can do, and what you would like (or not like) them to do. 

You are right, wihch is why I've tried not to take sides with bundlers, template literal fanatics, JSX objectors, etc. But I have to be careful not to break the original TEA promise, because then what would be the ideological benefit of using HyperApp over Mithril or Preact?

I think there's plenty of room for another tiny framework that is not contrained to React's API (like Preact) and can still coexist with Preact, but it is not HyperApp. 

I am putting all my money (figuratively) on #238. 😄 

> Something equivalent to Elm's ports would be nice.

IIRC ports exist to wrap 3rd party libraries and because Elm is strongly typed, not as a bridge to communicate with different apps all of which are running the entire Elm runtime. I could be wrong.

zaceno:
> ... more like a rope bridge

😆 I like that analogy!

>  I have to be careful not to break the original TEA promise

I get that. Just to be clear though: this PR would not break the TEA promise in any way. I'm not suggesting that you incorporate the `embed` example above into hyperapp. The diff is still just a one-liner. 

Yet it does open the door for *others* to break TEA if they so choose.

> IIRC ports exist to wrap 3rd party libraries

Good point. I haven't used them really, so I just made an assumption what they were for.

zaceno:
Actually, I just realized it's possible to use the `embed` patttern above even without this PR. 

This should work:

```
function embed (opts) {
  return {
    tag: 'div',
    children: [],
    data: {
      key: opts.key,
      oncreate: el => setTimeout(_ => {
        const fakeRoot = document.createElement('div')
        opts.root = fakeRoot
        app(opts)
        el.parentNode.replaceChild(fakeRoot.childNodes[0], el)
      })
    }
  }
}

```

... hence I am closing this PR

Issue #90, opened 2017-02-15
Should we care about requestAnimationFrame?


jbucaran:
The [initial commit](https://github.com/hyperapp/hyperapp/blob/b5cc01f06b5e62d1ad1a78b7d9b073369ce5dd05/index.js#L24) of hyperapp demonstrates how to use [requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame to batch view renderings, largely based in [Raynos/main-loop](https://github.com/Raynos/main-loop).

At the time, I was using Matt-Esch/virtual-dom, but when I moved to Snabbdom I dropped this in favor of whatever Snabbdom was doing. Now that we're using a in-house vdom algorithm, it is worth considering using requestAnimationFrame again.

Ideally this would come after https://github.com/hyperapp/hyperapp/issues/75.

dodekeract:
I think it's a good idea. Depending on the effects used the state might be updated more than 60 times per second, which would probably lead to high CPU usage, etc.

E.g. I planned to use hyperapp when re-writing @smartflat. It'd easily be possible to get more than 60 sensor updates per second in that particular project.

terkelg:
Always nice with performance optimisations and I think no matter what we should provide examples on how to do this in the documentation.

Personally I'd prefer not to have it included, because I often work on projects that involve a lot of canvas, WebGL, tweening etc. and to avoid multiple `requestAnimationFrame` all over the place I just have a single main loop, which all other code subscribe to.

That being said, I guess it's in most peoples interest to have stuff like this included out of the box and it's what most virtual dom implementations do. It could be fun to do some performance testing, and see how much it actually does for the performance.

*@dodekeract @smartflat looks really interesting! Nice one.*

jbucaran:
> It could be fun to do some performance testing, and see how much it actually does for the performance.

This. I haven't had time to work on new benchmarks since http://mathieuancelin.github.io/js-repaint-perfs/hyperapp/, but I also want to see how much it would mean for performance.

jbucaran:
Example: 
* http://codepen.io/anon/pen/GWKONK?editors=0010

/cc @tzellman

jbucaran:
Close in favor of #286.

PR #243, opened 2017-06-23
Return the state fragment to the caller if the action is a reducer.

### LINKED ISSUE #242: Access current state in onremove?
For the transitions solution I've got to work in hyperapp-apps, there needs to be a way to access the state of the app when onremove lifecycle methods are *called*.  Not the state that was when they were bound.

The onremove handler can call actions. Actions will have access to the state as it is at the time they are called. But you can't get anything out of them.

There are three solutions I can see:

A) Either the signature for on remove handlers is augmented to look like: `... onremove={(el, doRemove, state) => {...}}`

or B) We can add the concept of getters alongside actions. Getters would work just like actions, except that what they return goes back out to the caller, instead of modifying the state

or C) Along the lines of what @dodekeract has suggested before: We change the behavior of actions to return whatever they return without modifying the state, and give them access to a "setState" or whatever function they can use to perform state mutations.

From what I understand, the API is pretty much locked down so I don't have much hope for B or C. A could happen but it's the least elegant / general. It's very specific to my problem.

jbucaran:
Now actions.myAction will return whatever the original myAction function returns. Before, this was only true if myAction returned a Promise.

## Example

```jsx
app({
...
  actions: {
    getState: state => state
  }
...
})
```

Diff: 0 bytes.

Fix #242.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "hyperapp",
   "description": "1kb JavaScript library for building frontend applications.",
-  "version": "0.9.2",
+  "version": "0.9.3",
   "main": "dist/hyperapp.js",
   "jsnext:main": "src/index.js",
   "module": "src/index.js",
--- a/src/app.js
+++ b/src/app.js
@@ -44,11 +44,11 @@ export default function(app) {
             emit
           )
 
-          if (result == null || typeof result.then === "function") {
-            return result
+          if (result != null && typeof result.then !== "function") {
+            render((state = merge(state, emit("update", result))), view)
           }
 
-          render((state = merge(state, emit("update", result))), view)
+          return result
         }
       } else {
         init(namespace[key] || (namespace[key] = {}), action, name)
--- a/src/app.js
+++ b/src/app.js
@@ -6,12 +6,9 @@ export default function(app) {
   var node
   var element
 
-  for (var i = -1, mixins = app.mixins || []; i < mixins.length; i++) {
+  for (var i = -1, mixins = []; i < mixins.length; i++) {
     var mixin = mixins[i] ? mixins[i](app) : app
-
-    if (mixin.mixins != null && mixin !== app) {
-      mixins = mixins.concat(mixin.mixins)
-    }
+    mixins = mixins.concat(mixin.mixins || [])
 
     if (mixin.state != null) {
       state = merge(state, mixin.state)


jbucaran:
:tada: 💯

dodekeract:
So does this still update the state or did you just implement my suggestion regarding getting rid of the weird return to update functionality?

jbucaran:
@dodekeract This returns the result of an action to the caller, whatever that may be. 

This just helps with @zaceno's use case without breaking the API. 

Maybe @zaceno can explain it better?

zaceno:
@dodekeract: still updates state. I kind of still wish we could have getters, i.e. same as actions but *dont* update state -- just return out whatever the inner function returns from the state. 

... thinking maybe you'd be more ok with the behavior of actions if there were getters to use for when you *don't* want the state to update?

jbucaran:
@zaceno Those are the old effects no? Maybe we can come up with another clever way?

zaceno:
@jbucaran -- that's right! Hadn't thought of it before because I always thought of effects as a way of batching and/or deferring reducers -- not as a way of getting/processing data from the state -- but the old effects could work for that too.

I'm open to thinking about & discussing clever ways to get that behavior back. But to be fair you shouldn't give my opinons on hyperapp's state management too much weight, since I've switched my personal projects from away from hyperapp, to picodom + my home made state management solution (which I'm still very much experimenting with)

Anyhow, here are my thoughts: 

I feel like we've already  "violated" the reducer principle when we allowed actions to return a promise, and in that case *not* update the state. So we could always take that a step further, with other "magical" ways of returning in actions, that don't cause state updates.  Like perhaps if what you returned was an object `{update: false, result: x} then the action would return x without updating state.

I grant you, that's not very pretty. I'd much prefer something along the lines of what @dodekeract already suggested, which is that actions always return with the inner-returned value, and *don't* update state by default. Instead they are passed a `setState` or similar function to be used inside the action, when you actually want to set the state.

jbucaran:
@zaceno Thanks for the feedback! Definitely, something to think about.

dodekeract:
> without breaking the API — @jbucaran

Not true. In complex actions I'm often using `return actions.someOtherAction()` to end the action execution and forward it to a different action.

Now, this would update the state twice which may even lead to race conditions or hard to track down bugs.

jbucaran:
@dodekeract I see what you are saying. I can try to give priority to #90 (leaving #238 for later) and see if that helps, otherwise, we can reverse this.

PR #245, opened 2017-06-25
WIP suggestion for #238

naugtur:
I didn't exactly like where #238 was going with components looking like more hperapp apps with separate state and actions.

I've built something that's similar to how redux-react does single source of truth.
Actions would stay on the main app along with the state they're modifying.

The only issue I think has to be fixed before this is viable is finding a better place to put context in. It has to be rooted in the app, but I didn't find the right place yet. It'd probably eliminate the need to have Provider at all. 

see `/poc` for usage example.

### DIFF
--- a/poc/index.js
+++ b/poc/index.js
@@ -1,0 +1,32 @@
+// to run this, go to the poc folder and run
+// npm install -g beefy
+// npm install browserify babelify babel-preset-es2015 babel-preset-react
+// beefy index.js -- -t [babelify --presets [react es2015] ]
+
+import { app, h, Provider, connect } from '../src/index.js'
+// import { h, Provider, connect } from './connect.js'
+/** @jsx h */
+console.log(1)
+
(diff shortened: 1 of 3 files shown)


jbucaran:
@naugtur Can you modify this PR's title? How do you want to name this feature?

I have a few other questions:

1) What is connect?
2) What is context?
3) What is Provider?
4) Can you please update the POC example to make it a bit more readable? It's strange to call the counter tag `Controls`.

naugtur:
This is an old proposal and seeing the thread with Components evolve so much, I want to push it in a simpler direction.

I'll rename some things as I go.

jbucaran:
@naugtur Awesome thanks!

jbucaran:
@naugtur What's up! I think I'd prefer to close here and leave it up to you to reopen/create a new PR when the time is right. 😄

Issue #87, opened 2017-02-14
Add typings


jbucaran:
#### Resources
* https://github.com/typings/typings/blob/master/docs/faq.md
* http://stackoverflow.com/questions/12687779/how-do-you-produce-a-d-ts-typings-definition-file-from-an-existing-javascript
* http://stackoverflow.com/questions/34169314/are-typescript-type-definitions-required/34171621#34171621
* https://blog.mariusschulz.com/2014/05/19/using-typescripts-type-definition-files-to-get-tooling-support-for-plain-javascript

rbiggs:
Actually, I've already put some work into creating a TypeScript definition file so I could get intellisence, code completion, refactoring, etc. in IDE's that understand t.ds (TypeScript definition files). But that was for a much earlier version 0.0.8. A lot has changed. 🎢 A TypeScript declaration file is for use with JavaScript. As the user types, he/she gets information about a method's call signature, expected parameters, and the parameter's types. It will also show basically whatever documentation you provide for that in the IDE. It's a very helpful feature when you're not familiar with an API. Once I know an API, I dispense with it. Here's a video about [intellisense](http://code.visualstudio.com/docs/introvideos/intellisense)
That said, it doesn't make sense to deal with typings until you have a fairly stable API. That's why I stopped working on my t.ds. Things were changing to quickly so 🏖
By the way, I've written a number of TypeScript definition files and contributed to definitelytyped.org. I've got about 80,000 lines of TypeScript under my belt. But I'm currently just doing ES6.

jbucaran:
Thanks for the clarification @rbiggs! 👍 

>  I've got about 80,000 lines of TypeScript under my belt. 

Awesome! Please help me create the typings when the API has stabilized. Let's keep this issue open until then.

rbiggs:
No probs. I'd probably just make a fork for that. Hardest part doing the original one was discerning the return types in a few places. Geez, seems that always the hardest part of creating a typings file 🤔

rbiggs:
We'd want to also put together a typings test to run against actual code. You have to use every feature so that the TypeScript compiler and test types in all possible situations. It doesn't usually take that much code to pull that off.

rbiggs:
Just let me know when you feel you have your feature set stable. One thing, the plumbing can change quite a bit without affecting the public type interface, as long as call signatures and return types remain the same. Something to be aware of when you decide when you want to introduce types.

jbucaran:
[...]

PR #253, opened 2017-06-28
Added Flow type annotations

nitin42:
Initially I've added types to `h.js`. I wanted to know about the schema for app which include state, view, events, actions, namespace. Review this and then we can discuss further about adding types to `App.js`

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,9 @@
+[ignore]
+.*/node_modules/
+
+[include]
+.*/src/.*
+
+[libs]
+
+[options]
--- a/implementation.md
+++ b/implementation.md
@@ -1,0 +1,164 @@
+# Implementation Notes
+
+HyperApp's design is based on the [Elm Architecture](https://guide.elm-lang.org/architecture/) which enables you to create scalable browser-based applications using a functional paradigm.
+
+The logic of every program using this architecture will break into three parts :
+
+* **Model** - It represents the state of your application
+* **Update** - A way to update your state (actions)
+* **View** - View your state as HTML
+
+This pattern enforces you to write the code which is declarative and pure (immutable).
+
+## Virtual DOM
+HyperApp uses virtual DOM which provides a declarative way of representing the UI of your app. A virtual DOM tree is basically a data structure that consists of nested `vnodes` which describes a DOM tree. Whenever the state of your application changes, a virtual DOM tree is created to update the view. Then it ***diffs*** a `vnode` tree against its previous version and only modifies the DOM where there are changes.
+
+Read more about virtual DOM [here](https://github.com/Matt-Esch/virtual-dom).
+
+
+## Overview
+
+### Creating a `vnode` tree (hyperscript)
+
+#### `h(tag, properties, children)`
+
+`h()` takes a tag, an optional properties objects and an optional array of children or a single child that is a string. If you pass it an array of children, it will have child nodes.
+
+You can pass deeply nested tags and also components like this
+
+```javascript
+const Component = (data, children) => h("div", data, children)
+
+h(Component, { id: "foo" }, [h(Component, { id: "bar" })])
+```
+
+it will then create a `vnode` which consists of a `tag`, `props` and `children`.
+
+If you pass a number as a child, `h()` will parse it to string.
+
+### Creating an element
+
+#### `createElementFrom(node, isSVG)`
+
+It takes a `vnode` object and an optional `isSVG` option if the node is `svg`.
+
+It the type of a node is string, it will create a text node. Example -
+```javascript
+$ createElementFrom('Hello World!')
+'Hello World!'
+```
+
+If you provide a `vnode` object which includes a `tag` like this
+```javascript
+vnode = {
+  tag: 'div',
+  data: { id: 'foo' },
+  children: ["Hello World"]
+}
+```
+
+it would then create an element with that tag. Example -
+```javascript
+$ createElementFrom(h('div', { id: 'foo', style: { color: 'red'} }, ["bar"] ))
+<div id="foo" style="color: red;">bar</div>
+```
+
+More Examples -
+
+**Creating a `svg`**
+
+```javascript
+$ createElementFrom(h('svg', { width: "100", height: "100" }, h('circle', { cx: '50', cy: '50'})))
+
+<svg width="100" height="100">
+ <circle cx="50 cy="50"></circle>
+</svg>
+```
+
+**Lifecycle hook**
+
+In this example, style of an element is updated after it is created and added to DOM.
+
+```javascript
+$ createElementFrom(h('div', { oncreate: (element) => element.style = 'color: blue;' }, ["foo"]))
+
+<div style="color: blue;"></div>
+```
+
+Notice how all the data attributes on an element are appended to it after it is created and added to DOM. Let's take a look at it.
+
+### Setting data attributes on an element
+
+Earlier we created an element provided a `vnode` object with children as an array. Now let's take a look at how to set the data attributes on it.
+
+#### `setElementData(element, name, value, oldValue)`
+
+It takes an element, property name, property value and an option previous value of the property.
+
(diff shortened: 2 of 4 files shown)


Swizz:
Personally I vote in favor of Typescript

jbucaran:
We are going through a similar situation in picodom. Both @ngryman and @pspeter3 made compelling arguments against it and in favor respectively. It's a tough one.

- https://github.com/picodom/picodom/issues/10
- https://github.com/picodom/picodom/issues/7

Swizz:
I tried both, and both are similar but different.

The main purpose is what do you want types for?
* Expose a typed API
* Internal typing for ensure code
* Both

I think Flow is better for the second one, because it is able to catch typing trouble without explicit type. Typescript is just a transpiler without explicit typing.

For the first one, Typescript is better as long as your targeted devs use Typescript otherwise, Flow is better.

I prefer Typescript in everyday use, because Typecript is already an ES6 transpiler where Flow is not.

Both are similar but different; And this is a non-ending debate.
* Typescript is a compiler/transpiler with typing capabilities.
* Flow is only a type checker.

So you will need to ask your contributors or take a personal choice.

jbucaran:
@nitin42 Would you like to tackle #87 too?

nitin42:
Umm.. refer to this article http://thejameskyle.com/adopting-flow-and-typescript.html and I am sure you will come to know about both of them. There was a same discussion earlier with me and an another maintainer of a different library about adopting type system and we decided to use Flow rather than Typescript. The article is not biased but Flow will be much better option.

nitin42:
I will start working on creating the Flow declaration files if we choose Flow. Let me know!

SkaterDad:
For what it's worth, Typescript 2.3+ can type-check javascript files (no transpiling required).   In this way, it can be used similarly to flow, but only using JS-doc style comments.  This seems to be the least intrusive way to introduce type checking in my opinion (minifier already strips comments, so no intermediate JS file has to be created), so maybe it's an option?
https://github.com/Microsoft/TypeScript/wiki/Type-Checking-JavaScript-Files

Swizz:
Another reading : https://github.com/niieani/typescript-vs-flowtype

Another concern could be, can I take benefits of Type with Typescript if Hyperapp use Flow type ? 
Or do I need to write a `.d.ts` too ?

So, could you use Flow for internal type checking usage and provide a `.d.ts` file for the public API ?

jbucaran:
I think we should first do #87. 🤔

Swizz:
@jbucaran `.d.ts` is more than editor stuffs, lot of Typescript developer are one the ways to use the strict mode of typescript, that require library to have a declaration file.
Some of them, are regrouped in a unofficial (I mean not maintained by libraries authors) repo : [DefinitelyTyped/DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped)

I  repeat myself, but I think we need to know what is the main purpose of typing : To expose a typed API or for Internal typing to ensure code.

jbucaran:
> I repeat myself, but I think we need to know what is the main purpose of typing: To expose a typed API or for Internal typing to ensure code.

What would be a good compromise in your opinion?

Swizz:
If you are targeting developers experience : A declaration file will be sufficiant (Typescript, Flow, Both ?)
This will allow you to provide a typed API and communicated about function/api and types in a normalised way.

If your wish is to ensure code quality and testing, you can move into the way to add type into your code (Flow for just type checking, Typescript for both type checking and transpiling).

jbucaran:
@Swizz Parts of the code are optimized by hand, so I'd like to avoid transpilation. 

Flow, on the other hand may be worth adding to the project. 🤔

Swizz:
Only declaration file could be a great consensus. You will be able to type only the things that are publicly exposed without doing anything to your code base.

An example of `.d.ts` file for a `NotReact.createElement` implementation : https://github.com/Swizz/snabbdom-pragma/blob/develop/snabbdom-pragma.d.ts

whaaaley:
I vote for no change 😬 . I think the code base is small enough extra tooling would be... extra. If type checking must happen, I'd rather not convert everything to a new language.

jbucaran:
@nitin42 How do you feel about this? Would you like to help with #87 instead?

Thanks for igniting this discussion!

Swizz:
So declaration file without touching the actual code, this is just another file in the root directory 🎉 👍 

We only need declarations for exported functions and an interface for the `app()` parameter.

nitin42:
@jbucaran #87 refers to adding Typescript definition file. Right ? I am comfortable with both (I know Typescript and Flow). Will start working on this as soon as I make the notes about the codebase! Good ?

nitin42:
@whaaaley I think you're also right. API will change and size of the codebase is also small. We could work on type definitions instead and add types to the actual codebase later! Sounds good ?

jbucaran:
@nitin42 Sounds good!

pspeter3:
I think this is a great step! Many projects do not write tests that the type definitions are actually valid so perhaps adding a test for that would be useful?

nitin42:
Could you elaborate it further ?

selfup:
I would rather stick with Flow on this kind of project. Mostly for dev experience :+1:

pspeter3:
Basically, the DTS is similar to a caching issue. My thought is that we can write a simple TypeScript file that tests that the code compiles correctly and then check in the test runtime that the types are correct.

nitin42:
I am not sure but I think it would be unnecessary to do so. Write declaration files and test further ? Instead we could use a static checker and check the types (like Flow does).

pspeter3:
Are you proposing that Flow will check the validity of the DTS? If so, great.

nitin42:
Yes! Flow does this implicitly. 

The technique is that, `flow-typed` repo relies on `suppress_comment` configuration option in `.flowconfig`

```
[options]
suppress_comment=\\(.\\|\n\\)*\\$ExpectError
```

Now we can use this comment anywhere in declaration file for the module. If the line that follows the comment raises a flow error, it will be suppressed else flow will raise the error. Simple ?

nitin42:
@selfup is right ! Typescript could increase the complexity in the project (still not biased 😅 ).

pspeter3:
To be honest, I feel like the momentum for both hyperapp and picodom is against TypeScript and I'm surprised that the support for Flow is so high. That being said, I think both projects are great and have their use cases even if they don't have TypeScript. Therefore, maybe the best course of action is just accept neither will have TypeScript and move on.

nitin42:
It would be much better if both the typings are provided (seen in other libraries also). Then we can upload both the files on their respective repos (flow-typed and DefinitelyTyped). Ok ?

nitin42:
For a complete rewrite, I will stick with Flow!

Swizz:
> It would be much better if both the typings are provided (seen in other libraries also). Then we can upload both the files on their respective repos (flow-typed and DefinitelyTyped). Ok ?

We do not need flow-typed and DefinitelyTyped, we only need to write a `hyperapp.d.ts` file and add it to the package.json.
```json
{
  "main": "dist/hyperapp.js",
  "module": "src/index.js",
  "types": "hyperapp.d.ts",
}
```

I will work on it tomorrow to close this asap

Swizz:
Draft in #87

andyrj:
I didn't think the .d.ts to flow via flowgen was working yet.  Why would only the .d.ts be needed?  Curious if I am not aware of a new solution for that or something.

jbucaran:
> @Swizz In fact, personally I am only asking for a declaration file, not for a complete rewrite into Typescript

I agree.

@nitin42 Do you want to tackle #87 and close here?

nitin42:
👍🏻

nitin42:
Added implementation notes. Needs some work though!

jbucaran:
@nitin42 Thanks! I am still not sure if we'll adopt flow or not, so can you send your implementation notes as a different PR? I'd like to merge that and add it to the docs.

nitin42:
#261 Btwn I think we can close this issue!

PR #255, opened 2017-06-28
Rehydrate after SSR

andyrj:
This is a basic dom rehydration, I need to add test cases to cover the new code but will add that shortly, thought I would go ahead and make this PR to discuss.


andyrj:
sorry saw, yarn run format for future :(

andyrj:
need clarification about: https://github.com/hyperapp/hyperapp/blob/master/test/root.test.js#L29 
that seems like a strange special case to skip and look for first matching tag to render to instead of expecting users to explicitly target the right dom node to render into.

jbucaran:
@andyrj Could you explain what is the purpose of this PR?

andyrj:
When you render hyperapp on the server side and then render on the client side, it will double up the dom elements.  So you either need to empty the innerHTML you are going to render client side, or what this PR is trying to do is rehydrate the existing server side rendered dom instead of doubling or having to flicker empty dom to the user, and it "works" other than that test I need clarification on.

jbucaran:
@andyrj Can you or anybody present an example of SSR with hyperapp. It does not need to work, a fake example of how it should look like would be enough.

andyrj:
https://github.com/andyrj/hyperapp-starter  has SSR using JSDOM currently

jbucaran:
@andyrj Is this something we really need in core or can this be moved to SSR if we make some compromises? 🤔😅

andyrj:
I don't really see a good way to do this other than where I made my changes.  

If you didn't want hydrate() in core, you will need some kind of way for a mixin to hook into the render at that point.  I am happy to create more tests if it's a matter of not feeling confident in this having all the edge cases covered.  It really should have negligible impact on render, hydrate() is only ever called on first render, if the target dom element already has children...  If anyone has a way to make the method smaller we can obviously change it.

As for the tests, I still need clarification on https://github.com/hyperapp/hyperapp/blob/master/test/root.test.js#L29, being intended behavior?  Because as is what that test describes is against the idea of rehydrating dom that is server rendered.

andyrj:
moved the hydrate call...  it's going to be a little bit larger since I have to declare the root var in load :(

andyrj:
an alternate hydrate that I thought was going to be smaller could be like this...
```js
  function hydrate(elm) {
    var c = []
    if (elm !== undefined && elm.hasChildNodes()) {
      [].forEach.call(element.childNodes, function(child){c.push(hydrate(child))})
    } 
    return { tag: elm.tagName, data: {}, children: c }
  }
```
but for some strange reason that's 1 byte larger in my gzip bundle... ><

andyrj:
oh crap sorry I screwed up and rebased on the router removed branch already, so it pulled that into here, let me fix that

andyrj:
closing to open a new PR without the screwed up base I got this one into >< sorry

PR #256, opened 2017-06-29
Adds % char to router.match regex

lukejacksonn:
Currently the router will fail to match `/:search` against `/frontend%20libraries` because the `%` character is not valid according to the regex in _router.match_.

Related https://github.com/hyperapp/hyperapp/pull/230

Fixed by adding `%` to the the regex.


jbucaran:
Thanks @lukejacksonn.

PR #259, opened 2017-06-29
Allow messages to be sent to apps

zaceno:
We let the `app(...)` call return its emit-function. This allows
external agents to send messages to the app, and the app doesn't
need to know who they are and how they communicate. It only needs
to define its external interface in the `events` property.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -27,6 +27,8 @@ export function app(app) {
     addEventListener("DOMContentLoaded", load)
   }
 
+  return emit
+
   function init(namespace, children, lastName) {
     Object.keys(children || []).map(function(key) {
       var action = children[key]
--- a/docs/api.md
+++ b/docs/api.md
@@ -26,7 +26,7 @@ Type: ([tag](#h-tag), [data](#h-data), [children](#h-children)): [vnode]
 
 ## app
 
-Type: ([props](#app-props))
+Type: ([props](#app-props)): [emit](#emit)
 
 * <a name="app-props"></a> props
   * [state](#state)
--- a/docs/core.md
+++ b/docs/core.md
@@ -9,6 +9,7 @@
   * [Events](#events)
     * [Custom Events](#custom-events)
   * [Mixins](#mixins)
+* [Integration](#integration)
 
 ## Virtual Nodes
 
@@ -315,5 +316,33 @@ app({
   },
   mixins: [Logger]
 })
+```
+
+
+## Integration
+
+The `app(...)` call returns the [emit](/docs/api.md#emit) function, making it possible to trigger [custom events](#custom-events) from outside the [application](#applications) itself. This is useful in situations where your app is a part of a larger system.
+
+```js
+
+const tellApp = app({
+  ...
+  events: {
+    ...
+    'outside:data': (state, actions, newData) => actions.setNewData(newData),
+    ...
+  },
+  actions: {
+    ...
+    setNewData: (state, actions, newData) => ...
+    ...
+  }
+  ...
+})
+
+...
+
+tellApp('outside:data', someNewData)
+
 ```
 
--- a/test/app.test.js
+++ b/test/app.test.js
@@ -1,0 +1,19 @@
+import { h, app } from "../src"
+import { expectHTMLToBe } from "./util"
+
+beforeEach(() => (document.body.innerHTML = ""))
+
+test("send messages to app", () => {
+  const send = app({
+    view: state => h("div", {}, [state]),
+    state: "",
+    actions: {
+      set: (state, actions, str) => str
+    },
+    events: {
+      "info:set": (state, actions, str) => actions.set(str)
+    }
+  })
+  send("info:set", "testinfo")
+  expectHTMLToBe`<div>testinfo</div>`
+})


zaceno:
There are lots of situations you may find yourself in, where you may want your hyperapp-app to be it's own, self-contained thing, within a larger system of logic. Where moving all the rest of the logic into your hyperapp app is not feasible or desirable.

In such situations, want your app to be able to receive messages from the "outside" -- but you don't want to be dependent on what the outside is (because it might be changing). So you want to set up a "consumer-agnostic" interface for your app.

The great thing is: we don't need to add any new concepts! The `events` option of the app definition is the perfect place to define such an interface. 

This PR makes `app()` return the emit function, so that external agents are able to consume that interface, once your app has defined it.

zaceno:
Some situations where this would be useful:

- You want to add a small new hyperapp app to a webpage which is otherwise full jquery-spaghetti (and you're not ready to refactor the whole thing into a single hyperapp app)

- You want to port a large app written in some other framework to HyperApp. Needs to be done incrementally, by successively replacing parts of the old system with new Hyperapp-parts.

- You've made a chat-app for example, and you like how it works, but you want to play around with some alternative technologies for the network-communications without messing with the core app. So you break out the websocket-logic into a separate bit of logic, that communicates with your app over the interface. Now you can swap out the underlying technology, while never even having to touch the rest of your app.

- You want to run your app inside the context of some kind of debugger/profiler.

jbucaran:
@zaceno Thanks!

I'll wait until we've figured out components, since I am still not sure about how to implement those exactly.

zaceno:
@jbucaran No problem. Can't truly focus on more than one thing at a time :) 

But just to be clear: This PR doesn't overlap or have anything to with components.  Components are about what goes on 'inside' an app. This is about letting the app be aware of stuff going on *outside* it.

zaceno:
http://tech.noredink.com/post/126978281075/walkthrough-introducing-elm-to-a-js-web-app

Note the use of ports/signals - that's what I'm going for with this.

selfup:
This is GREAT! Exactly what I envisioned for cross app communication when sharing w common hyperapp bundle! 🎉

zaceno:
I'm a little bummed we have to wait for components before this. That Components thread is on 🔥 ...still...

src/app.js
```diff
@@ -258,4 +258,6 @@ export function app(app) {
 
     return element
   }
+
+  return emit
```
selfup: 🎉 So clean and simple 😛

jbucaran: @zaceno Can you put that `return` statement before all the functions?

jbucaran: Around here:
- https://github.com/zaceno/hyperapp/blob/49f23fd232e87149a7bf7a0c7d33bb643ef22fff/src/app.js#L29

jbucaran:
@zaceno If we return the emit function then we close the door to returning the element. Do you think we could regret that in the future?

lukejacksonn:
We could return an object `{ element, emit, ... }`

jbucaran:
@lukejacksonn Right. I shouldn't have said "close the door". There's always a way, but is not as elegant as having a single return value.

I am tempted to merge this now, but I'd like to hear more about what other people think? /cc @jamen @selfup @SkaterDad @lukejacksonn @MatejMazur @traducer

lukejacksonn:
The yo olde elegant vs extensible seesaw. 

We choose either `element` or `emit` now and in the future we decide we need/want the other so have to push a breaking change and resort to `{ element, emit, ... }` anyway.

OR

We choose either `{ element }` or `{ emit }` or both `{ element, emit, ... }` now and never have to worry about adding anything.. but you have to do `app.element` and `app.emit` forever more.

src/app.js
zaceno: Will do!

zaceno:
@jbucaran @lukejacksonn 

Good point about element. Both emit and element are probably useful/necessary in integration scenarios. 

The difference though, is that you can always find out element somehow. Like a top-level `id="my-hyperapp-app"` and then `document.querySelector('#my-hyperapp-app')`. Or just go get the element in whatever you set as the `root` of your app.

But the `emit` function will forever be hidden in the closure unless we expose it somehow.

Perhaps there are other aspects to integration that I haven't thought about, but I feel pretty confident that `emit` is the only one that there won't be a workaround for.

So for now, I say we go with just returning `emit`.

zaceno:
commit 249224b moves the return line to where @jbucaran wanted it.

jbucaran:
@zaceno Could add some explanation in core.md? Or do you think that's not necessary now?

traducer:
I really like this change. This makes custom events even more useful and makes any outside service or apps communicating with a `hyperapp app` predictable.

jbucaran:
@zaceno How would this look in practice?

zaceno:
@jbucaran sure I'll add some more explanation in the docs. 

as for "how this would look in practice"... well, lets go with the "I'm porting my app written in framework X to hyperapp"-scenario.

Like at work we have this app made with Backbone/Marionette. It's big and complicated so if we were going to port it, we'd do it one bit at a time. Lets say we go for the mediaplayer part of it. We're just converting our media-player with it's controls, to hyperapp.

So the app would need ways to signal out to the rest of the application things like: "user pressed pause" "current timestamp is ..." "I need more media to keep playing". Those are examples of *outgoing* messages. The app can be told how to send those, by defining callbacks in the events/actions of the app.

And the app would have to be able to take *in* messages like "here's your new media", "jump to this timestamp", and some other syncronization signals (we have multiple player windows that need to play back in sync, even though their media may come in at slightly different times) These are incoming signals. The "inner" mediaplayer app defines events for these (like `events: { 'heres more media': (state, actions, newMedia) => actions.addMedia(newMedia), ... }`)

And then the outer backbone app can :

```
tellMediaController = app(mediaControllerAppDef)
...
tellMediaController('heres more media', newMedia)
```

zaceno:
ebf3087 adds "integration" section to `core.md` with a very barebones example of how this could be used

SkaterDad:
I really like this idea.  :+1: 

Once this is landed, you could get more `hyperapp` advertising by writing an example app for FeathersJS for their docs.  They have a client lib which emits events, I believe.

Returning an object would certainly be the most flexible solution long-term.  I'm not sure I see the use case for returning the element, though?

jbucaran:
@SkaterDad For integrating a third party _with_ hyperapp, I think the best route is not using an emitter, but wrapping the library into a functional component (custom tags) and use the lifecycle events.

zaceno:
@jbucaran but featherjs (after just now learning about them and checking their website) seems like a backend-communication library. Nothing to render. So seems to me that events are the way to hook in (wether external or internal) -- not lifecycle methods

zaceno:
@SkaterDad Nah you're probably right. I had a PR before where the app would return the element, because I was trying to make app's work like stateful components, and I needed the element to inject it into the tree of the parent app. But again, you can always find the element anyway, so I closed that PR. 

I think just returnin `emit` is a safe bet for the future.

jbucaran:
@zaceno What if instead of returning emit, we return render()?

zaceno:
@jbucaran I don't think render() is enough. Like in my example of porting a part of a Backbone/Marionette app. There are different types of messages you want the app to recieve, and you want to be able to pass it a payload (like "here's new media for you to play")

zaceno:
As long as we expose emit somehow, I'm happy :) I don't really see the point in your `zap` function there, though. Seems like if forcing the app to rerender is something you want, you could define an action for that (`rerender: state => state`), and an event (`'force-rerender': (state, actions) => actions.rerender()`)... and then use the emit to tell the app to rerender: 

```
tellApp = app({...})
...
tellApp('force-rerender')
```

...sure it's more work than what you're suggesting. But I don't see that asking an app to render itself would be a common use case either.

jbucaran:
@zaceno Awesome thanks! Last question, how would you actually use this?

A module that runs hyperapp / calls app() and export the emit function?


```jsx
export const emit = app({ state, view, actions })
```

zaceno:
Could be in a module, sure. In the example I brought up about porting an app that seems sensible. But I can also imagine it as a way to start bringing som sanity into a web app made up of just jquery-spaghetti. In that case you wouldn't have modules and a build-system (probably) So it depends I guess.

jbucaran:
SGTM. What's the worst that could happen anyway.

dodekeract:
@jbucaran I'm calling it now — it will manage to break something for me. I don't know how, yet. 🙂

jbucaran:
@dodekeract Unlikely, app never returned anything. Now, https://github.com/hyperapp/hyperapp/pull/275 might break stuff.

zaceno:
🎉  Thanks, everyone involved!

@dodekeract from now on, if anything breaks, just assume it's my fault, ok? ;)

selfup:
OMG it is in! This is amazing, now I can use hyperapp for a side project I had in mind 👍 🎉 

Thank you so much @zaceno 🥇

PR #260, opened 2017-06-29
Add Tweetbox tutorial.

jbucaran:
I wrote a new tutorial to add to the tutorials section in the docs.

Please review. 🙏 

[CodePen](https://codepen.io/hyperapp/pen/bgWBdV?editors=0010)

### DIFF
--- a/docs/tweetbox.md
+++ b/docs/tweetbox.md
@@ -1,6 +1,6 @@
-# Tweetbox
+# TweetBox
 
-In this section we'll implement a simple Tweetbox clone and learn how to assemble a view using custom tags.
+In this section we'll implement a simple TweetBox clone and learn how to assemble a view using custom tags.
 
 [Try it online](https://codepen.io/hyperapp/pen/bgWBdV?editors=0010)
 
@@ -8,13 +8,16 @@ In this section we'll implement a simple Tweetbox clone and learn how to assembl
 const MAX_LENGTH = 120
 const OFFSET = 10
 
-const OverflowText = ({ text, offset, count }) =>
-  <p>
-    ...{text.slice(0, offset)}
-    <span class="overflow-text">
-      {text.slice(count)}
-    </span>
-  </p>
+const OverflowWidget = ({ text, offset, count }) =>
+  <div class="overflow">
+    <h1>Whoops! Too long.</h1>
+    <p>
+      ...{text.slice(0, offset)}
+      <span class="overflow-text">
+        {text.slice(count)}
+      </span>
+    </p>
+  </div>
 
 const Tweetbox = ({ count, text, update }) =>
   <main>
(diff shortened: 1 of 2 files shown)


jbucaran:
@zaceno Can you have a look? 😄

zaceno:
@jbucaran took a look at the codepen, and noticed that when you have 0 characters left, the tweet-button is disabled (but there is no error message as to why)

zaceno:
Going to look at the tutorial now

docs/tweetbox.md
```diff
@@ -0,0 +1,137 @@
...
+<Tweetbox
+  text={state.text}
+  count={state.count}
+  update={e => actions.update(e.target.value)}
+/>
+```
+
+When the user types in the input, we call `actions.update` to update the current text and calculate the remaining characters.
+
+To calculate the number of remaining characters, we subtract the new text length from the current text length and add it to `state.count`.
```
zaceno: I find this just a tad convoluted and hard to follow. Probably just because I'm not thinking about the problem in the same way. Anyhow, I might have worded this something like:

The subtracting the length of the current text, from the length of the previous text, tells us how the number of remaining characters has changed. Hence the new count of remaining characters is the old count plus the aforementioned difference.

docs/tweetbox.md
```diff
@@ -0,0 +1,137 @@
...
+
+```js
+state.count + state.text.length - text.length
+```
+
+When the input is empty, this operation is equivalent to `(MAX_LENGTH - text.length)`. As the text grows `state.count` shrinks and vice versa.
+
+When `state.count` becomes 0 or less, we know the `state.text` must be as long or longer than `MAX_LENGTH`, so we can disable the tweet button and display the OverflowText custom tag.
```
zaceno: "becomes less than 0..." "...must be longer than `MAX_LENGTH`..."

docs/tweetbox.md
```diff
@@ -0,0 +1,137 @@
...
+
+```jsx
+<button
+  onclick={() => alert(text)}
+  disabled={count >= MAX_LENGTH || count <= 0}
```
zaceno: `disabled={count >= MAX_LENGTH || count < 0}`

docs/tweetbox.md
```diff
@@ -0,0 +1,137 @@
...
+>
+  Tweet
+</button>
+```
+
+The Tweet button is also disabled when the input is empty, which we can tell by looking at `state.count`.
```
zaceno: The tweet button is also disabled when `state.count === MAX_LENGTH`, because that means we have not entered any characters.

jbucaran:
@zaceno Thanks! I committed all your suggestions.

zaceno:
@jbucaran You're welcome! Happy to help :)

But looks like you didn't change `disabled={count >= MAX_LENGTH || count < 0}` -- was that a mistake or intentional?

docs/tweetbox.md
```diff
@@ -0,0 +1,140 @@
+# Tweetbox
+
+In this section we'll implement a simple Tweetbox clone and learn how to assemble a view using custom tags.
+
+[Try it online](https://codepen.io/hyperapp/pen/bgWBdV?editors=0010)
+
+```jsx
+
+const MAX_LENGTH = 120
```
lukejacksonn: Aren't tweets 140 characters?

jbucaran:
Thank you all! I've fixed the other two mistakes and now we have a new tut.

PR #261, opened 2017-06-29
Implementation notes

nitin42:
Covers all the function. Let me know about the changes!

### DIFF
--- a/implementation.md
+++ b/implementation.md
@@ -15,7 +15,6 @@ HyperApp uses virtual DOM which provides a declarative way of representing the U
 
 Read more about virtual DOM [here](https://github.com/Matt-Esch/virtual-dom).
 
-
 ## Overview
 
 ### Creating a `vnode` tree (hyperscript)
--- a/implementation.md
+++ b/implementation.md
@@ -4,8 +4,8 @@ HyperApp's design is based on the [Elm Architecture](https://guide.elm-lang.org/
 
 The logic of every program using this architecture will break into three parts :
 
-* **Model** - It represents the state of your application
-* **Update** - A way to update your state (actions)
+* **State** - It represents the state of your application
+* **Actions** - A way to update your state (actions)
 * **View** - View your state as HTML
 
 This pattern enforces you to write the code which is declarative and pure (immutable).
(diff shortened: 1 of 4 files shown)


nitin42:
Wait! I'll need to rebase and delete the previous commits.

nitin42:
Done. You can merge now I think!

implementation.md
```diff
@@ -0,0 +1,163 @@
+# Implementation Notes
+
+HyperApp's design is based on the [Elm Architecture](https://guide.elm-lang.org/architecture/) which enables you to create scalable browser-based applications using a functional paradigm.
+
+The logic of every program using this architecture will break into three parts :
+
+* **Model** - It represents the state of your application
+* **Update** - A way to update your state (actions)
```
lukejacksonn: mixing `model/state` and `updates/actions` like this might confuse people?

implementation.md
```diff
@@ -0,0 +1,163 @@
...
+#### `h(tag, properties, children)`
+
+`h()` takes a tag, an optional properties objects and an optional array of children or a single child that is a string. If you pass it an array of children, it will have child nodes.
+
+You can pass deeply nested tags and also components like this
+
+```javascript
+const Component = (data, children) => h("div", data, children)
+
+h(Component, { id: "foo" }, [h(Component, { id: "bar" })])
```
lukejacksonn: This is passing a vnode as the tag name? Do you not mean..
```
Component({ id: 'foo' }, [
  Component({ id: "bar" }),
  Component({ id: "baz" }),
])
```

jbucaran: @lukejacksonn JSX like

```jsx
<MyTag id={foo} />
```

becomes

```js
h(MyTag, { id: foo }, null)
```

lukejacksonn: wtf? How does that even work?

jbucaran: Well, you know that the `h` function checks if the tag is a function, so in this case, it will see that MyTag is a function and return calling it.

jbucaran:
Thanks, @nitin42! I'll pull your changes locally, add my changes on top and possibly create or push to this PR. The content is going to grow/change considerably, but your contribution will be part of the project. 👍

implementation.md
lukejacksonn: duhh! TIL

nitin42:
Made some changes and I've removed the example for deeply nested tree (temporary).

jbucaran:
Closing in favor of #263.

PR #262, opened 2017-06-29
Add hash router capability to router.js

jrop:
Abstract out core routing functionality and construct two routers for export at module load:

* The usual `Router`
* `Router.Hash` that uses hash routing.

The interface for both is equivalent and can be used like the following:

```js
const {app, Router} = hyperapp
app({
  mixins: [Router]
})
// or
app({
  mixins: [Router.Hash]
})
```

The rollup file `hyperapp.js` increases by only 238 bytes with this addition:

```txt
-rw-r--r--  1 me  me   3621 Jun 29 11:39 hyperapp.js (before)
-rw-r--r--  1 me  me   3859 Jun 29 11:39 hyperapp.js (after)
```

Prior work:
* hyperapp/hyperapp/pull/79
* ...

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -1,70 +1,99 @@
-export default function(app, view) {
-  return {
-    state: {
-      router: match(location.pathname)
-    },
-    actions: {
-      router: {
-        match: function(state, actions, data, emit) {
-          return {
-            router: emit("route", match(data))
+function Router(getLocation, subscribe, goTo) {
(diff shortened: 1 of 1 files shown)


jbucaran:
@jrop We're in the process of splitting the router from the app! Let's move this PR to https://github.com/hyperapp/router. 🙏 🎉

PR #266, opened 2017-07-01
Remove router from core. Close #176.

### LINKED ISSUE #176: Remove Router from core?

jbucaran:
Moved to https://github.com/hyperapp/router.


jbucaran:
@lukejacksonn Do you mean [this](https://github.com/hyperapp/router/tree/master/test)? If so, then yes!

PR #267, opened 2017-07-02
Hydrate

andyrj:
PR corrected from #255 

sorry, opened new PR because I somehow screwed up the base in that last one...  didn't want to accidentally undo other changes.

If anyone has a way to make any of this code smaller/faster I'd love it...  this is as small as I could figure out  to hydrate existing dom.  It ends up adding 72 bytes gzipped.


andyrj:
down to 69B now I'm out of ideas for smaller/faster implementation

jbucaran:
@andyrj That's quite an accomplishment! Can you walk me through what's going on here? 🤔😀

andyrj:
Sure, the basic idea is that on load, if the target for render has existing dom, we make a minimal vdom representation for the diff prior to first render pass based on the existing nodes.  

This way the existing logic for render will see our minimal vdom and think all the tags exist but that they need their attributes added.  So it won't double up the dom elements like it would previously, and users who ssr won't have to empty target innerHTML prior to calling app() so they can avoid flicker.

It was the simplest way I could come up with.  I believe it has sufficient test cases as well, but happy to add more if anyone thinks of a case I haven't considered or something that won't work with this approach.

Sorry about having to open a new PR, I screwed up the base on last one somehow heh.

andyrj:
I do need clarification on the test "non-empty root", is that a valid use case?  It won't work with this code, and seems to be a special case where we implicitly assume we should render to the bodies last child?  Or something like that.

src/app.js
```diff
@@ -56,7 +56,25 @@ export function app(app) {
...
+      tag: elm.tagName,
+      data: {},
+      children: [].map.call(elm.childNodes, function(child){hydrate(child)})
+    }
+  }
+
   function load() {
+    var root = app.root || (app.root = document.body)
+    if (
+      node === undefined &&
```
jbucaran: @andyrj Why do you check if node/element are undefined here? `load` is only called once, so they must be undefined.

src/app.js
```diff
@@ -56,7 +56,25 @@ export function app(app) {
...
+      element === undefined &&
+      root.hasChildNodes !== undefined &&
+      root.hasChildNodes()
+    ) {
+      node = hydrate(root.children[0])
```
jbucaran: @andyrj Are you checking to see if root.children[0] is not null? Is there a way we can skip calling this function and just do nothing if root.children is null?

src/app.js
```diff
@@ -74,7 +92,7 @@ export function app(app) {
 
   function render(state, view) {
     element = patch(
-      app.root || (app.root = document.body),
+      app.root,
```
jbucaran: Looks like a bug. What if app.root is null. 🤔

src/app.js
```diff
@@ -56,7 +56,25 @@ export function app(app) {
     })
   }
 
+  function hydrate(elm) {
```
jbucaran: Maybe we need a vnode function that returns a vnode and share that with h. Perhaps we can save a few bytes that way and it'd better architecture-wise.

src/app.js
andyrj: Good point, that doesn't need to be there like it was when I had this in render()

src/app.js
andyrj: Will be handled in load() won't it?  I moved that to load so that I wasn't duplicating that logic

src/app.js
andyrj: Not certain I understand what you mean but changes/improvements welcome and if you think there is a better way to do this then I am all for it.

jbucaran: Well a vnode function would return a vnode like { tag, data, children }

```js
function vnode(tag, data, children) {
  return { tag: tag, data: data, children: children }
}
```

Now we can reuse it in h and app. Only makes sense if we use more than once, and it seems we will. Don't know if this would save any byte, but it's a good practice. 

Can be done later.

src/app.js
andyrj: Had that at one point, I agree returning early from hydrate would be better.

andyrj:
This is now down to a 49B increase in gzip size, after fixing the things mentioned above.

jbucaran:
@andyrj Can you rebase with master? There is no `load` function anymore.

andyrj:
I rebased and updated where the hydrate method and element/node initialization code goes...  but I am not great with the github ui for getting this PR to use the new rebased branch so I'm going to close this PR and open a new one with the rebased branch lol sorry...

PR #273, opened 2017-07-03
Add bundlesize.

jbucaran:
Add bundlesize.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,5 +36,11 @@
     "prettier": "~1.4.1",
     "rollup": "^0.41.6",
     "rollup-plugin-uglify": "^2.0.1"
-  }
+  },
+  "bundlesize": [
+    {
+      "path": "./dist/hyperapp.js",
+      "threshold": "1.5Kb"
+    }
+  ]
 }


jbucaran:
Hi @siddharthkp 👋😁,

Does bundlesize compare the size of the minified+gzip'd file or only minified?

```diff
+  },
 +  "bundlesize": [
 +    {
 +      "path": "./dist/hyperapp.js",
 +      "threshold": "1.5Kb"
 +    }
 +  ]
```

package.json
```diff
@@ -36,5 +36,11 @@
     "prettier": "~1.4.1",
     "rollup": "^0.41.6",
     "rollup-plugin-uglify": "^2.0.1"
-  }
+  },
+  "bundlesize": [
+    {
+      "path": "./dist/hyperapp.js",
+      "threshold": "1.5Kb"
```
Swizz: `1.5kb` ? Removing router didnt remove the overhead ?

jbucaran: @swizz What do you mean?

jbucaran:
I made this PR an image of [this one here](https://github.com/developit/preact/commit/8a729f99e50ec872410cacf7678aa3572ae0c8bb) in preact. I am not sure I got it right, though, since preact.min.js is `8.1 KB`.

Swizz:
Looking at the source code : [files.js#L12](https://github.com/siddharthkp/bundlesize/blob/master/src/files.js#L12)
```jsx
const size = gzip.sync(fs.readFileSync(path, 'utf8'))
```

jbucaran:
@Swizz Then we're good. Without the router, hyperapp is now 1400 bytes or so. I think 1.5 is a good compromise. I have a feeling that will go up to 1.8 or 2 at some point, but that's okay.

package.json
jbucaran: Ah gotcha. I left you a comment in the main thread! 😉

zaceno:
@jbcaran srsly youd be ok w 2kb? Wow man, you've grown  ;)

jbucaran:
@zaceno Under 2kb! Still, I'll aim at 1.4~1.5 max.

Swizz:
@jbucaran Did you miss bundlesize as dependency ?

PR #275, opened 2017-07-03
Make app() function synchronous.

jbucaran:
This will break your app if you were loading your bundle inside the `<head>` instead of at the bottom of `<body>`.

If your use case involves loading an app in the `<head>`, you can simply do follows:

```js
addEventListener("DOMContentLoaded", function() {
  app({ /* state, view, actions, events */ })
})
```

Also, `loaded` was renamed to `ready`. Now that app is sync and DOMContentLoaded is not being used,  events.ready makes more sense.


andyrj:
I like this, simplifies ssr won't need to add document.readyState.

jbucaran:
> @dodekeract  What's the recommended event to attach to when the script is loaded in the header? load?

I think this is good enough.

```js
addEventListener("DOMContentLoaded", () => app({ state, view, actions })
})
```

@SkaterDad Can you have a look? 🙏

src/app.js
```diff
@@ -21,13 +22,16 @@ export function app(app) {
     })
   }
 
-  if (document.readyState[0] !== "l") {
-    load()
-  } else {
-    addEventListener("DOMContentLoaded", load)
-  }
+  return emit("loaded", emit, render(state, view))
```
lukejacksonn: Is this guaranteed to fire only when `DOMContentLoaded`?

jbucaran:
@andyrj Yes, very likely.

src/app.js
jbucaran: No, it won't. You need to use `DOMContentLoaded` if you are unsure like if you are loading your app bundle inside the `<head>`. You don't often do that, yes?

andyrj: isn't defer supported by like 95%+ of browsers now?  http://caniuse.com/#search=defer

stupid Opera Mini dragging their feet on so much.

lukejacksonn: Haha, that is fine.. you are right. I don't do it often. Some people might, for them it will break so I guess some explicit *makes sure to put the script before the closing body tag* wording in the docs wouldn't go a miss.

andyrj: or put defer in your script tag if it's in the head, they become equivalent ;)  I really like putting scripts in head so I can just target document.body for render

jbucaran:
> @SkaterDad Will sites like Codepen require the DOMContentLoaded callback?

I don't know, but I don't think so. We used to have pens before this feature was originally added.

@lukejacksonn Do you approve of these changes?

jbucaran:
Thanks @jamen.

jbucaran:
> @dodekeract What's the recommended event to attach to when the script is loaded in the header? load?

The recommended way is to use:

```html
<script defer src="my-hyper-app-bundle.js"></script>
```

This little chart made it click for me.

script-defer-async

dodekeract:
@lukejacksonn Not true, I can't control the way the scripts I write will be embedded. Any restriction I add is one restriction too much.

I hope `DOMContentLoaded` will still be possible, otherwise we have to fork hyperapp.

jbucaran:
@dodekeract You can always use DOMContentLoaded in your own code. Like I said before you have tree options:

1) Use DOMContentLoaded and place your script in your preferred location.
2) Place your script at the end of the body.
3) Place your script in the head using `<script defer src="..."></script>`

PR #279, opened 2017-07-04
Add TypeScript declaration file

Swizz:
Following #87 : I present you my version of a TypeScript declaration only for the Hyperapp API.

The following define lot of interfaces as fixed one, but make it able to partially use them trough `Partial<Interface>` (Thanks to @nippur72).
All the interfaces need to be extended in order to be used.

Custom Events support is free by extending the Events interface, but without fixed typing yet.

---

This declaration file has been tested against the following features with the following implementation.

```tsx

import { h, app, Actions, Events, Application, emitFunction } from 'hyperapp'

interface State {
  readonly count: number,
  readonly name: string
}

interface ActionsInterface extends Actions<State> {
  inc(state: State): Partial<State>,

  subspace: {
    dec(state: State): Partial<State>
  },

  fail(_state: State, _actions: ActionsInterface, _data: any, emit: emitFunction<string>): string
}

interface EventsInterface extends Partial<Events<State, ActionsInterface>> {
  error(_state: State, _actions: ActionsInterface, error: string): never,

  loaded(): void
}

const emmiter: emitFunction<any> = app<State, ActionsInterface, EventsInterface>({
  state: {
    count: 0,
    name: "Jojo"
  },

  actions: {
    inc: state => ({ count: (state.count || 0) + 1 }),

    subspace: {
      dec: state => ({ count: (state.count || 0) - 1 })
    },

    fail: (_state, _actions, _data, emit) =>
      emit("error", "Fail")
  },

  view: (state: State, actions: ActionsInterface) => (
    <div>
      <h1>{state.name || ''}</h1>
      <h2>{state.count || 0}</h2>
      <div>
        <button onclick={actions.inc}>+</button>
        <button onclick={actions.subspace.dec}>-</button>
        <button onclick={actions.fail}>x</button>
      </div>
    </div>
  ),

  events: {
    error: (_state, _actions, error) => {
      throw error
    },

    loaded: () => {
      console.log('loaded')
    }
  }
})
```

options | value
------------ | -------------
noImplicitAny  | true
strictNullChecks  | true
noUnusedParameters  | true
jsx  | react
jsxFactory  | h

---

### todo
- [x] JSX Support
- [ ] Less naive VNodeData interface
- [ ] Custom Events support
- [ ] API documentation in the declaration file
- [ ] Automated tests against API changes

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -1,62 +1,149 @@
 declare module "hyperapp" {
-
   export interface VNodeData {
-    [key: string]: any;
+    [key: string]: any
   }
 
   export interface VNode {
-    tag: string;
-    data: VNodeData;
-    children: string | VNode | Array<VNode>;
+    tag: string
+    data: VNodeData
+    children: string | VNode | Array<VNode>
   }
 
-  export type viewFunction<State, Actions> = (state: State, actions: Actions) => VNode;
-  export type emitFunction<Data> = (event: string, data?: Data) => Data;
-  export type vnodeFunction = (props: VNodeData, children: string | VNode | Array<VNode>) => VNode;
+  export type viewFunction<State, Actions> = (
+    state: State,
+    actions: Actions
+  ) => VNode
+  export type emitFunction<Data> = (event: string, data?: Data) => Data
+  export type vnodeFunction = (
+    props: VNodeData,
+    children: string | VNode | Array<VNode>
+  ) => VNode
 
   export interface Actions<State> {
-    [key: string]: ( (state?: State, actions?: Actions<State>, data?: any, emit?: emitFunction<any>) => Partial<State> | any )
-                 | Actions<State>;
+    [key: string]:
+      | ((
+          state?: State,
+          actions?: Actions<State>,
+          data?: any,
+          emit?: emitFunction<any>
+        ) => Partial<State> | any)
+      | Actions<State>
   }
 
-  export interface Events<State, ActionsInterface extends Actions<State> | undefined> {
-    loaded: ( (state: State, actions: ActionsInterface, data: any, emit: emitFunction<any>) => void )
-          | Array< (state: State, actions: ActionsInterface, data: any, emit: emitFunction<any>) => void >;
-    action: ( <Data>(state: State, actions: ActionsInterface, data: Data, emit: emitFunction<any>) => Data )
-          | Array< <Data>(state: State, actions: ActionsInterface, data: Data, emit: emitFunction<any>) => Data >;
-    update: ( (state: State, actions: ActionsInterface, data: State, emit: emitFunction<any>) => State )
-          | Array< (state: State, actions: ActionsInterface, data: State, emit: emitFunction<any>) => State >;
-    render: ( (state: State, actions: ActionsInterface, view: viewFunction<State, ActionsInterface>, emit: emitFunction<any>) => viewFunction<State, ActionsInterface> )
-          | Array<(state: State, actions: ActionsInterface, view: viewFunction<State, ActionsInterface>, emit: emitFunction<any>) => viewFunction<State, ActionsInterface>>;
+  export interface Events<
+    State,
+    ActionsInterface extends Actions<State> | undefined
+  > {
+    loaded:
+      | ((
+          state: State,
+          actions: ActionsInterface,
+          data: any,
+          emit: emitFunction<any>
+        ) => void)
+      | Array<
+          (
+            state: State,
+            actions: ActionsInterface,
+            data: any,
+            emit: emitFunction<any>
+          ) => void
+        >
+    action:
+      | (<Data>(
+          state: State,
+          actions: ActionsInterface,
+          data: Data,
+          emit: emitFunction<any>
+        ) => Data)
+      | Array<
+          <Data>(
+            state: State,
+            actions: ActionsInterface,
+            data: Data,
+            emit: emitFunction<any>
+          ) => Data
+        >
+    update:
+      | ((
+          state: State,
+          actions: ActionsInterface,
+          data: State,
+          emit: emitFunction<any>
+        ) => State)
+      | Array<
+          (
+            state: State,
+            actions: ActionsInterface,
+            data: State,
+            emit: emitFunction<any>
+          ) => State
+        >
+    render:
+      | ((
+          state: State,
+          actions: ActionsInterface,
+          view: viewFunction<State, ActionsInterface>,
+          emit: emitFunction<any>
+        ) => viewFunction<State, ActionsInterface>)
+      | Array<
+          (
+            state: State,
+            actions: ActionsInterface,
+            view: viewFunction<State, ActionsInterface>,
+            emit: emitFunction<any>
+          ) => viewFunction<State, ActionsInterface>
+        >
 
     // [key: string]:( <Data>(state: State, actions: ActionsInterface, data: Data, emit: emitFunction<any>) => Data )
     //      | Array< <Data>(state: State, actions: ActionsInterface, data: Data, emit: emitFunction<any>) => Data >;
   }
 
-  export interface Mixin<State, ActionsInterface extends Actions<State> | undefined, EventsInterface extends Partial<Events<State, ActionsInterface>> | undefined> {
-    state: State;
-    actions: ActionsInterface;
-    events: EventsInterface;
-    mixins: Array< Partial<Mixin<State, ActionsInterface, EventsInterface>> >;
+  export interface Mixin<
+    State,
+    ActionsInterface extends Actions<State> | undefined,
+    EventsInterface extends Partial<Events<State, ActionsInterface>> | undefined
+  > {
+    state: State
+    actions: ActionsInterface
+    events: EventsInterface
+    mixins: Array<Partial<Mixin<State, ActionsInterface, EventsInterface>>>
   }
 
-  export interface Application<State, ActionsInterface extends Actions<State> | undefined, EventsInterface extends Partial<Events<State, ActionsInterface>> | undefined> extends Partial<Mixin<State, ActionsInterface, EventsInterface>> {
-    root: HTMLElement;
-    view: viewFunction<State, ActionsInterface>;
+  export interface Application<
+    State,
+    ActionsInterface extends Actions<State> | undefined,
+    EventsInterface extends Partial<Events<State, ActionsInterface>> | undefined
+  > extends Partial<Mixin<State, ActionsInterface, EventsInterface>> {
+    root: HTMLElement
+    view: viewFunction<State, ActionsInterface>
   }
 
-  export function h(tag: string | vnodeFunction, data?: VNodeData, children?: Array<number | string | VNode>): VNode;
-  export function h(tag: string | vnodeFunction, data?: VNodeData, ...children: Array<number | string | VNode >): VNode;
+  export function h(
+    tag: string | vnodeFunction,
+    data?: VNodeData,
+    children?: Array<number | string | VNode>
+  ): VNode
+  export function h(
+    tag: string | vnodeFunction,
+    data?: VNodeData,
+    ...children: Array<number | string | VNode>
+  ): VNode
 
-  export function app<State, ActionsInterface extends Actions<State> | undefined, EventsInterface extends Partial<Events<State, ActionsInterface>> | undefined>(props: Partial<Application<State, ActionsInterface, EventsInterface>>): emitFunction<any>;
+  export function app<
+    State,
+    ActionsInterface extends Actions<State> | undefined,
+    EventsInterface extends Partial<Events<State, ActionsInterface>> | undefined
+  >(
+    props: Partial<Application<State, ActionsInterface, EventsInterface>>
+  ): emitFunction<any>
 
-  global {
+  namespace global {
     namespace JSX {
-      interface Element extends VNode { }
+      interface Element extends VNode {}
       interface IntrinsicElements {
-        [elemName: string]: VNodeData;
+        [elemName: string]: VNodeData
       }
     }
   }
-
 }
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -50,4 +50,13 @@ declare module "hyperapp" {
 
   export function app<State, ActionsInterface extends Actions<State> | undefined, EventsInterface extends Partial<Events<State, ActionsInterface>> | undefined>(props: Partial<Application<State, ActionsInterface, EventsInterface>>): emitFunction<any>;
 
+  global {
+    namespace JSX {
+      interface Element extends VNode { }
+      interface IntrinsicElements {
+        [elemName: string]: VNodeData;
+      }
+    }
+  }
+
 }
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -1,0 +1,53 @@
+declare module "hyperapp" {
+
+  export interface VNodeData {
+    [key: string]: any;
+  }
+
+  export interface VNode {
+    tag: string;
+    data: VNodeData;
+    children: string | VNode | Array<VNode>;
+  }
+
+  export type viewFunction<State, Actions> = (state: State, actions: Actions) => VNode;
+  export type emitFunction<Data> = (event: string, data?: Data) => Data;
+  export type vnodeFunction = (props: VNodeData, children: string | VNode | Array<VNode>) => VNode;
+
+  export interface Actions<State> {
+    [key: string]: ( (state?: State, actions?: Actions<State>, data?: any, emit?: emitFunction<any>) => Partial<State> | any )
+                 | Actions<State>;
+  }
+
+  export interface Events<State, ActionsInterface extends Actions<State> | undefined> {
+    loaded: ( (state: State, actions: ActionsInterface, data: any, emit: emitFunction<any>) => void )
+          | Array< (state: State, actions: ActionsInterface, data: any, emit: emitFunction<any>) => void >;
+    action: ( <Data>(state: State, actions: ActionsInterface, data: Data, emit: emitFunction<any>) => Data )
+          | Array< <Data>(state: State, actions: ActionsInterface, data: Data, emit: emitFunction<any>) => Data >;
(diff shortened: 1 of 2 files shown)


jbucaran:
Thanks @Swizz! This is looking pretty awesome. 

Question: Can't `prettier` format this file? Try changing the pkg.scripts.format to accept `.ts` file extensions too.

Swizz:
@jbucaran Yes, prettier work with ts files and d.ts files.

I was waiting your comment to use it. Because it produce a more longer file.

jbucaran:
> @Swizz Because it produce a more longer file.

No worries.

Change package.json's format script to this:

```
"format": "prettier --semi false --write 'src/**/*.{js,ts}'",
```

and run `npm run format` for rock and roll🤘.

Swizz:
The declaration file is not under src folder but in the root folder.
And I dont know how to achieve that with glob pattern ^^'

jbucaran:
```
"format": "prettier --semi false --write 'src/**/*.js' '*.ts'",
```

jbucaran:
@Swizz I tried using the file in a project with vscode and intellisense didn't fire like I expected. Is there any additional work required so that vscode/$EDITOR/etc., picks up this file?

jbucaran:
@Swizz What about a typings.json file too?


```json
{
  "name": "hyperapp",
  "main": "src/hyperapp.d.ts"
}
```

Swizz:
@jbucaran From where are you using it ?

As it declared as a module, exported declarations will follow with
```tsx
import { h, app, Actions, Events, Application, emitFunction } from 'hyperapp'
```

For a local use, you will need to `npm link` your current hyperapp repo our explicitly import the declaration file

```tsx
/// <reference path="..." />
```

Isnt typings deprecated in favor of `package.json` ?
https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

jbucaran:
> @Swizz For a local use, ... our explicitly import the declaration file.

Sweet! How can I explicitly import the declaration file?

hyperapp.d.ts
```diff
@@ -0,0 +1,149 @@
...
+        ) => Partial<State> | any)
+      | Actions<State>
+  }
+
+  export interface Events<
+    State,
+    ActionsInterface extends Actions<State> | undefined
+  > {
+    loaded:
+      | ((
```
Swizz: I can remove the extra parentheses, it was for readability purpose but prettier do it as well

Swizz:
> Sweet! How can I explicitly import the declaration file?

```tsx
/// <reference path="./hyperapp.d.ts" />
```

hyperapp.d.ts
jbucaran: Let's give prettier the last word! :)

jbucaran:
@Swizz Sorry, but where do I put `<reference path="./hyperapp.d.ts" />`? I have no clue.

Swizz:
Sorry,  because it was defined as module, using it in a local environement is pretty tricky

```tsx
/// <reference path="./hyperapp.d.ts" />

import { h, app, Actions, Events, Application, emitFunction } from 'hyperapp'
```

hyperapp.d.ts
Swizz: It was not :/
Cause union types : 
`(a: number, b: string) => number | string` 
A function that take `a` as a `number` and `b` as a `string` and return a `number` or a `string`

**is not**
`((a: number, b: string) => number) | string`
A function that take `a` as a `number` and `b` as a `string` and return a `number` ; or a `string`

So let's let them here.

jbucaran: @Swizz So prettier is wrong?

Swizz: I was right by writing it ; prettier was right to keep i ; I was wrong at reading it again to take it as extra parentheses

jbucaran: omg 😆

hyperapp.d.ts
```diff
@@ -0,0 +1,149 @@
...
+  ) => VNode
+
+  export interface Actions<State> {
+    [key: string]:
+      | ((
+          state?: State,
+          actions?: Actions<State>,
+          data?: any,
+          emit?: emitFunction<any>
+        ) => Partial<State> | any)
```
Swizz: So, following https://github.com/hyperapp/hyperapp/pull/280#issuecomment-312893999

An action could return either a `Partial<State>`, `any` **or `void` or `never`** ?

Swizz:
If you are thinking about it this following implementation is also working as expected

```tsx
type State = number;

interface ActionsInterface extends Actions<State> {
  inc(count: State): State,
  dec(count: State): State
}
```
```tsx
interface ActionsInterface extends Actions<number> {
  inc(count: number): number,
  dec(count: number): number
}
```

hyperapp.d.ts
jbucaran: @Swizz An action can return an entire state (the same or a new one), a fragment of the state, a Promise, null or undefined.

Swizz: A Promise that resolve to a State (or fragment) ?

```tsx
=> Partial<State> | Promise<Partial<State>> | void | never
```

nippur72: I think it should be `Partial<State> | Promise<Partial<State>> | void`

jbucaran: A Promise that resolves to anything.

jbucaran:
I know @DanielRosenwasser expressed interest in helping out with the typings. Can you review this PR? 👋😏

hyperapp.d.ts
Swizz: We are on discussion about it above ^^'

Swizz: Something more like ?

```tsx
=> Partial<State> | Promise<Partial<State> | any> | void | never
```

nippur72: I don't understand the `never`, to my knowledge it's used for functions that do not return at all (e.g. they only `throw`). 
Promises may not return a partial state, they can just call other actions and do not return a value. So I am for:
```js
=> Partial<State> | Promise<Partial<State>|void|null> | void | null
```

jbucaran: @nippur72 Actually you can return a value from your promise and the receiver will get that value.

```jsx
app({
  view: (state, actions) =>
    <button
      onclick={() => {
        actions.do().then(data => alert(data))
      }}
    >
      Check
    </button>,
  actions: {
    do: state => new Promise(resolve => resolve("Hello"))
  }
})
```

Swizz: I was thinking about actions with infinite loop or something like.

And how about Promise that resolve to a Promise ? Is this even possible ?
```tsx
type ActionResult = Partial<State> | Promise<ActionResult> | void | null
```
```tsx
... => ActionResult 
```

jbucaran: Yes, I reckon that's possible too. So, that's why <samp>any</samp>.

Swizz:
```tsx
type ActionResult = Partial<State> | Promise<ActionResult> | any | void | null
```
```tsx
=> ActionResult 
```

jbucaran: Is void === undefined?

Swizz: void is the opposite of any both undefined or null.
But with strictNullChecks enable, it is needed to define null explicitly.

nippur72: in TypeScript `void` works a bit differently than other languages (C, Java etc); it's not just "no return value", but it's more like an alias for `undefined`. For example:
```ts
function a(): void {}
let b = a(); // strange, but this is allowed, and it's b === undefined
```
So I think it's ok to have `void` and left out `undefined` from our signature. I would also omit `any` because it makes the signature useless for type checking.

hyperapp.d.ts
```diff
@@ -0,0 +1,149 @@
+declare module "hyperapp" {
```
jbucaran: @Swizz Why not `declare namespace preact {`?

hyperapp.d.ts
```diff
@@ -0,0 +1,149 @@
...
+  export interface VNodeData {
+    [key: string]: any
```
jbucaran: Is this good? I ask because attributes can only be a string, number, function, null/undefined, an object in the case of key=`style`.

hyperapp.d.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  export interface VNode {
+    tag: string
+    data: VNodeData
+    children: string | VNode | Array<VNode>
+  }
+
+  export type viewFunction<State, Actions> = (
```
jbucaran: Why declare viewFunction's type here? and what is `vnodeFunction`?

hyperapp.d.ts
Swizz: Both are similar but different. declaring module are more ES6/CommonJS where declaring namespace is more like exposing a global object. 
This is why if you export and object you will need to export the namespace too.

````tsx
export = HyperApp

declare namespace HyperApp {
  ...
}

File: hyperapp.d.ts (hunk 4 above)
Comment (Swizz) [inline hyperapp.d.ts]: > - [ ] Less naive VNodeData interface

Here, as you wish, we can be really straight but generics :

```tsx
export interface VNodeData {
  [key: string]: string | number | Function | null
               | undefined | Object //as everything is object this is like all : `any`
}
```

Or we can dive into the horrific (like did in @Types/React), add a type to every attrs :-1: 
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L2101

hyperapp.d.ts
Swizz: It is requested into the Application interface 
```tsx
export interface Application<
    State,
    ActionsInterface extends Actions<State> | undefined,
    EventsInterface extends Partial<Events<State, ActionsInterface>> | undefined
  > extends Partial<Mixin<State, ActionsInterface, EventsInterface>> {
    root: HTMLElement
    view: viewFunction<State, ActionsInterface>
  }
```
and the render event 
```tsx
 export interface Events<
    State,
    ActionsInterface extends Actions<State> | undefined
  > {
 render:
      | (
          state: State,
          actions: ActionsInterface,
          view: viewFunction<State, ActionsInterface>,
          emit: emitFunction<any>
        ) => viewFunction<State, ActionsInterface>
  }
```

For `vnodeFunction` I wanted a way to express a function that return a vnode for `h`, and I wasnt sure about Component. But as the discution is closed, we can move into now.
```tsx
export function h(
    tag: string | vnodeFunction,
    data?: VNodeData,
    children?: Array<number | string | VNode>
  ): VNode
```

hyperapp.d.ts
jbucaran: @Swizz I don't know how to write this file, so I am only asking to learn. Please make the typings in the best way you know. Size doesn't matter.

hyperapp.d.ts
Swizz: So we are back on a circular type reference ?

````tsx
type ActionResult = Partial<State> | Promise<ActionResult> | void | null

export interface Actions<State> {
    [key: string]:
      | ((
          state?: State,
          actions?: Actions<State>,
          data?: any,
          emit?: emitFunction<any>
        ) => ActionResult)
      | Actions<State>
  }

File: hyperapp.d.ts (hunk 4 above)
Comment (Swizz) [inline hyperapp.d.ts]: Size doesn't matter, but I dont want you to be in charge of a monster. The react types is too too.

Here, I think we can allow the use of any. I dont want to maintain a complete list of HTML attribute, and how about custom ones ? And how about new ones ?
An attribute is nothing more a key and a value, and a value can be everything so `any`.

@nippur72 Do you agree ?

Comment (nippur72) [inline hyperapp.d.ts]: @Swizz attribute typing is a cool feature that lets you spot a lot of errors at compile time, so we should attempt to cover it. We don't have to implement it from scratch, we can easily adapt `react` or `preact` (their typings are MIT licensed). Custom attributes are not a problem: they do appear in custom elements which in turn are extensions of native HTML elements; their typings should be automatic if the component is written the right way.

The only problem I see is with events when an action is passed as event handler, e.g.
```js
<button onclick={actions.some}></button>
```
but I think that can be solved by something like `EventHanlder | Action`

jbucaran:
@Swizz How is this coming along? 😄 

Do you think you could help me a bit understand what's going on this file? It's really big and I am not
 familiar with any of the syntax. 😅

Swizz:
Let me check, what is needed to do right now.

* We need to conclude about Actions API
```tsx
type ActionResult = Partial<State> | Promise<ActionResult> | void | null

export interface Actions<State> {
    [key: string]:
      | ((
          state?: State,
          actions?: Actions<State>,
          data?: any,
          emit?: emitFunction<any>
        ) => ActionResult)
      | Actions<State>
}
```

* Decide about a naive or a complete VnodeData interface
```tsx
export interface VNodeData {
    [key: string]: any
}
```

* Perform some name polish, like `Component` instead of `vnodeFunction`

* And maybe add some documentations in the `.d.ts` file

---

Isnt [Typescript Guide](https://www.typescriptlang.org/docs/home.html) a good teacher ? ^^

Swizz:
Can I ask @nippur72 @rbiggs @LucianU and @DanielRosenwasser reviews and feedbacks on it ?

jbucaran:
> @Swizz Perform some name polish, like Component instead of vnodeFunction

That sounds correct. The term we use in the documentation is Component, so that's what it should be.

jbucaran:
@Swizz Actions/Events are no longer passed the `emit` function, so that argument should be removed too.

Swizz:
@jbucaran You are right, I need to rethink the API regarding recent changes

pspeter3:
Would it be helpful to have another reviewer? I made a PR for Picodom, Preact, and I've done a lot of the type definitions in DefinitelyTyped.

jbucaran:
@pspeter3 Welcome aboard! 👍

hyperapp.d.ts
pspeter3: I feel like you need to consider the various consumption cases for hyperapp

hyperapp.d.ts
pspeter3: I think you can use the example I have from Picodom. We should absolutely avoid using `any` and prefer `{}` when possible.

hyperapp.d.ts
pspeter3: In general, it's better to declare interfaces than types. You can make an interface like,

```ts
interface View<S, A> {
     (state: S, actions: A): VNode;
}
```

hyperapp.d.ts
```diff
@@ -0,0 +1,149 @@
...
+  export type emitFunction<Data> = (event: string, data?: Data) => Data
+  export type vnodeFunction = (
+    props: VNodeData,
+    children: string | VNode | Array<VNode>
+  ) => VNode
+
+  export interface Actions<State> {
+    [key: string]:
+      | ((
+          state?: State,
```
pspeter3: I feel like this is begging for a named interface. TypeScript produces better error messages against named interfaces rather than anonymous types.

Swizz:
@pspeter3 May I ask you to make a PR on my fork ? I am a little bit busy these days. Or I can make you as contributor as well

jbucaran:
/cc @Swizz 

@pspeter3 Can you create a new PR with your own implementation?

pspeter3:
Yeah, I can do that.

hyperapp.d.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  export interface Mixin<
+    State,
+    ActionsInterface extends Actions<State> | undefined,
+    EventsInterface extends Partial<Events<State, ActionsInterface>> | undefined
+  > {
+    state: State
+    actions: ActionsInterface
+    events: EventsInterface
+    mixins: Array<Partial<Mixin<State, ActionsInterface, EventsInterface>>>
```
andrewiggins: Per the [doc](https://github.com/hyperapp/hyperapp/blob/master/docs/mixins.md), does this need to be an array of functions that return `Partial<Mixin>`'s?

hyperapp.d.ts
andrewiggins: Is it VNodeData where [lifecycle methods](https://github.com/hyperapp/hyperapp/blob/master/docs/lifecycle-events.md) and [keys](https://github.com/hyperapp/hyperapp/blob/master/docs/keys.md) would go? Would be nice to demonstrate those here too.

hyperapp.d.ts
Swizz: Your Mixins can contains only `events` or `actions` for example. You do not need to define all the Mixin props. So, `Partial<Mixin>` ?

andrewiggins: True. Partial is definitely needed. I was suggesting, should the type of the mixins property be:
```typescript
mixins: Array<() => Partial<Mixin<State, ActionsInterface, EventsInterface>>>
```

Unless I'm misunderstanding the API doc

hyperapp.d.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  export type viewFunction<State, Actions> = (
+    state: State,
+    actions: Actions
+  ) => VNode
+  export type emitFunction<Data> = (event: string, data?: Data) => Data
+  export type vnodeFunction = (
+    props: VNodeData,
+    children: string | VNode | Array<VNode>
```
andrewiggins: Is this a type for the [Stateless Components doc](https://github.com/hyperapp/hyperapp/blob/master/docs/stateless-components.md)? If I'm reading the code correctly, the last line in h.js (`tag(data, children)`) seems to say the children type would always be an array:
```typescript
children: Array<VNode | string>
```

Swizz:
We reached a lot of comments here.
And I thought alot about it. How guys are you regarding closing this one an starting a fresh new one ?
Step by step, Together. We make an attempt to build the strongest one an discuss about it after. It will be more plesant and eficient to build it together.

---
> Would it be possible to create something simpler first?
> You know like skateboard -> bicycle -> bike -> car -> bus -> tractor -> boeing 777 -> tesla self-landing rocket
>
> @jbucaran - *Slack*

jbucaran:
@Swizz Sounds good to me! 👍

andrewiggins:
Happy to help out

Swizz:
Closed in favor of #311

PR #280, opened 2017-07-04
Simplify API surface.

jbucaran:
- [x] Don't pass `appOptions` to mixins.
- [x] Don't pass `emit` to events.
- [x] Don't pass `emit` to actions.
- [x] Update docs.
- [x] Pass `emit` as 2nd argument to mixins.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -47,20 +47,20 @@ Type: ([state](#state), [actions](#actions)): [vnode]
 ### actions
 #### <a name="actions-foo"></a>[namespace.]_foo_
 
-Type: ([state](#state), [actions](#actions), [data](#actions-data), [emit](#emit))
+Type: ([state](#state), [actions](#actions), [data](#actions-data))
 
 * <a name="actions-data"></a> data: any
 
 ### events
 #### ready
 
-Type: ([state](#state), [actions](#actions), _, [emit](#emit)) | Array\<[events](#ready)\>
+Type: ([state](#state), [actions](#actions)) | Array\<[events](#ready)\>
 
 Fired after the view is mounted on the DOM.
 
 #### action
 
-Type: ([state](#state), [actions](#actions), [data](#action-data), [emit](#emit)): [data](#action-data) | Array\<[action](#action)\>
+Type: ([state](#state), [actions](#actions), [data](#action-data)): [data](#action-data) | Array\<[action](#action)\>
 
 * <a name="action-data"></a>data
   * name: string
@@ -70,7 +70,7 @@ Fired before an action is triggered.
 
 #### update
 
-Type: ([state](#state), [actions](#actions), [data](#update-data), [emit](#emit)): [data](#update-data) | Array\<[update](#update)\>
+Type: ([state](#state), [actions](#actions), [data](#update-data)): [data](#update-data) | Array\<[update](#update)\>
 
 * <a name="update-data"></a>data: the updated fragment of the state.
 
@@ -78,7 +78,7 @@ Fired before the state is updated.
 
 #### render
 
-Type: ([state](#state), [actions](#actions), [view](#view), [emit](#emit)): [view](#view) | Array\<[render](#render)\>
+Type: ([state](#state), [actions](#actions), [view](#view)): [view](#view) | Array\<[render](#render)\>
 
 Fired before the view is rendered.
 
@@ -88,13 +88,13 @@ Type: Array\<[Mixin](#mixins-mixin)\>
 
 #### <a name="mixins-mixin"></a>Mixin
 
-Type: ([props](#app-props)): [props](#mixin-props)
+Type: ([emit](#emit)): [props](#mixin-props)
 
 * <a name="mixin-props"></a>props
-  * [mixins](#mixins)
   * [state](#state)
   * [actions](#actions)
   * [events](#events)
+  * [mixins](#mixins)
 
 ### root
 
--- a/docs/core.md
+++ b/docs/core.md
@@ -275,15 +275,14 @@ For a practical example see the implementation of the [Router](https://github.co
 
 #### Custom Events
 
-To create custom events, use the [emit(event, data)](/docs/api.md#emit) function. This function is passed as the last argument to actions/events.
+The [app](/docs/api.md#app) call returns the [emit](/docs/api.md#emit) function, making it possible to trigger [custom events](#custom-events).
 
 ```jsx
-app({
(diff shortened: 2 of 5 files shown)


jbucaran:
@zaceno There's a little issue when using `emit` inside the `ready` event. Since at that time `app` hasn't yet returned, emit is not available.

Swizz:
Stupid question but why ? 

It is all about purity, but, if I dont want to use this pattern : 
```jsx
cons emit = app({
  events: ...
  actions: ...
})
```

but directly 
```jsx
app({
  events: ...
  actions: ...
})
```

I wont be able to emit events ?

zaceno:
@Swizz the solution would be to define your actions and events in one or more mixins. Since mixins get emit passed to them.

So there is a way. Wether you like that way or not, though... I can see how this might not be liked by everybody. (I like it though)

zaceno:
For me: getting emit passed to mixins is the best thing about this. Cleaning up the api by removing the emit parameter to actions and events would be nice, but if people are generally against it it may not be such a good idea.

Swizz:
I am not fundamentally against (not angry here), I was just raising questions in the case, in a genius bloom, nobody was thinking about it

zaceno:
@Swizz It *is* a good question to raise. Because it does mean that if you intend to compose together your app-options, and you need to emit custom events, you *must* use mixins. (Or do something really hacky. Not sure how, but there's probably a way)

jbucaran:
I thin this is okay in preparation of a 1.0 release. If the world falls apart later, we can just add it again and it would not break anything, because:

download

docs/api.md
```diff
@@ -91,10 +91,10 @@ Type: Array\<[Mixin](#mixins-mixin)\>
 Type: ([props](#app-props)): [props](#mixin-props)
```
zaceno: Should read:

Type: ([props](#app-props), [emit](#emit)): [props](#mixin-props)

jbucaran: @zaceno Done.

Swizz:
From

```jsx

const state = {
  count: 0
}

const actions = {
  inc: (state) => ({ count: state.count + 1 }),
  done: (_state, _actions, _data, emit) => emit('done')
}

const view = (state, actions) => {
  return h('button', { onclick: actions.inc }, state.count)
}

const emit = app({ root: document.getElementById('app'), state, view, actions })
```

to
```jsx

const state =
  count: 0
}

const actions = (app, emit) => ({
  actions : {
    inc: (state) => ({ count: state.count + 1 }),
    done: () => emit('done')
  }
})

const view = (state, actions) => {
  return h('button', { onclick: actions.inc }, state.count)
}

const emit = app({ root: document.getElementById('app'), state, view, mixins: [actions] })
```

jbucaran:
@Swizz `emit` is considered a lower level function. You rarely need it in your app. Why would you trigger an event if you can call an action? Now, when you do need it, it's probably better to hide that complexity in a mixin.

zaceno:
One of the primary use-cases I've found for emit, is actually when I'm trying to decouple my app into separate mixins.

Where sometimes in mixin A, I'll need to call an action in mixin B. Now I may *know* the path to the action in mixin B. But it's not good practice to hardcode that sort of logic into mixin B. I'd like to be able to change mixin A without breaking mixin B.

So I'll use events for that. Rather than call the action in mixin B, I'll just emit an event, and have mixin A listen for it.

Swizz:
I will need to change my mind about events tho.
In a Vuex way, I was using actions as Mutators (things that change the state or call events) and events as Actions (things will do not pure stuffs).

jbucaran:
By the way, about passing the app into mixins. Other than the view is useless information because the `state` is the global state before other mixins have loaded, the actions are not yet initialized/wrapped.

zaceno:
@jbucaran I was just looking at that yesterday. Why do we do that? What's the envisioned use case? (Maybe that's for a slack-discussion, but anyway)

jbucaran:
@Swizz If you were using events in that way and this PR causes you to realize that was not the hyperapp-way, then I think we succeeded. While you can continue to do so at a certain extent with mixins, you should be using `actions` both as a reducers/mutators and side effects.

jbucaran:
> @zaceno Why do we do that? 

Because I didn't know what else to pass. Shall we knock it off too?

zaceno:
@Swizz: back in the day we used to have "reducers" = vuex mutators, and "effects" = vuex actions. Then we merged them together into hyperapp's actions which mostly resemble vuex mutators, but are able to call other actions, and do async stuff, so a little like vuex effects at the same time.

zaceno:
@jbucaran: why not, if there's no reason for it. It would certainly help the sinister alchemy I've got cooking.

jbucaran:
@Swizz With regards to that reason, it was a tradeoff (to me). By making actions both mutators/effects we were not really changing how anything worked internally, but it made the API much cleaner and easier to learn.

jbucaran:
@zaceno I've updated this PR so we don't pass the app to mixins either.

I also checked all the mixins I know in existence, ones I've made and ones that have been published as I was made aware of and none were using appOptions. Actually, the only one was the router, which I'll make sure to update after this is merged.

jbucaran:
@MatejMazur @jamen @andyrj have all authored mixins, so this PR is of your interest. Please have a look when you have a chance. 🙏

andyrj:
I don't believe I utilized emit in my mixins, so as long as actions and events are still going to be available inside a mixin I am ok with this change and can update how actions are initialized.  I like the cleaner api this provides.

jamen:
These changes seem okay to me.  I haven't personally used `appOptions` or `emit`in my mixins  yet, so it doesn't affect me.  But, I wasn't really a fan of `appOptions` to begin with.

MatejMazur:
I'm ok with it.

SkaterDad:
Forcing people to use mixins when they want custom events seemed like a strange idea at first, but I'm warming up to it.  Restricting how people use parts of the framework is certainly in line with the Elm philosophy.

It is nice that you can emit events directly from the view functions with this change.

The only downside is the way to use `emit` at the main app-level actions now is really awkward in my opinion.  It almost looks like it shouldn't work 😆 
```js
const emit = app({
  actions: {
    alldone:  () => emit('done')
  }
})
```

jbucaran:
@SkaterDad It will work in actions because they will run after the app() call has returned for sure.

It will not work for events.ready though, since that one runs before app returns.

zaceno:
🎉

PR #282, opened 2017-07-04
Hydrate

andyrj:
pulled over hydrate changes to new branch with clean history to stop spamming old comments from reverted commits in my master...

old PR's https://github.com/hyperapp/hyperapp/pull/267 and https://github.com/hyperapp/hyperapp/pull/255 were both closed in favor of this PR which cleaned commit history and implementation of this functionality.

Could use a review on this because with the sync change this may have some impact on the mixin initialization that I am not aware of and if so then we can move the node/element initialization lower in the app() method I wasn't 100% about where I put it and thought maybe init() might be better place if this causes any issues...


src/app.js
```diff
@@ -7,6 +7,17 @@ export function app(app) {
...
+  function hydrate(elm) {
+    return elm == null ? undefined : {
+      tag: elm.tagName,
+      data: {},
+      children: [].map.call(elm.childNodes, function(child){hydrate(child)})
+    }
+  }
+
+  element = root.firstChild
+  node = hydrate(element)
```
jamen: Does this need to be a function?  Why couldn't it just be an `if`?  Something like:

```js
if (element === null) {
  node = {
    tag: element.tagName,
    data: {},
    children: [].map.call(elm.childNodes, function(child){hydrate(child)}
  }
}

File: test/root.test.js [hunk 2]
```diff
@@ -45,6 +46,7 @@ test("non-empty root", () => {
     </main>
   `
 })
+*/
```
jamen: Hmm, I think think there is some instances we want this test, no?  Say for example, `document.body` at root, and you have a `<script>` inside of it.

andyrj: I don't really agree, this describes special casing render target selection and is really not going to be compatible with SSR in general.  If the user is placing <script> tags in the body they should also be providing a render target that does not include those script tags imo.

The described behavior reads to me that if no target is specified and body has children, then we skip to last node and render there?  or maybe it's target first node with matching name, I didn't track down the logic covered in this test?  Either way it is some kind of special case that I think makes more sense to be an explicit choice by the user.

src/app.js
andyrj: it's recursive so it would need to be function or it would need a loop and a stack to accomplish, there is discussion about moving this logic upto the SSR code and turning this into:
```js
node = JSON.parse(document.getElementById('nodes'))
```
Which would mean in the SSR code we would build the vnode object and serialize it to a script tag we send down from server like:
```html
<html>
<head>
  <script type='application/json'>
  {
    tag: 'div',
    data: {},
    children: [...]
  }
  </script>
</head>
</html>
```
This came up yesterday in the following discussion on hyperapp/server: https://github.com/hyperapp/server/issues/7

jamen: Ah my bad I didn't see that piece at first.

andyrj:
last commit was needed as I ran into bug where everything passed tests fine, but browser behaved differently than Jest...  The conditional is necessary for this code to work in browser and undom for server side render with hydrate :( might be able to remove the if statement if we implement a simple element.hasChildNodes, in our version of undom...  not 100% how hard that would be atm

src/app.js
```diff
@@ -7,6 +7,19 @@ export function app(app) {
...
+      data: {},
+      children: [].map.call(elm.childNodes, function(child){hydrate(child)})
+    }
+  }
+
+  if (root.hasChildNodes !== undefined && root.hasChildNodes()) {
+    element = root.children[0]
+    node = hydrate(element)
+  }
+
```
jbucaran: @andyrj What if the `root` is not empty. Like when appending to an element that already has children in it?

andyrj: That is why I have been asking for clarification on the non-empty root test case.  Is that the expected behavior, if root is not empty then append to children?  That is not what I would expect, and it makes hydration impossible just by contradicting the needed behavior I believe.  I am not aware of react/preact/inferno/Vue having this append to children if root isn't empty behavior, but they do hydrate existing dom from ssr to prevent flicker from removing existing dom before rendering.  

I think it will be clearer if I show a video of the behavior of hyperapp before and after this patch.  I am making one now to share for clarification.

jbucaran: > @andyrj  Is that the expected behavior, if root is not empty then append to children?

The current behavior is to render the tree in the given root or document.body if none is given. 

@ngryman Can you chime in here too? 👋 

> I think it will be clearer if I show a video of the behavior of hyperapp before and after this patch. I am making one now to share for clarification.

Thanks! 👍

andyrj: https://youtu.be/HI4TiBfWFNQ

ngryman: I never used SSR (bouh) but I think I understand the general problem: how to differentiate markup generated by SSR from "lambda" markup that lives on its own. One simple solution could be to add an attribute to the root of the SSR rendered DOM element (say `data-ssr` or whatever). So the the client HyperApp have info on how to hydrate it. Tell me if I misunderstood something.

jbucaran: @ngryman Can I have your input of:

> andyrj Is that the expected behavior, if root is not empty then append to children?

Basically, in order to hydrate the initial undefined node tree, the root must be non-empty. You'll recall that we didn't support that use case until you introduced the feature decades ago.

andyrj: I know it won't be the most elegant, but instead of requiring some dom attribute to differentiate, I think adding a flag for hydrate to the call to app is cleaner.

```js
app({state, views, mixins, hydrate: true})
```
This way by default the current behavior can be applied, and then if the user is using SSR they can opt-in to hydrate the dom.  However, I am open to whatever approach you all feel is best.  I just really want to be able to avoid duplicating dom or flickering on first client side renders for my use case.  :smile:

ngryman: @jbucaran Yes I remember. My point was, isn't it possible to know which is the root node to hydrate?

ngryman: Vuejs does that: https://github.com/vuejs/vue-ssr-docs/blob/master/en/hydration.md

andyrj: But even that link to Vue show that they force the user to target the div in that case with id='app'.  I don't like the implicit nature of this default behavior of appending to the end of the existing children and I feel like it will confuse end users that run into that behavior unexpectedly.

jbucaran:
@andyrj Can you rebase with https://github.com/hyperapp/hyperapp/tree/feat/hydrate?

PR #283, opened 2017-07-05
Add new lifecycle method: oninsert.

jbucaran:
- oninsert is a new lifecycle method that occurs after and
  element is created and mounted/inserted into the DOM.

- Also rename lifecycle events -> lifecycle methods to avoid
  creating new/unfamiliar terms.

- Update docs.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -106,6 +106,8 @@ export function app(app) {
       for (var i in node.data) {
         if (i === "oncreate") {
           node.data[i](element)
+        } else if (i === "oninsert") {
+          setTimeout(node.data[i], 0, element)
         } else {
           setElementData(element, i, node.data[i])
         }
--- a/test/lifecycle.test.js
+++ b/test/lifecycle.test.js
@@ -3,13 +3,31 @@ import { expectHTMLToBe } from "./util"
 
 beforeEach(() => (document.body.innerHTML = ""))
 
+const getElementByTagName = tag => document.getElementsByTagName(tag)[0]
+
 test("oncreate", done => {
   app({
-    state: 1,
-    view: state =>
+    view: () =>
       h("div", {
-        oncreate: e => {
-          expect(state).toBe(1)
+        oncreate: element => {
+          expect(element).not.toBe(undefined)
+          expect(getElementByTagName("div")).toBe(undefined)
+
+          setTimeout(() => {
+            expect(getElementByTagName("div")).toBe(element)
+            done()
+          })
+        }
+      })
+  })
+})
+
+test("oninsert", done => {
+  app({
+    view: () =>
+      h("div", {
+        oninsert: element => {
+          expect(getElementByTagName("div")).toBe(element)
           done()
         }
       })
@@ -21,7 +39,7 @@ test("onupdate", done => {
     state: 1,
     view: state =>
       h("div", {
-        onupdate: e => {
+        onupdate: element => {
           expect(state).toBe(2)
           done()
         }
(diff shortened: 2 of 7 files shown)


src/app.js
```diff
@@ -106,6 +106,8 @@ export function app(app) {
       for (var i in node.data) {
         if (i === "oncreate") {
           node.data[i](element)
+        } else if (i === "onmount") {
```
zaceno: `oninsert`... right?

test/lifecycle.test.js
```diff
@@ -3,13 +3,31 @@ import { expectHTMLToBe } from "./util"
...
+  })
+})
+
+test("onmount", done => {
+  app({
+    view: () =>
       h("div", {
-        oncreate: e => {
-          expect(state).toBe(1)
+        onmount: element => {
```
zaceno: `oninsert` ...

test/lifecycle.test.js
```diff
@@ -21,7 +39,7 @@ test("onupdate", done => {
     state: 1,
     view: state =>
       h("div", {
-        onupdate: e => {
+        onupdate: element => {
```
zaceno: `oninsert`...

zaceno: nevermind. ... not sure why I flagged this one :P

src/app.js
jbucaran: lol

jbucaran:
@zaceno Fixed! 🎉

zaceno:
... oh you already merged it. Thats why, Lol!

jamen:
Hmm, I am unsure about "Lifecycle Methods" though.  What is the confusion?

jbucaran:
@jamen I am borrowing the term from react. Newcomers are likely to be more familiar with that term. But more importantly, lifecycle events !== appOptions.events, which may actually cause confusion.

jamen:
Agree on the last point.

The first point is trivial imo, but I'll just say this:  "methods" seems like object-orientated wording to me.  Because in React they are [methods on a class](https://facebook.github.io/react/docs/react-component.html#the-component-lifecycle). Here I just thought of them as functions and more specifically events.  Especially with the usage `<h1 oninsert={fn}>`.

Swizz:
- Snabbdom call them (lifecycle events/methods) Hooks
- Preact is with Lifecycle methods too (It seem you are a lot inspired by Preact isnt it ?)
- Vue is in favor of Lifecycle Hooks
- Inferno use Lifecycle events

I think naming is not an issue ; if the doc is well written

jbucaran:
> @jamen "methods" seems like object-orientated wording

Good point. Oh, what to do...? 🤔

SkaterDad:
My vote would be "hooks" or "events".

Swizz:
Hooks looks good associated to a Lifecycle diagram.

e.g : [Vue - Lifecycle Diagram](https://vuejs.org/v2/guide/instance.html#Lifecycle-Diagram)
This diagram show us the instance lifecycle ; and in red there are some **hooks** to catch some moments of the instance lifecycle.

jbucaran:
So, I went with events for everything. I like the harmony it brings to the docs.

- app `events` property.
- custom events, created via `emit("myCustomEvent", data)`
- lifecycle events, oncreate, oninsert, onupdate, onremove

Swizz:
> - lifecycle events, oncreate, oninsert, onupdate, onremove

Lifecycle Events looks also great associated to DOM Events. Same API, one deal with Lifecycle, the other one with regular events : 

````jsx
<div
  oninsert={ () => {} }
  onupdate={ () => {} }
  onclick={ () => {} }
  onhover={ () => {} }
/>

PR #286, opened 2017-07-06
Use requestAnimationFrame to throttle DOM patches.

jbucaran:
## tl;dr
- Use rAF to throttle renders / batch render calls.
- Introduce new `init` event that run before the view has rendered the first time, but after the app internal state has been initialized.   

- Revert `ready` to `loaded` event. This event now fires once when the view has been attached to the DOM.

## Overview

Think of requestAnimationFrame (rAF) as a clever setTimeout. 

rAF tells the browser to call a specified function before the next screen repaint. This allow us to throttle DOM patches in  our render pipeline.

Imagine your app calls hundreds (or thousands) of actions within a very short time span, a likely scenario in real world apps.

This is how the story unfolds. We:

  1. Call your action to obtain a new/partial state and merge   it with the global state (or return immediately if the action doesn't return anything). At the end of this step we have a new global state.
  
  2. Call render() (which calls your view() function to get the vtree). The vtree is likely to be large and complex, composed of hundreds or thousands of h() calls.

  3. Call patch() with the new and old vtree. Patch walks through the new vtree comparing each vnode with its corresponding vnode in the old vtree and patches the real DOM where it needs.
Patch essentially means creating, updating, replacing or removing DOM nodes. This is usually the bottleneck.

  4. Finish. Step out of app.

How can we improve (2) and (3) other than by writing a more efficient patch algorithm? 

> TODO: Write a more efficient patch algorithm.

## Introducing rAF

What's going to happen now is that after (1) we are going to 
lock the render() call.

```js
function render() {
  if (!lockRender) {
    lockRender = true

    requestAnimationFrame(() => {
      PATCH
      lockRender = false
    })
  }
}
```

This mechanism allow us to immediately skip over to (4) and not block,i.e., not call view() and patch(). 

Actions will continue to run skipping (2) and (3) straight-to (4).

The global state will continue to be updated and action/update events will continue to fire.

At some point, and in good timing, the browser will run the function we scheduled using rAF. It's then, and only then when (2) and (3) will be finally called. 

Finally, inside the callback, we now can _unlock_ the render() function to allow the scheduling of future patches.


src/app.js
```diff
@@ -22,17 +26,31 @@ export function app(app) {
...
+
+      requestAnimationFrame(function() {
+        element = patch(
+          root,
+          element,
+          node,
+          (node = emit("render", view)(state, actions))
+        )
+
+        locked = !locked
```
SkaterDad: The "loaded" event could certainly trigger some user-defined actions, so this does seem like a potential issue.

SkaterDad: This code below produces 2 calls to `render` on app startup with this PR.   Both of the "loaded" event handlers get lumped into one animation frame, so you essentially have the initial render, then a "loaded" render.

Moving the `locked =! locked` after the `emit:loaded` breaks things, and the actions fired by the "loaded" event handlers doesn't trigger another render.  The state still updates, but nothing is rendered, which becomes clear when you click the "increment" button and see the number jump up.

```js
const {h, app} = hyperapp

const mixin = () => {
  return {
    events: {
      loaded: (state, actions) => {
        actions.inc()
      }
    }
  }
}

app({
  state: 1,
  actions: {
    inc: (state) => state +1
  },
  events: {
    loaded: (state, actions) => {
      console.log("loaded")
      actions.inc()
    },
    render: () => console.log("rendering")
  },
  view: (state, actions) => h('div', null, ['Hello ' + state, h('button', {onclick: () => actions.inc()}, "Increment")]),
  mixins: [mixin]
})
```

ngryman: > Simply moving the locked =! locked after the emit:loaded breaks things, and the actions fired by the "loaded" event handlers doesn't trigger another render.

Isn't it the actual expected behavor?

SkaterDad: > Isn't it the actual expected behavor?
Maybe I should have stated that `patch` never gets called, so the app state and view are out of sync.

ngryman: Ouch ok...

SkaterDad: I've just realized we may be thinking about this wrong. 

Anything which needs to happen before the first render should be in the "init" event.

It's probably a good thing that the "loaded" event causes another render, otherwise those 2 events would be identical.

In conclusion.... nevermind! :laughing:

jbucaran: @SkaterDad @ngryman Thank you both! This discussion made looked at this again, but I guess we all now can see that unlocking after emitting `loaded` would prevent actions called inside `loaded` to be schedule a render.

As a rule of thumb, I would use `init` to call actions and not `loaded`, since at that point the DOM has already been patched once, but it should be possible.

ngryman: I'm not sure we want as many renders as side effects produced by `loaded`, @jbucaran?

SkaterDad: @ngryman So long as your "loaded" event handlers are synchronous, they should all get batched into one render call.  At least that's what I saw in my example app pasted above.  Anything async would just render later on when resolved, as expected.

jbucaran: @ngryman If we unlock at the end of the block as you suggested, then whatever actions you may call in `loaded` will not be able to schedule any renders. So, the state will out of sync with the app.

ngryman: Alright!

jbucaran:
Merge? 😄

andyrj:
Looks good to me, will be nice to have debounced renders.

PR #289, opened 2017-07-07
Optimize h

ngryman:
The child of #187.

**Before**
```sh
NANOBENCH version 2
> node

# vnode with no child 1M times
ok ~319 ms (0 s + 319243686 ns)

# vnode with a single child 1M times
ok ~1.48 s (1 s + 481573267 ns)

# vnode with a single array child 1M times
ok ~1.67 s (1 s + 670888960 ns)

all benchmarks completed
ok ~3.47 s (3 s + 471705913 ns)
```

**After**
```sh
NANOBENCH version 2
> node

# vnode with no child 1M times
ok ~330 ms (0 s + 330196882 ns)

# vnode with a single child 1M times
ok ~954 ms (0 s + 953737873 ns)

# vnode with a single array child 1M times
ok ~990 ms (0 s + 990082693 ns)

all benchmarks completed
ok ~2.27 s (2 s + 274017448 ns)
```

I've committed a small facility to compare perfs, I'll strip them before squashing and merging.

Things to explore:
 - [ ] Better bench use cases. @jbucaran ideas?
 - [ ] Node constructors (https://github.com/hyperapp/hyperapp/pull/187#issuecomment-294316864)
 - [ ] Recycling.

Other `h` out there:
 - **preact**: https://github.com/developit/preact/blob/master/src/h.js
 - **react**: https://github.com/facebook/react/blob/5c6a496d98b80d19d851c2ec3a56d760b7118a50/src/isomorphic/classic/element/ReactElement.js#L180
 - **inferno**: https://github.com/infernojs/inferno/blob/master/packages/inferno-create-element/src/index.ts

Refs:
 - Why 2 for loops: https://github.com/hyperapp/hyperapp/pull/187#issuecomment-294339008

### DIFF
--- a/perf/h.js
+++ b/perf/h.js
@@ -1,0 +1,26 @@
+import bench from 'nanobench'
+import { h } from '../src/h'
+
+bench("vnode with no child 1M times", b => {
+  b.start()
+  for (let i = 0; i < 10000000; i++) {
+    h("div", {})
+  }
+  b.end()
+})
+
+bench("vnode with a single child 1M times", b => {
+  b.start()
+  for (let i = 0; i < 10000000; i++) {
+    h("div", {}, ["foo"])
+  }
+  b.end()
+})
+
+bench("vnode with a single array child 1M times", b => {
+  b.start()
+  for (let i = 0; i < 10000000; i++) {
+    h("div", {}, [["foo"]])
+  }
+  b.end()
+})
--- a/src/h.js
+++ b/src/h.js
@@ -1,22 +1,29 @@
-export function h(tag, data) {
-  var node
-  var stack = []
+function addChild(children, val) {
+  if (val == null || typeof val === 'boolean') return
+
+  if (Array.isArray(val)) {
+    for (var j = 0; j < val.length; j++) {
+      children.push(val[j])
+    }
+  }
+  else {
+    children.push(typeof val === 'number' ? val + '' : val)
+  }
+}
+
+export function h(tag, data, values) {
   var children = []
 
-  for (var i = arguments.length; i-- > 2; ) {
-    stack[stack.length] = arguments[i]
-  }
-
-  while (stack.length) {
-    if (Array.isArray((node = stack.pop()))) {
-      for (var i = node.length; i--; ) {
-        stack[stack.length] = node[i]
+  if (arguments.length > 2) {
+    if (Array.isArray(values)) {
+      for (var i = 0; i < values.length; i++) {
+        addChild(children, values[i])
       }
-    } else if (node != null && node !== true && node !== false) {
-      if (typeof node === "number") {
-        node = node + ""
+    }
+    else {
+      for (var i = 2; i < arguments.length; i++) {
+        addChild(children, arguments[i])
       }
-      children[children.length] = node
     }
   }
 
--- a/src/h.js
+++ b/src/h.js
@@ -1,28 +1,18 @@
-function addChild(children, val) {
-  if (val == null || typeof val === 'boolean') return
+var isArray = Array.isArray
 
-  if (Array.isArray(val)) {
-    for (var j = 0; j < val.length; j++) {
-      children.push(val[j])
-    }
-  }
-  else {
-    children.push(typeof val === 'number' ? val + '' : val)
-  }
-}
-
-export function h(tag, data, values) {
+export function createNode(tag, data, values) {
   var children = []
 
-  if (arguments.length > 2) {
-    if (Array.isArray(values)) {
-      for (var i = 0; i < values.length; i++) {
-        addChild(children, values[i])
+  for (var i = 0; i < values.length; i++) {
+    var val = values[i]
+    if (val && val !== true && val !== false) {
+      if (isArray(val)) {
+        for (var j = 0; j < val.length; j++) {
+          children.push(val[j])
+        }
       }
-    }
-    else {
-      for (var i = 2; i < arguments.length; i++) {
-        addChild(children, arguments[i])
+      else {
+        children.push(typeof val === 'number' ? val + '' : val)
       }
     }
   }
@@ -34,4 +24,15 @@ export function h(tag, data, values) {
         children: children
       }
     : tag(data, children)
+}
+
+export function h(tag, data, values) {
+  if (!isArray(values)) {
+    values = []
+    for (var i = 2; i < arguments.length; i++) {
+      values.push(arguments[i])
+    }
+  }
+
+  return createNode(tag, data, values || [])
 }
--- a/src/h.js
+++ b/src/h.js
@@ -5,7 +5,7 @@ export function createNode(tag, data, values) {
 
   for (var i = 0; i < values.length; i++) {
     var val = values[i]
-    if (val && val !== true && val !== false) {
+    if (val && val !== true) {
       if (isArray(val)) {
         for (var j = 0; j < val.length; j++) {
           children.push(val[j])
--- a/package.json
+++ b/package.json
@@ -7,7 +7,10 @@
   "module": "src/index.js",
   "license": "MIT",
   "repository": "hyperapp/hyperapp",
-  "files": ["src", "dist"],
+  "files": [
+    "src",
+    "dist"
+  ],
   "author": "Jorge Bucaran",
   "keywords": [
     "hyperapp",
@@ -24,7 +27,9 @@
     "build": "rollup -cm -n hyperapp -f umd -i src/index.js -o dist/hyperapp.js",
     "prepublish": "npm run build",
     "format": "prettier --semi false --write 'src/**/*.js'",
-    "release": "npm run build && npm test && git commit -am $npm_package_version && git tag $npm_package_version && git push && git push --tags && npm publish"
+    "release": "npm run build && npm test && git commit -am $npm_package_version && git tag $npm_package_version && git push && git push --tags && npm publish",
+    "bench": "rollup perf/h.js -f cjs | node",
+    "trace": "rollup perf/h.js -f cjs | node-irhydra"
   },
   "babel": {
     "presets": "es2015"
@@ -35,7 +40,9 @@
     "jest": "^20.0.4",
     "prettier": "~1.5.2",
     "rollup": "^0.43.0",
-    "rollup-plugin-uglify": "^2.0.1"
+    "rollup-plugin-uglify": "^2.0.1",
+    "nanobench": "^2.1.0",
+    "node-irhydra": "^1.0.0"
   },
   "bundlesize": [
     {


ngryman:
I'll increase coverage when we validate the code before.

src/h.js
```diff
@@ -1,22 +1,29 @@
...
+}
 
-  while (stack.length) {
-    if (Array.isArray((node = stack.pop()))) {
-      for (var i = node.length; i--; ) {
-        stack[stack.length] = node[i]
+export function h(tag, data, values) {
+  var children = []
+
+  if (arguments.length > 2) {
```
selfup: would doing `arguments[2] !== undefined` be faster here?

ngryman: I'll check this, but from what I remember, accessing an out of bounds array deopts the function because the VM has to do additional bounding checks.

selfup: Yes I think you are correct. Keeping everything within bounds will be a safer bet here. Let the VM _do the thing_ 🎉

perf/h.js
```diff
@@ -0,0 +1,26 @@
+import bench from 'nanobench'
+import { h } from '../src/h'
+
+bench("vnode with no child 1M times", b => {
+  b.start()
+  for (let i = 0; i < 10000000; i++) {
```
selfup: Would making a helper function for the 1M `for` loop be a bad idea in terms of the benchmark?

like:

```javascript
const ONE_MILLION = 1000000

function benchLoop(fn) {
  for (let i = 0; i < ONE_MILLION; i++) {
    fn()
  }
}
```

that way we can just do:

```javascript
b.start()
benchLoop(() => h("div", {}))
b.end()
```

Or am I just being picky with *DRY* 😂

ngryman: Sure! Benchs are crappy, if you have ideas of other scenarios?

selfup: Not at the moment :P

Just make sure to test the benches with the callbacks being called in the for loops vs. not abstracting the for loops. Curious to see if that would change the performance..

ngryman: If we want to be super precocinous here is how `jsperf` & cie do it: https://gist.github.com/ngryman/fd9bcd96910409f91af00d9fd4178a10

ngryman: Basically generate the function to it's not optimized and type feedback and all does leak to other tests.

selfup: oh wow ok that is a completely different ball game 😆

src/h.js
```diff
@@ -1,22 +1,29 @@
...
+    if (Array.isArray(values)) {
```
lukeed: This block may be better served as:

```js
var isArr = Array.isArray(values), 
  iter = isArr ? values : arguments,
  len = iter.length,
  i = isArr ? 0 : 2;

for (; i < len; i++) {
  addChild(children, iter[i])
}
```

This is because `for-loops` are always (currently) faster if `i` and `len` are declared early. Looks like this saves bytes too?

src/h.js
```diff
@@ -1,22 +1,29 @@
...
-  var node
-  var stack = []
-  var children = []
+function addChild(children, val) {
+  if (val == null || typeof val === 'boolean') return
 
-  for (var i = arguments.length; i-- > 2; ) {
-    stack[stack.length] = arguments[i]
+  if (Array.isArray(val)) {
+    for (var j = 0; j < val.length; j++) {
```
lukeed: Same as below:

```js
var j=0, len=val.length;
for (; j < len; j++) ...
```

src/h.js
ngryman: Factoring this piece is really tempting I admit 👀
But `iter = isArr ? values : arguments` is a honeypot. `arguments` is very different from a JS array and does not use the same stubs to iterate over. So if `iter` is either `arguments` or an array, the function is likely to never be optimized. Here is a comparison of the 2 versions: http://quick.as/qnGBfdboR

> This is because for-loops are always (currently) faster if i and len are declared early.

AFAIK code-motion takes this in charge: `i` is hoisted out of the loop, and `len` is probably too because the inner code is invariant and does not mutate the array, so the compiler can safely make the assumption `length` will never change.

src/h.js
selfup: Yes caching the length does seem to help quite a bit :smile: 

Found this thanks to reddit :joy: - http://i.imgur.com/XZzVKft.png

ngryman: I'm not against caching and will do for the cause, as said in an another comment JIT probably already does it though code-motion because the code in the loop does not mutates the array. JIT does ALOT of code motions :)

More generally on micro-benchmarks, this video is really a must see: https://www.youtube.com/watch?v=65-RbBwZQdU

ngryman: @selfup @lukeed A deep explanation why this is not necessary: http://mrale.ph/blog/2014/12/24/array-length-caching.html

lukeed: Hey, array caching has since changed. I've tested this extensively, for early `let i=0` too. It's more of a  V8 "bug" right now, but has been present for about a year & does not look to be changing soon.

src/h.js
lukeed: Didn't know the bit about arguments assignment. Will have to look more into it. Thanks!

src/h.js
ngryman: @lukeed Oh I'm not aware of this. Do you have a source that highlights this v8 issue? Because in my tests I really don't see any performance impact of not caching it.

lukeed: Nothing official, hence the quotation marks. Have mentioned it briefly to a few perf addicts on twitter & they're all aware of it too.

Here's a [simple gist](https://gist.github.com/lukeed/2629834a2d91a85885678ab561b3a202) of noticeable impact. Even with only 1000-item (empty) array, there's a 200ms difference.

ngryman: Hum, again be aware of micro benchmarks. If you replace `res = []` by `res = new Array(LEN)` (pre-allocating a dense array), you get from `~5000ms` to `~1700ms`, and the difference between a cache length becomes unoticable:

```
cached loop---
elapsed:  1646
op/s  607533414.3377886
non-cached loop---
elapsed:  1653
op/s  604960677.5559589
```

lukeed: Yes, that's an optimization for the _iterated_ function & that's why I intentionally left it out. If you replace the array-push with *any* other function, the cached loop is (still) always faster:

```js
  var i=0, len=data.length;
  for (; i < len; i++) {
    (6 + 1);
  }
```

```
cached loop---
elapsed:  440
op/s  2272727272.7272725
non-cached loop---
elapsed:  457
op/s  2188183807.439825
```

Again, these are still microbenchmarks, and mean nothing on their own. You should run the benchmark suite after applying the change & look for changes. This change has benefitted 99.999% of my real-world apps so I'm fairly confident in it. 😜

ngryman: Well I'm really sorry to insist but that's probably wrong and oversimplified (please read this excellent article by M. Egorov, V8 and Java VM guy).

```js
 var i=0, len=data.length;
  for (; i < len; i++) {
    (6 + 1);
  }
```

This loop does nothing. It's well explained in the video I linked above. `(6+1)` is removed by [loop invariant code motion](https://www.wikiwand.com/en/Loop-invariant_code_motion). The loop being empty, it's removed too thanks to [dead code elimination](https://www.wikiwand.com/en/Dead_code_elimination), ...and `i` and `len` assignments too.
So the above piece of code simply disappears after compilation and you end up measuring the outer benchmark loop, and only that 😉 

From M. Egorov:
> Don’t blindly trust microbenchmarks, they are often not measuring what you think they are measuring.

Like I mentioned in my previous comment, benchmarks are not deterministic in time and results may vary because of external factors (OS process scheduling, GC pauses, and many more). `457 / 440` is basically a 1% variation which is a way too small threshold to make any serious conclusion.
I even bet you sometimes get swapped digits like in the example I linked above, where non-cached was faster. And that's for one simple reason: they are doing the same thing.

JIT compilers are really complex, and depending on certain circumstances they optimize things or not. In the cases you actually got positive results, and if really related to `length` property, chances are that the length couldn't be cached by the compiler because it couldn't tell if the array would be mutated or not, so it keeps the access property subroutine.
In the bench you provided me (and in `h`), it's obvious to the compiler that the array will never be mutated, so it safely caches it for us (cf. the article I posted above). The results speaks for themselves: there is not any measurable difference in the benchmarks (and I do 1M loops, not 1K).

But then there is more, some excerpts from M. Egorov's article:

> Thing to realize here is simple: if compiler is unable to figure out that arr.length does not change within the loop, you can’t help it much by manually caching it. When it needs to access array element, compiler anyway has to emit a bounds check which requires loading length.

> Ok. We can’t reason about caching arr.length as an optimization that improves the code by reducing a number of arr.length accesses. [...] Well, in fact we do make some thing somewhat worse: the life of a poor register allocator. When we manually cache length we give it one more live value to care about and keep around in the loop.

> Understand before acting [...] Yes, surprisingly it has nothing to do with caching or not caching array.length access. Don’t derive your code style from microbenchmarks. Don’t base it on advices from esteemed elders (or mine for that matter). Instead invest into understanding where, how and why your application is spending CPU cycles.

Conclusion: do not optimize something until you analyzed compilation artifacts with a tool like IRHydra2 and highlighted sections of [SSA](https://www.wikiwand.com/fr/Static_single_assignment_form) that proves the compiler didn't do a good job.

Every piece of `h` was benched and analyzed with IRHydra2 to track deopts and optimal SSA. I'm not saying that the current version is perfect, it's not. But I did invest some time measuring stuff precisely, and `length` did not show up

Until I have some solid benchmark and traces that actually prove there is a bug in the current `v8` implementation under `h` circumstances (and I would be highly interested in it btw), I'll stick to this conclusion.

lukeed: Thanks!! I don't mind being wrong by any means 😸 

I'm going to look over all this great info & do what I can with it. Appreciate any & all new info 🙌

jamen: @ngryman That's some great info!!

I don't really have anything constructive to add here, because everyone gave a lot of objective points.  But, I'll just say I've noticed this redundancy in my own benchmarks.  Put simply:  I don't think caching the length is necessary because this type of loop is not uncommon, and it would make sense to me for V8 to optimize this sort of thing instead of forcing everyone to cache their lengths to get a perf gain.

Does this have other effects on other engines?  Curious how this performs cross-browser.

ngryman: @jamen The only answer would be: test it 🤓 I can only be sure for what I measure: `v8`.

But I will make the assumption (and I don't really like it) that as it's a cross language pattern, known way before JS became something serious, and as compilers exist for a long time and address those kind of issues since I had my first teeth, I'm pretty confident that any first-grade compiler will do it. But again, only tests can speak.

Also as pointed out in the article, accessing an array triggers a bound check, that itself requires a `length` lookup. I'm pretty the algorithm is detailed in the specs. If it's in the specs, every engine will do it. And if every engine lookup for `length` when accessing an array, end of story.

ngryman:
Are we ok with this implementation?
If so, I can add tests to get 100% coverage and merge it.

As discussed with @jbucaran node constructors and recycling will be introduced in another PR.

jbucaran:
@ngryman Thanks, but did you run these changes through the js-framework-benchmarks? That's the indicator I've used before making "big" changes that could affect performance. I know you are trying to improve `h` here, so I am optimistic.

Let me know the results if and when you have them! 👍

jbucaran:
@ngryman Did ever have a chance to run these through the benchmarks? 🤔 I'd really like to make a decision here! 👋😄

ngryman:
@jbucaran Hey I gave you the result on Slack 15 days ago 😋
They were approx 7% better, which is not a game changer, but I didn't expect more from `h`.

screen shot 2017-07-13 at 20 19 41

src/h.js
```diff
@@ -1,22 +1,19 @@
...
+export function createNode(tag, data, values) {
+  var children = []
 
-  while (stack.length) {
-    if (Array.isArray((node = stack.pop()))) {
-      for (var i = node.length; i--; ) {
-        stack[stack.length] = node[i]
+  for (var i = 0; i < values.length; i++) {
+    var val = values[i]
+    if (val && val !== true && val !== false) {
```
leeoniya: @ngryman `val !== false` is redundant here

ngryman:
@jbucaran We could also gain bytes and perf if this babel issue is resolved: https://github.com/babel/babel/issues/2034#issuecomment-294189475

The guy seemed open to a PR for this. I might get some free time next week, I think I'm gonna push this, it would also make lots of library authors happier.

leeoniya:
> They were approx 7% better, which is not a game changer,

7% is actually quite a lot

src/h.js
ngryman: @leeoniya Good catch! I don't know why I've done that 🙉

SkaterDad:
@ngryman  That JSX transform fix would be great.  When I first looked at 'h()', that arguments stuff struck me as really strange!

src/h.js
```diff
@@ -1,22 +1,19 @@
-export function h(tag, data) {
-  var node
-  var stack = []
-  var children = []
+var isArray = Array.isArray
```
jbucaran: @ngryman Why? It means nothing to uglify. Could this affect your benchmark results?

ngryman: I don't think so. I can remove it if you want.

jbucaran:
@ngryman Those results looked great, so I setup js-frameworks-benchmarks again from scratch and here are my results using the h.js currently on master VS your h.js (I made them both use the same app.js to avoid false positives).

screen shot 2017-07-27 at 21 44 13

But like @SkaterDad is saying, that JSX transform fix would be really great.

Not sure what else to do.

jbucaran:
I ran it --count 5, which took a while and the results were better:

screen shot 2017-07-27 at 21 56 22

ngryman:
@jbucaran I'm trying to understand, you are comparing `0.11.0` with `0.9.1`. Where is the optimized `h`? Those two versions have lots of other additions.

jbucaran:
@ngryman Ignore the labels:

- 0.11.0 === master app.js + nygrman's h.js
- 0.9.1 === master app.js + master h.js

ngryman:
@jbucaran Surprised, I'll take a look. The fact that the impact is not visible enough in this benchmark wouldn't surprise me, but your first batch of result is very... interesting :D

ngryman:
I've run several benchmarks, several times and got the same results for both implementations. I still can't explain why `0.11` is the same and `0.9` had a real gap between the two.

I have to admit that starting optimization passes with `h` was not the best idea. I've done it because it was quick and fun. But it does not really bring a lot to the table right now... I kind of expected that, but I was hoping for a slight perf boost like I measured in `0.9`.

I should have started elsewhere: normalizing `vnode`s.

The proposed implementation is faster because potentially optimizable by the JIT. Currently in a real world use case, we don't leverage that optimization because our `vnode`s are of several shapes, ... and we have bigger problems.
When we will have a "typed" `vtree`, we'll probably have a little performance boost from this PR, mainly in animations where `h` could potentially become hot. But again, this will be tiny.

I would propose to focus on normalizing our `vnode`s and recycle them (because we are generating tons of garbage right now) and revisit this PR later when those optimizations are done.

Issue #156, opened 2017-03-11
Is onUpdate correct?


jbucaran:
We're calling onUpdate _not_ inside the `if (shouldUpdate(node, oldNode)) {` clause, but the one afterwards. Essentially, onUpdate will fire for every element when its props/attrs are updated, which is basically on every cycle. I am not sure if the current behavior makes sense.

farism:
I think onUpdate should only fire inside shouldUpdate check, but right now shouldUpdate is not advanced enough (no deep checking on attr/props) to reliably fire onUpdate. Which makes me think that some type of immutable library with cheap referential checking would be helpful.

The confusing part may be how it's called `shouldUpdate` when it's more along the lines of something like `shouldRecreate`? Or perhaps I'm reading it incorrectly.

jbucaran:
The name choice may be bad and could change :)

dodekeract:
`onUpdate` is actually incorrect on a different level.

If you do a partial model change it will **not** make `nextModel` the full model. `nextModel` will be the partial update.

jbucaran:
@dodekeract Yeah, onUpdate works because it will get called when the view is rendered, but it's not correct. It should be called in shouldUpdate.

lukejacksonn:
Related: https://hyperappjs.slack.com/archives/C41ECC0V6/p1499698015602500

> The onupdate event is fired every time the element's data is updated.

Currently onupdate gets called every time the _state_ changes not every time the _element's data_ changes. It should probably behave as described in the docs, that would make sense.

@ngryman pointed toward [app.js#L178](https://github.com/hyperapp/hyperapp/blob/master/src/app.js#L178)

I have had a go at refactoring `updateElementData ` to only call `onupdate` when some attribute changes. Here is the function https://github.com/lukejacksonn/hyperapp/blob/onupdate-fix/src/app.js#L141

Here is a CodePen test with the old code replaced with the proposed  https://codepen.io/lukejacksonn/pen/JJeLre

Not sure how useful any of this is but it reopening until we fix the docs at least.

ngryman:
Also, does `onupdate` bubbles up to parents?

jbucaran:
https://github.com/hyperapp/hyperapp/commit/010172ca36596deef445ec2292478cd99c0c1a5b

PR #290, opened 2017-07-11
Only call onupdate if element data changed

lukejacksonn:
See https://github.com/hyperapp/hyperapp/issues/156#issuecomment-314251724

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -169,20 +169,20 @@ export function app(app) {
     }
   }
 
-  function updateElementData(element, oldData, data) {
-    var updated = false
+  function updateElementData(element, oldData, data, cb) {
     for (var name in merge(oldData, data)) {
       var value = data[name]
       var oldValue =
         name === "value" || name === "checked" ? element[name] : oldData[name]
 
-      if (value !== oldValue) {
+      if (value !== oldValue && name !== "onupdate") {
         setElementData(element, name, value, oldValue)
-        updated = true
+        cb = data.onupdate
       }
     }
-    if (updated && data["onupdate"]) {
-      data["onupdate"](element)
+
+    if (cb != null) {
+      cb(element)
     }
   }
 
--- a/src/app.js
+++ b/src/app.js
@@ -175,7 +175,7 @@ export function app(app) {
       var oldValue =
         name === "value" || name === "checked" ? element[name] : oldData[name]
 
-      if (value !== oldValue && name !== "onupdate") {
+      if (value !== oldValue) {
         setElementData(element, name, value, oldValue)
         cb = data.onupdate
       }
--- a/src/app.js
+++ b/src/app.js
@@ -170,16 +170,19 @@ export function app(app) {
   }
 
   function updateElementData(element, oldData, data) {
+    var updated = false
     for (var name in merge(oldData, data)) {
       var value = data[name]
       var oldValue =
         name === "value" || name === "checked" ? element[name] : oldData[name]
 
-      if (name === "onupdate" && value) {
-        value(element)
-      } else if (value !== oldValue) {
+      if (value !== oldValue) {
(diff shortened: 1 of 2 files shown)


jbucaran:
@lukejacksonn I think we can get away with a few bytes less using:

```js
  function updateElementData(element, oldData, data, cb) {
      var value = data[name]
      var oldValue =
        name === "value" || name === "checked" ? element[name] : oldData[name]

      if (value !== oldValue && name !== "onupdate") {
        setElementData(element, name, value, oldValue)
        cb = data.onupdate
      }
    }

    if (cb != null) {
      cb(element)
    }
  }
```

src/app.js
```diff
@@ -139,17 +139,20 @@ export function app(app) {
...
 
-      if (name === "onupdate" && value) {
-        value(element)
-      } else if (value !== oldValue) {
+      if (value !== oldValue) {
         setElementData(element, name, value, oldValue)
+        updated = true
       }
     }
+    if (updated && data["onupdate"]) {
```
SkaterDad: I really like this change (moving the `onupdate` call after the loop).  Should make its behavior more predictable.  :+1:

leeoniya: `&& data.onupdate` is shorter

lukejacksonn:
So I was going to ask.. what is the second part of this AND for..
```
if (value !== oldValue && name !== "onupdate")
```
Why do we not set "onupdate" if it has changed? Does this not also mean it will never be set :thinking:

If you remove `&& name !== "onupdate"` then all the tests pass.

test/lifecycle.test.js
```diff
@@ -38,15 +38,20 @@ test("oninsert", done => {
...
-      }),
+      h("div", { onupdate: mock },
+        h("div", {
+          class: state,
+          onupdate: element => {
+            expect(mock).not.toHaveBeenCalled()
+            expect(state).toBe(2)
+            done()
+          }
+        })
```
jbucaran: @lukejacksonn Are you sure about this test? If you call the action inside loaded, it will be called after the next cycle because of how we are faking rAF with setTimeout.

lukejacksonn: No I am not sure about it 😅 it is basically the same as the previous test and passes/fails under the right circumstances.

How would you test for this?

andyrj: this works for rAF stub as well...

```js
window.requestAnimationFrame = function(cb) {cb(Date.now())}
```

zaceno:
I just thought of something... What if your nodes are keyed (like in a list) and you change the *order* of the nodes. The data doesnt change, but the order did.

If we don't call onupdate for reordered lists, then for one thing, animated transitions (https://github.com/zaceno/hyperapp-transitions#move) will not be possible.

Will we at least call onupdate on the parent?

jbucaran:
@zaceno We are not calling onupdate for keys, but I can change that.

jbucaran:
> @zaceno My concern is correct: reordering keyed nodes will not, in itself ensure that onupdate gets called on them. (Side-note: reordering is not a thing unless you're nodes are keyed)

If the data of the nodes doesn't change, then how to call onupdate?

zaceno:
@jbucaran maybe onupdate is not the appropriate thing here. Maybe we should have a new lifecycle method `onmove` or `onreorder`. 

What I mean is that I'd like to run code on the element when it's position among siblings changes (which of course only makes sense for keyed nodes).

jbucaran:
@zaceno Ah, sure, that is an option. 👍

zaceno:
@jbucaran Cool, I'll look into a PR then

jbucaran:
@zaceno Maybe something around [here](https://github.com/hyperapp/hyperapp/blob/ef038cc4f3966c3434afc98a8e80b89b494c3302/src/app.js#L265)? 😄

PR #292, opened 2017-07-11
Don't patch DOM nodes if the element's text equals new node.

jbucaran:
When working with text nodes and content-editable, it is possible we end up patching a text node where element.nodeValue is the same as the new node text. In this situation, the state is already sync with the DOM.

If we ignore this and replace the old DOM, the content-editable element will run into focus / selection reset issues.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -199,7 +199,7 @@ export function app(app) {
   function patch(parent, element, oldNode, node) {
     if (oldNode == null) {
       element = parent.insertBefore(createElement(node), element)
-    } else if (node.tag && node.tag === oldNode.tag) {
+    } else if (node.tag != null && node.tag === oldNode.tag) {
       updateElementData(element, oldNode.data, node.data)
 
       var len = node.children.length
@@ -276,7 +276,11 @@ export function app(app) {
           removeElement(element, reusableChild[0], reusableNode)
         }
       }
-    } else if (node !== oldNode) {
+    } else if (
+      element != null &&
+      node !== oldNode &&
+      node !== element.nodeValue
+    ) {
       var i = element
       parent.replaceChild((element = createElement(node)), i)
     }
@@ -284,3 +288,4 @@ export function app(app) {
     return element
   }
 }
+


src/app.js
```diff
@@ -276,11 +276,16 @@ export function app(app) {
...
         }
       }
-    } else if (node !== oldNode) {
+    } else if (
+      element != null &&
+      node !== oldNode &&
+      node !== element.nodeValue
+    ) {
       var i = element
       parent.replaceChild((element = createElement(node)), i)
```
jbucaran: @ngryman Does that also work when `element` is _not_ a text node element? 🤔

ngryman: It's just to give the idea, but I guess it could become something like:
```js
```

jbucaran: Ah, that could work. What would be the advantage of this way though? Just curious.

ngryman: It would avoid to create new text nodes each time, good for memory and garbage collection.

jbucaran: @ngryman It would avoid creating a new text node when both node and oldNode are textnodes and only the text content has changed. Why `nodeValue` over `textContent`?

leeoniya: > Why nodeValue over textContent

speed.

jbucaran: @leeoniya Not what it says [here](https://stackoverflow.com/questions/21311299/nodevalue-vs-innerhtml-and-textcontent-how-to-choose), though. It says that nodeValue is faster than innerHTML, but textContent is the "faster".

leeoniya: beware of all perf-related conclusions reached > 3 months ago ;)

https://jsperf.com/innerhtml-vs-innertext/7.atom.atom.atom

the difference is *massive* in Edge and Firefox. Less so in Chrome, but still..

jbucaran: @ngryman I'd prefer `typeof node === "string" && typeof oldNode === "string"` and then do as you suggest. Can you make a PR? 😄

Swizz: I am coming from nowhere, but, isnt number treated like string in this case ?

jbucaran: `h` makes sure that nodes are either Vnodes or strings, so all numbers are already converted to strings at this point. 😄

Swizz: How, I am looking around patch to find the conversion ^^'
I am partially discovering hyperapp internal with questions. So, sometimes I miss things when reading at code that raise questions later.

jbucaran: Sure! I'm happy to answer all your questions and please do find all its secrets and help us make it better! 😅 

The number to string conversion is right [here](https://github.com/hyperapp/hyperapp/blob/master/src/h.js#L16).

PR #297, opened 2017-07-13
basic implementation of node recycling

andyrj:
Anyone see a point in patch that I should be grabbing nodes for recycling that I missed?  This basic implementation is based on the gist @ngryman provided in slack the other day. https://gist.github.com/ngryman/69717b9b6eb57da2a50a9faad02fb004  

I was curious how this would work so I put it together, seems to be working well for our tests but I'm not certain we don't have cases where recycling causes an issue that isn't in current tests...

This is mainly just for discussion, not expecting this to be merged as is without at least adding some recycling test cases.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -247,6 +247,9 @@ export function app(app) {
           if (null == oldKey) {
             patch(element, oldElement, oldChild, newChild)
             j++
+
+            // causes issues with test: insert children on top
+            //recoverNode(oldChild)
           }
           i++
         } else {
@@ -259,6 +262,15 @@ export function app(app) {
           } else {
             patch(element, oldElement, null, newChild)
           }
+
+          /* causes problems with the following tests:
+            reorder keyed
+            mixed keyed/non-keyed
+            style
+            update element data
+            svg
+          */
+          //recoverNode(reusableChild[1])
 
           j++
           newKeys[newKey] = newChild
@@ -281,6 +293,7 @@ export function app(app) {
           removeElement(element, reusableChild[0], reusableNode)
         }
       }
+      recoverNode(oldNode)
     } else if (node !== oldNode) {
       var i = element
       parent.replaceChild((element = createElement(node)), i)
--- a/src/app.js
+++ b/src/app.js
@@ -1,3 +1,5 @@
+import { getNode, recoverNode } from "./h"
+
 export function app(app) {
   var state = {}
   var actions = {}
@@ -41,13 +43,13 @@ export function app(app) {
   function hydrate(element, map) {
     return element == null
       ? element
-      : {
-          tag: element.tagName,
-          data: {},
-          children: map.call(element.childNodes, function(element) {
+      : getNode(
+          element.tagName,
+          {},
+          map.call(element.childNodes, function(element) {
             hydrate(element, map)
           })
-        }
+        )
   }
 
   function render() {
@@ -265,6 +267,7 @@ export function app(app) {
         var oldKey = getKey(oldChild)
         if (null == oldKey) {
           removeElement(element, oldElements[i], oldChild)
+          recoverNode(oldChild)
         }
         i++
       }
@@ -274,11 +277,13 @@ export function app(app) {
         var reusableNode = reusableChild[1]
         if (!newKeys[reusableNode.data.key]) {
           removeElement(element, reusableChild[0], reusableNode)
+          recoverNode(reusableNode)
         }
       }
     } else if (node !== oldNode) {
       var i = element
       parent.replaceChild((element = createElement(node)), i)
+      recoverNode(oldNode)
     }
 
     return element
--- a/src/app.js
+++ b/src/app.js
@@ -196,6 +196,7 @@ export function app(app) {
     function removeChild() {
       parent.removeChild(element)
     }
+    recoverNode(node)
   }
 
   function patch(parent, element, oldNode, node) {
@@ -267,7 +268,6 @@ export function app(app) {
         var oldKey = getKey(oldChild)
         if (null == oldKey) {
           removeElement(element, oldElements[i], oldChild)
-          recoverNode(oldChild)
         }
         i++
       }
@@ -277,7 +277,6 @@ export function app(app) {
         var reusableNode = reusableChild[1]
         if (!newKeys[reusableNode.data.key]) {
           removeElement(element, reusableChild[0], reusableNode)
-          recoverNode(reusableNode)
         }
       }
     } else if (node !== oldNode) {
--- a/src/app.js
+++ b/src/app.js
@@ -202,7 +202,9 @@ export function app(app) {
   function patch(parent, element, oldNode, node) {
     if (oldNode == null) {
       element = parent.insertBefore(createElement(node), element)
-    } else if (node.tag && node.tag === oldNode.tag) {
+    } else if (typeof node === "string" && typeof oldNode === "string") {
+      element.nodeValue = node
+    }else if (node.tag && node.tag === oldNode.tag) {
       updateElementData(element, oldNode.data, node.data)
 
       var len = node.children.length
--- a/src/h.js
+++ b/src/h.js
@@ -1,4 +1,5 @@
 var nodePool = []
+var nodePoolHead = -1
 
 function reuseNode(node, tag, data, children) {
   node.tag = tag
@@ -8,14 +9,14 @@ function reuseNode(node, tag, data, children) {
 }
 
 export function getNode(tag, data, children) {
-  return nodePool.length > 0 
-    ? reuseNode(nodePool.pop(), tag, data, children)
+  return nodePoolHead > -1
+    ? reuseNode(nodePool[nodePoolHead--], tag, data, children)
     : { tag: tag, data: data, children: children }
 }
 
 export function recoverNode(node) {
   if (typeof node !== "string") {
-    nodePool.push(node)
+    nodePool[++nodePoolHead] = node
   }
 }
 
--- a/src/h.js
+++ b/src/h.js
@@ -8,11 +8,9 @@ function reuseNode(node, tag, data, children) {
 }
 
 export function getNode(tag, data, children) {
-  if (nodePool.length > 0) {
-    return reuseNode(nodePool.pop(), tag, data, children)
-  } else {
-    return { tag: tag, data: data, children: children }
-  }
+  return nodePool.length > 0 
+    ? reuseNode(nodePool.pop(), tag, data, children)
+    : { tag: tag, data: data, children: children }
 }
 
 export function recoverNode(node) {
--- a/src/h.js
+++ b/src/h.js
@@ -1,3 +1,26 @@
+var nodePool = []
+
+function reuseNode(node, tag, data, children) {
+  node.tag = tag
+  node.data = data
+  node.children = children
+  return node
+}
+
+export function getNode(tag, data, children) {
+  if (nodePool.length > 0) {
+    return reuseNode(nodePool.pop(), tag, data, children)
+  } else {
+    return { tag: tag, data: data, children: children }
+  }
+}
+
+export function recoverNode(node) {
+  if (typeof node !== "string") {
+    nodePool.push(node)
+  }
+}
+
 export function h(tag, data) {
   var node
   var stack = []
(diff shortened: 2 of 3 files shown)


src/app.js
```diff
@@ -274,11 +277,13 @@ export function app(app) {
         var reusableNode = reusableChild[1]
         if (!newKeys[reusableNode.data.key]) {
           removeElement(element, reusableChild[0], reusableNode)
+          recoverNode(reusableNode)
```
andyrj: wasn't sure if this is safe to recoverNode here, since this is where keyed nodes are re-used...

src/h.js
```diff
@@ -1,3 +1,24 @@
...
+
+export function getNode(tag, data, children) {
+  return nodePool.length > 0 
+    ? reuseNode(nodePool.pop(), tag, data, children)
+    : { tag: tag, data: data, children: children }
+}
+
+export function recoverNode(node) {
+  if (typeof node !== "string") {
+    nodePool.push(node)
```
ngryman: For some extra perfs (and bytes?) I would use a "only growing stack" like in the initial gist.
Meaning that I would avoid the couple `push` / `pop` and only use `nodePool[poolHead]`.

Doing so reduce array internal re-allocations to zero.

ngryman: Oups not present in the gist, so here is the update:
```js
var nodePool = []
var nodePoolHead = -1

export function getNode(tag, data, children) {
  return nodePoolHead > -1
    ? reuseNode(nodePool[nodePoolHead--], tag, data, children)
    : { tag, data, children }
}

export function recoverNode(node) {
  if (typeof node !== "string") {
    nodePool[++nodePoolHead] = node
  }
}
```

src/app.js
ngryman: Moving `recoverNode` in `removeElement` would probably factorize things?

src/h.js
andyrj: If you leave nodes on an ever growing array, won't that essentially be a memory leak?

ngryman: Let’s do some quick analysis. Let’s define a `pass` being a `patch` call. At each call we have:
 - `n` nodes that will be created
 - `m` old nodes 
 - `t` dereferenced nodes that will be collected by the GC later
Let’s also say that a GC pass of `t` is linear (`O(t)`).

Here is a comparison of the current implementation, the PR one and the one using a “only-grow stack”:

|              | Avg space complexity | GC time complexity
|--------|-----------------------|--------------------
 Current | `O(n+m+t)`                 | `O(t)`
 PR         | `O(n+max(m, t))`        | `O(0)`
 OGS      | `O(n+max(m, max(t, t-1, … t0)))`        | `O(0)`

jbucaran:
@andyrj Can you please rebase with master?

src/h.js
```diff
@@ -1,3 +1,25 @@
+var nodePool = []
+var nodePoolHead = -1
```
jbucaran: @andyrj @ngryman Is it alright to share the nodepool with all apps that use `h`? 🤔

jbucaran:
@andyrj Awesome! Please clone https://github.com/krausest/js-framework-benchmark and set it up so you can benchmark your changes locally.

src/h.js
andyrj: should be fine, these are nodes that should no longer be used as we had been just throwing them away to GC

andyrj:
@jbucaran I think it's rebased now sorry about that

andyrj:
@jbucaran I would love to get numbers from that benchmark repo, but I can't get the master branch to build :(  failed in the webdriver-ts...  Had to prune a lot of broken build directories to even get to that point as well, is there a less fubar benchmark repo or should I use version-1 branch?

jbucaran:
@andyrj No need to install Java. Just get into the webdriver_ts/ directory and `npm run i && npm run compile`.

andyrj:
well I got past the first issue by installing maven, and downgrading node to v7.x, but I still get failure to build and thus can't collect my own data from local runs yet, a react typescript run is crapping out on me...

Is there a way for me to just run ours?

jbucaran:
No need to install Java. Just install the stuff inside webdriver_ts and to run only hyperapp's benchmarks do this:

```
yarn run selenium -- --framework hyperapp ; and yarn run results
```

Any else I could be missing @ngryman?

leeoniya:
here's the 1-2-3:

1. clone it.
2. in the main dir run `npm install http-server`
3. in `/webdriver-ts` run `npm install && npm run build-prod`
4. `/hyperapp-v0.9.1` run `npm install && npm run build-prod`
5. repeat step 4 for all other frameworks you wanna bench (like vanillajs)
6. in main dir run `npm start`
7. in `/webdriver-ts` run `npm run selenium -- --framework hyperapp vanillajs && npm run results`

andyrj:
I'm deleting and starting over from scratch, somehow it keeps trying to compile results from tests I have not run and fails in npm run results...  
@leeoniya thanks, going to run only those and hope I get results to work this time...

jbucaran:
@andyrj You could also modify the test runner code directly and hardcode 'hyperapp' so it runs only that test. I think @ngryman had to resort to that, but luckily I didn't have that problem. 😓

andyrj:
I surrender to the crap of that repo requiring so much bloat...  I am stuck either starting over again and wasting more time or figuring out now why their TS is throwing errors after having dug through the other BS...

ERROR in ./src/App.tsx
    (34,33): error TS2345: Argument of type '({ "framework": string; "benchmark": string; "type": string; "min": number; "max": number; "mean"...' is not assignable to parameter of type 'Result[]'.
      Type '{ "framework": string; "benchmark": string; "type": string; "min": number; "max": number; "mean":...' is not assignable to type 'Result'.
        Type 'undefined' is not assignable to type 'Result'.
    Child html-webpack-plugin for "table.html":
           [0] ../node_modules/html-webpack-plugin/lib/loader.js!./index.html 675 bytes {0} [built]
            + 4 hidden modules

I give up someone else will have to perf test this PR...  I can't get this crap to npm run results, even if I install the 99999199999191991919191919 npm packages it requires for all the broken projects left in that repo

leeoniya:
ping @krausest

src/app.js
```diff
@@ -197,11 +199,14 @@ export function app(app) {
     function removeChild() {
       parent.removeChild(element)
     }
+    recoverNode(node)
   }
 
   function patch(parent, element, oldNode, node) {
     if (oldNode == null) {
       element = parent.insertBefore(createElement(node), element)
+    } else if (typeof node === "string" && typeof oldNode === "string") {
```
jbucaran: @andyrj Does this have anything to do with recycling? 🤔

src/app.js
```diff
@@ -197,11 +199,14 @@ export function app(app) {
     function removeChild() {
       parent.removeChild(element)
     }
+    recoverNode(node)
```
jbucaran: @andyrj @zaceno What happens when you use `onremove`, and therefore must remove the element yourself using:

```js
element.parent.removeChild(element)
```

I guess we can assume (can we?) the user is going to remove the element, but is it okay to recycle the node this way?

ngryman:
There is a problem with the current implementation I didn't spotted yesterday.

Currently we are recycling only `vnode`s with the associated DOM node being removed. So it's only recycling `vnode`s within the same `patch` pass. The gain is pretty minimal and won't prevent the initial problem: GC pauses.

The aim of recycling as I meant it is to collect all `vnode`s from a previous `patch` pass instead of dereferencing them and letting the GC collect them. In other words, collect the whole old `vtree` before assigning to it a new value.

krausest:
@andyrj  Sad to hear. I'm afraid I haven't found an easy way to build 60 frameworks plus automated benchmark runner.
I haven't seen the error you reported before, please open an issue for it (and attach webdriver-ts-results/src/results.ts). It's failing when creating the interative results table, the static table should have been already created (http://localhost:8080/webdriver-ts/table.html). You just might remove the interactive table by editing webdriver-ts/package.json  to invoke just "results": "node dist/createResultTable.js" as a workaround.

Feel free to send me an incomplete/non-working PR and I'll merge it for you and report the results (just package.json or source code for hyperapp) or even just put them in an issue.

src/app.js
andyrj: Well, not recycling exactly but it's related.  Since we don't ever update textNodes currently GC is busy with collecting them, since recycling is about reducing GC I thought why not try this also.

src/app.js
andyrj: This is why I started the PR to ask if this was the correct place to hook into recycle.  I really don't know if it's wrong or right, but all of our tests still pass.

andyrj:
@ngryman I tried recursively recycling inside of the method that calls patch, on the oldNode, and things blew up, I will try again tonight when I have time

@krausest I apologize for my out burst above,  I appreciate that getting that many frameworks to benchmark is a difficult problem.  I will open an issue with steps to reproduce if that will be helpful.

src/app.js
andyrj: my initial attempt looked more like this but this :boom: 
```js
function render() {
    recoverNode(oldNode)
    oldNode = node
    node = emit("render", view)(state, actions)
    element = patch(
      root,
      element,
      oldNode,
      node
    )

    locked = !locked

    if (!loaded) {
      emit("loaded", (loaded = true))
    }
  }
```

andyrj: ^^ blows up on all the keyed tests, and some of the async tests...

I think the reusableNodes list will get recycled before they get a chance to be reused if I recycle up in render()...  which is what caused me to move the recoverNode down into the patch so I could work around the keyed interference...

krausest:
@andyrj @zaceno I tried to improve the README for js-framework-benchmark to create a step-by-step tutorial how to start building a single framework, add more frameworks and contribute a new framework without having to build everything. Doing so I think I found the typescript error you mentioned above. Hope that makes contributing much easier - if not please submit an issue I'll try to fix it.

*And I'd love to see a keyed implementation for hyperapp there.*

src/app.js
```diff
@@ -279,13 +296,15 @@ export function app(app) {
           removeElement(element, reusableChild[0], reusableNode)
         }
       }
+      recoverNode(oldNode)
```
andyrj: @ngryman @jbucaran I think this catches the oldNodes that need to be collected that I had missed by only collecting in the removeElement(),  I think this is easiest way to not step on reusableNodes for the keyed updates while still recycling.  Really my naive initial recursive recycle in render() that I commented on above just causes an extra iteration of all the nodes instead of hooking in during the patch to recycle like I tried in this PR.  But I did miss this recoverNode(oldNode)  I may have missed another, or there maybe a better place to do all of this, if this PR is closed/rejected I won't mind it's serving it's purpose by getting comments/review as far as I'm concerned :slightly_smiling_face:

andyrj:
Here is chrome dev tools performance timeline for current master branch: (notice green line increasing linearly from textNodes)
hyperapp-no-recycle

This is with the recycle and re-use of textNodes in this PR: (notice that the green line goes up on init and then is flat here)
hyperapp-recycle

Still need to figure out if there is elsewhere we could recoverNode() without causing issues with our keyed node implementation...

src/app.js
andyrj: after looking at the chrome devtools memory I think that these calls for recoverNode(oldChild) and recoverNode(reusableNode) are not proper and infact cause the extra js heap bloat in that memory graph I shared... heh

ngryman:
@andyrj Ref: https://github.com/hyperapp/hyperapp/pull/297#issuecomment-315312125.

If we want recycling to have a real benefit we need to recycle the whole `old` vtree.
If we do so, the blue sawtooth effect will "potentially" disappear or at least be less frequent with less amplitude.

The green one represent DOM nodes and should be addressed in a different issue. My guess is that it's mostly related to text nodes.

andyrj:
@ngryman my first attempt was recycling the whole tree from render() which causes much :boom: with both async and keyed nodes.  

If we find the right places to use recoverNode() from within patch it should be equivalent to recursively recycling the root node, only we don't have to iterate over the tree another time per render just for recycling.

I left commented out 2 recoverNode calls inside of patch with comments above them detailing which test they cause to fail.  Those same tests fail when recursively recycling from the root node.

jbucaran:
@andyrj Great work on this feature, but too bad it has kind of stalled. What do you think we should we do about this PR? Shall we close it for now or would you prefer to keep it open when in case you want to reactive it?

andyrj:
It's fine to close for now, I don't have time right now to push this forward :(

PR #304, opened 2017-07-17
changed 'data' to 'attributes' at getting-started.md

dividedbynil:
Avoid ambiguity.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -39,7 +39,7 @@ state => h("h1", {}, state) // <h1>Hi.</h1>
 [Hyperx]: /docs/hyperx.md
 [JSX]: /docs/jsx.md
 
-To compose the user interface, the [h(tag, data, children)](/docs/api.md#h) utility function returns a tree of [virtual nodes](/docs/virtual-nodes.md).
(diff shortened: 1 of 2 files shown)


jbucaran:
Hi @dividedbynil! 

I am not sure what to do here. [We use `data` internally](https://github.com/hyperapp/hyperapp/blob/436ed891f8d3d2df341b82b47f63fe2896cffb31/src/h.js#L26), so if I accept this change, shouldn't we actually change the code as well?

dividedbynil:
Hi @jbucaran, thanks for quick response, you mean the data is not just defining the HTML attributes?

jamen:
We use `node.data` several places inside `src/app.js` as well.  Are we only changing the signature here?  I would be more keen on approving this if it was changed across the entire library.

jbucaran:
To be honest I am not what the confusion is.

dividedbynil:
@jamen Hi, I will work on changing the entire library, by using 'attr', is it better instead of 'data'? @jbucaran

jbucaran:
@dividedbynil Thanks for the effort, but no need to. 👋  `.data` is part of the internal API and end users are unlikely to have to deal with this.

dividedbynil:
@jbucaran Okay, no problem.

PR #311, opened 2017-07-21
Add Typing (DTS file)

Swizz:
This follows and closes #279

The first attemps of a DTS file was made with a lot of stuffs in minds.
I chosen to Revoke completely my last attempt, to propose a more sanitized one.
The main goal of this PR is now to move with iteration. To easily discuss about declarations.

First, I propose you to talk about the basics, and the VNode, `h()` stuffs.

This declaration file, is 98% based on the Picodom one by @pspeter3 because the Vdom system is (almost) the same.

Here I have only overloaded `h()` and added JSX support. https://github.com/picodom/picodom/pull/17#issuecomment-317101607

I have also added the `Emit` interface and `app` function as starting point.

<details>
 <summary>Bonus : Emit use @andrewiggins awesome `keyof Events` : </summary>
 <br/>
 keyof Events
</details>

### Steps
- [x] Step 1 - Virtual DOM + JSX + Emit
- [x] Step 2 - View + Actions
- [x] Step 3 - Events
- [x] Step 4 - Application + Mixin

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -1,0 +1,165 @@
+export as namespace Hyperapp
+
+/** @namespace [Virtual DOM] */
+
+/** The Virtual DOM representation of an Element
+ *
+ * @memberOf [Virtual DOM]
+*/
+export interface VirtualNode<Data> {
+  tag: string
+  data: Data
+  children: VirtualNodeChild<{} | null>[]
+}
+
+/** In the Virtual DOM a Child could be either a Virtual Node or a string
+ *
+ * @memberOf [Virtual DOM]
+*/
+export type VirtualNodeChild<Data> = VirtualNode<Data> | string
+
+export interface Component<Data> {
+  /** A Component is a function that return a custom Virtual Node
+   *
+   * @memberOf [Virtual DOM]
+  */
+  (data: Data, children: VirtualNodeChild<{} | null>[]): VirtualNode<Data>
+}
+
+/** The soft way to create a Virtual Node
+ * @param tag       Either a tag name e.g. 'div'. Or a Component function
+ * @param data      Any valid HTML atributes, events, styles, and meta data
+ * @param children  The children of the VirtualNode
+ *
+ * @memberOf [Virtual DOM]
+*/
+export function h<Data>(
+  tag: Component<Data> | string,
+  data?: Data,
+  children?:
+    | VirtualNodeChild<{} | null>[]
+    | VirtualNodeChild<{} | null>
+    | number
+): VirtualNode<Data>
+
+/** @namespace [Application] */
+
+export interface State {}
+
+export interface Thunk {
+  (update: Function): any | null | void
+}
+
+export type ActionsResult<State extends Hyperapp.State> =
+  | Partial<State>
+  | Thunk
+  | {}
+  | null
+  | void
+
+export type VirtualActions<
+  State extends Hyperapp.State,
+  Actions extends Hyperapp.Actions<State>
+> = {
+  [P in keyof Actions]: (
+    data: any
+  ) => ActionsResult<State> | VirtualActions<State, Actions>
+}
+
+export interface Actions<State extends Hyperapp.State> {
+  [action: string]:
+    | Actions<State>
+    | ((
+        state: State,
+        actions: VirtualActions<State, Actions<State>>,
+        data: any
+      ) => ActionsResult<State>)
+}
+
+export interface VirtualEvents<
+  State extends Hyperapp.State,
+  Actions extends Hyperapp.Actions<State>
+> {
+  [action: string]: (
+    state: State,
+    actions: VirtualActions<State, Actions>,
+    data: any
+  ) => any
+}
+
+export interface Events<
+  State extends Hyperapp.State,
+  Actions extends Hyperapp.Actions<State>
+> {
+  load?: (
+    state: State,
+    actions: VirtualActions<State, Actions>
+  ) => void
+  resolve?: (
+    state: State,
+    actions: VirtualActions<State, Actions>,
+    result: ActionsResult<State>
+  ) => ActionsResult<State>
+}
+
+export interface View<
+  State extends Hyperapp.State,
+  Actions extends Hyperapp.Actions<State>
+> {
+  (state: State, actions: VirtualActions<State, Actions>): VirtualNode<{}>
+}
+
+export interface Mixin<
+  State extends Hyperapp.State,
+  Actions extends Hyperapp.Actions<State>,
+  Events extends Hyperapp.Events<State, Actions>
+> {
+  (): Application<State, Actions, Events>
+  (): (
+    emit: Emit<State, Actions, Events>
+  ) => Application<State, Actions, Events>
+}
+
+export interface Application<
+  State extends Hyperapp.State,
+  Actions extends Hyperapp.Actions<State>,
+  Events extends Hyperapp.Events<State, Actions>
+> {
+  state?: State
+  actions?: Actions
+  events?: Events
+  view?: View<State, Actions>
+  root?: HTMLElement | null
+  mixins?: Mixin<{}, {}, {}>[]
+}
+
+export interface Emit<
+  State extends Hyperapp.State,
+  Actions extends Hyperapp.Actions<State>,
+  Events extends Hyperapp.Events<State, Actions>
+> {
+  /** Call succesively each event handler of the specified event
+   * @param name  The name of the event to call
+   * @param data  Will be reduced by each event handler
+   *
+   * @memberOF [Application]
+  */
+  (name: keyof Events, data?: any): any
+}
+
+export function app<
+  State extends Hyperapp.State,
+  Actions extends Hyperapp.Actions<State>,
+  Events extends Hyperapp.Events<State, Actions>
+>(app: Application<State, Actions, Events>): Emit<State, Actions, Events>
+
+/** @namespace [JSX] */
+
+declare global {
+  namespace JSX {
+    interface Element<Data> extends VirtualNode<Data> {}
+    interface IntrinsicElements {
+      [elemName: string]: any
+    }
+  }
+}
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "dist/hyperapp.js",
   "jsnext:main": "src/index.js",
   "module": "src/index.js",
+  "typings": "hyperapp.d.ts",
   "license": "MIT",
   "repository": "hyperapp/hyperapp",
   "files": [
@@ -21,13 +22,13 @@
     "hyperscript"
   ],
   "scripts": {
-    "test": "jest --coverage --no-cache",
+    "test": "jest --coverage --no-cache && tsc --project test/ts",
     "build": "npm run bundle && npm run minify",
     "bundle": "rollup -i src/index.js -o dist/hyperapp.js -m -f umd -n hyperapp",
     "minify": "uglifyjs dist/hyperapp.js -o dist/hyperapp.js --mangle --compress warnings=false --pure-funcs=Object.defineProperty -p relative --source-map dist/hyperapp.js.map",
     "prepublish": "npm run build",
     "prepare": "npm run build",
-    "format": "prettier --semi false --write 'src/**/*.js'",
+    "format": "prettier --semi false --write 'src/**/*.js' 'hyperapp.d.ts'",
     "release": "npm run build && npm test && git commit -am $npm_package_version && git tag $npm_package_version && git push && git push --tags && npm publish"
   },
   "babel": {
@@ -38,6 +39,7 @@
     "jest": "^20.0.4",
     "prettier": "~1.6.0",
     "rollup": "^0.49.1",
+    "typescript": "^2.5.2",
     "uglify-js": "^2.7.5"
   }
 }
--- a/test/ts/index.tsx
+++ b/test/ts/index.tsx
@@ -1,0 +1,42 @@
+import { h, app } from 'hyperapp'
+
+interface State extends Hyperapp.State {
+  count: number
+}
+
+interface Actions extends Hyperapp.Actions<State> {
+  sub(state: State, actions: VirtualActions, value: number): Partial<State>
+  add(state: State, actions: VirtualActions, value: number): Partial<State>
+}
+
+interface VirtualActions extends Hyperapp.VirtualActions<State, Actions> {
+  sub(value: number): void
+  add(value: number): void
+}
+
+interface Events extends Hyperapp.Events<State, Actions> {
+  log: (state: State, actions: VirtualActions, value: string) => void
+}
+
+const emit = app<State, Actions, Events>({
+  state: {
+    count: 0
+  },
+  view: (state, actions: VirtualActions) => (
+    <main>
+      <h1>{state.count}</h1>
+      <button onclick={actions.sub(1)} disabled={state.count <= 0}>-</button>
+      <button onclick={actions.add}>+</button>
+    </main>
+  ),
+  actions: {
+    sub: (state, actions, value) => ({ count: state.count - value }),
+    add: (state, actions, value) => ({ count: state.count + value })
+  },
+  events: {
+    log: (state, actions, value) => { console.log(value + ' ' + state.count) }
+  },
+  root: document.getElementById("app")
+})
+
+emit("log", "State count : ")
--- a/test/ts/tsconfig.json
+++ b/test/ts/tsconfig.json
@@ -1,0 +1,18 @@
+{
+  "compilerOptions": {
+  "strict": true,
+  "noEmit": true,
+  "jsx": "react",
+  "jsxFactory": "h",
+  "baseUrl": "../..",
+    "paths": {
+      "classwrap": ["src", "hyperapp.d.ts"]
+    }
+  },
+  "include": [
+    "*.ts",
+    "*.tsx",
+    "**/*.ts",
+    "**/*.tsx"
+  ]
+}


hyperapp.d.ts
```diff
@@ -0,0 +1,45 @@
...
+  tag: Component<Data> | string,
+  data?: Data,
+  children?: VElement<{}>[]
+): VNode<Data>
+export function h<Data>(
+  tag: Component<Data> | string,
+  data?: Data,
+  ...children: VElement<{}>[]
+): VNode<Data>
+
```
jbucaran: @Swizz Why is this repeated?

jbucaran:
Great work @Swizz! Can we work on renaming a few things?

I was thinking instead of `VNode`, better `VirtualNode`.

And add types for `State`, `Actions`, `Mixins`, etc.

Note that actions allow nested actions so in flow:
```
{
  [action: string]:
    | Actions
    | (State, Actions, any) => ActionResult
}
```

We'll also need an `ActionResult` type:

> A partial state or Promise that resolves to a partial state.

Maybe also a definition for `CustomEvent`:

```
event(State, Actions, any): any
```

hyperapp.d.ts
jbucaran: @Swizz Is there any way to declare that using the same definition? If it's impossible, then can we leave it out? The fact that children can be `arguments` is an implementation detail needed by JSX, but users should definitely not use `h` like that.

hyperapp.d.ts
```diff
@@ -0,0 +1,45 @@
+export as namespace hyperapp
+
+// VNode part
+
+export interface VNode<Data> {
```
jbucaran: @Swizz Why does `VNode<Data>` mean?

hyperapp.d.ts
```diff
@@ -0,0 +1,45 @@
...
+  tag: string
+  data?: Data
+  children: VElement<{}>[]
+}
+
+export type VElement<Data> = VNode<Data> | string
```
jbucaran: @Swizz Is this possible:

```ts
export interface VNode<Data> {
  tag: string
  data?: Data
  children: VElement<{}>[]
} | string
```

Swizz:
@jbucaran 

> Great work @Swizz!
All credits goes to @pspeter3, seriously

> Can we work on renaming a few things? I was thinking instead of VNode, better VirtualNode.
Naming is open to discussion. I am just a bit curious, why is this a concern here and not in Picodom ?

> And add types for State, Actions, Mixins, etc.
Dont worry about. As the DTS will grow faster, I want to move side by side. Now I want to discuss about VNode, JSX, and the app basis. Once done, I will move forward.
Because there are a lot of thing to discuss. So I think is better to be scope minimal.

jbucaran:
Sounds good.

- VNnode → VirtualNode
- JSX 👍 

>  Why is this a concern here and not in picodom?

I didn't say picodom anywhere.

hyperapp.d.ts
Swizz: I am looking for a way to express children could be an array ov VElements or the rest of args could be too.

In Snabbdom-pragma, I am using the following
```tsx
type Children = VNode<{}>[] | VNode<{}> | string | number
type CircularChildren = Children | Children[]

export function h<Data>(
   tag: Component<Data> | string,
   data?: Data,
   ...children: CircularChildren 
 ): VNode<Data>
```

`CircularChildren` was about a quick fix regarding : 

```jsx
const arr = [
  [<div>Hello1.1</div>, <div>Hello1.2</div>],
  [<div>Hello2.1</div>, <div>Hello2.2</div>]
]
const vnode = <div> { arr } </div>
```

jbucaran: @Swizz Can we just leave out the `...children`? h() supports ...children just because of JSX. If JSX didn't exist, h() would be always Array of VNode.

Swizz: One more thing. 
Can I use `h()` with a number as children ?

plus: do you prefer the `Array<Data>` notation instead of the `Data[]` one ?

hyperapp.d.ts
Swizz: According to @pspeter3 DTS for Picodom, I am pretty sure this is to allow to define the Props interface of a vnode. And to help for JSX :tada:

Props in VNode

_(VS Code is awesome regarding TypeScript)_

jbucaran: Hmm, so this is only useful if the user defines an interface for the VirtualNode data? Hmm. Nice.

hyperapp.d.ts
jbucaran: I didn't know we could use both, but let's go on with `Array<Data>`.

> Can I use h() with a number as children ?

Yes, you can.

hyperapp.d.ts
Swizz: You want a VNode to be either a VNode or a string. I never heard it was possible. Union is only possible on type.

I guess, it is right to use a type.
A VirtualNode is a `VirtualNode` and a `VirtualElement` could be either a `VirtualNode` or a `string`.

jbucaran: I was just trying to remove the VirtualElement type, which is just confusing.

hyperapp.d.ts
Swizz: Yes using : 
```tsx
const link = h('a', {})
```

Will do not raise anything. Thats why we are using `VNode<{}>` as default. Instead of `any`.

hyperapp.d.ts
Swizz: Only in `h()` or we can tell a `VirtualElement` could be either a `VirtualNode`, a `string`, or a `number`.

The last one will affect : 
* `h()` children parameter
* `Component` children
* `VirtualNode` children

hyperapp.d.ts
Swizz: Not if you are thinking like the last sentence.

You cant tell a `VirtualNode` could be a `VirtualNode` or a `string`, isnt it ?
But its easy to reason about a `VirtualNode` **and** a `VirtualElement` that could be either a `VirtualNode` or a `string`.

The `VirtualNode` is the vdom thing. And the `VirtualElement` is in most case a children. And a children could really be another `VirtualNode` or a `string`.

hyperapp.d.ts
Swizz: Hum.. `Array<Data>` could be a little bit confusing.

```tsx
h(
  children: Array<VirtualElement<{}>>
)
```
```tsx
h(
  children: VirtualElement<{}>[]
)
```

jbucaran: @Swizz Why don't you just repeat `VirtualNode | string` everywhere?

Swizz: Do you prefer ?

I think is more correct to tell once a `VirtualNode` is `...` and then the user will be mind this.
And IntelliSens will yield about the right type when aliased. It still asking for `number`,  `string`, or `VirtualNode`. So this is just more DRY.

hyperapp.d.ts
jbucaran: Well, if it's always the children, why not call it `VirtualNodeChild`?

jbucaran: @Swizz ☝️

Swizz: Now it is used as child, but I dont think it is good to limit it at this role by naming.
I do really think `VirtualElement` is good. @pspeter3,did an awesome job there.

Swizz:
@jbucaran A giant plus DTS will be help to yield and suggest about types for VS Code user and atom-typescript user. 
Even if they are using JavaScript instead of TypeScript.

hyperapp.d.ts
```diff
@@ -0,0 +1,40 @@
...
+export type VirtualElement<Data> = VirtualNode<Data> | string
+
+export interface Component<Data> {
+  (data?: Data, ...children: VirtualElement<{}>[]): VirtualNode<Data>
+}
+
+export function h<Data>(
+  tag: Component<Data> | string,
+  data?: Data,
+  children?: VirtualElement<{}>[] | VirtualElement<{}> | number
```
jbucaran: @Swizz Can you make this just:

```VirtualElement<{}>[]```?

Comment (jbucaran) [inline hyperapp.d.ts]: People using `h` should use its best form (string, data, array of virtualnode).

Comment (Swizz) [inline hyperapp.d.ts]: Without it TSC will yield an exception if I want to use `h()` that way
```tsx
h('div', null, h('span', null, 'Hi.'))
```

This simple thing will not be possible to
```tsx
h('div', null, 'Hi.')

// ask for
h('div', null, ['Hi.'])
```

Do you really want to remove features by typing ?

jbucaran: @Swizz You should _not_ use it that way IMO.

@lukejacksonn What do you think?

andrewiggins: As a compromise, one option is to add a JSDoc @deprecated comment to steer people away from this function, saying it only exists for JSX support and to use the array version for better perf.

jbucaran: @andrewiggins Is the function info in vscode intellisense picked up from JSDoc comments?

andrewiggins: Correct, it is. Screenshot below shows an example of what a user would see for the argument spread version of h:

image

This screenshot shows an example of what a user would see for the array version of h:

image

Swizz: I dont know how this is a problem in fact. Performance ? 

You are saying `h('div', null, 'Hi.')` is less performant than `h('div', null, ['Hi.'])` ?
And what if I still want to use this feature like way, I will get typing/editor error/warning even if the function allow the use of it ? 
I think this is very effortless to types it. And let users to choose you prefered way to do as both exist. 

Let's write a performance warning into the doc and refer it in the JSdoc.
```jsx
/** Description...
 * @see {@link http://github.com/hyperapp/hyperapp/... | Performance issue}
 */
h(....
```

Swizz:
Discussed changes pushed, plus JSDoc !

Let's go to the Step 2 !

hyperapp.d.ts
```diff
@@ -0,0 +1,68 @@
+export as namespace hyperapp
+
+/** @namespace [Virtual DOM] */
+
+/** The Virtual DOM representation of an Element
```
andrewiggins: In all the examples I've seen (and on the [JSDoc website itself](http://usejsdoc.org/about-getting-started.html)) it appears the first line of a multi-line doc comment goes on the line after `/**`. It appears [JavaDoc](https://en.wikipedia.org/wiki/Javadoc) (which I assume JSDoc is inspired from) follows the same styling convention.

Not a big deal, but just thought I'd point it out.

hyperapp.d.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @memberOf [Virtual DOM]
+*/
+export interface VirtualNode<Data> {
+  tag: string
+  data: Data
+  children: VirtualNodeChild<{} | null>[]
```
andrewiggins: What's the reasoning for adding `| null` here?

hyperapp.d.ts
```diff
@@ -0,0 +1,68 @@
...
+
+/** A Component is a function that return a custom Virtual Node
+ *
+ * @memberOf [Virtual DOM]
+*/
+export interface Component<Data> {
+  (data?: Data, ...children: VirtualNodeChild<{} | null>[]): VirtualNode<Data>
+}
+
+/** The soft way to create a Virtual Node
```
andrewiggins: What does "soft" mean? I would just state that h creates VirtualNodes

hyperapp.d.ts
```diff
@@ -0,0 +1,68 @@
...
+  children: VirtualNodeChild<{} | null>[]
+}
+
+/** In the Virtual DOM a Child could be either a Virtual Node or a string
+ *
+ * @memberOf [Virtual DOM]
+*/
+export type VirtualNodeChild<Data> = VirtualNode<Data> | string
+
+/** A Component is a function that return a custom Virtual Node
```
andrewiggins: I'd put this comment inside the interface, on top of the method definition, and add param documentation. It provides better intellisense in VSCode

hyperapp.d.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @memberOf [Virtual DOM]
+*/
+export interface Component<Data> {
+  (data?: Data, ...children: VirtualNodeChild<{} | null>[]): VirtualNode<Data>
```
andrewiggins: I don't think the children args are spread out like this. I believe you just get a direct Array.

hyperapp.d.ts
```diff
@@ -0,0 +1,68 @@
...
+  data?: Data,
+  children?:
+    | VirtualNodeChild<{} | null>[]
+    | VirtualNodeChild<{} | null>
+    | number
+): VirtualNode<Data>
+
+/** @namespace [Application] */
+
+export interface Emit<Data, Events> {
```
andrewiggins: Should we swap the order of these generic parameters to match the order of the function parameters?

andrewiggins:
To maintain the quality of these typings over time I recommend we add some typings tests to the tests/ folder similar to what I did for picodom. It also makes it clearer to understand how to use the typings appropriately

hyperapp.d.ts
Swizz: It is my first time with JSDoc to be honest. I have let VSCode guide me with

hyperapp.d.ts
Swizz: In strict mode `null` need to be ensured, this is prerequisite if you are using : `h('tag', null)` node

hyperapp.d.ts
Swizz: You are right, I did it for `Emit`.

hyperapp.d.ts
Swizz: Already fixed I missed to push this

hyperapp.d.ts
Swizz: I was looking for a way to indicate "you can create `VirtualNode` without `h` too, but using `h` is the prefered way".

hyperapp.d.ts
Swizz: Is there a way to make a Generic parameter optional  ? Because we need to give generic a type even if data is optional :confused:

Swizz:
Thank you for your review @andrewiggins. Your help is a giant plus :+1:

To be honest this is 50% laziness. But are you okay to write tests ? 
It will ensure typing is understandable *and* intuitive. And this will help me a lot.

andrewiggins:
Absolutely! I'll copy what I started in picodom and expand it to include hyperapp features

hyperapp.d.ts
andrewiggins: Yup! You can specify a default type, which makes that generic param optional. For example:
```typescript
interface SomeInterface<T, U=any>
```

Swizz: Awesome thank you !

hyperapp.d.ts
pspeter3: Virtual Node Factory?

hyperapp.d.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @memberOf [Virtual DOM]
+*/
+export function h<Data>(
+  tag: Component<Data> | string,
+  data?: Data,
+  children?:
+    | VirtualNodeChild<{} | null>[]
+    | VirtualNodeChild<{} | null>
+    | number
```
pspeter3: Are numbers converted to strings?

lukejacksonn: I think so.. https://github.com/hyperapp/hyperapp/blob/master/src/h.js#L16

Swizz: Yes, a `h` children can be a number, and number will be converted to a string in the `VirtualNode` representation.

hyperapp.d.ts
pspeter3: As a general rule, prefer `{}` to `any`. Using `any` allows the consumer to call any property/method for `U` vs `{}` will allow the consumer to call none.

pspeter3:
After we merge this, doing one for the router would be great.

Swizz:
DTS typing of Hyperapp is a Masterpiece. I think typing the Router will be a lot more easy.

I am a little bit busy these days. I hope I'll push the step 2, within a week.

Swizz:
@pspeter3 Is there a way to ensure all api function to implements ... ?
```tsx
export interface API<State, Actions, DataIn, DataOut> {
    (state: State, actions: Actions, data: DataIn): DataOut
}
```

pspeter3:
@Swizz what API functions are you talking about?

jbucaran:
@Swizz @pspeter3 So #317 was merged! We can now move forward with this. 👍 

- See: https://github.com/hyperapp/hyperapp/blob/master/docs/api.md

mindplay-dk:
How can I install and try this?

jbucaran:
Pong @Swizz @pspeter3.

Swizz:
Quickest way : 

In Hyperapp repo
```sh
git checkout -b Swizz-fresh-typing master
git pull git://github.com/Swizz/hyperapp.git fresh-typing
npm link
```

Or clone my repository
```sh
git clone -b fresh-typing git://github.com/Swizz/hyperapp.git hyperapp
cd hyperapp
npm link
```

In you demo project, instead of installing hyperapp release use your local one
```
npm link hyperapp
```

mindplay-dk:
Thanks @Swizz but I'd rather not use links, as these wouldn't work for anyone else cloning the project. (would they?)

According to npm docs, this should work:

    "dependencies": {
      "hyperapp": "git://github.com/Swizz/hyperapp.git#fresh-typing",

It even displays the expected hash `93326df` on the console during installation - yet, I don't know what it's installing or why, but there are no typings in the installation folder afterwards.

Swizz:
To be honest, I always using in current dev DTS that way.
I know about `/// <reference path="..." />` but I never found how to use it.
Maybe @andrewiggins or @pspeter3 could help.

pspeter3:
They would not work. Using the reference path is an outdated way to tell the compiler about type definitions. You can now using the typings key in tsconfig.

jbucaran:
@Swizz Can you please update this with the new API? :)

Swizz:
Now synced with [0.12.0](https://github.com/hyperapp/hyperapp/releases/tag/0.12.0) 🎉

jbucaran:
@Swizz What do we need to merge this?

hyperapp.d.ts
```diff
@@ -0,0 +1,89 @@
...
+  (name: keyof Events, data?: Data): Data | any
+}
+
+export interface View<State, Actions> {
+  (state: State, actions: Actions): VirtualNode<{}>
+}
+
+export function app<State, Actions, Events>(app: {}): Emit<{} | null, Events>
+
+/** @namespace [JSX] */
```
okwolf: Why is there JSX typing in here? I didn't think we implement any of that within Hyperapp.

hyperapp.d.ts
```diff
@@ -0,0 +1,89 @@
...
+    | VirtualNodeChild<{} | null>
+    | number
+): VirtualNode<Data>
+
+/** @namespace [Application] */
+
+export type ActionsResult<State> =
+  | Partial<State>
+  | Promise<Partial<State>>
+  | Function // /!\ TODO: Type the thunk Function
```
okwolf: Mission Control: I believe we need a Thunk type.

hyperapp.d.ts
```diff
@@ -0,0 +1,89 @@
...
+   * @param name  The name of the event to call
+   * @param data  Will be reduced by each event handler
+   *
+   * @memberOF [Application]
+  */
+  (name: keyof Events, data?: Data): Data | any
+}
+
+export interface View<State, Actions> {
+  (state: State, actions: Actions): VirtualNode<{}>
```
okwolf: Might want to add a comment describing this.

hyperapp.d.ts
pspeter3: It's for the TypeScript compiler. It uses that to type check JSX when generating the equivalent JavaScript

Swizz:
You can now, please, reviews the View and Actions part

hyperapp.d.ts
```diff
@@ -0,0 +1,128 @@
...
+
+export interface Application<
+  State extends Hyperapp.State,
+  Actions extends Hyperapp.Actions<State>,
+  Events
+> {
+  state?: State
+  actions?: Actions
+  events?: Events
+  view?: View<State, Actions>
```
jbucaran: @Swizz what about root and mixins?

hyperapp.d.ts
```diff
@@ -0,0 +1,128 @@
...
+  State extends Hyperapp.State,
+  Actions extends Hyperapp.Actions<State>
+> {
+  (state: State, actions: ActionsCallers<State, Actions>): VirtualNode<{}>
+}
+
+export interface Events<
+  State extends Hyperapp.State,
+  Actions extends Hyperapp.Actions<State>
+> {}
```
jbucaran: @Swizz What about an event third `data` argument?

hyperapp.d.ts
```diff
@@ -0,0 +1,128 @@
...
+
+export type ActionsResult<State extends Hyperapp.State> =
+  | Partial<State>
+  | Thunk
+  | {}
+  | null
+  | void
+
+
+export type ActionsCallers<
```
jbucaran: @Swizz What if we call these VirtualActions? 

/cc @andyrj @pspeter3

hyperapp.d.ts
Swizz: This is only minimal, for testing `state`, `actions`, `events` and `view`. Application typing will come later

hyperapp.d.ts
Swizz: This is not arguments but the generics. Here is no interface for Events yet

hyperapp.d.ts
Swizz: For naming I'll let you choose

pspeter3:
What can I help with here?

hyperapp.d.ts
```diff
@@ -0,0 +1,128 @@
...
+ * @memberOf [Virtual DOM]
+*/
+export type VirtualNodeChild<Data> = VirtualNode<Data> | string
+
+export interface Component<Data> {
+  /** A Component is a function that return a custom Virtual Node
+   *
+   * @memberOf [Virtual DOM]
+  */
+  (data: Data, children: VirtualNodeChild<{} | null>[]): VirtualNode<Data>
```
okwolf: [React](https://facebook.github.io/react/docs/components-and-props.html#functional-and-class-components) and [Vue.js](https://vuejs.org/v2/guide/components.html#Props) refer to `data` as `props`, should we follow the same convention?

jbucaran: Could be a good idea! We'd need to change it in app and h too.

lazarljubenovic: Just to be complete, Angular refers to it as `inputs`, although in Angular's case, it's impossible for an input not to also become state inside the component after it's received. 

I'm also for renaming, especially since in Vue.js `data` refers to state inside the component.

Swizz: Give the names you want to. I am really bad at naming stuff. I followed the `API.md`

jbucaran: @Swizz Looks good, but what about the tests? We can't merge without tests.

Swizz: @andrewiggins  Are you always among us? Ready to help me about tests?

andrewiggins: I'm here :) Checkout what I did in picodom/picodom#19 if you want a simple example of some tests. In summary, add some typescript files that demonstrates how you would expect the types to be used, and add a call to `tsc` in the test npm script to make sure they compile correctly.

hyperapp.d.ts
Mytrill: h also supports array of numbers (all converted to strings) should it then be:
```typescript
export type VirtualNodeChild<Data> = VirtualNode<Data> | string | number
```
and 
```typescript
children?: 
  | VirtualNodeChild<{} | null>[] 
  | VirtualNodeChild<{} | null>
```

jbucaran:
@Swizz Is this ready? I'm ready to take over now. 😄

Swizz:
There is a lot of commits...
Did I fail my rebase ?

pspeter3:
🎆

JorgeBucaran:
Merged, merged, merged! 💥 🎉

PR #312, opened 2017-07-21
Add onmove lifecycle event

zaceno:
Lifecycle even that fires for elements when their position among siblings changes. Since elements don't correspond to virtual nodes, unless those virtual nodes are keyed, this lifecycle event only makes sense for (and only works with) keyed nodes.

Now that onupdate is only called when a node's props change, this is the only (non-hacky) way to detect reordering.

Why? One use case is for animated transitions (moving elements to their new position smoothly).

### DIFF
--- a/docs/lifecycle-events.md
+++ b/docs/lifecycle-events.md
@@ -20,6 +20,12 @@ Type: ([element](https://developer.mozilla.org/en-US/docs/Web/API/Element))
 
 The onupdate event is fired every time the element's data is updated.
 
+## onmove
+
+Type: ([element](https://developer.mozilla.org/en-US/docs/Web/API/Element))
+
+The onmove event is fired (at least) every time the element's position among siblings is changed. Note that [virtual nodes](/docs/virtual-nodes.md) do not correspond to specific elements unless using [keys](/docs/keys.md). Hence the onmove event will only work for keyed nodes.
+
 ## onremove
 
 Type: ([element](https://developer.mozilla.org/en-US/docs/Web/API/Element))
--- a/src/app.js
+++ b/src/app.js
@@ -236,6 +236,8 @@ export function app(app) {
       var i = 0
       var j = 0
 
+      var moved = false
+
       while (j < len) {
         var oldElement = oldElements[i]
         var oldChild = oldNode.children[i]
@@ -259,12 +261,16 @@ export function app(app) {
           i++
         } else {
           if (oldKey === newKey) {
+            if (moved && newChild.data.onmove) newChild.data.onmove(reusableChild[0])
             patch(element, reusableChild[0], reusableChild[1], newChild, isSVG)
             i++
           } else if (reusableChild[0]) {
+            moved = true
+            if (newChild.data.onmove) newChild.data.onmove(reusableChild[0])
             element.insertBefore(reusableChild[0], oldElement)
             patch(element, reusableChild[0], reusableChild[1], newChild, isSVG)
           } else {
+            moved = true
             patch(element, oldElement, null, newChild, isSVG)
           }
 
--- a/test/lifecycle.test.js
+++ b/test/lifecycle.test.js
@@ -98,3 +98,87 @@ test("onremove", done => {
     }
   })
 })
+
+test("onmove called when siblings moved", done => {
+  const called = {}
+  const handler = key => { called[key] = true }
+  app({
+    state: {
+      nodes: ["a", "b", "c", "d", "e", "f"]
+    },
+    actions: {
+      shuffle: state => {
+        state.nodes = ["a", "b", "e", "d", "c", "f"]
+        return state
+      }
+    },
+    events: {
+      loaded: (state, actions) => actions.shuffle(),
+      render: (state, actions) => {
+        setTimeout(_ => {
+          if (called.c && called.d && called.e && called.f) done()
+        }, 0)
+      }
+    },
+    view: state =>
+      h('div', {}, state.nodes.map(key =>
+        h('div', { key, onmove: el => handler(key) })
+      ))
+  })
+})
+
+test("onmove called when siblings removed", done => {
+  const called = {}
+  const handler = key => { called[key] = true }
+  app({
+    state: {
+      nodes: ["a", "b", "c", "d"]
+    },
+    actions: {
+      shuffle: state => {
+        state.nodes = ["a", "c", "d"]
+        return state
+      }
+    },
+    events: {
+      loaded: (state, actions) => actions.shuffle(),
+      render: (state, actions) => {
+        setTimeout(_ => {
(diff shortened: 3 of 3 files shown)


jbucaran:
Nice work @zaceno. I'd like to confirm with other libraries to know what they are doing to support this use case. If anyone knows, please add your comments.

jbucaran:
@zaceno What about an `onpatch` event that fires for every patch instead? 🤔💭

leeoniya:
domvm has node-level will/didReinsert hooks

jbucaran:
Thanks @leeoniya! That makes sense.

@zaceno Unlike domvm that uses something like _hooks:, we support these events as if there were other DOM events, which looks better on the surface, but it's more internal overhead. I am afraid we'll keep adding more and more events like this.

Isn't there anything else we can do?

leeoniya:
attrs diff/patch performance was definitely a core consideration, and why the hooks are grouped. in fact all special attrs in domvm are `_´ prefixed to allow for fast bypass.

when you guys work on hooks, check how much of the domvm demo you can replicate (i believe it's pretty complete):

http://leeoniya.github.io/domvm/demos/playground/#lifecycle-hooks

zaceno:
@jbucaran 

so `onpatch` would be like the old behavior of `onupdate` -- called basically every time the state changes, right?

That could work! My use-case only needs events when layout changes, so the proposed `onpatch` would be more calls than necessary. But it's probably an acceptable tradeoff to keep core lean.

jbucaran:
@zaceno It would be called every time patch is called. Ideally, you should be able to write logic inside onpatch to reduce waste.

zaceno:
@jbucaran That makes sense. But is it enough to pass element to `onpatch`, to enable that waste reduction logic? Trying to think what else one might need... Perhaps oldNode, newNode + oldelement

jbucaran:
Yeah, don't know. I am just trying to see what other options we have.

zaceno:
Actually, now that I think about it, I really think I need the old onupdate behavior (in some equivalent), because position of elements can change anyway, through CSS. So capturing just reorderings isn't actually enough for me.

Hence I will close this PR. 

I propose we introduce onpatch as a replacement of the old onupdate behavior, for those who miss it. Later we can look into what else we could pass to it. Will look into a PR for that.

jbucaran:
@zaceno What if we call onupdate for key changes? We are actually _not_ doing that right now. 

If that's no good, then `onpatch` might be the way to go. 😄

zaceno:
@jbucaran "when keys change" doesn't make sense to me. When keys are different on oldnode and node, we have that algorithm that reorders the nodes.

Either way, as I said before, just detecting reordering is not enough, since one could use CSS to change the layout even if reordering doesn't happen. So for my case I really need the old onupdate: I want to track the coordinates any time anything might have changed.

zaceno:
In fact, maybe I should look into making my lib a mixin, so I can hook into the render event. That would make usage a lot more complex (not even sure exactly how I'd do it), but maybe that's the right way to go...

jbucaran:
@zaceno In that case we might as well just bring back the old onupdate. The new onupdate is essentially the same as the old onupdate, except that it won't be called when you have attached a lifecycle event or for keys.

zaceno:
Well I think we should keep the new onupdate, since I think a lot of people want that behavior. But we could have both options. Onupdate works as it does today, and onpatch works as onupdate used to work. There's value in both, no?

jbucaran:
I am just looking for ways to eliminate stuff we don't need. An event that fires every time the tree is diffed is certainly needed, so if we are going to introduce a new one, then why is the other good for?

zaceno:
@jbucaran By "more useless than before" you're implying that there aren't many use cases for onupdate (old or new) anyway. So maybe I'm mistaken in thinking people care much about wether onupdate has this new optimized behavior or not? If that's so - then I agree. Then let's just revert onupdate to the old behavior. That would work for me.

cc: @lukejacksonn @ngryman ?

jbucaran:
@zaceno The original use case of onupdate is to communicate changes in the state to DOM nodes that are not handled by hyperapp. For example, when wrapping a third party component, e.g., CodeMirror. 

If setElementData is called, that means we are "rendering" the tree, that is, `render(view(state, actions))` was called.

Inside setElementData we only touch the DOM when oldValue != value, but _call_ onupdate only once. It seems to me that if we are "rendering" the most likely scenario is that at least one prop changed, not that none did.

zaceno:
> An event that fires every time the tree is diffed is certainly needed,

I agree. I think the new onupdate behavior was an attempt to optimize away needless calls without loosing any power. But it was too optimistic, and *did* drop some important abilities. Important to me anyway :) 

So if onupdate isn't frequently used, and this optimization isn't worth a lot to anyone in particular, then I think it should be reverted. If on the other hand it is importan, I'm fine keeping it as long as we reinstate the "every time the tree is diffed" lifecycle event also.

jbucaran:
What about renaming it to:
- onupdate
- onsetdata
- onsync
- ondiff
onpatch

zaceno:
@jbucaran 🤔 

Since you brought it up, I'm leaning toward that we should revert the recent optimization to `onupdate`. And yes, then maybe we should rename `onupdate` to perhaps `onrender` or `onpatch` 

...But @lukejacksonn and @ngryman (and others?) should probably weigh in.

jbucaran:
onrender makes the most sense, since that's exactly what's happening, but onupdate is more user friendly. Well, let's wait for everyone else to comment on this.

lukejacksonn:
I can see use cases for both the old behaviour and the new.. hypothetically:
- `onprops` - when a views prop(s) change (new `onupdate`)
- `oneval` - when a views virtual node is reconstructed (old `onupdate`)

Both of these resemble _app_ events `onupdate` and `onrender` but differ because they get passed reference to an specific `element` as supposed to `state, actions ,data`. Not sure if that means we should name them consistently with _app_ or completely differently as to avoid confusion.

If lifecycle events were good enough, then I can see app events becoming more redundant. As @jbucaran pointed out the other day.. every time I start to think this, people find a new use for app event *x* 😅

jbucaran:
Let's go back to the previous `onupdate` behavior: let's call it `onupdate`? `onreindeer`? `onpatch`?

The first argument to the function should be the `Element`, and a second argument could be a boolean flag `propsChanged` or some other fine strategy to enable users to act accordingly.

zaceno:
`onreindeer` clear winner! :)

lukejacksonn:
Calling it `onupdate` or `onrender` only really makes sense if it is synonymous with _app_ `update` or `render` events.. Reverting to the old behaviour is essentially app `update`.

A second argument like `propsChanged ` is news to me, I had not considered the possibility tbh.

zaceno:
how about second argument is just old props?

PR #313, opened 2017-07-23
replace onupdate with onrender

zaceno:
Onrender basically reinstates the old behavior of onupdate, with a more appropriate name.

If you have something you want to do *only* when some props (or some particular props) have changed, we now pass the old props as a second argument to `onrender` handlers, so you may use whatever logic is appropriate.

I thought this would let me shave off more bytes, but -10 bytes is at least something.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -157,10 +157,10 @@ export function app(app) {
       name === "key" ||
       name === "oncreate" ||
       name === "oninsert" ||
-      name === "onupdate" ||
+      name === "onrender" ||
       name === "onremove"
     ) {
-      return name
+      /* noop */
     } else if (name === "style") {
       for (var i in merge(oldValue, (value = value || {}))) {
         element.style[i] = value[i] || ""
@@ -180,22 +180,17 @@ export function app(app) {
     }
   }
 
-  function updateElementData(element, oldData, data, cb) {
+  function updateElementData(element, oldData, data) {
     for (var name in merge(oldData, data)) {
       var value = data[name]
       var oldValue = oldData[name]
 
-      if (
-        value !== oldValue &&
-        value !== element[name] &&
-        setElementData(element, name, value, oldValue) == null
-      ) {
-        cb = data.onupdate
+      if (value !== oldValue && value !== element[name]) {
+        setElementData(element, name, value, oldValue)
       }
     }
-
-    if (cb != null) {
-      cb(element)
+    if (data.onrender) {
+      data.onrender(element, oldData)
     }
   }
 
--- a/test/lifecycle.test.js
+++ b/test/lifecycle.test.js
@@ -37,48 +37,32 @@ test("oninsert", done => {
   })
 })
 
-test("fire onupdate if node data changes", done => {
-  app({
-    state: "foo",
-    view: state =>
-      h("div", {
-        class: state,
-        onupdate: done
-      }),
-    actions: {
-      change: state => "bar"
-    },
-    events: {
-      loaded: (state, actions) => {
-        actions.change()
+test("fire onrender each render with oldData", () => {
+  return new Promise((resolve, reject) => {
+
+    const dataA = {foo: "bar"}
+
+    const dataB = {
+      onrender: (el, oldData) => {
+        if (oldData && oldData.foo && oldData.foo === dataA.foo) {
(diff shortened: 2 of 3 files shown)


zaceno:
Why?

Because there is a need for a lifecycle event that calls each render, which was lost when `onupdate` behavior changed to be called only when props update. So we took it back, with a more appropriate name.

Rather than *add* `onremove` alongside the new `onupdate` we provide the old props so the user can diff, and decide wether to do an action. So the new `onupdate` behavior is still possible, with the added benefit of leaving it up to the user *how* to determine if props have changed. This is good for situations where props are functions. HyperApp previously had no choice but to assume it was different. But a user can make a much more informed call.

test/lifecycle.test.js
```diff
@@ -37,48 +37,32 @@ test("oninsert", done => {
...
-          class: state,
-          oncreate: noop,
-          onupdate: reject,
-          oninsert: noop,
-          onremove: noop
-        }),
+      state: true,
       actions: {
-        change: state => "foo"
+        change: _ => false
```
jbucaran: () => false

test/lifecycle.test.js
```diff
@@ -37,48 +37,32 @@ test("oninsert", done => {
...
       },
       events: {
         loaded: (state, actions) => {
           actions.change()
-          setTimeout(resolve, 100)
         }
-      }
+      },
+      view: state => h('div', (state ? dataA : dataB))
```
jbucaran: double quotes

src/app.js
```diff
@@ -180,23 +180,16 @@ export function app(app) {
...
 
-  function updateElementData(element, oldData, data, cb) {
+  function updateElementData(element, oldData, data) {
     for (var name in merge(oldData, data)) {
       var value = data[name]
       var oldValue = oldData[name]
 
-      if (
-        value !== oldValue &&
-        value !== element[name] &&
```
jbucaran: @zaceno You can't remove this, otherwise hell breaks loose.

src/app.js
```diff
@@ -180,23 +180,16 @@ export function app(app) {
...
-        cb = data.onupdate
+      if (value !== oldValue) {
+        setElementData(element, name, value, oldValue)
       }
     }
-
-    if (cb != null) {
-      cb(element)
-    }
+    data.onrender && data.onrender(element, oldData)
```
jbucaran: Please use `if`.

src/app.js
```diff
@@ -170,7 +170,7 @@ export function app(app) {
         element[name] = value
       } catch (_) {}
 
-      if (typeof value !== "function") {
+      if (typeof value != "function") {
```
jbucaran: Please use !== when comparing strings.

src/app.js
```diff
@@ -154,13 +154,13 @@ export function app(app) {
 
   function setElementData(element, name, value, oldValue) {
     if (
-      name === "key" ||
-      name === "oncreate" ||
-      name === "oninsert" ||
-      name === "onupdate" ||
-      name === "onremove"
+      name == "key" ||
```
jbucaran: Please use !== when comparing strings.

src/app.js
zaceno: Really? How? Tests all pass. What's it for?

zaceno: I didn't take away the `value !== oldValue` that's still there. And afaik `value !== element[name]` will always be the same as the first check (just slower). So it makes sense to me to get rid of the last check. But I could be missing something?

zaceno:
@jbucaran Thanks! Fixed according to your  comments -- except for the removed `value !== element[name]` I still don't see why removing that would be a problem. In fact I think it would help performance.

src/app.js
jbucaran: @zaceno Yep... because it's kind of difficult to test. My bad. We need this because there are times when the DOM is already synced with `value`, so we don't care if `value !== oldValue`. 

For example, an input text, a check box, etc.

zaceno: Ok, I follow - thanks for the explanation! I'll put it back :)

zaceno:
Ok all comments fixed according to your wishes, @jbucaran :)

zaceno:
Closing at @jbucaran's request

PR #314, opened 2017-07-24
Remove mixins inside mixins

jamen:
So I think this feature didn't go out as well in retrospect.

In order to make a mixin workable with this feature, you have to pass an option that changes the namespace of the actions, state, and events on the app.  Only way to avoid conflicts if it appears multiple times (e.g., you use it and another mixin uses it).  For example:

```js
// Takes an option:
actions[namespace].foo()

// Versus hardcoding it:
actions.myMixin.foo()
```

Only thing that would make this better is if there was some concept of local state and actions built-in.  For example `state.local` and `actions.local` resembling a unique namespace no matter where it is used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,10 @@
   "module": "src/index.js",
   "license": "MIT",
   "repository": "hyperapp/hyperapp",
-  "files": ["src", "dist"],
+  "files": [
+    "src",
+    "dist"
+  ],
   "author": "Jorge Bucaran",
   "keywords": [
     "hyperapp",
--- a/src/app.js
+++ b/src/app.js
@@ -2,7 +2,7 @@ export function app(app) {
   var state = {}
   var actions = {}
   var events = {}
-  var mixins = []
+  var mixins = app.mixins || []
   var view = app.view
   var root = app.root || document.body
   var node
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -7,7 +7,10 @@
   "module": "src/index.js",
   "license": "MIT",
   "repository": "hyperapp/hyperapp",
-  "files": ["src", "dist"],
+  "files": [
+    "src",
+    "dist"
+  ],
```
jamen: My bad including this, but `npm` always is reformatting this for me.  Might as well keep it this way? ¯\_(ツ)_/¯

lukejacksonn:
Has anyone been taking advantage of this feature? Personally I have not, mainly because I fear namespace conflicts causing instability.

Removing this would reduce the complexity of app and perhaps force us to come up with another solution like _scoped mixins_. Which I see being quite useful.

>  if there was some concept of local state and actions built-in.

**Conclusion:** I would not be sad to see mixins in mixins go.

SkaterDad:
I agree with @lukejacksonn regarding potential namespace conflicts with nested mixins.  Definitely not a feature I've used or have any interest in.

jbucaran:
The best use case for mixins in mixins is presets. For that reason, despite this feature opening the door to other problems, I think we should keep it now.

We should not encourage using mixins in mixins for the same reason you use mixins in an app, but rather for creating grouping mixins together and creating so-called "presets".

```js
mixins: [Logging, ThunkUtils, EventUtils, Scoped, HashRouter]
```

jbucaran:
Let's keep it! It's all for the greater good! 💯😉

PR #315, opened 2017-07-24
Give to afterAction event the same behavior as beforeAction

Swizz:
At the moment only `beforeAction` give the called `action` name.

It will be a better user experience to give to both event the same behavior.  Giving the action `name` will grant all common use cases.

Added tests coverage for `afterAction` too

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -78,9 +78,9 @@ export function app(app) {
 
       if (typeof action === "function") {
         namespace[key] = function(data) {
-          var result = emit(
-            "afterAction",
-            action(
+          var result = emit("afterAction", {
+            name: name,
+            data: action(
               state,
               actions,
               emit("beforeAction", {
@@ -88,7 +88,7 @@ export function app(app) {
                 data: data
               }).data
             )
-          )
+          }).data
 
           if (result == null) {
           } else if (typeof result === "function") {
--- a/test/events.test.js
+++ b/test/events.test.js
@@ -61,6 +61,30 @@ test("beforeAction", done => {
   })
 })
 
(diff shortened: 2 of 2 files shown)


zaceno:
Not exactly about this PR, but:

Why are we requiring users to `return {data: 'foo'}` from their `beforeAction` and `afterAction` handlers? Why not just `return 'foo'` and skip the `.data` in the implementations?

Swizz:
I guess here this is because we are excepting `emit` to return its input. And this input is `{ name, data }`.

But you are right, we will run into trouble if the user do not return `{ data }`

zaceno:
Hmm... I think you're right. But that seems a little strange too. Because if we're expecting the user to return the unmodified input, then we already have that `data`. And if we're expecting the user to modify it, why expect them to wrap it in `{data: modifiedOutput}`

Swizz:
Because `emit` return the given data if there is no event or the event `result` is null. 

But yes, this strange to rely on the user to respect the given interface.

zaceno:
Yes, hmm... I think the original implementation needs some work here. For that reason, unfortunately I can't approve this PR. Because it's building on the original, unsound looking, implementation.

Swizz:
Even if the first implementation is incorrect. It is also incorrect to do not provide to two twin events the same behavior.

What do you suggest ? 

* Merge this one and work right after on the rework of thoses
* Work on it instead of this merge

zaceno:
Agree the actions should have mirrored implementations. I think we should fix the implementation first before mirroring a "bad" implementation is what I'm saying. @jbucaran?

PR #316, opened 2017-07-25
enhanced the node checking expression

KhaledElAnsari:
- using `if(variable) {}` is a shortcut for checking if the value exist or not so values like `undefined`, `null`, `0`, `""`, `false` and `NaN` will not pass

- you were checking if `node` is not equal to `true` or `false` and those are `Boolean` values so instead it will easier to check the value type if it's `boolean` or not

### DIFF
--- a/src/h.js
+++ b/src/h.js
@@ -12,7 +12,7 @@ export function h(tag, data) {
       for (var i = node.length; i--; ) {
         stack[stack.length] = node[i]
       }
-    } else if (node !== null && typeof node !== "boolean") {
+    } else if (node != null && typeof node !== "boolean") {
       if (typeof node === "number") {
         node = node + ""
       }
--- a/src/h.js
+++ b/src/h.js
@@ -12,7 +12,7 @@ export function h(tag, data) {
       for (var i = node.length; i--; ) {
         stack[stack.length] = node[i]
       }
-    } else if (node && typeof node !== "boolean") {
+    } else if (node !== null && typeof node !== "boolean") {
       if (typeof node === "number") {
         node = node + ""
       }
--- a/src/h.js
+++ b/src/h.js
@@ -12,7 +12,7 @@ export function h(tag, data) {
       for (var i = node.length; i--; ) {
         stack[stack.length] = node[i]
       }
-    } else if (node != null && node !== true && node !== false) {
+    } else if (node && typeof node !== "boolean") {
       if (typeof node === "number") {
         node = node + ""
       }


leeoniya:
your version not logically equivalent. for example, failing for `0`.

lukejacksonn:
#### Existing (32 chars, worst case 3 comparisons, passes)
```
['', 0, 'x', true, false].map(x => !!(node != null && node !== true && node !== false))
=> [true, true, true, false, false]
```

#### Proposed (25 chars, worst case 2 comparisons, fails)
```
['', 0, 'x', true, false].map(node => !!(node && typeof node !== "boolean"))
=> [false, false, true, false, false]
```

I got a bit carried away there! As @leeoniya points out your proposal is not logically equivalent to what is in master at the moment. Adding `!= null` to the first check of `node` seems to make the check equivalent (unless I am missing an edge case).

#### Alternative (31 chars, worse case 2 comparisons, passes)
```
['', 0, 'x', true, false].map(node => !!(node != null && typeof node !== "boolean"))
=> [true, true, true, false, false]
```

Which results in one less byte and one less check in most cases. I am no perf guru but I see this boiling down to which is more performant:

`typeof node !== "boolean"` or `node !== true && node !== false`

🏎:fire::fire::fire::fire:

jbucaran:
Hi @KhaledElAnsari! 

This line is what makes conditional rendering possible:

```jsx
{true && <h1>Hello</h1>}
```

or

```jsx
{condition ? <h1>Okay</h1> : null}
```

and also why you must convert booleans to a string if you wish to render the name:

```jsx
<h1>{true + ""}</h1>
```

We can't remove the null check, so I guess this boils down to:

```jsx
} else if (node != null && node !== true && node !== false) {
```

or

```jsx
} else if (node != null && typeof node !== "boolean") {
```

So, which one is better? 🤔

jbucaran:
@lukejacksonn Which one is more performant you saying?

KhaledElAnsari:
Thank you guys for your points I'll edit my proposal now @lukejacksonn @leeoniya @jbucaran

Swizz:
Are `typeof` performance issues solved ?

src/h.js
```diff
@@ -12,7 +12,7 @@ export function h(tag, data) {
       for (var i = node.length; i--; ) {
         stack[stack.length] = node[i]
       }
-    } else if (node != null && node !== true && node !== false) {
+    } else if (node !== null && typeof node !== "boolean") {
```
jbucaran: @KhaledElAnsari 

```
} else if (node != null && typeof node !== "boolean") {
```

lukejacksonn:
@jbucaran yeh I have no clue which is more performant but I am going to assume that one less check is always going to win.

:boom:

jbucaran:
@lukejacksonn That is assuming `typeof` is only one check! 😆

SkaterDad:
@jbucaran @lukejacksonn The current & proposed logic are roughly equivalent perf in Chrome on my computer (adapted examples from earlier in this thread).  I'm sure Firefox/Edge would have different results.
https://jsperf.com/ha-pr316/1

leeoniya:
> I'm sure Firefox/Edge would have different results. https://jsperf.com/ha-pr316/1

original is 10-15% faster in Chrome and Firefox and 22% slower in Edge. at least on Win10x64.

KhaledElAnsari:
@jbucaran done

leeoniya:
as a tangent, this PR makes me wonder what Codecov means by "99%". the original logic of this PR surely would have failed at least the two cases i pointed out. i doubt that this specific line happens to be that missing 1%. so, what's the point?

SkaterDad:
> original is 10-15% faster in Chrome and Firefox and 22% slower in Edge. at least on Win10x64.

JS perf is so weird.  I'm seeing that 10% in my JSPerf now too (Chrome - Win7x64).

leeoniya:
if you're on a laptop, thermal throttling often affects the later runs more than the earlier. all benchmarks should really pre-heat the CPU. the effect is more evident in some tests more than others. i dunno if this applies here, but i've seen testcase inversion change the outcome :/

Swizz:
Tested on Chrome 59.0.3071 / Windows 7 0.0.0. The PR is always 5-10% slower than original.

Typeof suffer since while of performance issues.

jbucaran:
And there would be no difference in terms of bytes at the compressed level, so I think we are going to pass on this one.

Thanks anyway @KhaledElAnsari! 👋😄

lukejacksonn:
@leeoniya thanks for settling that one with hard evidence!

What is it they say "Never assume.." 😅

leeoniya:
> @leeoniya thanks for settling that one with hard evidence!

not sure if anecdotal [past] experience and a single jsperf run qualifies as "hard evidence", but it's a slightly better than pure speculation. if anyone deserves the credit here, it's @SkaterDad

lukejacksonn:
Ahh yeh sorry 😝 I thought you posted that `jsperf` link.. @SkaterDad nailed it 💯 @leeoniya your anecdotal experience is always insightful, so thanks regardless!

PR #317, opened 2017-07-26
Refactor events, lifecycle, no emit return, 💯 thunks, cut >100 bytes

jbucaran:
Work in progress.


charlike:
> - Simplify the merge function and enforce the use of a POJO
  (plain old JavaScript object) as your state property. This
  means primitive types are no longer valid states.

Do you remember, i was talking about that from the beginning? :)

> - Refactor thunks. Remove built-in promise support in favor of
  using thunk composition.

Don't like that if i understand it correctly. Didn't see it in the diff. can you show short example?

As about changing one more time the name of the app's argument... :laughing: Incoming isn't "props" in any sense.

jbucaran:
Yup.

Around [here](https://github.com/hyperapp/hyperapp/pull/317/files#diff-bd9c9dcd314f2d7df52935b3a6a4d504R78). 👈

lukejacksonn:
@charlike I wasn't sure on _Remove built-in promise support in favor of
using thunk composition._ wither but I came around to the idea once I saw the implementation and benefits it brings.. a summary (to reenforce my own understanding at least) IIRC:

Action: `(s,a,d) => ({ user: d })`
Thunk: `(s,a,d) => update => fetch('/users').then(x => update({ user: x }))`

Where `update` is a function that accepts a state fragment `actionResult`, merges that with the current `state` and calls `render`. Not dissimilar to `setState` in react.

@jbucaran @Swizz correct me if I am wrong!

SkaterDad:
Yeah, calling it "thunk composition" is confusing.  It does seem like a more powerful feature, though.

The built-in promise support really only handled the "happy path", right?  With these thunks you can more easily do state updates from Promise errors, too?

Also a fan of enforcing state to be an object :+1:

jbucaran:
@SkaterDad @lukejacksonn That works, but the beauty is that if you want "native" promise support like we had, **you can still totally have it**, not shown in your example, but in the following:

```jsx
events: {
  resolve(state, actions, result) {
    return result && typeof result.then === "function" 
      ? update => result.then(update)
      : result
  }
}
```

Certainly, you'll want to wrap that into a mixin, or better, group it with other things like support for observables or generators (I don't know) into a nice preset.

Swizz:
Using `thunk` basically allow you to operate an update of the state, whenever you want/need.
The given `update` function is no more than the internal one we use to trigger an update of the state.

For a sync action returning a partial state, the internal code will be the following : 
```jsx
const result = action(...)
update(result)
```

But, for an action returning a function, the internal code will be the following : 
```jsx
const result = action(...)
result(update)
``` 
Now, you understand that we are giving to the action result the internal `update` function.

By this way, it will be available to you to resolve to an update of the state whatever/whenever you want.

```jsx
(state, actions) => update => PROMISE.then(x => update(x))
```
```jsx
(state, actions) => update => OBSERVABLE.subscribe({ next: x => update(x) })
```

```jsx
(state, actions) => update => CALLBACK_FN(x => update(x))
```

jbucaran:
For the record, in case anyone is confused about anything. You don't need to use thunks anywhere.

Just like you were doing it before is okay (except for promises). For promises you'll have to use a thunk:

```js
(state,actions,data) => update => fetch('/users').then(x => update({ user: x }))
```

or if you add support for promises like in my [example above](https://github.com/hyperapp/hyperapp/pull/317#issuecomment-318394979), you can do this instead:

```js
(state,actions,data) => update => fetch('/users').then(x => { user: x })
```

charlike:
@jbucaran oh yea, examples clears things. It now reminds me the `freactal` which i really loved and thought for something like it, even before its release. :) So yea, i'm agree, gives power and flexibility.

Based on that i prototyped some ideas about that signatures of effects, reducers, render and hooks to be the same.

```js
foo({
  // `state` here would be empty object or undefined (not sure),
  // since they should not have access to it
  effects: {
    bar: ({ actions }, arg1, arg2) => {}
  },
  // forced synchronous actions
  reducers: {
    qux: ({ state, actions }, arg1, arg2) => {}
  },
  // here actions and state are actually "props",
  // so probably easy mocking. When stateless components
  // the `state` will be empty object, for no errors
  render: ({ state, actions, ...props }) => {},
  hooks: {
    init: ({ state, actions }) => {},
    error: ({ state, actions }, err) => {}
  }
})
```

I don't know if i'm the only one that i see one pretty fantastic ability of the destruction: allows enforcing conventions.

Anyway, above is just a concept of what i like.

docs/actions.md
```diff
@@ -1,182 +1,117 @@
 # Actions
 
-Use [actions](/docs/api.md#actions) to update the state.
+Actions are functions which take the current [state](/docs/state.md) and return a partial state or a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to a partial state. Actions represent the intent to manipulate the state tree.
```
Swizz: So is promise resolving will be part of the core or wont ?

docs/api.md
```diff
@@ -95,11 +97,20 @@ See also [Actions](/docs/actions.md).
 
 #### ActionResult
 
-A partial state or [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to a partial state.
+A partial state, [thunk](#thunk) or [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to a partial state.
```
Swizz: Now `ActionResult` can be whatever you want as the `resolve` event allows you to transform it.
Or you are telling about `ActionResult` is the only thing the core will understand do trigger an update of the state ?

docs/events.md
```diff
@@ -85,13 +95,17 @@ app({
 The `emit` function is available as the return value of the [`app`](/docs/api.md#app) function call itself.
```
Swizz: Not the truth at all, isnt it ?

docs/mixins.md
```diff
@@ -43,7 +44,7 @@ const ActionPerformance = (ignore = [], cache = []) => emit => ({
     afterAction() {
```
Swizz: Use `resolve` instead.

src/app.js
```diff
@@ -1,59 +1,55 @@
...
 
-  node = hydrate((element = root.querySelector("[data-ssr]")), [].map)
+  schedule(
+    (node = hydrate((element = root.querySelector("[data-ssr]")), [].map))
+  )
 
-  repaint(emit("init"))
-
-  return emit
+  return emit("load")
```
Swizz: Why not : 
```jsx
return emit("load", emit)
```

To return emit by default for non breaking change here. And emit will be given to `load` as data to permit to return emit even if you are returning something else.

```jsx
loads(state, actions, emit) {
  return { actions, emit }
}
```

docs/actions.md
charlike: The answer is "won't", but i believe it isn't the correct question. This description isn't very clear, because it don't mention that you may return a function which gets one argument called `update` which may/should be used to update the state, and also that `update` function returns the merged new object.

jbucaran: Yeah, haven't updated it yet.

docs/events.md
jbucaran: Hence wip 😆

src/app.js
```diff
@@ -63,64 +59,40 @@ export function app(app) {
...
-          } else if (typeof result.then === "function") {
-            result.then(update)
-          } else {
-            update(result)
-          }
+          emit("action", { name: name, data: data })
+
+          var result = emit("resolve", action(state, actions, data))
+
+          return typeof result === "function" ? result(update) : update(result)
```
MatejMazur: Is this sufficient for all situations? Because I have a problem with this example with async/await: https://codepen.io/MatejMazur/pen/eEJrJv?editors=0010 ..but maybe I'm wrong and I just don't fully get it yet. ;-)

src/app.js
zaceno: Yes! Please! Awesome idea @Swizz !

jbucaran:
For the first 100/100% coverage and 141B cut.

andyrj:
this PR is amazing!  :heart:   LGTM

docs/actions.md
dodekeract: > Actions represent the intent to manipulate the state tree.

I still disagree for apps that are more complex than a hello world. I very often use them to e.g. communicate via `socket.io`/`postMessage`/`http`/etc..

lukejacksonn:
@jbucaran I have only been keeping an eye on changes made to `app.js` but like @andyrj says it is looking good from here, some nice refactors and better naming of things 👍 💯

docs/actions.md
jbucaran: @dodekeract I'll mention that actions have others uses too.

selfup:
Wait why are we getting rid of emit? 🤔 I don't know if I like getting rid of such a powerful feature 😭

dodekeract:
@jbucaran Will we still support `beforeAction` and other events that aren't mentioned in your commit?

jbucaran:
> @dodekeract  I definitely need some place to catch actions being called.

Sure, `beforeAction` → `action`. There is no `afterAction` though.

jbucaran:
@selfup Sorry for the confusion, that commit message is a bit outdated, thanks for pointing this out.

selfup:
@jbucaran Ok thanks for clarifying! 😄

src/app.js
```diff
@@ -1,108 +1,71 @@
...
-  var view = app.view
-  var root = app.root || document.body
+  var view = props.view
   var node
-  var element
-  var locked = false
-  var loaded = false
+  var root = props.root
+  var host = (root && root.parentNode) || document.body
+  var lock = false
```
jamen: could we just do `var lock` with no `false`?  `undefined` will coerce to `true`, and it will just cycle between the booleans from there.

docs/actions.md
```diff
@@ -1,182 +1,129 @@
...
-    oneUp: (state, actions, index) => {
-      return R.over(
-        R.lensPath(["counters", index, "value"]),
-        value => value + 1,
-        state
-      )
-    }
+    Game: gameActions,
+    Score: scoreActions,
+    ...userActions
```
jamen: Another nitpick, but since when has it been convention to capitalize these?  If you were to ask me it just seems odd: `actions.Game.tick`, `actions.Score.add`.

docs/actions.md
```diff
@@ -1,182 +1,129 @@
...
+        autofocus
+        value={state.text}
+        oninput={e => setText(e.target.value)}
+      />
+    </div>,
   actions: {
-    up: ({ count, maxCount }) => ({
-      count: count + (maxCount > count ? 1 : -maxCount)
-    })
+    setText: (state, actions, text) => ({ text })
```
jamen: Just a tiny nitpick how this is written.  It feels inconsistent with everything below being written as:

```js
setText(state, actions, text) {
  return { text }
}
```

Does/can prettier have a preference? Or should they just coexist?

jbucaran: @jamen Will fix it right away.

docs/actions.md
jbucaran: @jamen There's no convention, it's just my current personal preference (but I can change my mind). What do you think is best and how can we find a way to agree on this?

jamen: @jbucaran If there is some rationale behind it, it would be nice to include a note.  Previously I just remember people writing them plainly: `actions.game.tick`, `state.score.count`, etc. and I had no major problems with this.

It does create a nice immediate distinction between your own stuff and stuff mixins bring in, but it is not really conveyed clearly in this example.

docs/actions.md
dodekeract: @jamen Prettier doesn't have a preference here. It'll use either version, depending on what the original code was.

src/app.js
jbucaran: Let me see if that works.

docs/events.md
```diff
@@ -12,71 +18,105 @@ app({
...
+```
 
-### beforeAction
+### resolve
 
-The beforeAction event fires before an action is called. This event can be useful to implement middleware, developer tools, etc.
+The [`resolve`](/docs/api.md#resolve) event fires after an action is called, allowing you to intercept its return value. Use this event to customize what types actions are allowed to return and modify the default update mechanism.
 
-### afterAction
+- Allow actions to return a [Promise](https://github.com/tc39/proposal-observable).
```
zaceno: url is wrong. Points to observable docs instead of promise docs

docs/lifecycle-events.md
```diff
@@ -1,59 +1,73 @@
...
+Lifecycle events are custom function handlers invoked at various points in the life of a [virtual node](/docs/virtual-nodes.md). They are used like any other [attribute](/docs/api.md#attributes).
 
 ## oncreate
 
-The oncreate event is fired when the element is created, but before it is inserted into the DOM. Use this event to start animations before an element is rendered.
-
-## oninsert
-
-The oninsert event is fired after the element is created and inserted into the DOM. Use this event to wrap third party libraries that require a reference to a DOM node, etc.
+The [oncreate](/docs/api.md#oncreate) event is fired after the element is created, but before it is attached to the DOM. Use it to start animations before an element is rendered, [adapt an external library](#adapting-an-external-library) that requires a reference to a DOM node, etc.
```
zaceno: We're changing this right? It's going to be called *after* element has been attached to the dom, right?

jbucaran: @zaceno Correct. 👍

docs/actions.md
jbucaran: @dodekeract How do you feel about this point?

docs/actions.md
```diff
@@ -1,182 +1,129 @@
...
-      <button onclick={() => actions.up(1)}>+</button>
-    </main>,
-  actions: {
-    up: ({ count }, actions, data = 0) => ({
-      count: count + data
-    })
-  }
-})
-```
+Actions are often called as a result of user events triggered on the [view](/docs/view.md) or inside application [events](/docs/events.md).
```
lukejacksonn: triggered **from** the view or from inside of application [events](/docs/events.md).

docs/actions.md
```diff
@@ -1,182 +1,129 @@
...
 
-## Side Effects
+## Thunks
 
-Actions are not required to have a return value. You can use them to call other actions, for example after an async operation has completed.
+Actions can return a function instead of a partial state. This function is called a _thunk_. They operate like regular actions but will not trigger a state update unless you call [`update`](/docs/api.md#update) from within the thunk function.
```
lukejacksonn: unless [`update`](/docs/api.md#update) is called from within the thunk function.

docs/actions.md
```diff
@@ -1,182 +1,129 @@
...
-      </h1>
-      <button onclick={actions.upLater}>+</button>
-    </main>,
   actions: {
-    up: ({ count }) => ({
-      count: count + 1
-    }),
-    upLater: (state, actions) => {
-      setTimeout(actions.up, 1000)
+    defer(state, actions, data) {
```
lukejacksonn: how come we are using this syntax here.. and not

```
defer: (state, actions, data) => update => {
  ...
  update(newData)
}
```

docs/actions.md
```diff
@@ -1,182 +1,129 @@
...
+          url: json[0].url
+        })
+      )
+    }
   }
 })
 ```
 
-Actions can be written as [async functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) too.
+Actions need not have a return value at all. You can use them to create side effects, call other actions, etc.
```
lukejacksonn: Perhaps worth mentioning explicitly that..
`Actions that return nothing will not trigger a state update`

docs/actions.md
lukejacksonn: Ok, I see we are using this syntax everywhere..

docs/actions.md
```diff
@@ -1,182 +1,129 @@
...
 
-## Complex State
-
-Suppose we have a complex state object and wish to update a given property avoiding mutation.
-
-Here is one way we could achieve this using [Ramda](https://github.com/ramda/ramda).
+## Namespaces
 
-[Try it online](https://codepen.io/hyperapp/pen/Zygvbg?editors=0010)
+We iterate over your action keys, allowing for namespaced actions. Namespaces help you organize your actions into categories or domains.
```
lukejacksonn: Hyperapp iterates over action keys recursively during setup, this means that actions can be _nested_ under developer defined namespaces.

docs/actions.md
```diff
@@ -1,182 +1,129 @@
 # Actions
 
-Use [actions](/docs/api.md#actions) to update the state.
+Actions are functions which take the current [state](/docs/state.md) and return a partial state or a [thunk](#thunks). Actions represent the intent to manipulate the state tree.
```
lukejacksonn: Is this link to thunks correct? Or should it be [Thunks](/docs/actions.md#thunks)

docs/events.md
```diff
@@ -85,34 +125,37 @@ app({
...
+const mixin = emit => ({
+  // ...
+})
 ```
 
-The `emit` function returns the supplied data reduced by successively calling each event handler of the specified event.
+The `emit` function returns the supplied data, reduced by successively calling each event handler of the specified event.
 
-### Interoperatiblity
+### Interop
```
lukejacksonn: Not convinced about this title 🤔

package.json
```diff
@@ -7,7 +7,10 @@
   "module": "src/index.js",
   "license": "MIT",
   "repository": "hyperapp/hyperapp",
-  "files": ["src", "dist"],
+  "files": [
```
lukejacksonn: What is this key used for (just out of curiosity)

docs/events.md
jbucaran: @lukejacksonn I thought it sounded okay. What else?

lukejacksonn: Not sure.. perhaps just `Interoperatiblity `

package.json
SkaterDad: It determines which files & folders get published to `npm`.
https://docs.npmjs.com/files/package.json#files

lukejacksonn: @SkaterDad thanks!

jbucaran:
Thank you everyone for helping to review this PR! Finally merging! 🎉

PR #321, opened 2017-08-03
Fix issue where two structures that differ only on a value attribute are not updated correctly

fsodano:
This PR aims to solve an issue I'm currently facing when the new html structure that needs to be rendered only differs in a single "value" attribute.

Bug example: https://codepen.io/fsodano/pen/YxGRZX

Here's where the error was introduced:
https://github.com/hyperapp/hyperapp/commit/010172ca36596deef445ec2292478cd99c0c1a5b a change was introduced in the way hypearpp identifies if an element needs to be updated or not.

This is the old way, where it would check if the oldValue was present in the element, if the attribute name was either "value" or "checked".
```javascript
  function updateElementData(element, oldData, data) {
      var value = data[name]
      var oldValue =
        name === "value" || name === "checked" ? element[name] : oldData[name]
      var onupdate

      if (name === "onupdate") {
      } else if (value !== oldValue) {
        setElementData(element, name, value, oldValue)
        onupdate = data.onupdate
      }
    }

    if (onupdate != null) {
      onupdate(element)
    }
  }
```

This is the new way, which removed that check, and also inverted the priority of the source of truth for the correct data (before it was the element, and then the oldData, now it's backwards)

```javascript
  function updateElementData(element, oldData, data, cb) {
      var value = data[name]
      var oldValue = oldData[name]

      if (
        value !== oldValue &&
        value !== element[name] &&
        setElementData(element, name, value, oldValue) == null
      ) {
        cb = data.onupdate
      }
    }

    if (cb != null) {
      cb(element)
    }
  }
```

### DIFF
--- a/test/app.test.js
+++ b/test/app.test.js
@@ -67,3 +67,47 @@ test("optional view", done => {
     }
   })
 })
+
+test("view values update upon state change", done => {
+  document.body.innerHTML = `<div id="app"></div>`
+  const emit = app({
+    root: document.getElementById("app"),
+    state: { changeInput: false },
+    view: state => h("input", { id: state.changeInput ? "foo" : "bar", "value" : state.changeInput ? "fizz" : "buzz" }),
+    actions: {
+      updateInput (app, state, actions) {
+        state.changeInput = true
+        return state
+      }
+    },
+    events: {
+      updateView (state, actions) {
+        actions.updateInput()
+      },
+      load () {
+        expect(document.body.innerHTML).toBe(
+          `<div id="app"></div>`
+        )
+      }
+    }
+  })
+
+  const firstRenderId = requestAnimationFrame(() => {
+    expect(document.body.innerHTML).toBe(
+      `<input id="bar" value="buzz">`
+    )
+
+    emit("updateView")
+
+    const secondRenderId = requestAnimationFrame(() => {
+      console.log(document.body.innerHTML)
+      expect(document.body.innerHTML).toBe(
+        `<input id="foo" value="fizz">`
+      )
+      done()
+      cancelAnimationFrame(secondRenderId)
+    })
+
+    cancelAnimationFrame(firstRenderId)
+  })
+})


jbucaran:
Thanks @fsodano! Can you share an example that shows why the new behavior is broken? I'd like to create a test or have the example at hand for when I shuffle things around.

fsodano:
I added a test! Can you let me know if it's enough? Otherwise I'll share an explanatory screencast tomorrow.

jbucaran:
@fsodano A screencast works too. I'll look at the test in a bit. If you can add a bit more of the problem you are having, that'd be great. I just want to know what's exactly happening. And yes, we'll merge this after that, because I think I know you are right! 😄

fsodano:
Here's a link to the screencast with the example:
https://drive.google.com/file/d/0BwFVXUo4ia4mUnljeE90MGFva2c/view

First I'm showing the incorrect behaviour: I enter the value "1" in the first input, and click on  "scan", and the next input (different one), maintains the value of the previous input. After that, I'm updating `app.js` with the code changes of this PR, reloading the page and trying again. The second time around it works, I enter "1" in the first input, and when i click on "scan", the next input is empty, as expected.

Here's a screenshot of the `oldValue` variable showing 
 "" (incorrect) vs the `element.value` showing "1" (correct). https://www.dropbox.com/s/m1scq5i4jr3g6qk/Screenshot%202017-08-03%2022.13.39.png?dl=0, this is using the code in this PR, which will then resolve the value to be "1" (correct), because it's fetching it from the element.

In my example, I have 2 inputs which are mutually exclusive, but will render in the same "position", so to speak. So hyperapp is treating it as the same input, and updating its id; the value gathered from h(), however, is empty (unsure why) and it doesn't get updated with the current code.

This is the relevant bit of my application's code.
```javascript
//...
  if (state.ui.scanBarcode) {
    const barcodeScanner = h('input', { type: 'text', id: 'ti-ts-barcode-input', value: state.currentIssue.item })
  }

  if (state.ui.scanLocation) {
    const locationScanner = h('input', { type: 'text', id: 'ti-ts-location-input', value: state.currentIssue.location })
  }
//...
```

Let me know if you need more details.

jbucaran:
@fsodano Thanks for creating that cast. I can definitely see there is a problem and a bug, but I don't understand your example very well. Now that we've come this far, would it be possible to have a CodePen or Fiddle with a minimal reproduction of the bug?

jbucaran:
@fsodano 


Here is a simple example:

```jsx
app({
  state: { changeInput: false },
  view: (state, actions) =>
    h("div", {}, [
      h("button", { onclick: actions.updateInput }),
      h("input", {
        id: state.changeInput ? "foo" : "bar",
        value: state.changeInput ? "fizz" : "buzz"
      })
    ]),
  actions: {
    updateInput(state, actions) {
      return { changeInput: !state.changeInput }
    }
  }
})
```

fsodano:
Hey there! Actually my test is working in the older versions. But here is a reproduction of the bug in codepen. https://codepen.io/fsodano/pen/YxGRZX

jbucaran:
@fsodano Thanks for all your work on reporting this bug! Looks good.

Let's rebase and merge after #317 🙏 👍

jbucaran:
@fsodano Or we could also simply do `value !== element[name]`. But we would be accessing the element's prop directly on each call to updateElementData which may be slower than looking the value on the oldData for most props (except for `value` and `checked`, which are the "stateful" ones).

leeoniya:
> But we would be accessing the element's prop directly on each call

this is the right way to do it.

jbucaran:
@leeoniya But isn't that expected to be slower? (Accessing the element's prop directly.)

jbucaran:
@leeoniya Or you mean like this PR proposes, incidentally, also the way preact does it.

```jsx
var oldValue =
  name === "value" || name === "checked" ? element[name] : oldData[name]
```

leeoniya:
comparing against the live value prevents lots of odd focus/caret position and value-reset bugs. the perf impact is immeasurable unless you plan on having thousands of inputs, and in that case it's not large anyhow. i have these props special-cased, too [1][2].

[1] https://github.com/leeoniya/domvm/blob/3.x-dev/src/view/patchAttrs.js#L56-L57
[2] https://github.com/leeoniya/domvm/blob/3.x-dev/src/view/utils.js#L19-L36

jbucaran:
@leeoniya Not sure we're talking about the same thing. Can you clarify?

Do you mean simply `value !== element[name]` like [I said](https://github.com/hyperapp/hyperapp/pull/321#issuecomment-320028081) in the comment you quoted me from, **or** doing it only for elements with `value` and `checked` props?

leeoniya:
the latter. in fact, you can see from the commented-out code in isDynProp that i removed the element type pre-check switch because it caused a significant perf drop. i only test the prop names now and compare against the DOM for `value`, `checked`, and `selected`.

jbucaran:
Also `selected`! 😥

jbucaran:
@leeoniya So now that we're on the same page. My original question was, what if you just did?

```js
var oval = vnode.el[key]
```

leeoniya:
it is probably slower, but mainly there are multiple incongruences between attrs and props [1], so diffing becomes more of a minefield that requires additional checks, which will eat into perf. i haven't gone down that route too much, so feel free to explore. Preact diffs to the real dom, so there might be some clues there.

[1] https://javascript.info/dom-attributes-and-properties#dom-properties-are-typed

fsodano:
@jbucaran do you need me to give you a hand with anything? I agree with @leeoniya's solution as I tried to fetch all the values from the element as well, but It was giving me some odd results.

jbucaran:
@fsodano Let's merge this after #317. I'll need you to rebase and update the code accordingly, but we're definitely getting this shipped soon.

jbucaran:
Hi @fsodano! 👋 Can you update this PR to add only the test case?

fsodano:
On it!

fsodano:
@jbucaran, I've included my tests and they are failing. I'm looking into it, but I find something very odd, as if the application is forgetting to include my rootElement when rendering.

Can you give this a go and tell me if you find it odd? Maybe I'm missing something from the last update?

```javascript
test('view values update upon state change', done => {
  document.body.innerHTML = `<div id="app"></div>`
  const emit = app({
    root: document.getElementById('app'),
    state: { changeInput: false },
    view: state => h('input', { id: state.changeInput ? 'foo' : 'bar', 'value': state.changeInput ? 'fizz' : 'buzz' }),
    actions: {
      updateInput (app, state, actions) {
        state.changeInput = true
        return state
      }
    },
    events: {
      test (state, actions) {
        actions.updateInput()
      },
      load () {
        expect(document.body.innerHTML).toBe(
          `<div id="app"></div>`
        )
      }
    }
  })

  requestAnimationFrame(() => {
    console.log('Expected: <div id="app"><input id="bar" value="buzz"></div>')
    console.log('Got: :' + document.body.innerHTML)
    expect(document.body.innerHTML).toBe(
      `<div id="app"><input id="bar" value="buzz"></div>`
    )
  })
})
```
Screenshot for comparison
https://www.dropbox.com/s/c4m7wwvb9ut0ms4/Screenshot%202017-08-10%2009.24.15.png?dl=0

andyrj:
In #317 the patch was changed to destroy render target, so you need to take that into account and add the div to your view, or change test to expect output without the render target now.

fsodano:
Thanks, I'll update the tests

fsodano:
@andyrj I've updated the test to reflect these changes. Thanks for the help!

jbucaran:
Thanks @fsodano! 🎉🎉

PR #323, opened 2017-08-10
adding initial attempt at flow type declarations. Need another pair o…

andyrj:
…f eyes to help with errors from flow...

screenshot from 2017-08-10 12-48-56

I'm really confused why flow is complaining about this module declaration can anyone spot what I'm doing wrong here?  I wouldn't call myself an expert with flow type declaration files exactly.  My google-fu is failing me with this error message but I feel like I am close to a workable module here...


andyrj:
kind of strange but the following code works just fine in their try flow webpage...

```js
declare type State = { [string]: any };

declare type Thunk = () => State;
declare type ActionResult = State | Thunk;

declare type Action = (
  state: State,
  actions: Actions,
  data?: any
) => ActionResult;
declare type Actions = { [string]: Action };

let state: State = { test: "foobar" };
let actions: Actions = { run: () => ({test: "baz"})};
let app = ({state, actions}) => {console.log("wtf")};

actions["run"] = "foobar";
```
shows the type error I would expect for trying to assign a string to a type that is expected to be a function...

jbucaran:
@andyrj Is it possible to rename the index.js.flow file to hyperapp.js.flow?

Also, what about .flowconfig, is that file not needed here?

andyrj:
@jbucaran yes on both counts, I didn't mean to commit my  .flowconfig, and this file should be named hyperapp.js.flow.  I will be updating this PR shortly.

jbucaran:
@andyrj Thanks for making those changes. So, after we merge this, how does one go about checking if this works? And how does this help a project using hyperapp?

package.json
```diff
@@ -24,7 +24,7 @@
   ],
   "scripts": {
     "test": "bundlesize && jest --coverage --no-cache",
-    "build": "npm run bundle && npm run minify",
+    "build": "npm run bundle && npm run minify && cp hyperapp.js.flow dist/",
```
jbucaran: Not sure, would have to check with other libraries first, but I think:

```
   "files": [
     "src",
     "dist",
     "hyperapp.js.flow"
   ],
```

should work too no? @andyrj

andyrj: your right, that would be fine as well

andyrj:
We could add flow types into the test suite, and then run flow on the test directory as an extra step for npm test, with 100% test coverage and with types on all the tests we should get a fully exercised flow library definition as well...

package.json
andyrj: actually no it won't, flow looks for the file named hyperapp.js.flow next to the hyperapp.js ...  infact it may need to be named hyerapp.min.js.flow for it to work seamlessly.

jbucaran: @andyrj Can't we add this file under src?

andyrj: I'm not sure, I think it has to be next to the file being imported, and in this case that is dist/hyperapp.js correct?  So I think dist/hyperapp.js.flow is the only default file that flow will check for otherwise you have to add the file into .flowconfig [libs]

jbucaran:
@andyrj I know I asked about tests before, but no, what I mean this time is how to use this new stuff? Like, okay we added flow, now what? :)

package.json
jbucaran: Of course you can. `cp src/hyperapp.js.flow` no? :P

andyrj:
oh gotcha, I think the quickest way to verify it's working is to install flow integration for your ide of choice, then in a project using hyperapp add a .flowconfig with this hyperapp.js.flow declared there and you should get inteli-sense code completion with type declarations, and if you add // @flow to your code which calls hyperapp.app you shouldn't get any errors for the types on app not being declared...

I didn't figure out how to provide type declarations for users writing mixins.  But basic usage of app, and h if used should come up type error clean if the code is correct. :)

andyrj:
I am still getting the error relating to Callable not found on object literal from the screen shot on this PR, will look into it some more when I get time, for now please don't merge this yet.

andyrj:
Should have no errors now, had some typos and a few errors in my types...  Could definitely still use some peer review in this all.

jbucaran:
Maybe @Swizz can have a quick look? 🙏 

@andyrj It might also be worth sharing our conversation the other day about how people can use this, who this benefits and add a before/after screenshot too.

andyrj:
I agree if a few people would add flow to there hyperapp projects to see if I mis-typed anything they are using it would be helpful.  I am fine with sharing that conversation from slack.  I can also post directions in this thread if need be.

andyrj:
If I can have a few volunteers to add flow type to their hyperapp's and report back on any issues it would be greatly appreciated.  Basic instructions for vscode are as follows:

1. install vs code flow plugin - [this one](https://marketplace.visualstudio.com/items?itemName=flowtype.flow-for-vscode)
2. in your project you want to test: npm install --save-dev flow-bin
3. flow init
4. copy the hyperapp.js.flow from [here](https://raw.githubusercontent.com/andyrj/hyperapp/e04faebe538c20e398ae05b3747f5d151e27fa81/hyperapp.js.flow) into your ./node_modules/hyperapp/dist/
5. Enable flow on your source file that calls h or app by add the following comment to the top: // @flow
6. You should see red and green squiggles appear in the source under flow errors and uncovered code.
7.  You can also run the check form the command line with simply: flow  or  flow check

If you could follow the above and report back any errors you find in my type declarations it would be awesome!

okwolf:
@andyrj this is still using `mixins` and `events`. Will this be updated in time for 1.0?

andyrj:
I was waiting for the dts update to finish before updating flow lib def based on it.  I think that is done so can look at it again.

JorgeBucaran:
@andyrj Would you like to revive this? :)

andyrj:
Ya I can take another stab at it and update for the current api.

JorgeBucaran:
@andyrj That would be awesome.

JorgeBucaran:
Hey @andyrj, thanks for your work on this PR, but for the looks of it, we are not really getting things done so I am just going to call it off and close here. If anyone else wants to contribute with flow type annotations, be my guest! 👍

PR #325, opened 2017-08-12
Fix typo

tu4mo:


jbucaran:
Thanks @tu4mo! This got fixed before 0.11.0 which I just published! 👍

PR #326, opened 2017-08-15
include an import in the hello world example

capaj:
sure it doesn't look as nice, but it's helpful to see it all together

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,9 @@ HyperApp is a JavaScript library for building frontend applications.
 [Try it Online](https://codepen.io/hyperapp/pen/zNxZLP?editors=0010)
 
 ```jsx
+import { h, app } from 'hyperapp'
+/** @jsx h */
+
 app({
   state: {
     count: 0


jbucaran:
Hi @capaj, thanks for your feedback.

The prominent "Try it Online" link should take you to a page where you can see the whole thing. 

Isn't that enough?

capaj:
@jbucaran while that shows how you use it via a global variable, for learning how to use it via ES6 import you have to dig deep into the docs. Yet this is a step every new user using ES6 import needs to do, basically forcing all of them to dig into the docs.

jbucaran:
What do you suggest we do about the other examples? Should we add import to those as well? 🤔 

I am not sure. React's examples also lack it.

- https://facebook.github.io/react
- https://github.com/facebook/react#examples

React is not my moral compass, but it's something they must've thought about as well.

What do you guys think?

/cc @lukejacksonn @jamen @zaceno

README.md
```diff
@@ -17,6 +17,9 @@ HyperApp is a JavaScript library for building frontend applications.
 [Try it Online](https://codepen.io/hyperapp/pen/zNxZLP?editors=0010)
 
 ```jsx
+import { h, app } from 'hyperapp'
+/** @jsx h */
```
Swizz: This is transpiler specific. As the import statement LGTM to understand where `h` and `app` come from. But I dont think exposing pragma is a good thing.

zaceno:
How to `import {h, app} from 'hyperapp'` is right there on the Getting Started page which is prominently linked to from the README.md. I feel like that's enough (it *barely* qualifies as "digging into the docs" imo)

So as far as I'm concerned, we're good.

jbucaran:
The example on the README is the entrée before you move on to the real nuts and bolts, for which you need to either dig into the docs one way or another or go through the examples on [CodePen](https://codepen.io/hyperapp) or [Glitch](https://hyperapp.glitch.me/).

I know @capaj's intentions are good and well-intended, but this change is not doing it for me. Still want to hear what others think.

andyrj:
I agree with @zaceno I think the import example in readme is sufficient.

jamen:
This is clear from our perspectives, and not something that is hard to figure out.  But, adding this only helps new people build an understanding faster, so I wouldn't mind this.  I agree with @Swizz's comment though.

jbucaran:
Tough call, but I'll keep the example the way it is for now. Thanks @capaj! 👋😉

PR #327, opened 2017-08-15
Correct link to vnode document

dylan:
Ensures vnode mentions in the various documents point to the correct place.


jbucaran:
Thanks @dylan!

PR #330, opened 2017-08-16
Add clarification on the reason for keys

zaceno:


docs/keys.md
```diff
@@ -17,7 +15,7 @@ const ImageGallery = images =>
   </ul>
 ```
 
-Don't use an array index as key. If the position and number of items in a list is fixed, it will make no difference, but if the list is dynamic, the key will change every time the tree is rebuilt.
+Don't use an array index as key, if the index also specifies the order of siblings. If the position and number of items in a list is fixed, it will make no difference, but if the list is dynamic, the key will change every time the tree is rebuilt.
```
jbucaran: @zaceno It took me a moment to understand this sentence. It makes sense now. Basically:

> Don't use an array index as key, if the index also specifies the order of items.

Do you think we'd lose preciseness, but gain clarity if we say "items" instead?

docs/keys.md
```diff
@@ -1,10 +1,8 @@
 # Keys
 
-Every time your application is rendered, a [virtual node](/docs/virtual-node.md) tree is created from scratch.
+Every time your application is rendered, a [virtual node](/docs/virtual-node.md) tree is created from scratch. This means that between renders, there is no definite mapping from a virtual node to a particular DOM element.
 
-Keys help identify which nodes were added, changed or removed in a list when comparing the old and new tree.
-
-Use keys to identify items in a list that can be re-ordered or where items can be added to / removed from.
+However, nodes with the same `key` accross renders will be interpreted as representing a particular DOM element, allowing you to force dom elements to be reordered, created or destroyed, rather than repurposed. This is sometimes important in order to avoid resetting data attached to DOM elements by the browser (for example the binary image data associated with an `<img src="..." />`), or in [lifecycle events](/docs/lifecycle-events.md).
```
jbucaran: Can you work a bit more on this sentence? It's quite verbose.

Also a couple of petpeeves:

- dom -> DOM (thought it would be nicer if you could find a way to not have to mention DOM so many times).
- lifecycle events -> VirtualDOM Events. 
  - https://github.com/hyperapp/hyperapp/blob/master/docs/vdom-events.md

jbucaran:
@zaceno Left you a few comments. Also, it seems you forgot to rebase before creating this PR? 🤔

zaceno:
@jbucaran thanks! Good comments

I made the PR within github (didn't check it out locally) and I had it up in a tab for a day or so (doing other stuff) so yeah, that could explain why it wasn't based on current master. I'll fix that too.

docs/keys.md
zaceno: Or perhaps just leave it at "specifies the order."

jbucaran: That sounds worse. Let's go with your "order of siblings" then. :

zaceno: Ok :)

zaceno:
So I pretty much rewrote the whole thing. A little more compact, and hopefully sounds better. Not quite as exact or comprehensive, but I think it'll do.

jbucaran:
@zaceno I'll probably make some tiny minor tweaks later, but this is great. Thanks! 🎉🎉

PR #332, opened 2017-08-18
Fix mixin init order. Close #329.

### LINKED ISSUE #329: Invert mixin initialization order.
Currently we are initializing the app, _then_ mixins. This makes it impossible for the app to overwrite a definition added by a mixin. 

Your app should have the highest precedence and not mixins.

jbucaran:
- Shave off presets (mixin mixins) in the process.
- Update tests & docs.

### DIFF
--- a/docs/mixins.md
+++ b/docs/mixins.md
@@ -113,14 +113,3 @@ app({
   ])]
 })
 ```
-
-## Presets
-
-Group mixins under the same category to create a preset. Then use it like any other mixin.
-
-```jsx
-const DevTools = () => ({
-  mixins: [Logger, Debugger, Replayer]
-})
-```
-
--- a/src/app.js
+++ b/src/app.js
@@ -2,26 +2,24 @@ var globalInvokeLaterStack = []
 
 export function app(props) {
   var appState
+  var appView = props.view
   var appActions = {}
   var appEvents = {}
-  var appMixins = []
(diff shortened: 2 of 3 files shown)


Swizz:
This will change the way events are reduced by mixin too ?

In the case I am using a mixin `PromiseResolve` to resolve each action using a promise to a thunk call.
```
App:actions:fetch -> PromiseResolve:events:resolve -> App:events:resolve
```

jbucaran:
> @Swizz This will change the way events are reduced by mixin too ?

Before the order started with app, now app is last.

With this PR.

```jsx
const A = () => ({
  events: {
    anyEvent(state, actions, data) {
      console.log(data) //=> Data:
      return data + "A"
    }
  }
})

const B = () => ({
  events: {
    anyEvent(state, actions, data) {
      console.log(data) //=> Data:A
      return data + "B"
    }
  }
})

const C = () => ({
  events: {
    anyEvent(state, actions, data) {
      console.log(data) //=> Data:AB
      return data + "C"
    }
  }
})

const emit = app({
  events: {
    anyEvent(state, actions, data) {
      console.log(data) //=> Data: ABC
    }
  },
  mixins: [A, B, C]
})

emit("anyEvent", "Data: ")
```

lukejacksonn:
LGTM :shipit:

PR #335, opened 2017-08-21
Support for iterable rendering

NoobLad:
Enable the use of iterable protocal on children.
So it will support generator, immutable.js collections, etc...
Example : 
```jsx
import { List } from "immutable.js"
const technos = List(["Hyperapp", "Preact", "Inferno"])
const technosList = () => <ul>
  {
    technos.map(name => <li>{name}</li>)
  }
</ul>
```

### DIFF
--- a/src/h.js
+++ b/src/h.js
@@ -1,9 +1,13 @@
 var i
 var stack = []
+var ITERATOR_SYMBOL = typeof Symbol === 'function' && Symbol.iterator || '@@iterator'
 
 export function h(tag, data) {
   var node
   var children = []
+  var iterator
+  var next
+  var subStack
 
   for (i = arguments.length; i-- > 2; ) {
     stack.push(arguments[i])
@@ -14,6 +18,13 @@ export function h(tag, data) {
       for (i = node.length; i--; ) {
         stack.push(node[i])
       }
+    } else if (node != null && typeof node === "object" && typeof node[ITERATOR_SYMBOL] === "function") {
+      iterator = node[ITERATOR_SYMBOL]()
+      subStack = []
+      while(!(next = iterator.next()).done) {
+        subStack.push(next.value)
+      }
+      stack.push(subStack)
     } else if (node != null && node !== true && node !== false) {
       if (typeof node === "number") {
         node = node + ""
--- a/test/h.test.js
+++ b/test/h.test.js
@@ -100,3 +100,59 @@ test("components", () => {
     ]
   })
 })
+
+test("iterables children", ()  => {
+  function* iterable() {
+    yield "Hello"
+    yield " world"
+    yield "!"
+  }
+
+  const expectedSimple = {
+    tag     : 'div',
+    data    : {},
+    children: [
+      'Hello',
+      ' world',
+      '!'
+    ]
+  }
+
+  expect(h("div", {}, iterable())).toEqual(expectedSimple)
+
+  function* nestedIterable() {
+    yield h(
+      "div",
+      {},
+      iterable()
+    )
+  }
+
+  const expectedNested = {
+    tag     : 'div',
+    data    : {},
+    children: [
+      {
+        'children': [
+          'Hello',
+          ' world',
+          '!',
+        ],
+        'data'    : {},
+        'tag'     : 'div',
+      }
+    ]
+  }
+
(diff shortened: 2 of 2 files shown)


jbucaran:
@NoobLad This is a good start, thanks! Can you explain what you are doing and why? If you can share a more elaborate code example or pseudo code that would help speed things up as well.

/cc @dodekeract @jamen @andyrj

NoobLad:
Javascript from ES6/ES2015 has iterable protocol
It means that there is a way to make things iterable without being an array.
But now more and more libraries add support for it, immutable.js is one of them.
In immutable for example they have done it so you can use it without having to call `.toArray()` which cost a lot if your are using Immutable.js Lists in React slowing down renders.
For example for-of work with iterables, when you do a for-of on an Array it uses it's iterator implementation too, string implements iterator protocol too.

Since iterable protocole use _Symbol.iterator_ to be compatible with old browser a common fallback is to us _'@@iterator'_

a complete example without this PR : 
```
import {List} from "immutable"

app({
    state: {
        todos: List()
    },
    view(state, actions) {
        return <div>
            <h1>Todos</h1>
            <form onsubmit={actions.todos.save}>
                <input type="text" name="todo"/>
                <button>Add</button>
            </form>
            <ul>
                {
                    state.todos.map(todo => <li>
                        {todo}
                    </li>)/* without the PR needs to add .toArray() here */
                }
            </ul>
        </div>
    },
    actions: {
        todos: {
            save(state, actions, event) {
                event.preventDefault()
                const todo = event.target.todo.value
                return {
                    todos: state.todos.push(todo)
                }
            }
        }
    }
})
```
In the code the effect is minor but for performance the difference is huge, since you dont need to transform the data structure at all.

You could use generators too.

The part i'm not really confident is : 
```
      subStack = []
      while(!(next = iterator.next()).done) {
        subStack.push(next.value)
      }
      stack.push(subStack)
```
we can't go from end to start with iterators as they are, so i create an intermediate array...

andyrj:
Can you help me understand why that change is needed for h.  In your example what h ends up with is an array of h("li", ...).  So I don't think it hits the code in this PR from this example.  I think to hit that code, the List itself without jsx/h, would need to be one of the children or what am I misunderstanding here?  And this just appears to push equivalent logic of that toArray() into h, so I am not 100% sold on this solving any performance issues, but it does offload where the conversion occurs, which would be more convenient for immutable.js users that want to use immutable.js structures as children.  I am just not sure that is going to be frequent enough to warrant adding this code to core.  

My opinion at this point would be adding this to a replacement h library is preferable to adding this to core.

NoobLad:
I tried a quick and dirty version of what it could be : https://gist.github.com/NoobLad/2a14649b84cddea8b15cf90fe22c260a

If you want to play with it : https://codepen.io/nooblad/pen/MvGgVK

dodekeract:
With generators getting more and more popular, I think this does fit into core. Having to completely copy `h()` is not a "viable alternative" IMHO.

jbucaran:
> @NoobLad In the code the effect is minor but for performance the difference is huge, since you dont need to transform the data structure at all.

What do you mean by that?

jbucaran:
Hi all, I am not going to support iterables in core, but I'd definitely like to see libraries adding this on top of h. I'm going to close here just because this will not be merged, but I'd like to continue the discussion.

@NoobLad Do you have an example that takes advantage of these changes, but does not use immutable.js?

lazarljubenovic:
>  Do you have an example that takes advantage of these changes, but does not use immutable.js?

@jbucaran Well, JS has other iterables other than `Array`, such as `Map`, `Set`, `WeakMap`, `WeakSet`, `FormData`.

> In the code the effect is minor but for performance the difference is huge

@NoobLad Is it? You still create an axiliary array yourself and keep calling `.next` until you've iterated over everything. How's that different than calling `.toArray()` or using spread operator?

NoobLad:
@lazarljubenovic Using spread operator will do the exact same thing.
I've just run tests. It doesn't boost anything...
So i should have re-checked before with real values. So it's completely false, it has no performance value. Sorry about that.

@jbucaran Yes, i can. But in vanilla javascript it has low value, since Map/Set/FormData interface is poor for this use case (you'll have to manipulate an array at some point). After talking about other features with everyone, it's probably not that useful XD.

In the end, it's about writing a more concise code with some external libraries.

PR #338, opened 2017-08-23
Call thunks inside update function.

jbucaran:
Before we would call thunks inside wrapped actions, now thunks are called inside the update function.

This allows you to pass also a thunk to update and it will be called.

```js
actions: {
  someAction(state, actions, data) {
    return update => {
      update(update => {
        update(update => {
          update(update => {
            update({ value })
          })
        })
      })
    }
  }
}
```

Either way, thunks are called immediately after an action is called.

This also reduces the bundlesize by 5 bytes ;)

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -36,10 +36,7 @@ export function app(props) {
       if (typeof action === "function") {
         actions[key] = function(data) {
           emit("action", { name: name, data: data })
-
-          var result = emit("resolve", action(appState, appActions, data))
-
-          return typeof result === "function" ? result(update) : update(result)
+          return update(emit("resolve", action(appState, appActions, data)))
         }
       } else {
         initialize(actions[key] || (actions[key] = {}), action, name)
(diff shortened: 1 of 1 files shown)


Swizz:
Only looking at the source code. The `resolve` will stay unchanged isnt it ?

jbucaran:
@Swizz Yep.

@Andyrj Here is how I used it. I was using `Events.resolve` to check the return type of an action and if it is a Promise, wrap it in a thunk, effectively allowing actions to return a promise that resolves to a partial state in order to update the state. I've shown how this is possible before. See [the docs](https://github.com/hyperapp/hyperapp/blob/master/docs/events.md#resolve).

Then I imagined, what if I want a promise that resolves to a thunk? I don't know why I would want it, but _thunks_ is a thing is hyperapp so I wanted to try it out... That'd would be impossible in the current design because the promise is wrapped inside a thunk and we can't resolve to another thunk.

With this, the update function is now able to process thunks and will in fact exhaust thunks of infinite depth.

The reason for this change is basically that I realized that I could call the thunk inside update instead of inside the action and by doing so make update arguably a bit useful. And at the same time cut some bytes.

andyrj:
@jbucaran alright sounds good, thank you for the explanation.

jbucaran:
Thank you all guys.

PR #339, opened 2017-08-27
Add ActionInfo to resolve and update events

okwolf:
In today's API the `ActionInfo` object is made available to the `action` event, but not to the `resolve` or `update` events, even though both of these had an original action that caused them to happen. It can be helpful to have access to the action that triggered a state update in larger/complex apps. It can also be useful to know the original action that was fired when deciding what to do when it resolves.

Now `update` has a new parameter tracking the action that caused it, including forwarding it on in the event of thunks/async. Also `emit` has a new parameter for the action that eventually led to this event, and it calls the callback with this action added. I'm open to rearranging these, but I wanted to start with a change that was backward compatible in adding the new feature.

Test coverage was updated including async. Docs were updated. The gzipped bundle grew by 19 bytes 🤷‍♂️

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -159,8 +159,8 @@ See [Events](/docs/events.md).
 <a id="load"></a>load(<a href="#state">State</a>, <a href="#actions">Actions</a>, <a href="#root">Root</a>): <a href="#vnode">VNode</a>
 <a id="render"></a>render(<a href="#state">State</a>, <a href="#actions">Actions</a>, <a href="#view">View</a>): <a href="#view">View</a>
 <a id="action"></a>action(<a href="#state">State</a>, <a href="#actions">Actions</a>, <a href="#actioninfo">ActionInfo</a>): <a href="#actioninfo">ActionInfo</a>
-<a id="resolve"></a>resolve(<a href="#state">State</a>, <a href="#actions">Actions</a>, <a href="#actionresult">ActionResult</a>): <a href="#actionresult">ActionResult</a>
-<a id="update_event"></a>update(<a href="#state">State</a>, <a href="#actions">Actions</a>, <a href="#state">State</a>): <a href="#state">State</a>
+<a id="resolve"></a>resolve(<a href="#state">State</a>, <a href="#actions">Actions</a>, <a href="#actionresult">ActionResult</a>, <a href="#actioninfo">ActionInfo</a>): <a href="#actionresult">ActionResult</a>
+<a id="update_event"></a>update(<a href="#state">State</a>, <a href="#actions">Actions</a>, <a href="#state">State</a>, <a href="#actioninfo">ActionInfo</a>): <a href="#state">State</a>
 </pre>
 
 #### CustomEvent
--- a/src/app.js
+++ b/src/app.js
@@ -35,8 +35,12 @@ export function app(props) {
 
       if (typeof action === "function") {
         actions[key] = function(data) {
-          emit("action", { name: name, data: data })
-          return update(emit("resolve", action(appState, appActions, data)))
+          var actionInfo = { name: name, data: data }
+          emit("action", actionInfo)
+          return update(
+            actionInfo,
+            emit("resolve", action(appState, appActions, data), actionInfo)
+          )
         }
       } else {
         initialize(actions[key] || (actions[key] = {}), action, name)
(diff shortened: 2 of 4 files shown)


jbucaran:
Hi @okwolf, thank you for thinking about this problem. It never occurred to me to modify emit to share ActionInfo with other events, although I had thought about this a lot.

I don't think we need to share ActionInfo with events.update, but I can understand why you'd want events.resolve to have it.

Having said that, you can still access ActionInfo from events.resolve right now if you push ActionInfo to an array inside events.action and pop it inside events.resolve.

I'll think about it and see if I can come up with a similar set of changes that satisfy your need, but keeps the API tidy.

okwolf:
@jbucaran thanks for the reply! For sync actions and updates, I agree that using an array as a stack would work without needing to change the API. However, once you have async actions that resolve or update out of order this won't work anymore. If having access to this information isn't a common enough use case then so be it.

jbucaran:
@okwolf I'd like to see a valid use case for sharing ActionInfo with events.update, so I didn't take that event into consideration, but yes, I'd really like to have ActionInfo with events.resolve. Because of how emit works, the current API is the best compromise I could make, but your idea of passing another argument to event is worth considering.

I've tried to implement the following in the past, but it is challenging because how do we share the updated ActionInfo with other listeners of the same event?

```js
events: {
  resolve(state, actions, { action, data }) {
    return data // <-- not ActionInfo
  }
}
```

okwolf:
@jbucaran the simplest possible use case I can think of for access to `ActionInfo` from `events.update` would be to implement a [redux-logger](https://github.com/evgenyrodionov/redux-logger) style mixin. Today we have access to the prev and next state, but we don't get the action that led to the state change if it was emitted asynchronously before the latest action.

andyrj:
@okwolf  have you tested using a stack and triggering actions async?  https://github.com/andyrj/hyperapp-redux-devtools/blob/master/index.js#L39   I haven't seen any issues with actions getting out of order...  but maybe I just haven't run into it yet.

okwolf:
@andyrj [this](https://codepen.io/anon/pen/xLaayy?editors=0010) is a simple example of three async actions completing at different speeds than the order they are called in (fast, slow, medium). The `firedActions` used in `events.update` seems to think they completed in medium, slow, fast order when you can clearly see from the state changes the actual order was fast, medium, slow.

jbucaran:
@okwolf Sorry, @andyrj was probably referring to events.action & events.resolve, not events.update.

okwolf:
@jbucaran I'm not sure that's the case since the code example from @andyrj shows using using a stack of `firedActions` in `events.update` for exactly the use case that my pen demonstrates a race condition exists with multiple simultaneous async actions.

The problem with using `events.resolve` for async actions is having the `ActionResult` as a thunk doesn't give you access to either the action that returned it nor any updates that happen inside it when called.

andyrj:
@okwolf you are right that is a bug in my package...  when I was writing that I asked about for this exact same feature, and thought the stack had solved it but didn't properly test async cases, thanks for making the PR.

jbucaran:
Correct, it makes sense to have ActionInfo inside the function, what I am not sure about is the API.

okwolf:
After some creative thinking it turns out there's a hacky solution without changing the API. You maintain two stacks of actions, one for the originally fired actions and another for actions that are about to trigger an update. In `events.resolve` sync actions go immediately onto the second stack but async actions are wrapped in an extra thunk that calls the original thunk and then calls back into the resolve code before calling the update of the new extra wrapper thunk. I'm not quite sure yet if this is ugly or elegant.

Take a look: https://codepen.io/anon/pen/RZYeqp?editors=0012

jbucaran:
@okwolf Very nice! 😄 

See: https://github.com/hyperapp/hyperapp/issues/340

jbucaran:
Here is a shorter version:

```jsx
const Logger = (emit, stack = []) => ({
  events: {
    action(state, actions, action) {
      // Start log.
      stack.push(action)
    },
    resolve(state, actions, result) {
      if (typeof result === "function") {
        const action = stack.pop()
        return update =>
          result(result => {
            stack.push(action)
            return update(result)
          })
      }
    },
    update() {
      const action = stack.pop()
      // End log.
    }
  }
})
```

jbucaran:
@okwolf Good to close?

okwolf:
@jbucaran close it. Thanks for posting the simplified version.

PR #343, opened 2017-08-27
Remove Mixinception from Docs

okwolf:
Currently mixins are not allowed within other mixins. There will be a separate flame war on the merits of changing that. This just updates the docs to reflect the current reality.


jbucaran:
Thanks @okwolf! 🎉

PR #345, opened 2017-08-30
[DOCS] Fix "resolve" example for promises

SkaterDad:
Promise handling example wasn't returning a function, and would not work.

Used this code last night in a project, and it works nicely!

Though I do have a question.  Why return the `&& result`?  The observable example doesn't do that.

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -76,7 +76,7 @@ app({
   events: {
     resolve(state, actions, result) {
       if (result && typeof result.then === "function") {
-        return result.then(update) && result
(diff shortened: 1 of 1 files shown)


jbucaran:
> @SkaterDad Why return the && result? The observable example doesn't do that.

To return the Promise back to the caller and so the caller can do:

```js
action.doThis().then(...)
```

Does that make sense?

SkaterDad:
Ah, okay!  Thanks for the explanation.

PR #347, opened 2017-08-31
Allow update to take a reducer.

jbucaran:
A reducer is a function that takes the state and returns a new state or a reducer.
    
```js
update(state => {
}
```    
    
This allows you to get the current & latest state inside the callback of some async call.
    
The alternative is using the state that was passed to the action that originated the async call. The problem is that this reference to the state may have gone out-of-date if other actions run and finished during the async process and changed the state.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ yarn.lock
 package-lock.json
 
 # Misc
-sandbox
+example
 node_modules
 coverage
 *.xml
--- a/docs/api.md
+++ b/docs/api.md
@@ -21,6 +21,7 @@
   - [Root](#root)
 - [Emit](#emit)
 - [Update](#update)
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -8,7 +8,7 @@ yarn.lock
 package-lock.json
 
 # Misc
-sandbox
+example
```
okwolf: We're not using either of these are we?

jbucaran: @okwolf I am locally, but it's not committed to the repo.

jbucaran:
Updated docs and added 100% coverage.

I'll merge here and now work on a substantial docs update.

jbucaran:
Related https://github.com/zaceno/hyperapp-partial/issues/3

PR #355, opened 2017-09-08
#354 forward compatible prepare script

### LINKED ISSUE #354: Seems that we should use both `prepublish` and `prepare` for npm5 and npm3
As the title indicates, to be forward and backward compatible we should use both script keys.

```bash
hyperapp (master) $ npm install
npm WARN prepublish-on-install As of npm@5, `prepublish` scripts are deprecated.
npm WARN prepublish-on-install Use `prepare` for build steps and `prepublishOnly` for upload-only.
npm WARN prepublish-on-install See the deprecation note in `npm help scripts` for more information.
```

selfup:
closes #354 

@jbucaran Let me know if this makes sense!

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "bundle": "rollup -i src/index.js -o dist/hyperapp.js -m -f umd -n hyperapp",
     "minify": "uglifyjs dist/hyperapp.js -o dist/hyperapp.js --mangle --compress warnings=false --pure-funcs=Object.defineProperty -p relative --source-map dist/hyperapp.js.map",
     "prepublish": "npm run build",
+    "prepare": "npm run build",
     "format": "prettier --semi false --write 'src/**/*.js'",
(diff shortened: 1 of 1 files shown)


selfup:
Also, why is CI not running? 🤔 

Is it because I assigned it to you?

selfup:
Turns out @lukejacksonn mentioned that Travis and GitHub Webhooks have been in a weird state as of late. I have a friend who's Webhook based project is currently not working. So this should resolve on its own.

I'll amend this commit in a few hours to re-trigger a push event 👍

jbucaran:
@selfup No idea, but LGTM.

When do you think it will be safe to remove `prepublish` from package.json?

selfup:
That is a good question! I assume people contributing to hyperapp will be using npm5 but some might still use npm3. Then again you are the one that publishes 😂

jbucaran:
@selfup Alright then, so let's keep it for a while.

PR #358, opened 2017-09-10
Fix components documentation example.

lambdatastic:
`todo` in this example is undefined as written, due to the destructuring. This patch fixes it.

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -39,7 +39,7 @@ If you don't know all the properties that you want to place in a component ahead
 ```jsx
 <ul>
   {state.todos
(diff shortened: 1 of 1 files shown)


jbucaran:
@lambdatastic Looks good, thanks! 👋

selfup:
@lukejacksonn @jbucaran 

Does Travis CI only trigger when JS files change? It should fire on any change 🤔 

Example here: [I only changed the README and Travis CI still triggered](https://github.com/selfup/syncr32/pull/6/files)

I'll open up an issue (maybe we hit a limit of some sort?)!

Issue #360, opened 2017-09-10
No Protections Against XSS Attacks Using innerHTML


okwolf:
If someone were building an app with Hyperapp and wanted to allow for formatting of user-supplied data with HTML instead of Markdown or other such syntax, using the `innerHTML` attribute adds a XSS vulnerability that we should probably at least warn about if not prevent entirely.

[Example pen](https://codepen.io/okwolf/pen/XavOKy?editors=0010).

In React this would be prevented with the message:

```
Warning: Directly setting property `innerHTML` is not permitted. For more information, lookup documentation on `dangerouslySetInnerHTML`.
```

As the warning implies, the [dangerouslySetInnerHTML](https://facebook.github.io/react/docs/dom-elements.html#dangerouslysetinnerhtml) attribute is React's way of allowing the user to still do this while acknowledging how dangerous it is.

zaceno:
Hmm... My personal feeling: It's not hyperapp's responsibility to protect users from shooting themselves in the foot when they chose to work directly with the DOM. Perhaps a note somewhere in the lifecycle event docs (which is where we expose the DOM directly)

React's approach sounds a lot like it started as a secure-coding-policy internally at Facebook, which they decided to just build into "their" framework as a convenience for themselves.

(Maybe I'm being a little cynical. But really -- innerHTML is not any more "dangerous" then anything else. As always, you just have to know what you're doing when dealing with user supplied data)

okwolf:
@zaceno I'm all for not trying to idiot-proof the API, but it is a very real concern that beginner developers might not be aware of. At a minimum I think we should document the dangers of this gotcha if we don't add any special handling. Especially for users converting from React who rely on the safety features it provides.

selfup:
Yea in the docs I would follow the same warning as Vue does:

https://vuejs.org/v2/api/#v-html

Simple, clear warning 👍

jbucaran:
So, what solutions have we come up with so far?

- Disable innerHTML and give you dangerouslySetInnerHTML to do the same.
- Document using innerHTML is dangerous.

...

okwolf:
@jbucaran we could also choose to disable `innerHTML` and not provide any work-around, but that would probably be too disruptive.

My vote right now is for documenting along the lines of: **Here Be Dragons** 🐲

SkaterDad:
My vote is documentation.  No need to add the extra complexity in my
opinion.

It's a good candidate for "Dev mode" console warnings in the future also.

[...]

PR #361, opened 2017-09-11
Docs Update

okwolf:
Did this in many parts in case any are not desired and should be reverted:
- ~~[Markdownlint](https://github.com/DavidAnson/markdownlint) to sanity check the content for consistency as part of the build~~
- Update inconsistent use of languages in code fences. If we want to pick one standard we could use `jsx` for all JavaScript code since it's a superset of `js`. For now I picked the one that fits closest based on content used.
- Some additional helpful links on Getting Started page.
- ~~Follow GitHub's [standard for the code of conduct](https://help.github.com/articles/adding-a-code-of-conduct-to-your-project).~~
- Some OCD changes where there were deviations in grammar, punctuation, and use of `inline code`.
- A warning about using `innerHTML`. See #360.


jbucaran:
@okwolf Looks good.

Pet peeve: Don't split the CoC and CONTRIBUTION.md

jbucaran:
2nd pet peeve: Remove markdown lint and .markdownlintrc.

okwolf:
> @jbucaran Don't split the CoC and CONTRIBUTION.md

Do you disagree with https://opensource.guide/code-of-conduct? I noticed they also recommend linking it from CONTRIBUTING or README.

okwolf:
@jbucaran reverted the Markdownlint changes.

jbucaran:
My views on CoC are still very mixed. I originally added the CoC because there was a user trolling hyperapp, but since that did not stop the person I blocked them.

I don't think a CoC will make people suddenly behave well, and if it does, then I'd be wary of those people (but then how do you tell!) 

It's not as if humans need a moral code to be moral.

jbucaran:
So, I'd say let's not have a CoC for now. If we find a need for it in the future, then we can add it.

okwolf:
> @jbucaran So, I'd say let's not have a CoC for now. If we find a need for it in the future, then we can add it.

I reverted my changes for CoC.

okwolf:
@jbucaran who else should we assign as reviewers?

jbucaran:
Ref: https://github.com/markdownlint/markdownlint/issues/179

jbucaran:
@okwolf Whoever wants to take a stab. I'll merge later today.

jbucaran:
Thanks @okwolf! 💯🎉

Issue #362, opened 2017-09-12
Support for components that return arrays


infinnie:
We often have use cases where our components need to return arrays, e.g.
```jsx
var SomePageComponent = function (props) {
    props = props || {};
    return [<Header />,
        <SideNav />,
        <div class="panel panel--main">...</div>
    ];
};
```

However, it isn’t easy for us to really do so in HyperApp. I’ve written a polyfill for exactly that (using jQuery, along with other polyfill code for backward compatibility):
```js
function patch(parent, element, oldNode, node, isSVG, nextSibling) {
    if ($.isArray(node)) {
        return $.map(node, function (t, i) {
            return patch(parent, element, i === node.length - 1 ? oldNode : null, t, isSVG, nextSibling);
        });
    }

    if ($.isArray(element)) {
        var remaining = element.slice(1);
        element = element[0];
        $(remaining).remove();
    }

    // original code ...
}
```

I do not quite know whether there would still be issues with components returning arrays.

Will HyperApp officially support that?

leeoniya:
in my experience, robust, optimal and fast dom reconciliation of fragments is non-trivial. i'd be interested to see what you guys come up with and how much code it adds.

https://github.com/developit/preact/pull/703#issuecomment-322791946

jbucaran:
@infinnie @leeoniya By the way, I am sure you all already know this, but hyperapp does not aim to be compatible with React in any way, so IMO there is no pressure to introduce this feature. Or at least it is not remotely as critical as it is in preact.

@zaceno @lukejacksonn @selfup @andyrj How do you guys feel about the idea of returning arrays from components?

leeoniya:
fragments are not about React parity; they do have actual uses in both rendering optimizations (sCU the wrapper separately from the children) and flexbox cases as mentioned in the original React issue [1]. domvm also has a user that had a use-case of expanding a table row to one or multiple adjacent rows without having to diff/reconcile all siblings that could number in the thousands [2].

[1] https://github.com/facebook/react/issues/2127#issuecomment-90171492
[2] https://github.com/leeoniya/domvm/issues/124

jbucaran:
@leeoniya All I am saying is that if you have a component in React written this way, it will not work on Preact unless they support this new feature in React 16. _That_ problem does not exist in Hyperapp.

andyrj:
[...]

PR #366, opened 2017-09-15
Make it possible for top-level components to return arrays.

infinnie:
See also #362 .


JorgeBucaran:
@infinnie Thanks for your work on this interesting option! 

It's good to point out that Hyperapp **components can return arrays**, we just can't from the root element.

I think the increase in complexity introduced by this change does not justify the overall usefulness of the feature. Not to say this feature is not useful, it's just not very useful _in_ Hyperapp-land where our entire app is just "one big component".

Issue #34, opened 2017-02-06
Browser support? 👾


jbucaran:
What browsers should we aim to support?

**>= IE11** ?
**>= IE10** ?

dodekeract:
IE10 would be great. That'd allow me to use this. ;)

jbucaran:
Yeah, that makes sense.

dodekeract:
@jbucaran I used this for an app I had to write for university. I'll check if anything doesn't work properly on it. Would be nice if it's an official goal to support IE10 though.

jbucaran:
Not a lot of feedback here. I think IE11 is a must-support, but I'll aim at IE10.

VinSpee:
Hi,

I realize this issue is old, but I still don't see any out-right declaration of browser support in the docs. Should I assume IE10+?

jbucaran:
@VinSpee Correct! :)

okwolf:
@jbucaran this should probably make its way into the docs for clarity.

jorgebucaran:
###### Summary

IE10 and IE11 are still officially supported, but you may need to add your own polyfills to fix some odd issue here and there. #690, https://github.com/jorgebucaran/superfine/pull/140

<img width="957" alt="screen shot 2018-08-31 at 16 07 43" src="https://user-images.githubusercontent.com/56996/44897833-0c3e8d80-ad38-11e8-8fbe-b9922190c5d0.png">

PR #372, opened 2017-09-18
add browser support table to readme

VinSpee:
support table includes latest FF, Chrome, IE 10 & 11, Edge, and Safari.

this was mentioned in #34


README.md
```diff
@@ -33,6 +33,12 @@ app({
   }
 })
 ```
+## Browser Support
+Firefox | Chrome | IE   | Edge | Safari
+-------|---------|------|------|-------
+55 ✔️   | 61 ✔️    | 11 ✔️ | 15 ✔️ | 10 ✔️
```
dodekeract: Why Safari 10? There is no way that Safari 9 is worse than IE10.

VinSpee: I just went with latest everything except known exceptions, I'm sure it works in chrome 35 too. Does anyone have access to safari 9 to test?

jbucaran: @VinSpee Could we add this to the readme in the / after or before the installation instructions? 🤔

jbucaran: BTW, >=IE10  😄

VinSpee: 👍🏻 got it to 1 line now

> Could we add this to the readme in the / after or before the installation instructions?

My personal preference is to have it in the repo index - i find it's one of the first questions people will ask when looking for things like this. I'm happy to move it if you like however.

jbucaran: @VinSpee Yeah, I understand, but the reason I haven't added it until now is because I haven't been able to find a good spot for it yet.

I cam up with this line:

> Hyperapp supports all ES5-compliant browsers, >=IE10.

But I am not sure where to put in on the README that won't break the balance we have at the moment.

Swizz:
Maybe off topic but there is a way to automatically test browser support ?

JorgeBucaran:
BTW I am going to add a  new `overview.md` page to the docs with their own "Browser Support" section so this story has a happy ending.

jacktuck:
@Swizz There is https://saucelabs.com - they have nice badges too - you can see them on preact's README. They only have paid plans, though, as far as i can tell :(

Swizz:
@jacktuck I don't think we need something as complex as an automation browser testing.
I meant HA is under 300 LOC, there is no way to determine browser compatibility only by reading the code ?

JorgeBucaran:
Thanks @VinSpee! See feb3ce2.

👋

PR #376, opened 2017-09-18
Default mixin.events to empty object

sghall:
Cool library. Just taking a look at the source.  I don't think this matters from a code execution standpoint (you'll get an empty array of keys with [] or {}) but it caused me to do a double take.  I was looking at the logger mixin source https://github.com/hyperapp/logger/blob/master/src/index.js#L16 where events is an object to see how it gets handled and spotted this.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -30,7 +30,7 @@ export function app(props) {
   return emit
 
   function initialize(actions, withActions, lastName) {
-    Object.keys(withActions || []).map(function(key) {
+    Object.keys(withActions || {}).map(function(key) {
       var action = withActions[key]
       var name = lastName ? lastName + "." + key : key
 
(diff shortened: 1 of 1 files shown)


okwolf:
I'm ok with this. Although `Object.keys({}) == Object.keys([])` the former probably makes more sense than the latter.

sghall:
Suppose to be totally consistent on this would change this line too for the actions...
https://github.com/hyperapp/hyperapp/blob/master/src/app.js#L33

Just makes it clearer you're expecting an object. If you want to add that to the PR or I can add it tomorrow.

JorgeBucaran:
Hi @sghall,

Looks good to me. Can you also change: 
- https://github.com/hyperapp/hyperapp/blob/master/src/app.js#L33 ? 🙇 

I'll merge then.

sghall:
Done.  This one is ready to go.

JorgeBucaran:
Thanks @sghall! 🎉

PR #377, opened 2017-09-19
Fix Events definitions in the DTS

Swizz:
As we discussed on slack, here is a more relevant Events definition.

User will define custom Events following the Events interface and the App will both use the user Events defined interface and the DefaultEvents interface.

Side effect : Emit will not show `keyof Events` anymore in VS Code intelisense

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -73,21 +73,21 @@ export interface Actions<State extends Hyperapp.State> {
       ) => ActionResult<State>)
 }
 
-export interface CustomEvents<
-  State extends Hyperapp.State,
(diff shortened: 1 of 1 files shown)


JorgeBucaran:
> @Swizz Side effect : Emit will not show keyof Events anymore in VS Code intelisense

What was that?

Swizz:
This feature : 
keyof InteliSense

JorgeBucaran:
@Swizz We want to have that, but only for our own `Events` that extend `Hyperapp.Events`. Any way we can have it?

Swizz:
I think we dont need it at the moment. I will read a bit more about that and make a new change

Issue #373, opened 2017-09-18
How to implement shouldComponentUpdate to increase Performance of complex components?


rajaraodv:
In React, there is a lifecycle event called *shouldComponentUpdate*. It is called on every component and sub component before they get updated. If any of the component returns false, then the it kills re-rendering that component and all it's child components. 

The idea is simple, if the app-developer knows that there is no need to diff vdom any further due to some business case, they can simply ask/force React to stop vdom diffing.

See:
https://facebook.github.io/react/docs/optimizing-performance.html#shouldcomponentupdate-in-action

https://medium.com/modus-create-front-end-development/component-rendering-performance-in-react-df859b474adc

In Hyperapp, there is only app-level lifecycle (and not component-level or sub-component-level) and every component are stateless. So how to achieve that? Please provide an example.

andyrj:
just to keep links with this issue for reference, from slack.

https://addyosmani.com/blog/faster-javascript-memoization/

``js
/* app.js - at top of patch */
if (oldNode === node) {
  return element
}
```

rajaraodv:
Here is the  [Memoize React components ](https://medium.com/@planttheidea/memoize-react-components-33377d7ebb6c) article for stateless components.

SkaterDad:
Relevant links for elm's implementation, `Html.lazy`.  It's not keeping a cache of previously computed values, but in the v-dom diffing algorithm, they check that the function args are the same, and if so, return the existing node.

http://package.elm-lang.org/packages/elm-lang/html/2.0.0/Html-Lazy
https://github.com/elm-lang/virtual-dom/blob/master/src/Native/VirtualDom.js#L572-L585

rajaraodv:
Thanks @SkaterDad. That means the end users dont have to worry about Memoizing on their end and it's out-of-the-box! #sweet!

JorgeBucaran:
@SkaterDad We would need to change `h` for this though. Is it really worth it? 🤔 

I am leaning on no, but it would be nice to see if it's possible and what it looks like.

SkaterDad:
@rajaraodv elm doesn't apply that optimization by default.  You have to specifically wrap your function in Html.lazy.  It's not really memoization since there's no cache, but just skips reproducing the DOM nodes if the model passed to the view function was the same last time it rendered.

[...]

PR #378, opened 2017-09-20
add short circuit for patching memoized vnodes

andyrj:
this adds a short circuit to patch for memoized components...

example memoize:
```js
/*
* memoize.js
* by @philogb and @addyosmani
* with further optimizations by @mathias
* and @DmitryBaranovsk
* perf tests: http://bit.ly/q3zpG3
* Released under an MIT license.
*/
function memoize( fn ) {
    return function () {
        var args = Array.prototype.slice.call(arguments),
            hash = "",
            i = args.length;
        currentArg = null;
        while (i--) {
            currentArg = args[i];
            hash += (currentArg === Object(currentArg)) ?
            JSON.stringify(currentArg) : currentArg;
            fn.memoize || (fn.memoize = {});
        }
        return (hash in fn.memoize) ? fn.memoize[hash] :
        fn.memoize[hash] = fn.apply(this, args);
    };
}
```

Users would wrap their components/view with this method to skip patching vnodes which have not changed.

This PR satisifies the following issue: https://github.com/hyperapp/hyperapp/issues/373

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -190,7 +190,8 @@ function app(props, container) {
   }
 
   function patch(parent, element, oldNode, node, isSVG, nextSibling) {
-    if (oldNode == null) {
+    if (oldNode === node) {
+    } else if (oldNode == null) {
       element = parent.insertBefore(createElement(node, isSVG), element)
     } else if (node.type != null && node.type === oldNode.type) {
       updateElement(element, oldNode.props, node.props)


src/app.js
```diff
@@ -180,6 +180,10 @@ export function app(props) {
   }
 
   function patch(parent, element, oldNode, node, isSVG, nextSibling) {
+    if (oldNode === node) {
+      console.log("memoized...")
```
okwolf: I love `console.log` as much as the next guy, but perhaps it's not needed here?

test/app.test.js
```diff
@@ -67,3 +67,41 @@ test("optional view", done => {
...
+          hash += (currentArg === Object(currentArg)) ?
+          JSON.stringify(currentArg) : currentArg;
+          fn.memoize || (fn.memoize = {});
+      }
+      return (hash in fn.memoize) ? fn.memoize[hash] :
+      fn.memoize[hash] = fn.apply(this, args);
+  };
+}
+
+test("skips patching memoized vnodes", done => {
```
okwolf: Does this actually check that the same VNode is returned?

src/app.js
andyrj: was just to help me verify that it was actually running that memoize case, trying to rack down the uncovered line #266 in coverage report LOL

test/app.test.js
andyrj: trying to figure out how to write the test for that...

andyrj:
if anyone can help me see why line 265 is not covered after adding the un-connected conditional at the top of patch, it would be appreciated...  gonna have another look in the morning after some coffee.

test/app.test.js
JorgeBucaran: I think another approach to this would be to create a simple demo app with one memoized component and one component not. Then somehow benchmark it, and once we understand the problem well, take another shot at this?

src/app.js
```diff
@@ -181,7 +181,6 @@ export function app(props) {
 
   function patch(parent, element, oldNode, node, isSVG, nextSibling) {
     if (oldNode === node) {
-      console.log("memoized...")
```
okwolf: 👍

andyrj:
ok so I have screenshots of chrome devtools performance tab with non-memoized, memoized without patch shortcut, and memoized with patch shortcut.

Non-memoized:
no-memo-table-no-shortcut

Memoized-no-shortcut:
memoized-table-perf-no-shortcut

Memoized-with-shortcut:
table-memo-with-shortcut

You can see the JS (yellow) in the cpu time decreasing between the screenshots with each part of the optimization enabled...

This was made via this codepen, https://codepen.io/andyrj/pen/yzzmLo, which is updating state every 16ms to increment a counter with a table that is based on static data below it...  If this codepen is not sufficient to demonstrate the js overhead associated with a component that would benefit from memoization I don't really know what's going to satisfy a demo outside of a real app with expensive components.

test/app.test.js
pspeter3: You can call the memoized function twice and check for identity equality.

pspeter3:
What is required to make this approved?

JorgeBucaran:
@pspeter3 Just a little time. 🙏

andyrj:
There was a branch missing in coverage last I looked, and I still don't like the test I added...  Just couldn't think of a better one at the time.

lukejacksonn:
I was just going through issues and came from https://github.com/hyperapp/hyperapp/issues/373#issuecomment-341914817 to here.. when I first read this PR description (and graphs :100: ) I thought wow this is awesome, but assumed for some reason that we would need to include the memoize function in core.. but this is not the case, right? Just an equality check will suffice. Which will always return false unless you have wrapped your component with memoize in userland. If I am understanding this correctly then what have we got against the idea?

src/app.js
```diff
@@ -196,7 +196,9 @@ export function app(props, container) {
   }
 
   function patchElement(parent, element, oldNode, node, isSVG, nextSibling) {
-    if (oldNode == null) {
+    if (oldNode === node) {
```
JorgeBucaran: @andyrj If this is all we need, we can simplify this down a bit more as follows.

```js
if (oldNode===node){
    } else 
```

andyrj: Good call I wasn't taking the return at bottom into account for bytes.

andyrj:
@lukejacksonn ya the memoize doesn't need to be apart of core.  I just copy pasted that memoize() in comments above as an example someone could use.

src/app.js
JorgeBucaran: Okay, please rebase. 🙏

lukejacksonn:
@andyrj 👌 and although I appreciate not every app/developer will take advantage of this feature nor will you need to memoize most of the time.. it does make sense to enable memoizing for when components get really big (more precisely, when the cost of diffing args and cacheing becomes cheaper than a straight up re-render, right?). If this is all that it takes then I'd say it is worth the minimal extra bytes and added complexity!

JorgeBucaran:
Thanks @andyrj! 🎉 🌮 🎉

PR #384, opened 2017-09-22
Update typescript definitions for Component and h

Mytrill:
The VNode returned by a component doesn't (in general) contain the props that are passed to this component, so the return type should not be linked to Props.

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -23,7 +23,7 @@ export interface Component<Props> {
    *
    * @memberOf [VDOM]
   */
-  (props: Props, children: VNodeChild<{} | null>[]): VNode<any>
+  (props: Props, children: VNodeChild<{} | null>[]): VNode<{}>
 }
 
 /** The soft way to create a VNode
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -1,4 +1,4 @@
-export as namespace Hyperapp
+export as namespace Hyperapp;
 
 /** @namespace [VDOM] */
 
@@ -7,23 +7,23 @@ export as namespace Hyperapp
  * @memberOf [VDOM]
 */
 export interface VNode<Props> {
-  tag: string
-  props: Props
-  children: VNodeChild<{} | null>[]
+  tag: string;
+  props: Props;
+  children: VNodeChild<{} | null>[];
 }
 
 /** In the VDOM a Child could be either a VNode or a string
  *
  * @memberOf [VDOM]
 */
-export type VNodeChild<Props> = VNode<Props> | string
+export type VNodeChild<Props> = VNode<Props> | string;
 
 export interface Component<Props> {
   /** A Component is a function that return a custom VNode
    *
    * @memberOf [VDOM]
   */
-  (data: Props, children: VNodeChild<{} | null>[]): VNode<Props>
+  (data: Props, children: VNodeChild<{} | null>[]): VNode<any>;
 }
 
 /** The soft way to create a VNode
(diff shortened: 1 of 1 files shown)


hyperapp.d.ts
```diff
@@ -23,7 +23,7 @@ export interface Component<Props> {
    *
    * @memberOf [VDOM]
   */
-  (data: Props, children: VNodeChild<{} | null>[]): VNode<Props>
+  (props: Props, children: VNodeChild<{} | null>[]): VNode<any>
```
JorgeBucaran: @Mytrill Looks good. What does `XX<YY>` mean in TypeScript? 🤔 😅

selfup: I think it is similar to Rust where you have a vector of 32bit integers `Vec<i32>`

And in Go: `var a [10]int`

So in TS an Array (pretty much a Vector because it is a heap allocated array that has an unknown size at compile time) you would define an Array of numbers like: `Array<number>`

JorgeBucaran: @selfup Thanks! So, it means Array? 😄

okwolf: @JorgeBucaran I think it's a type parameter, or perhaps more like a higher-order type. `VNode` gets to decide what to do with `<any>` it can use that type for parameters/return values to functions it defines, or whatever it wants really.

selfup: @JorgeBucaran it is a way to define what a certain type can contain. I was just using an `Array` as an example to narrow down the scope of examining the syntax 😛

selfup: So like say you have a:

`Car` and a `BigTruck`

To narrow down scope, let's assume all `Car`s only run on gas and all `BigTrucks` only run on diesel.

We would say something like:

```
const myCar: Car<gas> = new Car();

const myBigTruck: BigTruck<diesel> = new BigTruck();
```

My examples might be awful but I hope this helps 😂

Swizz: This is called Generic : https://www.typescriptlang.org/docs/handbook/generics.html

This is more like C++ Template. 

In this case we give user the ability to use `Props` as its own type, this is the way we give to user to type a component, h, vnode props. Like this is possible in Vue or React.

```ts
interface Props {
  min: number
  max: number
}

const Slider: Component<Props> = (props, children) => WHATEVER
```

Here we want now props object to contain a `min`  and a `max` value both as a number and required.

selfup: This is the correct answer for the scope of this PR ^^

🎉 

Thanks @Swizz!

Swizz: The most simple example about Generics.

We want a function that take a value and return the same : `identity`. 

```ts
function identity(arg: any): any {
    return arg;
}
```
But here we will lose the given type, so this is not really an identity function because the type is not saved.

Using generic
```ts
function identity<T>(arg: T): T {
    return arg;
}
```

Now we have an `identity` function for a `T` type that take a `T` argument and return a `T` result where `T` is a given type.

`identity<number>` is the same identity function for `number`
`identity<string>` is the same identity function for `string`
`identity<MyType>` is the same identity function for `MyType`
etc...

Swizz:
So now lets talk about the purpose of this PR ^^'

You are right about this catch. I dont know what I thinked about at this time.
But may I ask you to use `VNode<{}>` instead of `VNode<any>` ? This will ensure VNode Props to be considered as object/interface ?

hyperapp.d.ts
Swizz: Please use `VNode<{}>`

PR #385, opened 2017-09-22
Introduce fragment state / state slices. 🍰

### LINKED ISSUE #333: [question] deep state, destructuring & action verbosity/DRY
hey guys,

i don't believe there is something like concise notation for deep destructuring:

```js
const x = ({a.b.c}) => c.foo;
```

most examples with hyperapp use a flat state, so they look simple. but the app size will grow rather quickly and become less DRY as you have to repeat the state structure in multiple places. adding just one level to the counter example results in a lot more repetitive action code, or any code that can return partial state (but must be aware of the entire ancestor state structure):
https://codepen.io/anon/pen/ZJrrGW?editors=0010

attempts to make it smaller [understandably] don't quite work:
https://codepen.io/anon/pen/MvQQje?editors=0010

thoughts?
thanks!

### LINKED ISSUE #357: onremove() should return a promise for actually removing the element
I wish the function `onremove()` were to return a promise to tell the library to actually remove the element, instead of our having to manually do so in our own implementation.

### LINKED ISSUE #374: Improve Mixins doc
Mixins are very versatile in Hyperapp. [Mixins doc](https://github.com/hyperapp/hyperapp/blob/master/docs/mixins.md#mixins ) are used to perform various things says the following: "Use mixins to encapsulate your application behavior into reusable modules, to share or just to organize your code." But there are not much examples of how to use it to:

1. Encapsulate application's behavior into reusable modules
2. Integrate with 3rd party (May be something [like this](https://github.com/lukejacksonn/hyperapp-firebase-auth/tree/master/src))

### LINKED ISSUE #387: Rendered view is appended to the `root` element instead of replacing its contents
I noticed that when a hyperapp app initializes, the rendered view is appended to the `root` element, leaving any existing content in place. I kind of expected it to replace the existing content. Doing that would allow stuff like showing the user an informative message if the app code fails to run:

```html
<!-- index.html -->
<div id="root">
Please ensure you've enabled JavaScript and your browser is up to date.
</div>
```

```javascript
// main.js
app({
  root: document.getElementById('root'),
  /* ... */
});
```

Any chance the rendering behavior can be changed so the view completely replaces the content of the `root` element?

**EDIT**: Currently, I'm working around this by doing

```javascript
events: {
    load: function() {
      document.getElementById('root').innerHTML = ''
    }
  }
```

JorgeBucaran:
Hi everyone! 👋 

> ⚠️ This is a massive breaking change and **the last one** before 1.0. After this is merged and details are sorted, I'll publish 1.0 and then we can celebrate! 

## What's new? 

This PR bids farewell to the following concepts/features:

- Mixins (Close #374)
- Events & Custom Events
- `emit`

and adds:

- Hydration **by default** (no more append by default case, fix #387)
- State/Action Slices
- `onremove` with Thunks

### Fragmented State & Actions

Fragmented means your actions now receive a fragment (think _slice_ 🍰) of the state tree that corresponds to the leaf/namespace in which they are declared.

```js
actions: {
  hello(state) {
  // The state is the global `state`.
  },
  foo: {
    bar: { 
      howdy(state) {
      // The state is: `state[foo][bar]`
      }
    }
  }
}
```

With this feature, you can update deeply nested state more easily. Close #333.

For example, if you had the following state:
```js
state: {
  foo: {
    bar: {
      value: 0,
      anotherValue: 1
    }
  }
}
```

...and wanted to update `value`, you'd end up updating an entire record (siblings too), since there is no way to single out `value` from a nested state tree.

In other words, you'd need to write something like this in order to be inmutable.

```js
actions: {
  updateValue(state) {
    return {
      foo: {
        bar: {
          value: state.foo.bar.value + 1,
          anotherValue: state.foo.bar.anotherValue
        }
      }
    }
  }
}
```

With fragmented State & Actions, it's possible to update `value` immutably and easily:

```js
state: {
  foo: {
    bar: {
      value: 0,
      anotherValue: 1
    }
  }
}
```

And the corresponding action:

```js
actions: {
  foo: {
    bar: {
      updateValue(state) {
        // State is `state[foo][bar]`
      }
    }
  }
}
```

Here is contrived, but complete example using components.

```jsx
const counter = {
  state: {
    value: 0
  },
  actions: {
    up(state, actions) {
    }
  }
}

function Counter(props) {
  return (
    <main>
      <h1>{props.value}</h1>
      <button onclick={props.up}>1UP</button>
    </main>
  )
}

app({
  state: {
    counter: counter.state
  },
  actions: {
    counter: counter.actions
  },
  view: (state, actions) => (
    <Counter value={state.counter.value} up={actions.counter.up} />
  )
})
```

### Life after events

What's life going to look like after `events.load` is gone?

```jsx
const actions = app({})

// Register global DOM event listeners using document.addEventListener
// Do things you would normally use events.load for here.

actions.doThat()
```

### Hydration by default

Before to enable hydration you'd return a hydrated node from `events.load`. Now that events are no more, you just sit down and do nothing. We hydrate your stuff for you. 

How? Well, we just check if there are children in your root and we assume you rendered them on the server. If that's not the case you will need to provide a different root element that you own. 🎉 

### `onremove` with thunks

You don't have to remove elements inside `onremove` handlers by yourself now. Return a function from `onremove` to access a remover `remove` function and call it when you are done with your business. Close #357 

```js
function AnimatedDiv() {
  return <div onremove={element => 
     remove => fadeout(element).then(remove)}
  />
}
```

### Life after Mixins

Mixins introduce implicit dependencies when abused, and they are easy to abuse.

Say you have two mixins, `mixinA` and `mixinB`. In one or more of `mixinB`'s actions you call an action exposed by a `mixinA`. This is possible, because currently, you can call any action from within an action. After this PR, you will be able to call only actions inside the same namespace.

So, in this way, `mixinB` depends on `mixinA`. This is an implicit dependency. If you remove `mixinA`, `mixinB` will break. Not to mention, how hard it is to test and debug `mixinB` and how opaque this relation is.

But with fragmented state and actions, what harm can mixins cause? Good question! The answer is that we could actually keep them, since fragmented state and actions would make inter-mixin communication _impossible_.

```js
const mixinA = {
  state: {
    A: {
      value: 0
    }
  },
  actions: {
    A: {
      setValue(state, actions, value) {
        return { value }
      }
    }
  }
}

app({
  mixins: [mixinA]
})
```

The problem with writing code this way is that authors are forced to define the mixin's namespace and nest their state, which is prone to name collisions and complicated state tree updates. It also means you will have to check `mixinA`  docs or source code to figure out what was the namespace it uses.

Now without mixins, how do we encapsulate `mixinA`'s state and actions, which is what we are really after anyway?

Easy!

```js
const A = {
  state: {
    value: 0
  },
  actions: {
    setValue(state, actions, value) {
      return { value }
    }
  }
}

app({
  state: {
    A: A.state
  },
  actions: {
    A: A.actions
  }
})
```


okwolf:
@JorgeBucaran I fixed all the tests except for hydration. Coverage is almost complete again.

src/app.js
```diff
@@ -8,7 +8,7 @@ export function app(props, oldNode) {
   var appActions = {}
   var appState = props.state
   var appView = props.view
-  var appSubs = props.subscriptions
+  var appSubs = props.subscriptions || []
```
okwolf: Without the default there were errors leaving out `subscriptions`.

okwolf:
So to me, after this PR:

Hyperapp = [Elm](https://guide.elm-lang.org) (same architecture) - [Effects as Data](https://guide.elm-lang.org/architecture/effects) (built-in support for avoiding side effects in app code) + auto Redux [`combineReducers`](http://redux.js.org/docs/api/combineReducers.html) (scoped `state`) + [Redux Thunk](https://github.com/gaearon/redux-thunk) (return functions from actions for async) + Redux [`applyMiddleware`](http://redux.js.org/docs/api/applyMiddleware.html) (subscribe and respond to app lifecycle events)

src/app.js
```diff
@@ -1,59 +1,121 @@
...
+      withState &&
+      (withState = setPath(path, merge(partialState, withState), appState))
+    ) {
+      requestRender((appState = withState))
+    }
+    return appState
+  }
+
+  function getPath(path, source) {
+    return path.length === 0 ? source : getPath(path.slice(1), source[path[0]])
```
okwolf: @JorgeBucaran I think there's a bug here, or at least an inconsistency. Based on this code in `setPath` below(line 100):
```
source != null && name in source
  ? source[name]
  : path[1] >= 0 ? [] : {}
```
I would expect an update for a `state` fragment that doesn't exist yet to work, however this line will throw an error. One possible fix would be to write:
```
return path.length === 0 ? source : getPath(path.slice(1), source[path[0]] || {})
```

instead. I came across the issue because the condition in `setPath` is currently not covered by the tests and I was trying to add coverage.

src/app.js
```diff
@@ -1,104 +1,140 @@
...
-      (appEvents[name] || []).map(function(cb) {
-        var result = cb(appState, appActions, data)
-        if (result != null) {
-          data = result
-        }
-      }),
-      data
-    )
+  function getPath(path, source) {
+    return path.length === 0 ? source : getPath(path.slice(1), source[path[0]])
```
okwolf: @JorgeBucaran I think there's a bug here, or at least an inconsistency. Based on this code in setPath above (line 109):

```
source != null && name in source ? source[name] : {}
```

I would expect an update for a state fragment that doesn't exist yet to work, however this line will throw an error like cannot read property 'foo' of undefined from getPath. One possible fix would be to write:
```
return path.length === 0 ? source : getPath(path.slice(1), source[path[0]] || {})
```

instead. I came across the issue because the condition in setPath is currently not covered by the tests and I was trying to add coverage.

SkaterDad:
@JorgeBucaran @okwolf 
Can you clarify how the `subscriptions` are used?  The name is pretty confusing, since it appears more like middleware/hooks.  When I see "subscribe", I think of observables or attaching handlers to browser events.

Looking through the code in app.js, this is what I _think_ they do:

```js
subscriptions: [
  (state, actions) => {
    // this gets called on app load
    // replacement for "load" event
    return ({name, data}) => {
      // this gets called before an action is performed
      // replacement for "action" event
      return (result) => {
        // this gets called with the result of an action
        // replacement for "resolve" event
      }
    }
  }
]
```

JorgeBucaran:
@SkaterDad Edited my original post. The name is going to be **hooks**.

SkaterDad:
Thanks for clarifying!

lukejacksonn:
@JorgeBucaran out of curiosity.. why `hooks` and not `subscriptions` like elm?

JorgeBucaran:
@lukejacksonn Because while the concept is "similar" to Elm's subscriptions, I've come to learn that what we are doing is different. 

Our subscriptions are a simplified one-size-fits-it-all Elm's init function, their subscription model and our former events.

Unlike Elm, we are not inside a bubble (I mean that in a good way) and can simply use the platform directly; use `addEventListener` to subscribe to GlobalEvents, use WebSocket, setTimeout, setInterval, fetch, etc. We don't need a special module, we have no use for this abstraction. 

We also don't have ports. In Elm, subscriptions are also the high-level interface to the lower level ports.

Finally, our subscriptions are filling the gap for our former `action` and `resolve` events, which were very much "hooks".

VinSpee:
Have you considered cutting an alpha of this so we can easily use it in codepen/codesandbox/whatever to play around with? I've been copying and pasting so far 🤡

JorgeBucaran:
@VinSpee `npm i hyperapp@alpha`

test/actions.test.js
```diff
@@ -8,24 +8,50 @@ beforeEach(() => {
...
+            expect(state).toEqual({ baz: "minimal baz" })
             expect(data).toBe("foo.bar.baz")
-            done()
+            return { baz: "only the baz will do" }
           }
         }
+      },
+      fizz: {
+        buzz: {
+          fizzbuzz: () => ({ fizzbuzz: "fizzbuz" })
```
okwolf: So glad to see my creative names made it into the tests 🤣

okwolf:
@JorgeBucaran so is it just docs changes left before shipping this?

docs/state.md
```diff
@@ -25,3 +25,5 @@ app({
 ```
 
 The notion of representing the application state as a single source of truth is known as single state tree and the tree is manipulated using [actions](/docs/actions.md).
+
+## State Slices
```
okwolf: I feel like this section could be made shorter. Or longer, if you're into that...

dodekeract:
Is there any way to opt-out of this? It'll make Hyperapp completely useless for me/@attach-live. Otherwise, I'll be forced to fork Hyperapp before this commit is merged.

I don't see any way to work around this major limitation. The only remaining pattern would be to never, ever use namespaces and always use `actions.someRandomComplicatedName` again, like in `v0.0.1`.

JorgeBucaran:
Hey @dodekeract! Yes, you'll need to do some amount of rewriting, can you address what's specifically limiting for you?

lukejacksonn:
Or give an example of a mixin that you don't think can be re-written to work with the new arch? It doesn't have to be actual code if it is business sensitive.. but it would be useful to see what pattern you are using so that we can help you get from A to B.

SkaterDad:
I've been playing around with the alpha release on npm, and overall I'm liking the new fragmented/sliced state & actions.  This will make stuff that looks & feels like stateful components a lot easier.

I think @dodekeract has a point, that the stuff we were coding in mixins previously will not just work out of the box, because all of the `(state, actions)` were being passed the global `{ state, actions}`.  I'm not saying it's good or bad, but the architecture of our apps will need to change because of it.

Here's a quick example I just made on Webackbin. https://www.webpackbin.com/bins/-KvcpYjjax8hztTms_PF

You'll see 3 separate modules which could have been mixins previously.  The "challenge" module needed to post some alert messages on success or failure.  With the old mixins, I could have directly accessed `actions.alerts.new(...)`.  With this new scoped state/actions paradigm, I had to pass in those functions as parameters.  This is similar to how we're passing actions down into view components.  Making the external dependencies of the action explicit is a nice pattern, I think. :+1: 

I haven't tried replacing custom events yet, though...  That may be more painful.  @lukejacksonn Do you have an example of how we might replace the handling of custom events, like the 'route' event of the official router?  I was using it to fetch data.

VinSpee:
@SkaterDad thanks for sharing the really great example. I think it does a great job of showcasing the beauty of this pattern.

I think the missing like for you is the new `hooks` feature. Inside of a hook for `route`, you could listen for an action by name and trigger another action in response.


```js
hooks: [
  (state, actions) => {
    return ({name, data}) => {
      if(name === 'routeChange') {
        actions.fetchData(data.pathname)
     }
    }
  }
]
```

zaceno:
@VinSpee @SkaterDad 
... except now that hooks are out too that won't work either.

VinSpee:
@zaceno woah I missed that - a lot can happen in 7 hours

lukejacksonn:
What @VinSpee said.. awesome demo @SkaterDad 💯 and great question too! We are all in the same boat here, testing out new patterns. What I would have suggested 12 hours ago would have been something similar to what was suggested, using hooks. You have prompted me to think about it further now.. and routers in general (especially a component based routers).

If hooks disappear and app returns actions then you can change the apps state from outside of app by calling actions. In this particular case we want to subscribe to `window.popstate` events and call `router.go` every time one is triggered. How about something like..

```js

const routes = {
  '/adventure': {
    request: '/api/data',
    callback: call.backpack.fill
  }
}

const call = app({
  state: {
    backpack: {
      items: []
    },
    router: {
      path: '',
      fetching: false,
    }
  }, 
  actions:  {
    backpack: {
      fill: (state, actions, items) => ({ items })
    },
    router: {
      fetching: (_, __, fetching) => ({ fetching }),
      go: (state, actions, destination) => {
        const { request, callback } = routes[destination]
        request && actions.router.fetching(true) && fetch(request)
          .then(data => callback(data) && actions.router.fetching(false))
        history.pushState(null, null, destination)
        return { path: destination }
      }  
    }
  },
  view: (state, actions) =>
    h('main', {}, !state.router.fetching && [
      state.router.url === '/'
        ? h('h1', { onclick: actions.router.go('/adventure') }, 'Go on an adventure..'),
        : backpack.items.map(item => h('li', {}, item)),
    ])
})

const route = e => call.router.go(location.pathname)
window.onpopstate = route
route()
```

This is just off the top of my head late night pseudo code.. but hopefully you get the idea 😅

okwolf:
I wish we had kept this branch/PR focused on the original topic of state slices/fragments 🍰  so we could merge that update in separately while still discussing the merits of the other updates independently.

JorgeBucaran:
@okwolf Sorry, man. I agree, but now it's too late! :)

okwolf:
@JorgeBucaran water under the bridge, my friend. A lesson for future us ⏳

IamCarbonMan:
Wait so there's now no way to do middleware in hyperapp? events.action was removed and replaced with hooks, and now hooks are removed. Is action/state middleware going to be added back in any way?

JorgeBucaran:
@IamCarbonMan With HOAs. 

HOAs were not removed, what we removed was some "special built-in HOA support" that let you write HOAs in a particular way that, after careful consideration, was not any better (in fact worse and less flexible) than the way you would write a higher order function of any function in JavaScript.

IamCarbonMan:
@JorgeBucaran so I'm guessing the suggested way to implement something like events.action is just to write an HOA that wraps each action with whatever logic you need?

JorgeBucaran:
@IamCarbonMan Correct, that's the way. The logger is being rewritten like that and here is the PR with the changes.

https://github.com/hyperapp/logger/pull/7

/ cc @OKWolf

IamCarbonMan:
That's actually really smart. Especially with the PoC terminal UI it really feels like Elm, you're slowly going from a virtual DOM library to a simple state machine that maps one piece of data (a JS object) to another (DOM or some other view). Really excited to see where this is going!

okwolf:
@IamCarbonMan shameless plug for my [`hyperapp-middleware`](https://github.com/okwolf/hyperapp-middleware) that you can use to build such HOAs.
😊  🔌

dodekeract:
> can you provide an example

I can't share the source code, but the app I'm writing is pretty complex since it needs to manage a lot of things like WebRTC, WebSockets, postMessage communication, localStorage, etc..

I need a reliable way to split code between multiple files and so far, mixins were perfect for this. I could call stuff like `actions.message.send()` from any mixin that needed to communicate via `postMessage` with a different iframe. I don't see any way to do this without basically using a giant hacky workaround that doesn't fit into the current hyperapp usage at all. It'd probably be easier to just switch to Preact and write a custom state management library that imitates hyperapp pre-merge at this point.

zaceno:
@dodekeract While i completely agree with the need for modules  to be able to communicate beyond top-down, I disagree with the pattern of calling an action directly in another module (other than top down, which is ok), because of the dependency mess it can create. (Like what if you rename  `message` module to `com`, how many places will you need to change that?). 

This is why the custom events of yore were so valuable:  They allowed dependency-free cross module communications. 

Of course, you know your situation better than me -- but from what you've described and If I were you, I would rather integrate an event bus and refactor all `actions.someothermodule.xxx(...)` to `emit` calls. Not just for the sake of sticking with hyperapp, but more importantly to decouple my modules.

dodekeract:
@zaceno I'm not needlessly tying modules together. The whole app needs to work with all other parts of the app and there is no sensible way to do it differently, at least none that I'm aware of.

Converting actions to `emit` is, IMHO, exactly the same with a different implementation. It doesn't make them any less connected.

> how many places will you need to change that

For `actions.message.send`, 19 places. It's one of the "core" actions I have that are needed to connect the apps across different iframes, and no, it can't be easily removed from hyperapp as it contains logic that accesses the state, and other actions. After all, the `postMessage` channel isn't trusted in my case, so there needs to be some crypto applied to it and I also have async/await support for `postMessage` actions.

I completely agree that it would be better to have no dependencies between mixins, but honestly, that's unrealistic in my case. I outsource everything I can into as small/well-defined as helpful mixins. This works quite well for this particular project. The flexible mixin system is one of the reasons why I decided to upgrade hyperapp instead of porting the app to Preact when breaking updates came out. It was helpful, and still is (as I didn't upgrade yet, for obvious reasons).

IamCarbonMan:
@dodekeract so if there was a way to continue using mixins, with the same API as before, using the latest hyperapp, would you be able to make use of that? It should be trivial to reimplement mixins using an HOA.

dodekeract:
@IamCarbonMan if there's something that passes the global actions/state to all actions then all problems I have with the recent changes should be solved, yes.

The thing I don't like about using a HOA for this is that dropping official support means and outsourcing it means that it won't be considered for future API changes. So, it might become more and more hacky/impossible over time.

IamCarbonMan:
That is true. It would probably be advisable to try and migrate the project
to the newer API or simply stay on the old API.

dodekeract:
@IamCarbonMan I'm pretty sure it's practically impossible to do this project in the new API. Of course, I could always completely circumvent all of hyperapp's state management solutions, but then I might as well write my own state manager. 🙂

> stay on the old API

Well, for now that's the only thing I can do. It's not a long-term solution though and this'd probably lead to me having to hard-fork hyperapp.

JorgeBucaran:
> @dodekeract The thing I don't like about using a HOA for this is that dropping official support and outsourcing it means that it won't be considered for future API changes.

This is incorrect. We didn't drop official HOA support, just for the same reason we can't drop official support for the `if` statement. You could also create HOAs before, we just never explored that path. 

---

I am pretty sure that if you were using mixins for code organization, it should be relatively easy to port everything to the new API and I encourage you to try it. On the other hand, you were abusing mixins to take advantage of implicit dependencies, then I am afraid you may need a HOA. 😉 

I suggest you use this opportunity to improve your application architecture and adopt the new API.

zaceno:
@dodekeract I didn't meant to imply that you were carelessly creating a messy structure of mixins. Hope it didn't sound like that. I'm sure your design is carefully and well thought out.

There *are* principal differences between events and direct-action-calling. One of them is, that if the action you are calling goes a way (or is renamed), your whole app breaks then and there. If the event you are emitting no longer has listeners, you don't get the desired effect, but the rest of the app keeps running. Because emitting events creates no dependencies on listeners (and vice versa).

Mixins/modules communicating via events are loosely coupled.

The bottom line is the sameof course: your app doesn't work. In fact you might prefer your app to stop dead in its tracks any time something isn't as it should be (easier to catch and track down bugs).

But I prefer to have the loose coupling. To me it is in line with the idea of "optimize for deletablility" -- structure your app so you could take a big chunk out and rewrite it with minimal effect on the rest of the code.

olstenlarck:
Sounds and looks pretty fantastic in first look. But while i'm reading about the case of @dodekeract i'm thinking that he have a right. It is interesting scenario. Both architectures have Pros and Cons.

And if you are thinking to fork.. you may consider contacting me. I had few implementations born around when Hyperapp was born. I was reimplementing such architectures/apis several times to found my vision, soon after that `freactal` popped up, which was exactly what i was thinking. I'm trying to find time to release such library for few months, but i'm busy and sadly i messed my account and workflow totally recently :D 

But in any way, @JorgeBucaran great ideas is happening here. I'm just curious why you don't use `picodom` directly? And just wrap it which the state management?

PR #389, opened 2017-09-25
Update typescript definitions: Actions and h()

Mytrill:
I always found it a bit strange that the definition of Action is the definition of the implementation (only used once when declaring the implementation of actions) rather than the definition of the interface (used everywhere else).
So the aim of this PR is to rename ActionCallers -> Actions and Actions -> ActionsImpl

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -26,6 +26,11 @@ export interface Component<Props> {
   (props: Props, children: VNodeChild<{} | null>[]): VNode<{}>
 }
 
+export type VNodeChildren =
+  | Array<VNodeChild<{} | null> | number>
+  | VNodeChild<{} | null>
+  | number
+
 /** The soft way to create a VNode
  * @param tag       Either a tag name e.g. 'div'. Or a Component function
  * @param props     Any valid HTML atributes, events, styles, and meta data
@@ -36,7 +41,7 @@ export interface Component<Props> {
 export function h<Props>(
   tag: Component<Props> | string,
   props?: Props,
-  children?: VNodeChild<{} | null>[] | VNodeChild<{} | null> | number
+  children?: VNodeChildren
 ): VNode<Props>
 
 /** @namespace [App] */
@@ -54,23 +59,32 @@ export type ActionResult<State extends Hyperapp.State> =
   | null
   | void
 
-export type ActionCallers<
-  State extends Hyperapp.State,
-  Actions extends Hyperapp.Actions<State>
-> = {
-  [P in keyof Actions]: (
-    data: any
-  ) => ActionResult<State> | ActionCallers<State, Actions>
+export interface Action<State extends Hyperapp.State, Data> {
+  (data: Data): ActionResult<State>
 }
 
-export interface Actions<State extends Hyperapp.State> {
+export interface Actions<
+  State extends Hyperapp.State & Partial<Record<keyof Actions<State>, any>>
+> {
   [action: string]:
-    | Actions<State>
-    | ((
-        state: State,
-        actions: ActionCallers<State, Actions<State>>,
-        data: any
-      ) => ActionResult<State>)
+    | Actions<State[keyof Action<State, any>]>
+    | Action<State, any>
+}
+
+export interface InternalAction<
+  State extends Hyperapp.State,
+  Actions extends Hyperapp.Actions<State>
+> {
+  (state: State, actions: Actions, data: any): ActionResult<State>
+}
+
+export type InternalActions<
+  State extends Hyperapp.State & Partial<Record<keyof Actions, any>>,
+  Actions extends Hyperapp.Actions<State>
+> = {
+  [P in keyof Actions]:
+    | InternalAction<State, Actions>
+    | InternalActions<State[P], Actions[P] & Hyperapp.Actions<State[P]>>
 }
 
 export interface Events<
@@ -78,7 +92,7 @@ export interface Events<
   Actions extends Hyperapp.Actions<State>
 > {
   [action: string]:
-    | ((state: State, actions: ActionCallers<State, Actions>, data: any) => any)
+    | ((state: State, actions: Actions, data: any) => any)
     | undefined
 }
 
@@ -86,10 +100,10 @@ export interface DefaultEvents<
   State extends Hyperapp.State,
   Actions extends Hyperapp.Actions<State>
 > extends Hyperapp.Events<State, Actions> {
-  load?: (state: State, actions: ActionCallers<State, Actions>) => void
+  load?: (state: State, actions: Actions) => void
   resolve?: (
     state: State,
-    actions: ActionCallers<State, Actions>,
+    actions: Actions,
     result: ActionResult<State>
   ) => ActionResult<State>
 }
@@ -98,11 +112,11 @@ export interface View<
   State extends Hyperapp.State,
   Actions extends Hyperapp.Actions<State>
 > {
-  (state: State, actions: ActionCallers<State, Actions>): VNode<{}>
+  (state: State, actions: Actions): VNode<{}>
 }
 
 export interface Mixin<
-  State extends Hyperapp.State,
+  State extends Hyperapp.State & Record<keyof Actions, any>,
   Actions extends Hyperapp.Actions<State>,
   Events extends Hyperapp.DefaultEvents<State, Actions>
 > {
@@ -111,12 +125,12 @@ export interface Mixin<
 }
 
 export interface App<
-  State extends Hyperapp.State,
+  State extends Hyperapp.State & Record<keyof Actions, any>,
   Actions extends Hyperapp.Actions<State>,
   Events extends Hyperapp.DefaultEvents<State, Actions>
 > {
   state?: State
-  actions?: Actions
+  actions?: InternalActions<State, Actions>
   events?: Events
   view?: View<State, Actions>
   root?: HTMLElement | null
@@ -138,7 +152,7 @@ export interface Emit<
 }
 
 export function app<
-  State extends Hyperapp.State,
+  State extends Hyperapp.State & Record<keyof Actions, any>,
   Actions extends Hyperapp.Actions<State>,
   Events extends Hyperapp.Events<State, Actions>
 >(app: App<State, Actions, Events>): Emit<State, Actions, Events>
--- a/test/ts/index.tsx
+++ b/test/ts/index.tsx
@@ -1,38 +1,97 @@
 import { h, app } from "hyperapp"
 
-interface State extends Hyperapp.State {
+// Module 1
+
+interface Module1State extends Hyperapp.State {
   count: number
 }
 
-interface Actions extends Hyperapp.Actions<State> {
-  sub(state: State, actions: ActionCallers, value: number): Partial<State>
-  add(state: State, actions: ActionCallers, value: number): Partial<State>
+interface Module1Actions extends Hyperapp.Actions<Module1State> {
+  sub(value: number): Partial<Module1State>
+  add(value: number): Partial<Module1State>
 }
 
-interface ActionCallers extends Hyperapp.ActionCallers<State, Actions> {}
+const module1Actions: Hyperapp.InternalActions<Module1State, Module1Actions> = {
+  sub: (state, actions, value: number) => ({ count: state.count - value }),
+  add: (state, actions, value: number) => ({ count: state.count + value })
+}
+
+// Module 2
+
+interface Module2State {
+  count: number
+}
+
+interface Module2Actions extends Hyperapp.Actions<Module2State> {
+  reset(): Partial<Module2State>
+  add(value: number): Partial<Module2State>
+}
+
+const module2Actions: Hyperapp.InternalActions<Module2State, Module2Actions> = {
+  reset: (state, actions) => ({ count: 0 }),
+  add: (state, actions, value: number) => ({ count: state.count + value })
+}
+
+// app
+
+interface State extends Hyperapp.State {
+  module1: Module1State
+  module2: Module2State
+  // just to demonstrate there can be other attributes than the ones defined in actions
+  unused?: number
+  unused2: {
+    foo: string
+  }
+}
+
+interface Actions extends Hyperapp.Actions<State> {
+  module1: Module1Actions
+  module2: Module2Actions
+}
+
+const actions: Hyperapp.InternalActions<State, Actions> = {
+  module1: module1Actions,
+  module2: module2Actions
+}
 
 interface Events extends Hyperapp.Events<State, Actions> {
-  log: (state: State, actions: ActionCallers, value: string) => void
+  log: (state: State, actions: Actions, value: string) => void
 }
 
 const emit = app<State, Actions, Events>({
   state: {
-    count: 0
+    module1: {
+      count: 0
+    },
+    module2: {
+      count: 0
+    },
+    unused2: {
+      foo: "I have to set a value here, otherwise compilation error."
+    }
   },
-  view: (state, actions: ActionCallers) => (
+  // no need to set the types here
+  view: (state, actions) => (
     <main>
-      <h1>{state.count}</h1>
-      <button onclick={() => actions.sub(1)}>-</button>
-      <button onclick={() => actions.add(2)}>+</button>
+      <h1>Typescript Demo</h1>
+      <h2>Module 1</h2>
+      <p>
+        <button onclick={() => actions.module1.sub(-1)}>-</button>
+        {state.module1.count}
+        <button onclick={() => actions.module2.add(2)}>+</button>
+      </p>
+      <h2>Module 2</h2>
+      <p>
+        <button onclick={() => actions.module2.reset()}>Reset</button>
+        {state.module2.count}
+        <button onclick={() => actions.module2.add(1)}>+</button>
+      </p>
     </main>
   ),
-  actions: {
-    sub: (state, actions, value) => ({ count: state.count - value }),
-    add: (state, actions, value) => ({ count: state.count + value })
-  },
+  actions,
   events: {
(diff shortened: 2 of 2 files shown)


JorgeBucaran:
Could @andrewiggins and @pspeter3 have a look? 👋 🏄

JorgeBucaran:
@Mytrill Can you please add examples of ActionsImpl to index.tsx?

Also, I think we should come up with a more "pleasant to type" name for ActionsImpl. I am open to ideas:

- ActionsObject
- ActionsModel
- ActionModule

etc.

hyperapp.d.ts
```diff
@@ -54,56 +56,58 @@ export type ActionResult<State extends Hyperapp.State> =
...
 
+export type ActionsImpl<
+State extends Hyperapp.State,
+Actions extends Hyperapp.Actions
+> = {
+[P in keyof Actions]: (
+  state: State,
+  actions: Actions,
+  data: any
+) => ActionResult<State> | ActionsImpl<State, Actions>
```
andrewiggins: Can you add a test in index.tsx for action namespaces? I think that this type as it stands says that "`ActionImpl` is an object whose values are functions that return an `ActionResult`. Or the function can return an `ActionsImpl`" as opposed to what I think you want to say: "`ActionImpl` is an object whose values are either a function that returns an `ActionResult`. Or the object value can be an `ActionsImpl`"

andrewiggins: Can you add a test in index.tsx for action namespaces? I think that this type as it stands says that "`ActionImpl` is an object whose values are functions that return an `ActionResult`. Or the function can return an `ActionsImpl`" as opposed to what I think you want to say: "`ActionImpl` is an object whose values are either a function that returns an `ActionResult`. Or the object value can be an `ActionsImpl`"

andrewiggins: I think this bug existed previously but was on the `ActionCallers` interface

test/ts/index.tsx
```diff
@@ -4,31 +4,29 @@ interface State extends Hyperapp.State {
...
+interface Actions extends Hyperapp.Actions {
+  sub(value: number): Partial<State>
+  add(value: number): Partial<State>
 }
 
-interface ActionCallers extends Hyperapp.ActionCallers<State, Actions> {}
-
 interface Events extends Hyperapp.Events<State, Actions> {
-  log: (state: State, actions: ActionCallers, value: string) => void
+  log: (state: State, actions: Actions, value: string) => void
```
andrewiggins: I like how this change provides more types to the `actions` parameter here. Before this change, `actions.sub` wasn't the most helpful:
image

This change increases the help the intellisense gives:
image

hyperapp.d.ts
```diff
@@ -54,56 +56,58 @@ export type ActionResult<State extends Hyperapp.State> =
...
-    | Actions<State>
+    | Actions
     | ((
-        state: State,
-        actions: ActionCallers<State, Actions<State>>,
         data: any
       ) => ActionResult<State>)
 }
 
+export type ActionsImpl<
```
andrewiggins: As @JorgeBucaran suggested, a more pleasant name would be nice. Though (if I understand this PR correctly) since this is merely and internal type, it may not be necessary.

andrewiggins: Actually, do we need to export this type at all? I believe this type is only used by the `App` interface to define the structure of the `actions` option prop and not intended to be directly extended by users. @Mytrill, is that correct?

Swizz:
This is the kind of contribution I was looking for. That seem to solve the biggest problem I was stuck about : Actions .

Will do a complete review.

hyperapp.d.ts
Swizz: You are right. Missing parentheses. The `|` is hard to play with ^^

test/ts/index.tsx
Swizz: Yeah. This is very smart to take the problem from the other side. I am really happy that it solve this issue !

hyperapp.d.ts
andrewiggins: Hmm nested actions seem to be a little tricky. I think the body of `ActionsImpl` needs to include the parentheses as well as a Partial:
```ts
[P in keyof Actions]: ((
  state: State,
  actions: Actions,
  data: any
) => ActionResult<State>) | Partial<ActionsImpl<State, Actions>>
```
This fix would allow nested actions I think at least.

hyperapp.d.ts
Mytrill: You will need it to implement the actions:

```typescript
const actions. ActionsImpl<State, Actions> = {...}
```

I will add it in index.tsx.

andrewiggins: Oh good call - that makes sense.

andrewiggins:
Other ideas for names:

* ActionsDefinition
* ActionsDef

JorgeBucaran:
@andrewiggins What if we call them ActionCallers, it's only confusing if you remember how we used to call them before, but for someone looking at this for the first time it might not be so bad?

ActionCallers has a nice rhythm that I like. What do you think? @Mytrill

hyperapp.d.ts
```diff
@@ -36,96 +41,91 @@ export interface Component<Props> {
...
-  [action: string]:
-    | Actions<State>
-    | ((
-        state: State,
-        actions: ActionCallers<State, Actions<State>>,
-        data: any
-      ) => ActionResult<State>)
-}
+// this is done to allow ActionsImpl to reference State[P]
+export type State = { [S in keyof Actions]?: any };
```
Mytrill: This is very questionable, but it seems to do the trick, anyone has any better idea?

JorgeBucaran:
@Mytrill Please run `npm run format` / `yarn format`! 🙏

Mytrill:
I don't know, I am not very good when it comes to naming.
I'll bend to the public opinion.

JorgeBucaran:
Okay, so if no one else is opposed to the idea let's go with ActionCallers?

/cc @andrewiggins @Swizz  @pspeter3  @andyrj

hyperapp.d.ts
Swizz: Can you explain a bit more ?

Swizz:
This is the name I used so..

@mytrill is adding semicolon everywhere necessary ?

Mytrill:
No no, I messed up the formating... I'll push another commit with the proper format tonight.

One remark regarding the name, before ActionCallers meant "The API for the actions", now this is just Actions, and ActionsImpl means "The type for the implementation of the actions".

Swizz:
Juste thinking but sometimes is better to name things by their simple names, as those actions are the internal used ones : `InternalActions`

JorgeBucaran:
@Mytrill Gotcha, so to clarify, I meant to say that `Actions` is perfect, but I am not so thrilled about `ActionsImpl` so, if no one is against `ActionCallers`, let's go with that.

`ActionsImpl` → `ActionCallers`

hyperapp.d.ts
pspeter3: You should prefer `{}` to `any`. I'm not sure `State` here is the right name. You're getting the union type of Action names.

test/ts/index.tsx
```diff
@@ -1,41 +1,94 @@
...
+  module1: Module1State;
+  module2: Module2State;
+  // just to demonstrate there can be other attributes than the ones defined in actions
+  unused?: number;
 }
 
-interface Actions extends Hyperapp.Actions<State> {
-  sub(state: State, actions: ActionCallers, value: number): Partial<State>
-  add(state: State, actions: ActionCallers, value: number): Partial<State>
+interface Actions extends Hyperapp.Actions {
```
pspeter3: This looks weird to me. I feel like this should be a generic of state since actions run over state.

pspeter3:
The Vuex type definitions seem to use `XTree` for any of these maps. I think in that case they're nested. My proposal is to suffix with `Tree` for any data structure which could be nested and `Map` for anything that is flat. For example `StateTree` and `ActionsMap`.

test/ts/index.tsx
Mytrill: Yeah, it is, I am trying to do `Action<State>` but I get a lot of issues, and I think it comes from the following truth: State may have any key, which has in turn (almost) any type (VERY easy to write in function of Action). Action has 2 different types of keys: if `key in keyof State`, then it's either a sub-tree of actions, or a library of actions, otherwise it can be any key, in which case it must be of type `(data: any) => ActionResult<...>` (MUCH harder to write in function of State, AND, cannot be an interface, has to be a type...).

Here is the snippet (I think it will be easier to show than try to explain):

```typescript
// represents the external actions exposed to the world
// represents an object where each key (any key) is an action
export type ActionsFunctions<State extends Hyperapp.State> = {
  [action: string]: (data: any) => ActionResult<State>
}

// represents the external actions exposed to the world
// represents an object where each key (MUST be a key in the state object) can be either a sub-tree or a set of actions
export type ActionsTree<State extends Hyperapp.State> = {
  [S in keyof State]?: ActionsTree<State[S]> | ActionsFunctions<State[S]>
}

// represents the external actions exposed to the world
// represents an object where each key is either: an action, a sub-tree or a set of actions
export type Actions<State extends Hyperapp.State> = ActionsFunctions<State> &
  ActionsTree<State>

// represents the internal actions: their implementations
// represents an object where each key (any key) is an action
export type InternalActionsFunctions<
  State extends Hyperapp.State,
  Actions extends Hyperapp.ActionsFunctions<State>
> = {
  [A in keyof Actions]: (
    state: State,
    actions: Actions,
    data: any
  ) => ActionResult<State>
}

// represents the internal actions: their implementations
// represents an object where each key (MUST be a key in the state object) can be either a sub-tree or a set of actions
export type InternalActionsTree<
  State extends Hyperapp.State,
  Actions extends Hyperapp.ActionsTree<State>
> = {
  [S in keyof State]?:
    | InternalActionsTree<State[S], Actions[S]>
    | InternalActionsFunctions<State[S], Actions[S]>
}

// represents the internal actions: their implementations
// represents an object where each key is either: an action, a sub-tree or a set of actions
export type InternalActions<
  State extends Hyperapp.State,
  Actions extends Hyperapp.Actions<State>
> =
  | InternalActionsFunctions<State, Actions>
  | InternalActionsTree<State, Actions>
```

This, however does not work when actually using it:
screenshot from 2017-09-26 19-59-04

PS: in the screenshot above, even without the error, it gives another error because Module1Actions is an interface and Hyperapp.Actions is a type..

With the approach in this PR, everything just works!

If anyone has any suggestion, feel free to help me out :(

andyrj:
I agree ActionCaller is better than ActionImpl.  When this is in a good place I will try running flowgen on it again to see if it will give reasonable output for flow library def based on it.

hyperapp.d.ts
Mytrill: @Swizz When we declare ActionsImpl:
```typescript
export type ActionsImpl<
  State extends Hyperapp.State,
  Actions extends Hyperapp.Actions
> = {
  [P in (keyof Actions)]:
    | ((state: State[P], actions: Actions[P], data: any) => ActionResult<State>)
    | ActionsImpl<State, Actions>
};
```
We can do `state: State[P]` because we declared the state this way, otherwise this line would give an error.

And you may have properties in the state and not in the actions because when declaring state, you extends `Hyperapp.State<Actions>`.

test/ts/index.tsx
pspeter3: Yeah, I think you would want to write that as:

```ts
const module1Actions: Hyperapp.Actions<Module1State> = {
  add: (value: number) => value + 1,
  sub: (value: number) => value - 1,
}
type Module1Actions = typeof module1Actions;
```

Mytrill: Check out the new commit, I modified the InternalActions/ActionCallers:
```typescript
export type InternalActions<
  State extends Hyperapp.State & Partial<Record<keyof Actions, any>>,
  Actions extends Hyperapp.Actions<State>
> = {
  [P in keyof Actions]:
    | InternalAction<State, Actions>
    | InternalActions<State[P], Actions[P] & Hyperapp.Actions<State[P]>>
}
```

This `& Partial<Record<keyof Actions, any>>` make it possible to do `InternalActions<State[P], Actions[P] & Hyperapp.Actions<State[P]>>`.

This seems to allow to define actions simply (see index.tsx).

hyperapp.d.ts
```diff
@@ -26,6 +26,11 @@ export interface Component<Props> {
   (props: Props, children: VNodeChild<{} | null>[]): VNode<{}>
 }
 
+export type VNodeChildren =
+  | Array<VNodeChild<{} | null> | number>
+  | VNodeChild<{} | null>
+  | number
```
pspeter3: Why is this not included with `VNodeChild`?

hyperapp.d.ts
```diff
@@ -54,42 +59,48 @@ export type ActionResult<State extends Hyperapp.State> =
   | null
   | void
 
-export type ActionCallers<
+export interface Action<State extends Hyperapp.State> {
+  (data: any): ActionResult<State>
```
pspeter3: Can we make this generic? I feel like the typing for `data` here is too loose.

hyperapp.d.ts
```diff
@@ -54,42 +59,48 @@ export type ActionResult<State extends Hyperapp.State> =
...
+}
+
+export interface Actions<State extends Hyperapp.State> {
+  // should be Actions<State[action]>, but this is not allowed by typescript
```
pspeter3: Perhaps we could link to an issue with the TypeScript compiler so we can fix this in the future?

hyperapp.d.ts
```diff
@@ -54,42 +59,48 @@ export type ActionResult<State extends Hyperapp.State> =
...
+  [action: string]: Actions<any> | Action<State>
+}
+
+export interface InternalAction<
```
pspeter3: Is this a `ModuleAction`?

hyperapp.d.ts
Mytrill: This type is just an input type to the `h()` function, whereas `VNodeChild` is the type on the node itself (the `h()` functions accepts arbitrarily nested array of `VNode | string | number` and outputs a `VNode` with `VNode.children: VNodeChild[] (= Array<VNode | number>)`

hyperapp.d.ts
Mytrill: This is the type that represents the implementation of the action (i.e. `(state, actions, data) = ...`.
I liked the name `InternalAction` that Swizz suggested, but it seems everyone prefers `ActionCaller` (even though it it not something that calls an action, it's the code that implements the actions).

pspeter3: Is there a link to the JavaScript documentation for this? I'm not sure where I would use this right now.

hyperapp.d.ts
JorgeBucaran: @Mytrill

hyperapp.d.ts
Mytrill: Sorry I don't understand your comment: documentation for actions?

hyperapp.d.ts
Mytrill: Done, but this is not super useful since `Action<>` is only used in `Actions<>` and I have to set the type to `any` there (it may be something else than an object).

hyperapp.d.ts
Mytrill: Changed to:
``` typescript
export interface Actions<
  State extends Hyperapp.State & Partial<Record<keyof Actions<State>, any>>
> {
  [action: string]:
    | Actions<State[keyof Action<State, any>]>
    | Action<State, any>
}
```

hyperapp.d.ts
pspeter3: I meant this https://github.com/hyperapp/hyperapp/blob/master/docs/actions.md

Mytrill:
@Swizz @pspeter3 @andrewiggins @andyrj @JorgeBucaran I would like to go ahead and merge this, it seems to me there are no open issues as of now? Could you please react to this message if you are OK with the merge or otherwise reply to this?

Once this is merged I want to update #391 to match the latest fragments/hooks developments.

Thanks!

hyperapp.d.ts
Mytrill: I am sorry but I still don't understand what you want me to change now? Could you explicitly say what is wrong with the code at this point?

Thanks!

Swizz:
So this is still necessary to define both `Actions` and `InternalActions` ? there are no way to move forward of this by defining only `Actions` ?

Mytrill:
IMHO no, you define Action as the API to call actions, but you still need to define the implementation somewhere, I don't see any issue with this though users don't have any verbose code to type and they get typechecking in their action implementation:
```typescript
// API
interface Module1Actions extends Hyperapp.Actions<Module1State> {
  sub(value: number): Partial<Module1State>
  add(value: number): Partial<Module1State>
}
// implementation
const module1Actions: Hyperapp.InternalActions<Module1State, Module1Actions> = {
  // state and actions are properly type checked thanks to InternalActions
  sub: (state, actions, value: number) => ({ count: state.count - value }),
  add: (state, actions, value: number) => ({ count: state.count + value })
}
```

pspeter3:
I'm looking over the test to see if it is idiomatic TypeScript code.

pspeter3:
So just to make sure I'm correct in understanding, the actions passed into render are actually just a wrapper around the InternalActions?

Mytrill:
Yes, the actual Actions are a wrapper around the InternalActions, BUT in the type system it is inverted (i.e. InternalActions<Actions>) since it is much better to defined the interface for Actions (since that what you use everywhere) rather than the one for InternalActions.

pspeter3:
I guess that makes sense

JorgeBucaran:
@Mytrill @pspeter3 @andrewiggins @Swizz Thank you all! 🎉🏄‍♀️

PR #390, opened 2017-09-25
Add VS Code Editor config file to help debug and run Tests from within the Editor

rajaraodv:
VS Code provides excellent support to run and debug tests. i have added two configurations to the .vscode and anyone who opens Hyperapp will get the following:
1.  ***Debug Tests (Current File)*** - This helps us to simply open one of the Test files and put break point to debug.
2. ***Debug All Tests*** -  This is to simply run all the tests.

Following pictures shows how it works and will serve as  documentation:

1. **Debugging Tests (Current File)**

Debugging Current file getting started:
debugging-single-test-file

Debugging file in action:
debugging-single-test-file-in-action

Output of a single test file:
output-single-file

2. **Debug All Tests**
Use this to simply run all tests and ensure everything works. If you see any errors in any file, switch back to the above option and debug that one file.

Output of running all files:
output-all-files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 # IDE
 .idea/
-.vscode/
 
 # Logs
 *.log*
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,30 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+    {
+        "name": "Debug Tests (current file)",
+        "type": "node",
+        "request": "launch",
+        "runtimeArgs": [   
+            "--inspect-brk",
+            "${workspaceRoot}/node_modules/jest/bin/jest",
+            "--runInBand",
+            "${file}"
+        ],
+        "console": "integratedTerminal",
+        "internalConsoleOptions": "neverOpen"
+        },
+      {
+        "name": "Debug All Tests",
+        "type": "node",
+        "request": "launch",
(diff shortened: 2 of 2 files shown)


Swizz:
This is a great resource for learning but I am not really in favor of adding editor specifics things in the Hyperapp main repository.

JorgeBucaran:
This is definitely a super useful resource and I will copy those settings right away, but I am not sure about adding it to the project.

It still would be nice to share this info with the community somehow, via a separate repo for example, or in a wiki, etc.

lukejacksonn:
I would agree with @Swizz and @JorgeBucaran here.. it is a very useful resource but not essential for all hyperapp users.

rajaraodv:
OK, makes sense. Can we have add this to Docs? (under testing section)?

lukejacksonn:
@rajaraodv how about you make a PR showing where you would put it and what it would look like?

rajaraodv:
OK. Closing this.

okwolf:
Someone should publish an npm package for creating/adding launch configs that you could just run with `npx` instead of manually copying this every time you wanted it.

selfup:
oh that would be pretty easy using `fs` 🤔

okwolf:
@selfup wouldn't it just be a Jest launch config? I could use this on non HA projects.

selfup:
Oh right, that makes sense as well 😂

PR #394, opened 2017-09-29
Update hyperx url

aeosynth:
https://github.com/substack/hyperx -> https://github.com/choojs/hyperx


JorgeBucaran:
@aeosynth Thanks! I think I fixed this in #385.

PR #400, opened 2017-10-05
Make update function more explicit

rajaraodv:
Make update function more explicit so people don't wonder what that `update` is or don't think that is Hyperapp's special function.

### DIFF
--- a/docs/thunks.md
+++ b/docs/thunks.md
@@ -7,7 +7,7 @@ app({
   actions: {
     setData(state, actions, data) {
       return update => {
-        asyncProcess(data, ({ value }) => update({ value }))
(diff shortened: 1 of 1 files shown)


JorgeBucaran:
@rajaraodv Well, [`update`](https://github.com/hyperapp/hyperapp/blob/8425fd22507f97151a6927646f5cf849b151d961/src/app.js#L73) is Hyperapp's special function. 😄 

As you know, thunks are async; we don't know when they are done or if done at all, so we give you an `update` function that you use to tell us your task is complete and we should update the state with some data. 

This idea is similar to how the `onremove` lifecycle allows you to return a function that receives a `remove` function, so you can defer removing an element (e.g., for animations.)

rajaraodv:
Ah!

PR #401, opened 2017-10-06
Update Docs

okwolf:
Completed:
- [x] Fix typos
- [x] Expand `state` documentation including slices 🍰 
- [x] Document `modules`
- [x] Explain what the callback in [`onremove`](https://github.com/hyperapp/hyperapp/blob/docs/docs/lifecycle.md#onremove) does

Deferred Updates:
- [ ] Document HOAs (maybe?)
- [ ] More examples?
- [ ] Reorganize to separate Virtual DOM from `state`/`actions`?


docs/README.md
```diff
@@ -65,7 +68,7 @@ app(
 
 ### State
 
-The state describes the data model in your application. In this example, it consists of a single property: `count` which is initialized to 0.
+The state describes the data model in your application. The state must always be an object. In this example, it consists of a single property: `count` which is initialized to 0.
```
RikuVan: A little punctuation error here: In this example, it of consists of a single property, `count`, which is initialized to 0.

docs/initialization.md
```diff
@@ -0,0 +1,11 @@
+# Initialization
+
+The `init` function is run when the app starts. Use this for subscribing to mouse / keyboard events, starting timers, fetching resources, etc. This is also available on [`modules`](modules.md).
+
```
RikuVan: better style: avoid etc. and add an 'and' to the list of examples (we know they are examples not a complete list).

docs/modules.md
```diff
@@ -0,0 +1,39 @@
+# Modules
+
+Modules allow you encapsulate your application behavior into reusable parts, to share or just to organize your code. Modules are scoped to a state / action slice and there is no way to use a module without supplying the slice key.
+
```
RikuVan: grammar: ...allow you **to** encapsulate (or allow encapsulation of your...)

docs/modules.md
```diff
@@ -0,0 +1,39 @@
...
+app({
+  init(state) {
+    console.log(state) // => { foo: { value: 1 } }
+  },
+  modules: { foo }
+})
+```
+
+Modules make it easy to share and modularize your app, but without losing transparency or any of the benefits of single state tree architecture.
+
```
RikuVan: style: ...modularize your app without losing...

docs/modules.md
```diff
@@ -0,0 +1,39 @@
...
+app({
+  init(state) {
+    console.log(state) // => { foo: { bar: { value: 1 } } }
+  },
+  modules: { foo }
+})
+```
+
+Inside your module's actions, you can only call actions your actions or those exposed in modules under you. This is similar to how a parent component can pass props down to a child component or how a top level action can access the entire state.
```
RikuVan: something odd here 'you can only call actions your actions...'?

docs/state-slices.md
```diff
@@ -0,0 +1,84 @@
+# State Slices
+
+Hyperapp uses a single state tree — that is, this single object contains all your application level state and serves as the single source of truth. This also means, if you are coming from Redux/Flux, that you have only one store for each app. A single state tree makes it straightforward to locate a specific piece of state, and allows for incredibly easy debugging.
```
RikuVan: sort of odd use of dash. why not omit ' - that is'?

docs/state-slices.md
```diff
@@ -0,0 +1,84 @@
...
+
+A single state tree is not free from struggle. It can be daunting to update a part of the state deeply nested in the state tree immutably _and_ without resorting to functional lenses / setters or advanced martial arts.
+
```
RikuVan: don't need 'and' before without

docs/state-slices.md
```diff
@@ -0,0 +1,84 @@
...
+State slices attempt to address this issue by giving you via actions, a slice of the state tree that corresponds to the namespace where both state and action are declared.
+
```
RikuVan: style: State slices address this issue with actions that access a slice of the state tree that corresponds to the namespace where both state and action are declared.

docs/state-slices.md
```diff
@@ -0,0 +1,84 @@
...
+    bar: {
+      value: 0,
+      anotherValue: 1
+    }
+  }
+}
+```
+
+...and wanted to update `value`, you had to update an entire record (**including siblings**), since there is no way to single out `value` from a nested state.
+
```
RikuVan: Technically a new sentence. Then, when you wanted to update, you...

docs/state-slices.md
```diff
@@ -0,0 +1,84 @@
...
+    bar: {
+      value: 0,
+      anotherValue: 1
+    }
+  }
+}
+```
+
+And have a corresponding action inside a namespace that matches the state you want to update.
+
```
RikuVan: sentence fragment: You will have a corresponding...

docs/README.md
```diff
@@ -65,7 +68,7 @@ app(
...
```
RikuVan: the first sentence feels a bit circular, maybe something like: The **state object** describes the entire data model for...

ngryman:
I think @RikuVan did a pretty good review on this 👍

docs/README.md
okwolf: I feel like the first comma may be unneeded.

docs/modules.md
okwolf: I'm guessing "you can only call your actions..." was meant.

okwolf:
@RikuVan thanks for all your comments! I've incorporated most of them into the latest commit, along with some other wordsmithing of my own.

infinnie:
HOAs? Higher order apps?

okwolf:
@infinnie yes Higher-Order Apps. It takes some tricks to do them right (one of these days I'll learn them) so we should definitely have those documented somewhere. I'm just not sure if that belongs in the core docs or not.

RikuVan:
I just had a look at the language but didn't really think too much about what might be missing or need clarification. At some point, I hope HOA's are included. I don't really buy the the idea that devs should be discouraged from using some part of the api or functionality through silence about it. A caution could be added. Would be happy to proof new stuff as added.

RikuVan:
Not in this PR, but would love to see a bit of the Typescript declarations documented. An example of how you might import and use them. The tests are useful examples. But yeah they are in the tests.

zaceno:
@RikuVan HOA aren't technically part of the api, which is why they're not documented I think.  It's just basically a function that passes your app-definition along to the real `app(...)`, giving you a chance to do some pre- and post-processing, in order to splice in behaviors that extend what plain `app()` gives you.

That said, it certainly would be very useful to have some documentation of the dos and don'ts, because every person who comes asking "does hyperapp do X" will get the response. "No, but you could make a HOA for it"

zaceno:
@RikuVan Agree with you on modules and typedefs though

infinnie:
Higher-order apps are not the currently recommended practice even in the React community. See https://cdb.reacttraining.com/use-a-render-prop-50de598f11ce.

JorgeBucaran:
@infinnie Higher order apps are not higher order components.

infinnie:
@JorgeBucaran So? Could you give me a practical example?

JorgeBucaran:
@infinnie The HOA pattern is only needed when you want to alter or enhance how the app works, the logger is a good example. In practice, you most like don't need a HOA, ever. In React, abusing the HOC pattern is too easy, not so in Hyperapp.

Notice that hyperapp is not doing anything to enable/disable this pattern, is the user's problem.

docs/CONTRIBUTING.md
```diff
@@ -20,7 +20,7 @@ Thank you for taking the time to read our contribution guidelines. You can contr
 
 - We use [Babel](https://babeljs.io) and [Jest](http://facebook.github.io/jest) to run the tests.
 - Feel free to create a new `test/*.test.js` file if none of the existing test files suits your test case.
-- Tests usually start by creating a small application, and using a feature, then check if `document.body.innerHTML` matches some string. The app() call is async, so we often use [oncreate](/docs/vdom-events.md#oncreate) or [onupdate](/docs/vdom-events.md#onupdate) events to detect when the view has been rendered.
+- Tests usually start by creating a small application and using a feature, then check if `document.body.innerHTML` matches some string. The app() call is async, so we often use [oncreate](lifecycle.md#oncreate) or [onupdate](lifecycle.md#onupdate) events to detect when the view has been rendered.
```
okwolf: Fixed broken links.

README.md
```diff
@@ -10,7 +10,7 @@ Hyperapp is a JavaScript library for building frontend applications.
 - **Functional**: Hyperapp's design is based on [The Elm Architecture](https://guide.elm-lang.org/architecture). Create scalable browser-based applications using a functional paradigm. The twist is you don't have to learn a new language.
 - **Batteries-included**: Out of the box, Hyperapp combines state management with a VDOM engine that supports keyed updates & lifecycle events — all with no dependencies.
 
-[Read the Docs](/docs/README.md)
+[Read the Docs](/docs/README.md#documentation)
```
okwolf: Jump directly to documentation content.

docs/lifecycle.md
```diff
@@ -40,14 +40,14 @@ function Textbox({ placeholder }) {
...
-This event is fired before the element is removed from the DOM. Use it for cleaning up resources, creating slide out animations, etc.
+This event is fired before the element is removed from the DOM. Use it for cleaning up resources, creating slide out animations, etc. A function returned from `onremove` receives a callback argument that once called will remove the element for you when you are done with your business. Otherwise you must also remove the element yourself.
 
 ```jsx
 function MessageWithFadeout({ title }) {
   return (
     <div
       onremove={element => done => fadeout(element).then(done)}>
-      <h1>{title}<h1>
+      <h1>{title}</h1>
```
okwolf: Typo in JSX.

docs/README.md
```diff
@@ -1,16 +1,18 @@
...
 - Getting Started
   - [Hello World](#hello-world)
   - [Installation](#installation)
 - Concepts
-  - [Keys](/docs/keys.md)
-  - [Lifecycle](/docs/lifecycle.md)
-  - [Components](/docs/components.md)
-  - [Sanitation](/docs/sanitation.md)
-  - [Hydration](/docs/hydration.md)
+  - [State Slices](state-slices.md)
```
JorgeBucaran: Let's call it just "Slices"? Since it includes actions as well.

okwolf: Renamed.

JorgeBucaran:
@okwolf Can we merge here?

okwolf:
@JorgeBucaran We can merge as soon as you are happy. I just wanted to get up to date with 0.16 first.

JorgeBucaran:
@okwolf Would you like to add anything else?

okwolf:
@JorgeBucaran not for this round. I moved the other ideas to a "Deferred" section.

JorgeBucaran:
@okwolf Can we delete the branch?

okwolf:
@JorgeBucaran sure, keep things nice and tidy.

PR #403, opened 2017-10-07
Typo causing  'Cannot read property 'key'' error

kenota:
It seems that there is a typo in app.js which causing 1 instead of i to be used as a index lookup in keyed value.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -287,7 +287,7 @@ export function app(props) {
 
       for (var i in oldKeyed) {
         var keyedNode = oldKeyed[i]
-        var reusableNode = keyedNode[1]
+        var reusableNode = keyedNode[i]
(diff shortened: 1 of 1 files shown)


JorgeBucaran:
@kenota Can you please share an example, preferably a CodePen that demonstrates the error?

kenota:
Hm. It seems I've misunderstood the code. I will try to create a CodePen to demonstrate the error.

JorgeBucaran:
@kenota Make sure you this is not the problem: https://github.com/hyperapp/html/issues/2

kenota:
@JorgeBucaran, thanks! I am not using hyperapp/html, only hyperapp itself and h function. Trying to create reproducible example on codepen now.

kenota:
@JorgeBucaran created codepen and raised an issue #404

PR #405, opened 2017-10-07
Update typescript definitions for release 0.14

Mytrill:
Update typescript definitions:
 - update actions definition to reflect the fragmented state
 - delete mixins and events definitions
 - update the test to utilize fragmented state

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -5,7 +5,7 @@ export as namespace Hyperapp
 /** The VDOM representation of an Element
  *
  * @memberOf [VDOM]
-*/
+ */
 export interface VNode<Props> {
   tag: string
   props: Props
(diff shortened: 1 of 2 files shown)


pspeter3:
One thing that would also be good to show an example of is a type safe higher order app.

Mytrill:
The typescript definitions are missing several other pieces to be complete: 
- a proper definition for Thunk
- a proper definition for the jsx namespace (intrinsic elements + their props, maybe we could take them from preact? It's MIT so should be fine...)
- examples with HOA + above points
- I probably forgot a bunch.

I may open an issue to gather all those and fix them little by little..

Mytrill:
Opened #409 to track these items. Can we merge this?

PR #408, opened 2017-10-08
Set default props when calling h(Component)

### LINKED ISSUE #407: Props null when calling components with Jsx
Imagine the following code:
```javascript
const Hello = ({ name = 'world' }, children) => {
  return <div>Hello {name}!<div>
}

app({
  view: () => <Hello />
})
```
When ran, it throws an error because undefined cannot be destructured into `{ name = 'world' }`.

This can be fixed by changing `h.js`:
```javascript
return typeof tag === "string"
    ? // ...
    : tag(props = {}, children)
```

Is this a bug or is it an intentional design decision?

Mytrill:
This PR fixes #407.

### DIFF
--- a/src/h.js
+++ b/src/h.js
@@ -25,5 +25,5 @@ export function h(tag, props) {
         props: props || {},
         children: children
       }
-    : tag(props, children)
+    : tag(props || {}, children)
 }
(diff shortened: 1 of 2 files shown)


JorgeBucaran:
@Mytrill LGTM, shall we merge?

JorgeBucaran:
Thanks @Mytrill! 👋🔥

Issue #404, opened 2017-10-07
Uncaught TypeError: Cannot read property 'key' of undefined


kenota:
Here is  [codepen](https://codepen.io/anon/pen/oGEQBY) triggering the issue.

Steps to reproduce:
1. Launch code pen
2. Open Chrome developer tools console
3. Type one character in the textarea => observe error log in chrome developer tools console.

Seems to be happening only for first character, but I am having similar problems with textfields. 

Update:
This error is happening on Chrome Version 61.0.3163.100 (Official Build) (64-bit), on Mac

johanalkstal:
I got `x.props is undefined` on first keystroke.

kenota:
@johanalkstal which browser did you use?

johanalkstal:
@kenota Firefox

lukejacksonn:
I am looking now 👀  not sure I have permissions to update the pen if I fix it though

lukejacksonn:
Humm.. that is weird. Couldn't pin down exactly what it was.. I created a fork that does (similar) to what I thought you were trying to achieve. Similar pattern of updating an item in a nested state object. I took the `map` level out so it is just a list, no errors https://codepen.io/lukejacksonn/pen/LzQqGa. Let me know if you figure anything out. 

@JorgeBucaran codepen wasn't picking up sourcemaps which makes it hard to debug this sort of thing, not 100% sure why they aren't working but I will try to look into it.

lukejacksonn:
Weird.. if you change the `textarea` to an input with a `value` then it works as expected. 

- https://codepen.io/lukejacksonn/pen/LzQqGa

kenota:
@lukejacksonn Good catch! I am able to narrow it down, it seems that issue happening when you update text in tag from being empty. See updated codepen:

* [https://codepen.io/anon/pen/oGEQBY](https://codepen.io/anon/pen/oGEQBY)

If you remove code `{comment.body}` inside `<p>` then error disappears

JorgeBucaran:
@kenota The good news is that this problem does not occur with [0.12.1](https://unpkg.com/hyperapp@0.12.1/dist/hyperapp.js), which means the fix must not be so difficult (I hope).

If you use 0.12.1 in that CodePen and change the action provisionally to:

```js
state.comments.map[id].body = newText;
console.log('updating text to ', newText);
return state
```

There's no error. That's how I know that the code works in 0.12.1.

kenota:
[...]

PR #411, opened 2017-10-08
Introduce init and modules. Close #406.

### LINKED ISSUE #406: Introduce init and modules.
We don't have an elegant way to call any actions, subscribe to global events or kickstart ourselves when our app loads. We can call actions after we return from `app()` but packages like the router, a mouse / keyboard interop interface, etc., could be exposed as **modules** and for their initialization needs, we can introduce a new **init** function. 🎉 

> By the way, modules are not like mixins as they would be scoped to a state & action slice, that you can't opt out from.

```jsx
import { whopper } from "./whooper"
import { mouse } from "./mouse"

import state from "./myState"
import actions from "./myActions"
import view from "./myView"

app({
  init(state, actions) {
    // Subscribe to global events, start timers, fetch resources & more!
  },
  state,
  actions,
  view,
  modules: { whopper, mouse }
})
```

The modules `whopper` and `mouse` of this example can definitely expose their own `init` function. They would be called after the top-level `init` with a slice of the state and actions, e.g., `state["whopper"]` and `actions["whopper"]`,  just how it works with actions.

In @Zaceno's words, what's the difference between modules and mixins and how do we explain their overlap with HOAs? 

> This actually gets at the core of the problem I saw with mixins (and why I wanted them renamed from plugins as they once were called): they basically served _two_ purposes: either to A) augment hyperapp’s features, or B) to modularize your app. **Now**, modules is only for (B), and we have HOA for (A).

/cc @lukejacksonn @zaceno @andyrj @Swizz

JorgeBucaran:
## tl;dr

- Introduce `init` function (essentially the original `events.load`).
- Introduce `modules`, similar to the old mixins, but without all their issues. See #406 for a full explanation.
  - Modules can have an init function too.
  - Modules can have nested modules.

### `init(state, actions) { ... }`

We don't have an elegant way to call any actions, subscribe to global events or kickstart ourselves when our app loads.

```jsx
app({
  init(state, actions) {
    // Subscribe to global events, start timers, fetch resources & more!
  }
})
```

### `modules: { foo, bar, baz  }`

We also need a way to encapsulate code without all the issues mixins had.

```jsx
app({
  modules: { router, whopper, bazinga }
})
```

Desugars into:

```jsx
app({
  init(state, actions) {
    // Close enough. Actual order is non-deterministic. 
    // Also, router, whooper and bazinga `init` function runs BEFORE the top level `init`.
    // This is consistent with sub-modules. Deepest first. Inside-out.

    router.init(state.router, actions.router)
    whopper.init(state.whopper, actions.whopper)
    bazinga.init(state.bazinga, actions.bazinga)
  },
  state: { 
    router: router.state,
    whopper: whopper.state,
    bazinga: bazinga.state,
  },
  actions: { 
    router: router.actions,
    whopper: whopper.actions,
    bazinga: bazinga.actions
  }
})
```

Quoting @zaceno on the difference between mixins and modules and how they fit together with HOAs.

> This actually gets at the core of the problem I saw with mixins (and why I wanted them renamed from plugins as they once were called): they basically served _two_ purposes: either to A) augment hyperapp’s features, or B) to modularize your app. **Now**, modules is only for (B), and we have HOA for (A).

Close #406, #404.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -1,25 +1,25 @@
 import { h } from "./h"
 
-var callbackStack = []
+var callbacks = []
 
 export function app(props) {
   if (typeof props === "function") {
     return props(app)
   }
 
-  var skipRender
+  var appState
+  var appActions
   var appView = props.view
-  var appState = {}
-  var appActions = {}
   var appRoot = props.root || document.body
   var element = appRoot.children[0]
-  var node = hydrate(element, [].map)
+  var node = elementToVNode(element, [].map)
+  var skipRender
 
-  requestRender(flush(setup(props, appState, appActions)))
+  repaint(flush(initFromModule(props, (appState = {}), (appActions = {}))))
 
   return appActions
 
-  function requestRender() {
+  function repaint() {
     if (appView && !skipRender) {
       requestAnimationFrame(render, (skipRender = !skipRender))
     }
@@ -38,67 +38,10 @@ export function app(props) {
   }
 
   function flush(cb) {
-    while ((cb = callbackStack.pop())) cb()
+    while ((cb = callbacks.pop())) cb()
   }
 
-  function setup(module, state, actions) {
-    copy(state, module.state)
-
-    setupActions(actions, state, module.actions)
-
-    if (module.init) {
-      callbackStack.push(function() {
-        module.init(state, actions)
-      })
-    }
-
-    if (module.modules) {
-      Object.keys(module.modules).map(function(key) {
-        setup(module.modules[key], (state[key] = {}), (actions[key] = {}))
-      })
-    }
-  }
-
-  function setupActions(actions, withState, withActions) {
-    Object.keys(withActions || {}).map(function(name) {
-      if (typeof withActions[name] === "function") {
-        actions[name] = function(data) {
-          return typeof (data = withActions[name](withState, actions, data)) ===
-          "function"
-            ? data(update)
-            : update(data)
-        }
-      } else {
-        setupActions(
-          (actions[name] = {}),
-          withState[name] || (withState[name] = {}),
-          withActions[name]
-        )
-      }
-    })
-
-    function update(data) {
-      return (
-        typeof data === "function"
-          ? update(data(withState))
-          : data && requestRender(copy(withState, data)),
-        withState
-      )
-    }
-  }
-
-  function copy(target, source) {
-    for (var i in source) {
-      target[i] = source[i]
-    }
-    return target
-  }
-
-  function merge(target, source, result) {
-    return copy(copy((result = {}), target), source)
-  }
-
-  function hydrate(element, map) {
+  function elementToVNode(element, map) {
     return (
       element &&
       h(
@@ -107,10 +50,58 @@ export function app(props) {
         map.call(element.childNodes, function(element) {
           return element.nodeType === 3
             ? element.nodeValue
-            : hydrate(element, map)
+            : elementToVNode(element, map)
         })
       )
     )
+  }
+
+  function initFromModule(module, state, actions) {
+    if (module.init) {
+      callbacks.push(function() {
+        module.init(state, actions)
+      })
+    }
+
+    assign(state, module.state)
+
+    initActions(state, actions, module.actions)
+
+    Object.keys(module.modules || {}).map(function(i) {
+      initFromModule(module.modules[i], (state[i] = {}), (actions[i] = {}))
+    })
+  }
+
+  function initActions(state, actions, source) {
+    Object.keys(source || {}).map(function(i) {
+      typeof source[i] === "function"
+        ? (actions[i] = function(data) {
+            return typeof (data = source[i](state, actions, data)) === "function"
+              ? data(update)
+              : update(data)
+          })
+        : initActions(state[i] || (state[i] = {}), (actions[i] = {}), source[i])
+    })
+
+    function update(data) {
+      return (
+        typeof data === "function"
+          ? update(data(state))
+          : data && repaint(assign(state, data)),
+        appState
+      )
+    }
+  }
+
+  function assign(target, source) {
+    for (var i in source) {
+      target[i] = source[i]
+    }
+    return target
+  }
+
+  function merge(target, source) {
+    return assign(assign({}, target), source)
   }
 
   function createElement(node, isSVG) {
@@ -122,23 +113,23 @@ export function app(props) {
         : document.createElement(node.tag)
 
       if (node.props && node.props.oncreate) {
-        callbackStack.push(function() {
+        callbacks.push(function() {
           node.props.oncreate(element)
         })
       }
 
-      for (var i = 0; i < node.children.length; ) {
-        element.appendChild(createElement(node.children[i++], isSVG))
+      for (var i = 0; i < node.children.length; i++) {
+        element.appendChild(createElement(node.children[i], isSVG))
       }
 
       for (var i in node.props) {
-        setProp(element, i, node.props[i])
+        setElementProp(element, i, node.props[i])
       }
     }
     return element
   }
 
-  function setProp(element, name, value, oldValue) {
+  function setElementProp(element, name, value, oldValue) {
     if (name === "key") {
     } else if (name === "style") {
       for (var name in merge(oldValue, (value = value || {}))) {
@@ -160,18 +151,17 @@ export function app(props) {
   }
 
   function updateElement(element, oldProps, props) {
-    for (var name in merge(oldProps, props)) {
-      var value = props[name]
-      var oldValue =
-        name === "value" || name === "checked" ? element[name] : oldProps[name]
+    for (var i in merge(oldProps, props)) {
+      var value = props[i]
+      var oldValue = i === "value" || i === "checked" ? element[i] : oldProps[i]
 
       if (value !== oldValue) {
-        setProp(element, name, value, oldValue)
+        value !== oldValue && setElementProp(element, i, value, oldValue)
       }
     }
 
     if (props && props.onupdate) {
-      callbackStack.push(function() {
+      callbacks.push(function() {
         props.onupdate(element, oldProps)
       })
     }
--- a/src/app.js
+++ b/src/app.js
@@ -65,9 +65,9 @@ export function app(props) {
 
     initActions(state, actions, module.actions)
 
-    Object.keys(module.modules || {}).map(function(i) {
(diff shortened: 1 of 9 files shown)


src/app.js
```diff
@@ -1,21 +1,21 @@
 import { h } from "./h"
 
-var lifecycleCallbackStack = []
+var callbackStack = []
```
pspeter3: Just curious, why stack as opposed to queue?

src/app.js
```diff
@@ -1,21 +1,21 @@
...
 
 export function app(props) {
+  if (typeof props === "function") {
```
pspeter3: This is what provides HOA syntax right?

src/app.js
```diff
@@ -25,107 +25,92 @@ export function app(props) {
...
+    while ((cb = callbackStack.pop())) cb()
+  }
+
+  function setup(module, state, actions) {
+    copy(state, module.state)
+
+    setupActions(actions, state, module.actions)
+
+    if (module.init) {
+      callbackStack.push(function() {
```
pspeter3: To get some bytes back, maybe use bind here?

src/app.js
```diff
@@ -25,107 +25,92 @@ export function app(props) {
...
+          : data && requestRender(copy(withState, data)),
+        withState
+      )
     }
   }
 
-  function set(prop, value, source) {
-    var target = merge(source)
-    target[prop] = value
+  function copy(target, source) {
```
pspeter3: Does hyperapp target browsers where this cannot be `Object.assign`?

zaceno:
I thought of a question: Is there any danger in reusing a module? Like what if I say:

```js

const counterModule = {
  state: {value: 0}
  actions: {
    increment: state => ({value: state + 1}),
    decrement: state => ({value: state - 1})
  }
}

app({
  modules: {
    counterA: counterModule,
    counterB: counterModule,
  },
  ...
})
```

does that mean, since I initially defined the two counter modules with the same object reference, that their state will be entangled somehow? Should I instead have written:


```js

const counterModule = _ => {
  state: {value: 0}
  actions: {
    increment: state => ({value: state + 1}),
    decrement: state => ({value: state - 1})
  }
}

app({
  modules: {
    counterA: counterModule(),
    counterB: counterModule(),
  },
  ...
})
```

or is the first option safe? I'm suspecting it is, but I wanted to ask to be sure :)

src/app.js
JorgeBucaran: @pspeter3 A queue would imply we execute callbacks in the order we found them.

```counter() -> subcounter() -> subsubcounter()```

But instead we execute inside out, the deepest callback first and unwind.

```subsubcounter() -> subcounter() -> counter()```

So, AFAICT, the correct terminology is stack.

File: src/app.js (hunk 4 above)
Comment (JorgeBucaran) [reply src/app.js]: @pspeter3 Yep, so >=IE10.

File: src/app.js (hunk 3 above)
Comment (JorgeBucaran) [reply src/app.js]: @pspeter3 Thanks, I'll see if I can squeeze this any further. 🤓

**Comment by JorgeBucaran** [comment]
@zaceno In your first example, there will be effectively two `{ value: 0 }`: counterA.value and counterB.value. So, it should be fine.

This was a problem during the mixin era since there was no fragmented state and you had to hardcode the namespace inside the mixin definition (bringing with it all the fun of namespace collisions).

With modules, you must specify the namespace to use it: `modules: { counter }`.

File: src/app.js [hunk 5]
```diff
@@ -25,107 +25,92 @@ export function app(props) {
...
-        element.tagName.toLowerCase(),
-        {},
-        map.call(element.childNodes, function(element) {
-          return element.nodeType === 3
-            ? element.nodeValue
-            : hydrate(element, map)
-        })
-      )
-    )
+  function flush(cb) {
```
johanalkstal: When does `flush` actually receive a `cb`?

JorgeBucaran: To avoid declaring the variable inside the function with `var`. Just a silly technique to save a few bytes here.

src/app.js
pspeter3: Yeah, makes sense. I think we should document that as people may rely on the execution order.

src/app.js
```diff
@@ -25,107 +25,92 @@ export function app(props) {
...
-          paths.length > 1
-            ? setPath(
-                paths.slice(1),
-                value,
-                source && name in source ? source[name] : {}
-              )
-            : value,
-          source
-        )
+  function merge(target, source, result) {
```
okwolf: It looks like this is only ever called with the `target` and `source`, and `result` is always set to `{}`. Will this work just as well?

```js
function merge(target, source) {
  return copy(copy({}, target), source)
}
```

All the tests pass, but for some reason the gzip output is a byte larger 😕

src/app.js
JorgeBucaran: Execution order within siblings is undetermined though. I think this is not a problem, as it was with mixins because with modules the user, not the author, is responsible to give it a key / scope.

src/app.js
JorgeBucaran: Good idea, this should work.

test/modules.test.js
```diff
@@ -0,0 +1,29 @@
...
+    }
+  }
+
+  app({
+    init(state, actions) {
+      expect(state.foo.value).toBe(0)
+      expect(actions.foo.up().value).toBe(1)
+      done()
+    },
+    modules: { foo }
```
okwolf: There should probably be tests for nested modules and `init` on a module.

src/app.js
```diff
@@ -25,107 +25,92 @@ export function app(props) {
...
-    }
-    return result
+  function hydrate(element, map) {
+    return (
+      element &&
+      h(
+        element.tagName.toLowerCase(),
+        {},
+        map.call(element.childNodes, function(element) {
+          return element.nodeType === 3
```
okwolf: I'm guessing we can't use `Node.TEXT_NODE` here because uglify won't know to replace it with the constant? I also thought we used to have a comment here about the meaning of this magic number.

SkaterDad: Could also copy what Preact does, and check if `element.splitText !== undefined`.   In a conference talk, `developit` mentioned the perf win was substantial, since `nodeType` is a getter function.

https://youtu.be/LY6y3HbDVmg?t=19m53s
https://github.com/developit/preact/blob/master/src/vdom/diff.js#L76

zaceno:
I've been experimenting with this branch, and think there might be a bug in the code that enables the simplified HOA syntax (`app(x)(y)(z)({...})`)

I have two HOA `A` and `B`. When I start my app with `A(B(app))({...})` it works. With `B(A(app))({..})` it works.

However, when I write `app(A)(B)({...})`, `A` fails to do it's thing. And when I write `app(B)(A)({...})` it's `B` that fails.

Am I, as  HOA-author, expected to do anything special in my HOA to enable that "chained" syntax?

test/modules.test.js
JorgeBucaran: @okwolf Added them both! Please check 🙏

JorgeBucaran:
@okwolf @Swizz @andyrj @zaceno Yes, the correct way to create multiple HOAs that take into account each other's enhancements follow.

```jsx
function B(app) {
  function enhancedApp(props) {
    return app({
      ...props,
      init() {
        console.log("B")
        props.init()
      }
    })
  }

  return props =>
    typeof props === "function" ? props(enhancedApp) : enhancedApp(props)
}

function A(app) {
  function enhancedApp(props) {
    return app({
      ...props,
      init() {
        console.log("A")
        props.init()
      }
    })
  }

  return props =>
    typeof props === "function" ? props(enhancedApp) : enhancedApp(props)
}

app(A)(B)({
  init() {
    console.log("Main")
  }
})
```

pspeter3:
I feel like the HOA abstraction is great but the API feels really awkward when you need to consider all the other higher order apps.

JorgeBucaran:
@pspeter3 What else do you have in mind?

JorgeBucaran:
Thanks everyone! I am ready to merge.

We should address changes to HOA support on a different issue.

/cc @andyrj @dodekeract Comments & feedback please 🙏

pspeter3:
That's exactly what I'm thinking. It seems even less necessary now that there are modules.

JorgeBucaran:
@pspeter3 I hear you. Let's talk about it in https://github.com/hyperapp/hyperapp/issues/413.

src/app.js
```diff
@@ -45,115 +50,86 @@ export function app(props) {
...
+      callbacks.push(function() {
+        module.init(state, actions)
+      })
+    }
+
+    assign(state, module.state)
+
+    initActions(state, actions, module.actions)
+
+    Object.keys(module.modules || {}).map(function(i) {
```
Swizz:
```js
for (i in module.modules || {}) {
```

Using it here will save some bytes and give us better perf, I guess.

src/app.js
```diff
@@ -45,115 +50,86 @@ export function app(props) {
...
+      init(module.modules[i], (state[i] = {}), (actions[i] = {}))
+    })
+  }
+
+  function initActions(state, actions, source) {
+    Object.keys(source || {}).map(function(i) {
```
Swizz:
```js
for (i in source || {}) {
```

src/app.js
JorgeBucaran: We will need a function to wrap the expression anyway because we would be accessing a shared variable `i` inside the new action definition `source[i]`. If we could use `let` instead of `var` it would work though.

JorgeBucaran:
Init & modules landed in master! 🎉

src/app.js
Swizz: `bind` is adding more byte than `function(){ ...` here Oo

vdsabev:
Loving the changes here! And the coverage 😉 
I'll use `modules` for my `router` implementation, and `init` to put some code inside the app definition where I think it'll look better (instead of after).

SkaterDad:
@JorgeBucaran Any idea on when this will get to npm?  Will you wait for the typescript defs to be updated?  I'd love to start refactoring my 0.12.x app now, but will probably wait for the typescript defs for editor tooling support.

JorgeBucaran:
Maybe if we can get @Mytrill @pspeter3 @Swizz to help with the types I could publish 0.15.0 sooner than later? 🤔

pspeter3:
Sure, though I'll likely need some help making sure I understand how modules and init are supposed to work. Perhaps you can modify the TypeScript test and we can make the type definitions work?

Issue #409, opened 2017-10-08
Enhance typescript definitions


Mytrill:
Several pieces are missing from the typescript definitions:
- [x] proper definition for Thunk (with types for arguments & return type) + example (done in #412)
- [ ] jsx namespace: intrinsic elements + example
- [ ] lifecycle events on VNode + example
- [x] proper definition of module + example
- [x] proper definition of init() + example

If anyone has any other idea, I'll add to this list.

Update: removed HOA, added modules and init.

pspeter3:
Where would you like to work on this?

Mytrill:
Thanks for #412!

I will work next on the definition for HOAs + an example.
Hopefully I will have some time tomorrow/Wednesday.

JorgeBucaran:
Bring it on! 🎰 😄

pspeter3:
I think at this point we can drop HOAs. I also think the JSX typing can be done another time. Init and Modules are more important.

PR #412, opened 2017-10-09
Enhanced type definitions

pspeter3:
Address the thunk definitions for #409

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -31,7 +31,7 @@ export interface Component<Props> {
 }
 
 /**The type for the children parameter accepted by h().
- * 
+ *
  * @memberOf [VDOM]
  */
 export type VNodeChildren =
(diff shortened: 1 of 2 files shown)


hyperapp.d.ts
```diff
@@ -55,40 +55,44 @@ export function h<Props>(
...
 
 /** The application state.
- * 
+ *
  * @memberOf [App]
  */
 export interface State {}
 
+export interface Update<State extends Hyperapp.State> {
+  (value: Partial<State>): void
```
Swizz: How would it work with the new namespace feature ? `Update<State[namespace]>` ?

pspeter3: I'm not sure. That's actually one of things that makes me nervous about hyperapp.

Mytrill: Actually I think it just works: in each module you define `interface ModuleState` and `interface ModuleActions extends Hyperapp.Actions<ModuleState>` and in each of the actions, you may return a `Thunk<ModuleState>`. There is an example of this (without thunk) in https://github.com/hyperapp/hyperapp/blob/master/test/ts/index.tsx.

PR #424, opened 2017-10-20
Don't lock render before calling view function.

JorgeBucaran:
Basically any actions called inside the view function or inside "functions called inside the view" (e.g., Components) can invalidate the state, and when the view function returns, the state is left "out of sync" with the new vnode.

The idea is to compute the vnode, but before patching the DOM, determine if one or more actions were called (i.e., detect if a re-render was scheduled) and if that's the case return early (don't patch, since render will be called again soon anyway), otherwise proceed as usual.

Fix #424.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -1,10 +1,10 @@
 import { h } from "./h"
 
 export function app(props, container) {
+  var lock
+  var callbacks = []
   var root = (container = container || document.body).children[0]
   var node = toVNode(root, [].map)
-  var callbacks = []
-  var skipRender
   var globalState
   var globalActions
 
@@ -13,21 +13,16 @@ export function app(props, container) {
   return globalActions
 
   function repaint() {
-    if (props.view && !skipRender) {
-      requestAnimationFrame(render, (skipRender = !skipRender))
+    if (props.view && !lock) {
+      requestAnimationFrame(render, (lock = !lock))
     }
   }
 
-  function render() {
-    flush(
-      (root = patchElement(
-        container,
-        root,
-        node,
-        (node = props.view(globalState, globalActions)),
-        (skipRender = !skipRender)
-      ))
-    )
+  function render(next) {
+    lock = !lock
+    if ((next = props.view(globalState, globalActions)) && !lock) {
+      flush((root = patchElement(container, root, node, (node = next))))
+    }
   }
 
   function flush(cb) {
@@ -107,7 +102,7 @@ export function app(props, container) {
         ? document.createElementNS("http://www.w3.org/2000/svg", node.type)
         : document.createElement(node.type)
 
-      if (node.props && node.props.oncreate) {
+      if (node.props.oncreate) {
         callbacks.push(function() {
           node.props.oncreate(element)
         })
@@ -155,7 +150,7 @@ export function app(props, container) {
       }
     }
 
-    if (props && props.onupdate) {
+    if (props.onupdate) {
       callbacks.push(function() {
         props.onupdate(element, oldProps)
       })
@@ -223,7 +218,6 @@ export function app(props, container) {
         }
 
         var newKey = getKey(newChild)
-
         var keyedNode = oldKeyed[newKey] || []
 
         if (null == newKey) {
--- a/test/app.test.js
+++ b/test/app.test.js
@@ -7,33 +7,35 @@ beforeEach(() => {
 })
 
 test("throttling", done => {
+  var count = 0
+
   app({
-    view: state =>
-      h(
-        "div",
-        {
-          oncreate() {
-            expect(document.body.innerHTML).toBe("<div>5</div>")
-            done()
-          }
-        },
-        state.value
-      ),
+    view(state, actions) {
+      if (0 >= count) {
+        actions.fire()
+      }
+
+      count += 1
+
+      return h("div", {}, state.value)
+    },
     state: {
-      value: 1
+      value: "foo"
     },
     actions: {
-      up(state) {
-        return {
-          value: state.value + 1
-        }
+      render(state) {
+        return state
       },
       fire(state, actions) {
-        actions.up()
-        actions.up()
-        actions.up()
-        actions.up()
+        actions.render()
+        actions.render()
+        actions.render()
       }
     }
   }).fire()
+
+  setTimeout(() => {
+    expect(count).toBe(2)
+    done()
+  }, 10)
 })


JorgeBucaran:
@SkaterDad Can you please explain your concern with these changes? This is very much still open to debate.

/cc @zaceno

SkaterDad:
Properly managed, I'm warming up to the ideas which this PR is supporting.

I was expecting an infinite loop if you just called an action directly in the view  (action -> state update -> render -> action -> ....), but was pleasantly surprised & slightly confused that it was not the case. 😆 

The first code I tried in details below.  Why does `actions.toggle()` only trigger the redraw once?
<details>

```js
  app({
    state: { test: true, other: 1 },
    actions: {
      toggle(state) {
        console.log('Toggle called')
        return { test: !state.test }
      },
      otherThing(state) {
        console.log('Other action called')
        return {
          other: state.other + 1
        }
      }
    },
    view(state, actions) {
      actions.toggle()
      return h('div', null, [
        h('button', { onclick: actions.otherThing }, 'Click Me'),
        h('pre', null, JSON.stringify(state, null, 2))
      ])
    }
  })

```
</details>
-

I did manage to create an infinite loop while implementing data fetching with a simple `loading` flag, but that was just bad logic on my part.  This seems to work okay, and triggers `toggle()` once on each update.

<details>

```js
  const { h, app } = hyperapp

  app({
    state: { test: true, thing: 1, data: null, requestState: 'NOT_ASKED' },
    actions: {
      toggle(state) {
        console.log('Toggle called')
        return { test: !state.test }
      },
      otherThing(state) {
        return { thing: state.thing + 1 }
      },
      fetchData: (state) => (update) => {
        console.log('Fetch action called')
        update({ requestState: 'LOADING' })
        setTimeout(() => {
          update({requestState: 'SUCCESS', data: [1,2,3,4]})
        }, 1000)
      }
    },
    view(state, actions) {
      actions.toggle()

      if (state.requestState === 'NOT_ASKED') {
        actions.fetchData()
      }

      return h('div', null, [
        h('button', { onclick: actions.otherThing }, 'Click Me'),
        h('pre', null, JSON.stringify(state, null, 2))
      ])
    }
  })
```

</details>

okwolf:
I'm not sure we can afford the bloat of a 1402 byte bundle...
<br />
<br />
<br />
<br />
<br />
<br />
<br />
<br />
<br />
<br />
<br />
<br />
... I'm joking 😆 :shipit:

lukejacksonn:
@SkaterDad me too! I was playing around all weekend with this PR's version of hyperapp.. it has prompted me to think quite differently (in a good way) about components.

@JorgeBucaran mentioned he is going to put this on hold until a solution to #425 has been suggested; fearing that they might conflict or be mutually exclusive. I hope that this is not the case because I am really liking what the proposed changes here, allow you to do with hyperapp.

JorgeBucaran:
@lukejacksonn I don't think they are mutually exclusive, so we'll probably have this one too eventually / soon. 👍 

>  it has prompted me to think quite differently (in a good way) about components

I know what you mean! That was a really illuminating chat we had the other day!

src/app.js
```diff
@@ -13,21 +13,16 @@ export function app(props, container) {
...
-        container,
-        root,
-        node,
-        (node = props.view(globalState, globalActions)),
-        (skipRender = !skipRender)
-      ))
-    )
+  function render(next) {
+    lock = !lock
+    if ((next = props.view(globalState, globalActions)) && !lock) {
```
zaceno: Do we need `&& !lock` here? Unless I'm mistaken, `render` is only called from L17, which is already guarded by a lock. So we will never unlock until `render` is actually called. So `!lock` on L23 will always be true. 
...again: unless I'm mistaken. (it does happen occasionally ;) )

zaceno: In fact.. I suspect the whole `if` statement around the `next = props.view...` may be superfluous

zaceno: Just a suggestion: maybe `lock` could be named `lockRepaint` just to make it a little clearer what we are locking and unlocking?

JorgeBucaran: This is all new behavior. Before, updating the state in a view via actions was possible, but left the state out of sync with the rendered VNode. I'll explain what's going on now.

The lock starts false. Then we call repaint to schedule the first render and the lock is now true. fAF kicks in and calls render. We enter render and the first thing we do now is toggle the lock. Now it's false, like at the start. This is new business.

Now we call the view with your state and actions. The view computes the next VNode. During the process, if the view or a component calls an action, we update the state and _call repaint_ as usual. The twist is that now, the lock, which is false at this point, allow us to schedule a new render. rAF is async so we return immediately, [toggle the lock again](https://github.com/hyperapp/hyperapp/blob/4b5708d83924462e3a86021169ad3ed9da8459bc/src/app.js#L17) and eventually return from the view. We're here now. `&& !lock`.

We know the lock was false, so `&& !lock` checks if we are still false and then proceeds to patch the DOM with the new VNode. Otherwise, the view (or maybe an async action from above and beyond...) toggled the lock. Then we know that another render has been scheduled, so we skip patchElement and get out knowing that we'll be back soon. 😎

When rAF calls render again, the lock is true, so we're like at the start and the whole process  starts over.

@zaceno @SkaterDad @lukejacksonn @okwolf

zaceno: @JorgeBucaran Ah right. Although the first thing `render` does is unlock, then calling `props.view` may cause `update` -> `repaint` -> locked again. So before we go ahead and patch, we make sure that didn't happen. If it did happen: skip patching. Well be right back soon anyway. 

Thanks for the illuminating explanation! I withdraw my comment.

SkaterDad: @JorgeBucaran Thanks for the detailed explanation.

Can you enlighten me on why this doesn't cause an infinite cycle?  Are actions called from within the view somehow different than actions called in response to a user event, like clicks?

```js
    view(state, actions) {
      actions.toggle() // <-- THIS ONLY GETS RUN ONE TIME
      return h('div', null, [
        h('button', { onclick: actions.otherThing }, 'Click Me'),
        h('pre', null, JSON.stringify(state, null, 2))
      ])
    }
```

lukejacksonn: I was struggling to get my head around this bit too :P perhaps that is a sign that it could be refactored for readability. I suspect this foo has some size/perf justification behind it but now that you mention it it looks like the assignment of `next` is going to happen every call regardless of the state of `lock`.

Are you suggesting:

```
function render(next) {
  lock = !lock
  next = props.view(globalState, globalActions)
  if(!lock) { ... }
}
```

lukejacksonn: @SkaterDad I believe this is something to do with `requestAnimationFrame` but I can not for the life of me explain why. I just know if you remove rAF and call render directly then it breaks things.

JorgeBucaran: @SkaterDad It will totally cause an infinite loop. Actions called within the view are no different that actions called in resp to user events, etc. To avoid an infinite loop, you will need to add logic.

```js

app({
  view(state, actions) {
    if (state.name === "") {
      actions.populate() && return
      // You don't need to return here, but you could, because we don't call
      // patchElement when the VNode is undefined.
    }

    return h("div", null, [
      h("button", { onclick: actions.up }, "Click Me"),
      h("pre", null, JSON.stringify(state, null, 2))
    ])
  },
  state: {
    value: 0,
    name: ""
  },
  actions: {
    up(state) {
    },
    populate(state) {
      return { name: "Hyperapp" }
    }
  }
})
```

SkaterDad: @JorgeBucaran I agree, it should cause an infinite loop.  However, I when I cloned this branch and tried that code, I only saw the action getting called once.  If you can't reproduce, I'll give it another try this evening if I get time.

JorgeBucaran: @SkaterDad If you get time and do try it, please post the entire code so I can cop&paste to check. Thanks!

SkaterDad: @JorgeBucaran Just tried it again (clone, build), and now I do get an infinite loop.  Not sure what I did wrong the first time around!

SkaterDad: Maybe my test the other day was accidentally on 0.15.1.  This doesn't loop.
https://www.webpackbin.com/bins/-Kx9XsEEufXxqggvDDYz

JorgeBucaran: @SkaterDad Odd!

default

lukejacksonn:
What is the status of this at the moment? I know that there is a lot of discussion around shaking up actions etc. but don't think that this PR would affect or be affected by any of those decisions.. am I overlooking something?

JorgeBucaran:
@lukejacksonn Walking on thin ice! rAF needs to go, but maybe we can keep this if we replace rAF with setTimeout, instead of making app the sync.

lukejacksonn:
I'm not too fussed about the implementation details, certainly not precious about keeping raF but am still of the opinion that this feature is both sensical and useful, as well as non-breaking. Everyone has had the chance to scrutinised it and no one has come up with a good reason as to why we shouldn't merge (3/3 approved reviews).

I guess my point was, instead of bundling _all_ the proposed changes into one big release, can we do this incrementally and see how it goes? I can't see it causing any issues!

JorgeBucaran:
@lukejacksonn I've had this on hold after you mentioned you were having problems with some audio API because of rAF.

JorgeBucaran:
Honestly, I'd rather not keep this. In the future, as we start building on top of this, we may realize that calling actions inside the view is not problem-free and we may need to change something else as a result of wanting this.

zaceno:
@JorgeBucaran 
> as we start building on top of this, we may realize that calling actions inside the view is not problem-free

But we'll never know unless we try. ;)

lukejacksonn:
I was indeed.. but that is also a separate issue. That could be addressed in this PR if we like but just as easily done incrementally if this gets merged. How I see it:

- This PR ensures actions called mid-render update the state and trigger a re-render.
- Swapping `rAF` for `setTimeout` ensures renders occur when `document.hidden` is true.

That is all. Fixing one does no fix or break the other and vice versa. Both omit unexpected behaviour (that would otherwise need to be documented) and are non-breaking changes.

JorgeBucaran:
> Swapping rAF for setTimeout ensures renders occur when document.hidden is true.

Do we want to ensure render even if document.hidden is true?

lukejacksonn:
That is a separate issue. But if you intend on removing `rAF` then render will happen even if document.hidden is true.

JorgeBucaran:
Yes, I know, but if we remove rAF and call this off, we can also make the app sync which has other benefits as well.

lukejacksonn:
Making the app sync will mean that renders happen when `document.hidden` is true so you would fix that issue by proxy. Doing that would make this feature much less obtainable though (I think).

JorgeBucaran:
It's here #445. If this solution is good, then why not. And if turns out to be a headache, then we'll just remove it for good.

Issue #425, opened 2017-10-21
The state should be immutable


Mytrill:
# Current state
Since the addition of modules, the state is no longer immutable: when an action, or a nested action, or an action in a module is called, its result is merged with the current state (or state slice for nested actions/modules).

This prevent users from using memoize functions on the view with simple `===` checks for equality checks and forces to use deep equality.

For example, the following pen should work with immutable state: https://codepen.io/mytrill/pen/WZmVWQ?editors=0010

I believe for large apps, the benefit of having very cheap memoize functions based on immutable state far outweight the cost of maintaining an immutable state (both in terms of performance and source code size).

# Expected behavior
When an action gets executed, the result that it returns should get merged to the state slice/module's state this action operates on and create a new instance of the state and re-use the part of the state that hasn't been touched.
Here is an example: https://upload.wikimedia.org/wikipedia/commons/thumb/5/56/Purely_functional_tree_after.svg/438px-Purely_functional_tree_after.svg.png

In this case, when the node "f" is modified ("e" gets added to it) the update function recursively goes up the state tree, creating new nodes when modified and re-using un-touched nodes.

## Advantages
 - Allow debug tools to keep the state history and implement time travel
 - Allow very performant memoization of components/views on large app (deep equality on state can be checked with `===`)

## Drawbacks
 - small performance penalty when executing nested action (need to duplicate a whole branch of the state-tree
 - extra logic/bundle size

# Implementation
PR #426 addresses this issue

JorgeBucaran:
We'll need the three musketeers to help with this, so I'll leave them here for the time being.

```js
function setPath(path, value, source) {
  if (path.length === 0) {
    return value
  }

  var key = path[0]

  return setProp(
    key,
    path.length > 1
      ? setPath(
          path.slice(1),
          value,
          source != null && key in source ? source[key] : path[1] >= 0 ? [] : {}
        )
      : value,
    source
  )
}

function setProp(prop, value, source) {
  var target = {}

  if (prop >= 0 && Array.isArray(source)) {
    target = source.slice()
  } else {
    for (var key in source) {
      target[key] = source[key]
    }
  }
  target[prop] = value

  return target
}

function getPath(path, source) {
  return path.length === 0 ? source : getPath(path.slice(1), source[path[0]])
}
```

okwolf:
[...]

Issue #418, opened 2017-10-14
Issue with actions re-using the state


Mytrill:
The following unit test fails due to infinite recursion:
```javascript

test("Recursive state", done => {
  app({
    view: state =>
      h(
        "div",
        {
          oncreate() {
            console.log(document.body.innerHTML)
            expect(document.body.innerHTML).toBe(`<div>{"child":{}}</div>`)
            done()
          }
        },
        JSON.stringify(state)
      ),
    state: {},
    actions: {
      recurse(state) {
        return { child: state }
      }
    }
  }).recurse()
})
```

```
Uncaught TypeError: Converting circular structure to JSON
    at JSON.stringify (<anonymous>)
    at Object.view (bundle.js:283)
    at render (bundle.js:47)
```

This is because the update assigns `{ child: state }` to `state` making it reference itself.

Is it meant to work like this?

JorgeBucaran:
@Mytrill What were you expecting to happen and why are you doing this? 

> The following unit test fails due to infinite recursion...

Is the infinite recursion caused by Hyperapp or by `JSON.stringify(state)`?

Mytrill:
Hi Jorge. My use case is: I am implementing a UI where I can recursively split the view in 2. So my state is a tree, and I have a split() action that takes a path in the tree (dot separated string) and insert a node at the given path. The unit test you see here is the equivalent of calling this split() action on the root node.

I already changed my code to use Object.assign({}, state) before opening this issue, but I wanted to know if this was intended behavior (users do the assign in this case, saves perf and bytes to HA) or a bug (HA should always copy state before assigning).

It took me some times to debug this, if there ever is a section "common issues" in the doc, this should be there!

Thanks

Thanks for the answer!

PR #426, opened 2017-10-22
Make the app's state immutable

Mytrill:
This PR addresses #418 and #425.

The approach here is to recursively modify the `update` function as we initialize the actions/modules.

EDIT1: Gzipped size (without the console.log...): 1476 bytes
EDIT2: after commit 7b839a5 size is: 1492 bytes (but the code is easier to understand IMHO)
EDIT3: 1487 bytes (with fix to #438)

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -5,17 +5,19 @@ export function app(props, container) {
   var node = toVNode(root, [].map)
   var callbacks = []
   var skipRender
-  var globalState = {}
-  var globalActions = {}
+  var globalState
+  var globalActions
 
   repaint(
     flush(
       initModule(
         props,
-        globalState,
-        globalActions,
+        (globalState = {}),
+        (globalActions = {}),
         updateGlobalState,
-        getGlobalState
+        function() {
+          return globalState
+        }
       )
     )
   )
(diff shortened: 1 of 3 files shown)


Mytrill:
Other way to write this PR (by passing a path down the init()/initActions() functions + using getPath()/setPath() from #425) at this branch: https://github.com/Mytrill/hyperapp/tree/immutable-state-2

lukejacksonn:
It is going to take me a while to comprehend this.. do you have a branch/commit with comments included at all? 😅

JorgeBucaran:
Comments or just add here/below an explanation of what's happening there! 🙏 
/cc @Mytrill

src/app.js
```diff
@@ -58,33 +70,63 @@ export function app(props, container) {
...
+        (state[i] = {}),
+        (actions[i] = {}),
+        updateFor(update, get(getState, i), i),
+        get(getState, i)
+      )
+    }
+  }
+
+  function globalUpdate(result) {
+    console.log("Calling global update with result: " + JSON.stringify(result))
```
lukejacksonn: I guess this is here as a clue! 🕵️😅

lukejacksonn:
I haven't checked the diff.. but I think I prefer the approach here https://github.com/Mytrill/hyperapp/tree/immutable-state-2

src/app.js
Mytrill: Will remove this.

Mytrill:
https://github.com/Mytrill/hyperapp/tree/immutable-state-2 is much easier to understand (pass the current path to init() and initActions() and always get(globalState, path) to get the current slice or set(globalState, path, result) to merge in the actions results.

I will play around a bit more with both versions but probably end up doing another PR with the other branch, this one is too complicated, it's not worth the ~50 bytes difference (unless someone disagree, in which case I'll try to write down how the code in this PR works.).

JorgeBucaran:
@Mytrill Can you just push on top of this PR? Don't push force, just add another commit with the changes. 🙏

src/app.js
```diff
@@ -5,10 +5,20 @@ export function app(props, container) {
...
 
-  repaint(flush(init(props, (globalState = {}), (globalActions = {}))))
+  repaint(
+    flush(
+      initModule(
+        props,
+        globalState,
+        globalActions,
+        updateGlobalState,
+        getGlobalState
```
Pyrolistical: saves 3 bytes if we inline `getGlobalState` with
```
function () {
  return globalActions
}
```

Uncompressed
3346 bytes vs 3349

Mytrill: Done in the next commit.

src/app.js
```diff
@@ -58,33 +83,104 @@ export function app(props, container) {
...
-      init(module.modules[i], (state[i] = {}), (actions[i] = {}))
+    initModuleActions(state, actions, module.actions, update, getState)
+
+    for (var key in module.modules) {
+      initModule(
+        module.modules[key],
+        // do not override state is already exist in current module
+        state[key] || (state[key] = {}),
+        // do not override actions is already exist in current module
+        actions[key] || (actions[key] = {}),
```
Mytrill: Should solve #438.

JorgeBucaran: @Mytrill My PR doesn't have this, is it necessary?

JorgeBucaran:
Thanks @Mytrill! 🎉

PR #429, opened 2017-10-23
Remove duplicate comparison

### LINKED ISSUE #428: Don't perform the same comparison twice
Do you need to check `value !== oldValue` twice?
https://github.com/hyperapp/hyperapp/blob/a06b17ea2c851d6e873e38c63128d97d23d22378/src/app.js#L153-L155

pockethook:
Closes #428

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -151,7 +151,7 @@ export function app(props, container) {
       var oldValue = i === "value" || i === "checked" ? element[i] : oldProps[i]
 
       if (value !== oldValue) {
-        value !== oldValue && setElementProp(element, i, value, oldValue)
+        setElementProp(element, i, value, oldValue)
       }
     }
 


src/app.js
```diff
@@ -151,7 +151,7 @@ export function app(props, container) {
       var oldValue = i === "value" || i === "checked" ? element[i] : oldProps[i]
 
       if (value !== oldValue) {
-        value !== oldValue && setElementProp(element, i, value, oldValue)
```
lukejacksonn: Leaving this line but removing the whole `if` would have had the same effect (minus 2 bytes perhaps? ⛳️) but it doesn't really conform with the style of other binary checks in the codebase. I am happy with either really, up to you @JorgeBucaran

JorgeBucaran: @lukejacksonn Actually uglifyjs already takes care of it. This is the minified result without the repeated code.

```js
for (var r in l(n, t)) {
  var o = t[r],
    i = "value" === r || "checked" === r ? e[r] : n[r]
  o !== i && a(e, r, o, i)
}
```

lukejacksonn: Well what do you know! Good job uglify 🥇

Swizz: So, can we replace all binary check tricks with a real `if` block or in some cases this is optimised by hand ?

JorgeBucaran: @Swizz Yup, that's why I keep all the `if`s in there. We don't need to optimize those. 🎉

JorgeBucaran:
Thanks @pockethook! 🎉

Issue #432, opened 2017-10-25
Remove update & thunks.


JorgeBucaran:
Obviously this issue needs more info, but the core idea is to remove the concept of update and thunks unless we can demonstrate that using only actions is not enough to do everything that can be accomplished with actions _and_ thunks.

I'll update this issue with a bit of history about why thunks were introduced and why I think they are dispensable now that events and mixins are not part of our architecture.

An counter-alternative to this proposal: https://github.com/hyperapp/hyperapp/issues/433

EDIT: This is **just** a proposal and it may not happen! A different proposal may come out of this or nothing at all. We are just brooding together! ☝️🤓❤️

zaceno:
The two things I find thunks good for are:

A) Keeping the list of actions neatly limited to just actions that will be called from the UI (less need for "utility actions" that will only be called from other actions). However, I think making a generic `update` action in each module is a quite acceptable alternative.

B) Ability to get values computed from the current state/actions in the view. Like:

```js
actions: {
  calcSum: state => _ => state.a + state.b
}
```
This is not a good enough reason to keep thunks, but I would really like to see some replacement-way of creating getters -- if not through thunks, then some other way -- as it constitutes a significant convenience (and is not "cheating" with the purity IMO)

JorgeBucaran:
From the Slack.

## With thunks

```js
(state, actions) => update => {
  update({ fetching: true })
  fetch(...).then(data => update({
    data, fetching: false
  }))
}
```

## Without thunks!
```js
(state, { update }) => {
  update({ fetching: true })
  fetch(...).then(data => update({
    data, fetching: false
  }))
}```

/cc @Swizz @lukejacksonn

**Comment by JorgeBucaran** [comment]
Getters are another typical use case of thunks.

Say you are writing the next generation Warcraft-spinoff in Hyperapp, and your orcs module needs to compute some values based on your state slice. 

### With thunks as getters

```js
// orcs.js

export default {
  state { ... },
  actions: {
    getWorth(state) {
      return () => state.gems * state.gold * state.oil
    },
    attackHumans(state, actions) {
      // ...
      const worth = actions.getWorth()
      // ...
    }
  }
}
```

### Without thunks!

```js
// orcs.js

function getWorth({ gems, gold, oil }) {
  return gems * gold * oil
}

export default {
  state { ... },
  actions: {
    attackHumans(state, actions) {
      // ...
      const worth = getWorth(state)
      // ...
    }
  }
}

[...]

Issue #380, opened 2017-09-20
[RFC] What if hyperapp was just a bit less magical and a bit more transparent?


JorgeBucaran:
This is not a proposal to rewrite hyperapp, but just an attempt to gather some feedback from the community about our API.

If this were to happen it would be a major breaking change. This would be so major that one **might as well** publish it as a different "framework". 😉 

I'd just like to share the API I thought up with all of you so help me decide if this could ever be for us or if I should explore this idea elsewhere. 🤔 

Here it comes...

```js
import { h, app } from "hyperapp"
const { render, call } = app({ count: 1 })

document.body.appendChild(
  render(state => (
    <main>
      <h1>{state.count}</h1>
      <button onclick={() => call(state => ({ count: state.count + 1 }))}>
        +
      </button>
      <button
        onclick={() =>
          setTimeout(() => {
            call(state => ({ count: state.count - 1 }))
          }, 1000)}
      >
        -
      </button>
    </main>
  ))
)
```

Please give it 5 minutes and think about all the implications. See how by decoupling `render` and `update` from the app call that means we no longer need _actions_ or events. And mixins, well those can be implemented  in a manner similar to Redux enhancers, as the second argument to the `app()` call.

On the other hand you are responsible for initializing the app yourself and merging the state, actions and other stuff exported by mixins or component-like mixins.

I am not in favor or replacing the current API for this (**yet**), but I am starting to like it a lot. Like @lukejacksonn said, food for thought. 🍝 

EDIT: Corrected as pointed out by @okwolf.

okwolf:
> ```js
> <button
>   update(state => {
>     setTimeout(() => {
>       update(state => ({ count: state.count - 1 }))
>     }, 1000)
>   })}
> >
>   -
> </button>
> ```

@JorgeBucaran is the outer `update()` call even needed? Shouldn't calling `update` from inside `setTimeout` already do the job? 🤔

JorgeBucaran:
@okwolf It's needed so we can get a copy of the state.

JorgeBucaran:
BTW I called the functions `render` and `update`, but that's just a name. I think `render` works well, but `update` could be `dispatch`, `action`, `resolve`, `command`, `call`, etc.

zaceno:
I think it would be worth exploring, and I'd like to experiment with it. Perhaps we could create a branch for it? Publish it as hyperapp2000 or something.

[...]

PR #440, opened 2017-10-28
Add curried API for actions data

Swizz:
Some weeks ago, we were discussing about the magics behind Hyperapp and the ways to be more transparent. Here I made a proposition https://github.com/hyperapp/hyperapp/issues/380#issuecomment-330901225 now this on the way to be a real PR.

To sum up (a long long long war on slack), the proposition is about removing the comfusion regarding the actions you will write/define and the actions you will get/use.

At the moment, you are defining actions that way.
```ts
(state, actions, data) => function
```
But, when it will be time to use you action, you will use action with this signature
```ts
(data) => function
```
And I am pretty sure here is a lot of troubles here.

How many times have you answered a question about someone trying to use their actions that way ?
```jsx
actions.myaction(state, actions, data)
```

This is why I am coming at you with this proposition : 
```ts
(state, actions) => (data) => function
```

The following will be better on three points : 
1. The api will look the same all arround the app/entries 
```ts
(state, actions) => function
```
2. Users will understand more easily what they need to use
```ts
(data) => function
```
3. Actions are now able to use more than 1 argument
```ts
(foo, bar, baz, ... fizz) => function
```

in 11 bytes

---

```jsx
app({
  state: {
    count: 0
  },
  view: (state, actions) =>
    <main>
      <h1>
        {state.count}
      </h1>
      <button onclick={() => actions.subLater(1, 3000)} disabled={state.count <= 0}>ー (3 sec)</button>
      <button onclick={() => actions.sub(1)} disabled={state.count <= 0}>ー (now)</button>
      <button onclick={() => actions.add(1)}>＋ (now)</button>
      <button onclick={() => actions.addLater(1, 3000)}>＋ (3 sec)</button>
    </main>,
  actions: {
    sub: (state, actions) => (value) => ({ count: state.count - value }),
    add: (state, actions) => (value) => ({ count: state.count + value }),
    subLater: (state, actions) => (value, time) => update => {
      setTimeout(() => {
        update(state => state.count <= 0 ? {} : { count: state.count - value })
      }, time)
    },
    addLater: (state, actions) => (value, time) => update => {
      setTimeout(() => {
        update(state => ({ count: state.count + value }))
      }, time)
    }
  }
})
```

Now time for questions :smiley:

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -69,9 +69,11 @@ export function app(props, container) {
     Object.keys(source || {}).map(function(i) {
       if (typeof source[i] === "function") {
         actions[i] = function(data) {
-          return typeof (data = source[i](state, actions, data)) === "function"
-            ? data(update)
-            : update(data)
+          return typeof (data = source[i](state, actions)) !== "function"
+            ? update(data)
+            : typeof (data = data.apply(null, arguments)) !== "function"
+              ? update(data)
+              : data(update)
         }
       } else {
         initActions(state[i] || (state[i] = {}), (actions[i] = {}), source[i])
--- a/src/app.js
+++ b/src/app.js
@@ -69,11 +69,11 @@ export function app(props, container) {
     Object.keys(source || {}).map(function(i) {
       if (typeof source[i] === "function") {
         actions[i] = function(data) {
-          return typeof (data = source[i](state, actions)) !== "function"
-            ? update(data)
-            : typeof (data = data.apply(null, arguments)) !== "function"
-              ? update(data)
-              : data(update)
+          return typeof (data = source[i](state, actions)) === "function"
+            ? typeof (data = data.apply(0, arguments)) === "function"
+              ? data(update)
+              : update(data)
+            : update(data)
         }
       } else {
         initActions(state[i] || (state[i] = {}), (actions[i] = {}), source[i])
--- a/test/actions.test.js
+++ b/test/actions.test.js
@@ -33,10 +33,19 @@ test("slices", done => {
     actions: {
       foo: {
         bar: {
-          baz(state, actions, data) {
-            expect(state).toEqual({ baz: "minimal baz" })
-            expect(data).toBe("foo.bar.baz")
-            return { baz: "only the baz will do" }
+          baz(state, actions) {
+            return data => {
+              expect(state).toEqual({ baz: "minimal baz" })
+              expect(data).toBe("foo.bar.baz")
+              return { baz: "only the baz will do" }
+            }
+          },
+          buz(state, actions) {
+            return (data1, data2, data3) => {
+              expect(data1).toBe("foo")
+              expect(data2).toBe("bar")
+              expect(data3).toBe("baz")
+            }
           }
         }
       },
@@ -49,6 +58,7 @@ test("slices", done => {
   })
 
   actions.foo.bar.baz("foo.bar.baz")
+  actions.foo.bar.buz("foo", "bar", "baz")
   actions.fizz.buzz.fizzbuzz()
 })
 
@@ -98,15 +108,17 @@ test("async updates", done => {
       value: 2
     },
     actions: {
-      up(state, actions, byNumber) {
-        return {
+      up(state, actions) {
+        return byNumber => ({
           value: state.value + byNumber
-        }
-      },
-      upAsync(state, actions, byNumber) {
-        mockDelay().then(() => {
-          actions.up(byNumber)
         })
+      },
+      upAsync(state, actions) {
+        return byNumber => {
+          mockDelay().then(() => {
+            actions.up(byNumber)
+          })
+        }
       }
     }
   }).upAsync(1)
@@ -132,8 +144,8 @@ test("thunks", done => {
       value: 3
     },
     actions: {
-      upAsync(state, actions, data) {
-        return update => {
+      upAsync(state, actions) {
+        return data => update => {
           mockDelay().then(() => {
             update({ value: state.value + data })
           })
@@ -163,8 +175,8 @@ test("thunks", done => {
       value: 4
     },
     actions: {
-      upAsync(state, actions, data) {
-        return update => {
+      upAsync(state, actions) {
+        return data => update => {
           mockDelay().then(() => {
             update(state => ({ value: state.value + data }))
           })


JorgeBucaran:
@Swizz Thanks! I am not sure if I am going to move forward with this, but I still need more time to come up with my own proposal.

Just to address one of your questions.

> How many times have you answered a question about someone trying to use their actions that way ?

```js
actions.myaction(state, actions, data)
```

I haven't, but I imagine not too few people must have found this a bit confusing, I admit that.

Now, does that mean the same people that found the above confusing would be calling actions like this now?

```js
actions.myaction(state, actions)(data)
```

olstenlarck-old:
Just enforce destructuring. It is just fantastic language feature/tool which gives very big options. Once you have `({ state, actions }, ...args)` you are ready to go everywhere, and currying won't be needed cuz it is small enough. Not to mention that it enforce conventions which is awesome. One more thing is that it leads to lower number of arguments (mAx-ParaAmMMMS MUSsST bEe 3! !1! ;d)

JorgeBucaran:
@Swizz In the original draft of Hyperapp, we had two app props `reducers` and `effects` instead of one app prop `actions` like now. So, it was not a big jump to think `actions` were more that just a copy of your reducers and effects in an object.

When the API changed to only actions, I was able to simplify things around, but created another problem, the one you are describing.

I agree that a way to deal with this is to use destructuring across the documentation. Another option is to call the actions we give you something else, for example: `dispatch`, `send` or `call`.

vdsabev:
Hey @charlike-old, funny to bump into your here :)

Personal observation - whenever someone uses the word _just_ in a sentence, I feel like there is hidden complexity that is _just_ obscured by that word 🙃 

In the case you described, `({ state, actions }, ...args)` doesn't really address the issue that when you define your app, you have a signature of `(state, actions, data)` for your actions, and when you actually use them, the signature is just `(data)`. Example:
```jsx
app({
  state: { name: '' },
  actions: {
    setName: (state, actions, name) => ({ name }),

    // Contrived example, I know I could `return { name: theNewValue }` instead
    setNameToInputValue(state, actions, e) {
      // Wrong! But this is what I expected at first, and what TypeScript or Flow can infer at best
      actions.setName(state, actions, e.currentTarget.value); // Oops
      // That's right, the parameter signature was different than in the definition
      actions.setName(e.currentTarget.value);
    },
    clearName(state, actions) {
      // Wrong! But this is what I expected at first, and what TypeScript or Flow can infer at best
      actions.setName(state, actions, ''); // Oops
      // That's right, the parameter signature was different than in the definition
      actions.setName('');
    }
  },
  view: (state, actions) =>
    <div>
      <input type="text" oninput={actions.setNameToInputValue} />
      <button type="button" oninput={actions.clearName}>Clear Name</button>
      <div>Hello {state.name}!</div>
    </div>
});
```

I'll now add in some points from discussions in Slack.

## Renaming the `actions` argument of functions
One way we discussed we could address this is if we rename `actions` to something else when using them, for example `dispatch`:

```jsx
app({
  state: { name: '' },
  actions: {
    setName: (state, dispatch, name) => ({ name }),

    setNameToInputValue(state, dispatch, e) {
      dispatch.setName(e.currentTarget.value);
    },
    clearName(state, dispatch) {
      dispatch.setName('');
    }
  },
  view: (state, dispatch) =>
    <div>
      <input type="text" oninput={dispatch.setNameToInputValue} />
      <button type="button" oninput={dispatch.clearName}>Clear Name</button>
      <div>Hello {state.name}!</div>
    </div>
});
```

This way it would be clearer that the actions we're defining are somewhat different from what we get afterwards. We could also name the argument `do`, `call` or `bamboozle` (I think that one's going to be the community favorite 😆).

However, that alone isn't going to be enough to enable type safety in TypeScript or Flow - someone please correct me if I'm wrong, but we can't really map `(state, dispatch, name: string): Partial<State>` to `(name: string): Partial<State>`.

## Destructuring the `actions` argument of functions
To avoid having to name that argument, we can also use destructuring (in a different way than @charlike-old meant):
```jsx
app({
  state: { name: '' },
  actions: {
    // Except we don't need any actions here, so no destructuring,
    // and the question of what to name the second parameter remains
    setName: (state, _, name) => ({ name }),

    setNameToInputValue(state, { setName }, e) {
      setName(e.currentTarget.value);
    },
    clearName(state, { setName }) {
      setName('');
    }
  },
  view: (state, { setNameToInputValue, clearName }) =>
    <div>
      <input type="text" oninput={setNameToInputValue} />
      <button type="button" oninput={clearName}>Clear Name</button>
      <div>Hello {state.name}!</div>
    </div>
});
```

Some observations here:
1. As you can see, this doesn't avoid the issue of naming the argument when an action is _not_ used, like in `setName`
2. It doesn't really solve the issue that something that is _only_ called `setName` has 2 different signatures - `(state, actions, data)` and `(data)`. I think the distinction between `actions.setName(state, actions, data)` and `dispatch.setName(data)` is much clearer
3. Destructuring actions would prevent `F2 -> Rename` from working in TypeScript due to Microsoft/TypeScript#6312 and Microsoft/TypeScript#15787
4. It's still not type safe

## @Swizz's proposal
Combining this PR's proposal with renaming the `actions` argument in functions would resolve the naming discrepancy _and_ enable type safety for actions - in TypeScript after Microsoft/TypeScript#6606 is done, and in Flow using [`$Call`](https://github.com/facebook/flow/commit/ac7d9ac68acc555973d495f0a3f1f97758eeedb4). Example:
```jsx
app({
  state: { name: '' },
  actions: {
    setName: (state, dispatch) => (name: string) => ({ name }),

    setNameToInputValue: (state, dispatch) => (e: Event) => {
      dispatch.setName(1);     // Oops, TypeScript error, `setName` expects a string
      dispatch.setName((e.currentTarget as HTMLInputElement).value);
    },
    clearName: (state, dispatch) => () => {
      dispatch.setName(null); // Oops, TypeScript error, `setName` expects a string
      dispatch.setName('');
    },
    // Potentially confusing that there are no args here? Worried someone might do this?
    // Well, then Hyperapp could throw an error if it doesn't get a function
    clearNameWrongDefinition: (state, dispatch) => {
      dispatch.setName('');
      // We're not returning a function, so Hyperapp throws an error
    }
  },
  view: ...
});
```

To conclude, this looks like the best of both worlds. So...I am @vdsabev and I support this PR 😄

JorgeBucaran:
@vdsabev Unfortunately, there is another issue. I want to distinguish reducers from effects, but I don't want to add a new app prop. The solution we've been discussing lately on Slack fixes #432 and involves not passing actions to actions, but to thunks.

```js
actions: {
  // Reducers
  increment: state => ({ count: state.count + 1 }),
  incrementBy: (state, data) => ({ count: state.count + data })
  
  // Side Effects
  incrementLater: state => actions => setTimeout(() => actions.increment(), TIMELAPSE)
}
```

I think the separation of reducers and effects is also important and perhaps more than users getting confused about how to call actions which can be ameliorated in several ways as @vdsabev explains and better documentation.

Tough one.

JorgeBucaran:
@Swizz @vdsabev I have another issue with this proposal, that I hadn't noticed before. 

```js
const actions = {
  increment: state => ({ count: state.count + 1 }),
  incrementBy: state => data => ({ count: state.count + data }),
}
```

Leaving effects aside, there are apparently two kind of actions above. Why would it be obvious to anyone that increment is invoked `actions.increment()` and incrementBy `actions.incrementBy(10)`.

vdsabev:
@JorgeBucaran does `(state, data) => (actions)` enable some performance optimization vs `(state, actions, data)` or `(state, actions) => (data)`? Is there something other than conceptual separation of reducers and effects?

What I've noticed is I can still make pseudo reducers and pseudo side effects that don't really do anything:
```jsx
actions: {
  // Pseudo reducer :)
  justLogIt: (state) => {
    console.log('Not a reducer, not doing anything to the state or calling actions');
  },
  
  // Pseudo side effect
  justLogItLater: (state) => (actions) => {
    console.log('Not a side effect, not doing anything to the state or calling actions');
  },
  
  // Maybe both reducer and side effect? Not really different from how we do things right now
  logAndIncrement: (state) => (actions) => {
    actions.justLogIt();
    return { count: state.count + 1 };
  }
}
```

Or would any of these be "forbidden" / discouraged?

vdsabev:
@JorgeBucaran regarding your second point, I think with a `(state, actions) => (data)` signature, action definitions should _always_ return a function. Еven if the action doesn't need data, we should define it like `(state, actions) => ()`. As I said in my long-wall-of-text comment, Hyperapp can enforce this by checking whether the result of the function in `actions` is a function or not:
```
app({
  state: { name: '' },
  actions: {
    setName: (state, dispatch) => (name) => ({ name }),
    clearName: (state, dispatch) => () => ({ name: '' }),
    clearNameIncorrectly: (state, dispatch) => ({ name: '' }) // Hyperapp throws an error here
  },
  view: ...
});
```

JorgeBucaran:
> @vdsabev Is there something other than conceptual separation of reducers and effects?

It's conceptual at this point. But a good foundation is built on good concepts.

In your code: `justLogIt` and `justLogItLater` are side effects because the first one returns undefined and the second one returns a function. The first one is just a "quirk" of JavaScript, which is not a FL and really a non-issue to everyone. The second one is how I think effects should be implemented from now on. It's a user friendly compromise between the JavaScript most people know and FP/Elm.

A more extreme way involves triggering effects more awkwardly:  `[newState, actions.incrementBy, 100]`.

Finally, `logAndIncrement` is actually incorrect and does not work in the API described [here](https://github.com/hyperapp/hyperapp/pull/440#issuecomment-340241946).

JorgeBucaran:
@vdsabev To address your follow post, I feel enforcing that creates a new problem. It makes the API awkward.

```js
app({
  actions: {
    incrementBy(state) {
      return data => ({
        count: state.count + data
      })
    }
  }
})
```


VS 


```js
app({
  actions: {
    incrementBy(state, data) {
      return {
        count: state.count + data
      }
    }
  }
})
```

JorgeBucaran:
What about the perception of the initial state versus the state you receive in the view and actions?

I have never heard anyone find fault in the fact the state that you pass to the app() is only the initial state and the state you receive inside the view and actions is the _current_ state.

vdsabev:
@JorgeBucaran I've been thinking about `initialState` vs `state`, but the difference is it's an object, not a function, so it doesn't have a different parameter signature, and you use `state` the exactly same way you would use `initialState`.

The only potential issue I see there is someone thinking they have to mutate the state to make actual changes. Which means they've completely missed the point of Hyperapp.

vdsabev:
After some investigation, it looks like we won't be able to achieve perfect automatic typings, even if we had `(s, a) => (d)` and Microsoft/TypeScript#6606 was resolved.

If you take a look at [this pen](https://codepen.io/vdsabev/pen/eeJaor) where I've tried writing Hyperapp's types from scratch, you'll see a couple of challenges that I couldn't overcome:
1. Action names can't be type checked (or reliably renamed with F2), so you could type `actions.avadaKedavra()` and the compiler can't detect it's an invalid action name
2. Additional action parameter types can't be type checked, so you could type `actions.incrementBy('a')` and the compiler can't detect it's an invalid parameter type

Perhaps someone with more TypeScript experience than me could resolve those issues, but I don't know how to right now. It's possible Microsoft/TypeScript#5453 could help to some extent, but then writing types for `(s, a, d)` would probably be just as easy as writing types for `(s, a) => (d)`.

I still think `(s, a) => (d)` makes sense, and @lukejacksonn put it best:
```
   hyperapp          user
       v              v
(state, actions) => (data) => ({ ... })
```
But as far as types go, I don't see an advantage right now, or in the near future.

olstenlarck:
@vdsabev Дооооооооооооообре, just ме нави с първият си коментар. :D 

Ahh, types. Is Flow have same problems as TS, cuz can't try now?

olstenlarck:
I think all of this reducers should work

```js
const reducer = (state) => (actions) => (data) => {}

const reducer = (state) => (actions) => (data) => ({ foo: 123 + state.num })
const reducer = (state) => (actions) => ({ foo: 123 + state.num })
const reducer = (state) => ({ foo: 123 + state.num })

const reducer = (state) => (actions, data) => ({ foo: 123 + state.num })
const reducer = (state) => (actions) => ({ foo: 123 + state.num })

const reducer = (state, actions, data) => ({ foo: 123 + state.num })
const reducer = (state, actions) => ({ foo: 123 + state.num })
```

Really i still don't getting what's so hard to understand reducers/effect by the most of users.

**Reducers** for me:

**1)** absolutely pure functions;
**2)** absolutely synchronous;
**3)** must return an object;
**4)** it's good to assume that they return whole state (not partial/chunk) and give to user a helper that accept partial and merge it with the whole state - that's easier to implement and looks good too.
**5)** in rare cases you'll call another reducers, and never effects;
**6)** even to not have any access to effects. 


| rules        | reducers | effects |
|:------------- |:-------------:|:-----:|
| update/replace state | :heavy_check_mark: | not directly |
| access to state | :heavy_check_mark: | :heavy_check_mark: |
| access to reducers | :heavy_check_mark: | :heavy_check_mark: |
| access to effects | :x: | :heavy_check_mark: |
| synchronous | :x: | :heavy_check_mark: |
| pure functions | :heavy_check_mark: | :x: |
| must return value | :heavy_check_mark: | :x: |
| must return object | :heavy_check_mark: | :x: |
| are callers | :x: | :heavy_check_mark: |
| are called | :heavy_check_mark: | :heavy_check_mark: |

Reducers are folks that are called, not _callers_.

I have always been for `reducers/effects` approach instead of one single object `actions` and @JorgeBucaran nows that, imo. And i (still) think it was wrong move. I totally love that concept of sliced states and actions, but it is a bit harder - not so, but a bit bigger abstraction, user should think first and even more must realize that there are relation between the state keys and action keys and etc. It's brilliant idea, definitely.

If you have separate reducers and effects you don't need thunks too. Once you need the state to be updated you just call the specific reducer, this works for promises and everything. That way everything is absolutely clear and i don't see any possible problems.

We still can support nested reducers and nested effects so to have some namespaces.

JorgeBucaran:
Accepted! Thanks @Swizz! 🎉

PR #441, opened 2017-11-01
Allow app to be loaded as a ES6 module

SahAssar:
This is the only change needed to be able to use hyperapp as a ES6 module in a browser, while it still remains compatible with node and bundlers.

It changes the path to the h file to "./h" from "./h.js", since browsers will not auto-append the .js suffix like many bundlers/transpilers do.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "jest": "^21.2.1",
     "prettier": "~1.7.4",
     "rollup": "^0.50.0",
-    "typescript": "^2.5.2",
+    "typescript": "~2.5.2",
     "uglify-js": "^2.7.5"
   }
 }
--- a/src/app.js
+++ b/src/app.js
@@ -1,4 +1,4 @@
-import { h } from "./h"
+import { h } from "./h.js"
 
 export function app(props, container) {
   var root = (container = container || document.body).children[0]


SahAssar:
Hmm, I don't really understand the test failure though, could someone point me in the right direction to fix it (if it is related to this change)?

JorgeBucaran:
Thanks @SahAssar! This looks deceptively simple. 🎉 

The declarations file is complaining. @Swizz @Mytrill can you have a look?

Mytrill:
Ok so, it seems to be due to typescript version. Even though typescript has the version `^2.5.2` in the `package.json`, it seems the latest `2.6.1` is used to compile `tests/ts/index.tsx` and looks like this version has broken the typings. 
I'll take a look when I have time but it will not likely be before Saturday/Sunday, @Swizz feel free to have a look...

Mytrill:
Separate issue to track the typings problems: https://github.com/hyperapp/hyperapp/issues/442

JorgeBucaran:
@SahAssar Do you know any other libraries doing something similar?

SahAssar:
@JorgeBucaran You mean suffixing `.js` to allow for module loading? In that case, no, but I'm not that involved in other communities, so I wouldn't know.

JorgeBucaran:
@SahAssar We also need to change index.js to add the .js extension.

Issue #421, opened 2017-10-18
Minified UMD source maps broken


zaceno:
I served the following simple app on `localhost:8082` (using the CDN distributed minified umd bundle of hyperapp):

```
<!doctype html>
<html>
    <body>
        <script src="https://unpkg.com/hyperapp@latest/dist/hyperapp.js"></script>
        <script>
            const {h, app} = hyperapp
            app({
                state: {value: 0},
                actions: {
                    up: state => ({value: state.value + 1}),
                    down: state => ({value: state.value - 1}),
                },
                view: (state, actions) => h('div', {}, [
                    h('button', {onclick: actions.down}, ['-']),
                    state.value,
                    h('button', {onclick: actions.up}, ['+']),
                ])
            })
        </script>
    </body>
</html>
```

The app works of course. But looking in the sources tabs is interesting. Chrome says "source maps detected" and gives me two "hyperapp.js" sources. One which is the regular, minfied file, and one which is claimed to be "source mapped" -- but * it is identical to the minified file*.

So I think source maps are just mapping to the minified bundle, not the true source, giving the *appearance* that that no source mapping is taking place at all.

The normal, not mapped minified umd file:

<img width="815" alt="screen shot 2017-10-18 at 16 16 59" src="https://user-images.githubusercontent.com/2061445/31724085-0a787b70-b421-11e7-9277-7d78c35ec9fe.png">

The mapped source:

<img width="808" alt="screen shot 2017-10-18 at 16 17 27" src="https://user-images.githubusercontent.com/2061445/31724113-23a70bac-b421-11e7-9bc6-e8cab9d2c10a.png">

PR #443, opened 2017-11-01
fix sourcemaps

zaceno:
Of course, we'll have to wait for a new release of hyperapp, after this PR is merged, before source maps are usable in cdn-scenarios like codepen. But now `npm run build` generates a proper source map at `dist/hyperapp.js.map`

fixes: https://github.com/hyperapp/hyperapp/issues/421

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "test": "jest --coverage --no-cache && tsc --project test/ts",
     "build": "npm run bundle && npm run minify",
     "bundle": "rollup -i src/index.js -o dist/hyperapp.js -m -f umd -n hyperapp",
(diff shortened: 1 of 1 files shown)


zaceno:
Ok this is wierd. Travis fails, because of something to do with the type tests? (Which I obviously haven't touched). Is this related to the thing @SahAssar and @Mytrill were talking about on the slack?

SahAssar:
@zaceno Yeah, as soon as my PR is merged it will probably pass. The problem is TSC released 2.6 and hyperapp requires ^2.5.X instead of ~2.5.X so it'll pull in 2.6 which breaks the tests currently (I think due to some change behavior in regards to `undefined`)

zaceno:
@SahAssar thanks! Ok I'll sit tight and wait for your PR then debase.

JorgeBucaran:
Thanks @zaceno! 🎉

Issue #442, opened 2017-11-01
Typings are broken with Typescript 2.6.X


Mytrill:
Looks like the latest version of typescript broke the typings.

Here is the error when trying to compile `test/ts/index.tsx` with typescript `2.6.1`:
```
test/ts/index.tsx(48,7): error TS2322: Type '{ state: { count: number; }; actions: { sub: (state: Module1State, actions: Module1OwnActions & A...' is not assignable to type 'Module<Module1Stat
e, Module1Actions, Module1OwnState, Module1OwnActions>'.
  Types of property 'modules' are incompatible.
    Type '{ submodule: Module<SubModuleState, SubModuleActions, Partial<SubModuleState>, Partial<SubModuleA...' is not assignable to type 'Modules<Partial<Module1State> & Record<string, any>,
Partial<Module1Actions> & Actions<Module1Sta...'.
      Type '{ submodule: Module<SubModuleState, SubModuleActions, Partial<SubModuleState>, Partial<SubModuleA...' is not assignable to type 'Modules<Partial<Module1State> & Record<string, any>
, Partial<Module1Actions> & Actions<Module1Sta...'.
        Types of property 'submodule' are incompatible.
          Type 'Module<SubModuleState, SubModuleActions, Partial<SubModuleState>, Partial<SubModuleActions>>' is not assignable to type 'Module<SubModuleState | undefined, SubModuleActions & A
ctions<SubModuleState | undefined>, Partia...'.
            Type 'Module<SubModuleState, SubModuleActions, Partial<SubModuleState>, Partial<SubModuleActions>>' is not assignable to type 'Module<SubModuleState | undefined, SubModuleActions &
 Actions<SubModuleState | undefined>, Partia...'.
              Type 'SubModuleState | undefined' is not assignable to type 'SubModuleState'.
                Type 'undefined' is not assignable to type 'SubModuleState'.
test/ts/index.tsx(124,7): error TS2322: Type '{ async: InternalActions<AsyncModuleState, AsyncModuleActions>; }' is not assignable to type 'InternalActions<AppState, AppActions>'.
  Property 'async' is incompatible with index signature.
    Type 'InternalActions<AsyncModuleState, AsyncModuleActions>' is not assignable to type 'InternalAction<AppState, AppActions> | InternalActions<any, Actions<any>> | InternalActions<any, ...
'.
[...]

PR #445, opened 2017-11-03
0.16.0

### LINKED ISSUE #422: Don't lock render before calling view function.
Recently I have been playing around with modules that map external state of things like `mouse`, `scroll` and `location` to hyperapp state. I have observed that it certain cases when `update` is called in rapid succession; that the data used in the last render is not the data in the state.

If I remove the skip render check (but leave rAF), the behaviour goes away:
```
function repaint() {
  if (props.view) { requestAnimationFrame(render) }
}
```

I started looking at [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) as I haven't used it since I experimented with WebGL years ago. From what I can gather:

- Requests that the browser call a specified function before the next repaint
- You should call this method whenever you're ready to update your view onscreen

It doesn't say anything about skipping renders 🤔 and it makes me wonder why we are doing this? I have always thought of `requestAnimationFrame` as like a smart debounce/throttle which you throw whatever you want at, and it makes sure not to over compute.

Looking at the current setup it seems that a state update `assign(state, data)` could be made, which calls `repaint()` when `skipRender` is true meaning that `render` is never called. This would explain the behaviours I have witnessed.

Interestingly removing all trace of `skipRender` seems to have no adverse effects at all on hyperapp even when challenged with rapid state updates. Do we need it for some reason I am not remembering?

### LINKED ISSUE #425: The state should be immutable

### LINKED ISSUE #432: Remove update & thunks.

JorgeBucaran:
- Remove init in favor of using the app() returned actions. 
- Immutable state. Close #425. Thanks @Mytrill. 🎉 
- No-thunks. Close #432.
- Curry actions' data argument. Thanks @Swizz. 🎉 
- Allow calling actions in the view. Close #422.
- Actions now return what the action implementation returns.


okwolf:
Size impact?

zaceno:
@okwolf @JorgeBucaran  Yes all PR's are gonna need a fix to #442. @SahAssar has the fix in his: https://github.com/hyperapp/hyperapp/pull/441 (that's why his PR has a green check, and ours don't 😄

JorgeBucaran:
@okwolf 1499 B

zaceno:
@JorgeBucaran It grew?! I thought removing thunks would make it much smaller? Oh but immutable state... Is that where the bulk of the increase is?

JorgeBucaran:
@zaceno Yup.

src/app.js
```diff
@@ -49,54 +41,89 @@ export function app(props, container) {
...
+                actions
+              )) === "function"
+            ) {
+              state = state(data)
+            }
+
+            if (
+              state &&
+              !state.then &&
+              state !== (data = getObject(path, globalState))
```
Mytrill: No more getters? :(

JorgeBucaran: 🤕 No more getters.

src/app.js
```diff
@@ -151,11 +178,11 @@ export function app(props, container) {
       var oldValue = i === "value" || i === "checked" ? element[i] : oldProps[i]
 
       if (value !== oldValue) {
-        setElementProp(element, i, value, oldValue)
+        value !== oldValue && setElementProp(element, i, value, oldValue)
```
lukejacksonn: Double checking for `value !== oldValue` this was perviously fixed by https://github.com/hyperapp/hyperapp/pull/429

JorgeBucaran: 😅 Whoops. Thanks!

src/app.js
```diff
@@ -1,37 +1,29 @@
...
-  repaint(flush(init(props, (globalState = {}), (globalActions = {}))))
+  repaint(init(props, globalState, globalActions, []))
 
   return globalActions
 
   function repaint() {
-    if (props.view && !skipRender) {
-      requestAnimationFrame(render, (skipRender = !skipRender))
+    if (props.view && !lock) {
+      setTimeout(render, (lock = !lock))
```
SkaterDad: What was the reasoning for changing `requestAnimationFrame` to `setTimeout`?

src/app.js
Swizz: You can make a getter that return a state slice without triggering a render isn't it ?

src/app.js
Swizz: From a discussion on slack @lukejacksonn found some drawbacks to `requestAnimationFrame` (not working on hidden tabs) plus the fact we needed to mock it for testing.

src/app.js
JorgeBucaran: @Swizz Yes.

src/app.js
JorgeBucaran: setTimeout does not have that problems and still allow us to "debounce" re-renders after calling multiple actions in rapid-fire.

src/app.js
```diff
@@ -49,54 +41,89 @@ export function app(props, container) {
...
-            ? data(update)
-            : update(data)
+    function create(state, actions, from, path) {
+      Object.keys(from || {}).map(function(i) {
+        if (typeof from[i] === "function") {
+          actions[i] = function(data) {
+            if (
+              typeof (state = from[i](
+                getObject(path, globalState),
+                actions
```
Swizz: Is this the slice actions ?

JorgeBucaran: Yes.

PR #449, opened 2017-11-07
Remove unused variable

vekat:
I believe a cheeky var managed to elude linting. She's gone now.


JorgeBucaran:
Thanks @vekat! Nice catch. 👍

PR #450, opened 2017-11-08
Anonymous actions

Swizz:
### tl;dr : anonymous actions are actions not living in the actions object

Another proposal wanting to add ease about the use of actions. Async ones.

You all know, when thunk disappeared, it do not let us without any way to perform
async actions needing the most up to date state.  
This is because we were still, since a while, able to call sync actions from our
async actions.

Lets use a basic fetch action that fetch an array of users an add it to the current
users array.

```js
actions: {
  addUsers: state => users => ({ users: state.users.concat(users) })
  fetchNewUsers: (state, actions) => 
    fetch("/newusers")
      .then(res => res.json())
      .then(data => actions.addUser(data.users))
}
```

So as I said, we need two actions to perform this operation, the effects that
make the fetch and the reducer to update the state.

The proposition as you will do with function, is to allow you to write action in place : anonymous actions
So, you will dont have to write the `addUsers` function.

Mh. Yes, you will have to do this but only the implementation.

```js
actions: {
  fetchNewUsers: (state, actions) => 
    fetch("/newusers")
      .then(res => res.json())
      .then(data => state => ({ users: state.users.concat(data.users) }))
}
```

If you did not catch it, the anonymous action is right here :

```js
data => state => ({ users: state.users.concat(data.users) })
```

This one is exacly the same as the previous `addUsers` but anonymously.

Except one thing, the data position, because here you deal with the Promise resolving
data, exacly like you did with ... (the data live in the closure)

```js
.then(data => actions.addUser(data.users))
```

This PR include one more feature, a way to update with a constant partial (partials that dont need state).

**Why ?**

Anonymous actions allow you to catch the most up do date state, that is why they exist
and why you earlier use another action as reducer.

But sometime, you will just want to use the Promise resloving data as partial.

Actual :
```js
actions: {
  setUser: () => (user) => ({ user: user }),
  fetchUser: (state, actions) => 
    fetch("/user")
      .then(res => res.json())
      .then(user => actions.setUser(user)) // or .then(actions.setUser)
}
```

With anonymous actions :
```js
actions: {
  fetchUser: (state, actions) => 
    fetch("/user")
      .then(res => res.json())
      .then(user => ({ user: user }))
}
```
Following my implementation, you will notice one thing, as we consider anonymous 
actions as regular actions, constant actions also exist in the named world.

So, this will add as side effect another way to deal with

```js
actions: {
  reset: () => ({ counter: 0 })
}
```

By simply giving a constant partial 
```js
actions: {
  reset: ({ counter: 0 })
}
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -55,18 +55,30 @@ function app(props, container) {
   function initDeep(state, actions, from, path) {
     Object.keys(from || {}).map(function(key) {
       if (typeof from[key] === "function") {
-        actions[key] = function(data) {
-          var result = from[key]((state = get(path, appState)), actions)
+        function dispatch(action, data) {
+          var result = action((state = get(path, appState)), actions)
 
           if (typeof result === "function") {
             result = result(data)
           }
 
-          if (result && result !== state && !result.then) {
-            repaint((appState = setDeep(path, merge(state, result), appState)))
+          if (result && result !== state) {
+            if (result.then) {
+              result = result.then(dispatch)
+            } else if (result.subscribe) {
+              result = result.subscribe({ next: dispatch })
+            } else {
+              repaint(
+                (appState = setDeep(path, merge(state, result), appState))
+              )
+            }
           }
 
           return result
+        }
+
+        actions[key] = function(data) {
+          return dispatch(from[key], data)
         }
       } else {
         initDeep(
--- a/src/index.js
+++ b/src/index.js
@@ -56,7 +56,12 @@ function app(props, container) {
     Object.keys(from || {}).map(function(key) {
       if (typeof from[key] === "function") {
         function dispatch(action, data) {
-          var result = action((state = get(path, appState)), actions)
+          var result
+          if(typeof action !== "function") {
+            result = action
+          } else {
+            result = action((state = get(path, appState)), actions)
+          }
 
           if (typeof result === "function") {
             result = result(data)
--- a/test/actions.test.js
+++ b/test/actions.test.js
@@ -2,6 +2,8 @@ import { h, app } from "../src"
 
 const mockDelay = () => new Promise(resolve => setTimeout(resolve, 50))
 
+const mockObservable = (fn) => ({ subscribe: ({ next }) => { setInterval(next, 50, fn()) } })
+
 beforeEach(() => {
   document.body.innerHTML = ""
 })
@@ -93,8 +95,61 @@ test("async updates", done => {
       ),
     actions: {
       up: state => data => ({ value: state.value + data }),
-      upAsync: (state, actions) => data =>
+      upAsync: (state, actions) => data => {
         mockDelay().then(() => actions.up(data))
+      }
     }
   }).upAsync(1)
+})
+
+test("anonymous actions promise", done => {
+  app({
+    state: {
+      value: 2
+    },
+    view: state =>
+      h(
+        "div",
+        {
+          oncreate() {
+            expect(document.body.innerHTML).toBe(`<div>2</div>`)
+          },
+          onupdate() {
+            expect(document.body.innerHTML).toBe(`<div>3</div>`)
+            done()
+          }
+        },
+        state.value
+      ),
+      actions: {
+        upAsync: (state, actions) => data =>
+          mockDelay().then(() => (state) => ({ value: state.value + data }))
+      }
+  }).upAsync(1)
+})
+
+test("anonymous actions observable", done => {
+  app({
+    state: {
+      value: 2
+    },
+    view: state =>
+      h(
+        "div",
+        {
+          oncreate() {
+            expect(document.body.innerHTML).toBe(`<div>2</div>`)
+          },
+          onupdate() {
+            expect(document.body.innerHTML).toBe(`<div>3</div>`)
+            done()
+          }
+        },
+        state.value
+      ),
+    actions: {
+      upObserve: (state, actions) => data =>
+        mockObservable(() => (state) => ({ value: state.value + data }))
+    }
+  }).upObserve(1)
 })
--- a/test/actions.test.js
+++ b/test/actions.test.js
@@ -153,3 +153,51 @@ test("anonymous actions observable", done => {
     }
   }).upObserve(1)
 })
+
+test("constant", done => {
+  app({
+    state: {
+      value: 2
+    },
+    view: state =>
+      h(
+        "div",
+        {
+          oncreate() {
+            expect(document.body.innerHTML).toBe(`<div>3</div>`)
+            done()
+          }
+        },
+        state.value
+      ),
+      actions: {
+        set: (state, actions) => ({ value: 3 })
+      }
+  }).set()
+})
+
+test("async updates promise", done => {
+  app({
+    state: {
+      value: 2
+    },
+    view: state =>
+      h(
+        "div",
+        {
+          oncreate() {
+            expect(document.body.innerHTML).toBe(`<div>2</div>`)
+          },
+          onupdate() {
+            expect(document.body.innerHTML).toBe(`<div>3</div>`)
+            done()
+          }
+        },
+        state.value
+      ),
+    actions: {
+      setAsync: (state, actions) =>
+        mockDelay().then(() => ({ value: 3 }))
+    }
+  }).setAsync()
+})


olstenlarck:
Or more easily said, the proposed actions are functions that returns a function which is passed with state -
 currently that function is passed with `update`.

You are getting more and more view what i had before several months and around when [freactal](https://github.com/FormidableLabs/freactal) appeared.

Swizz:
But we do not give `update` to users anymore, they will need to give us an action we call for them.
And as bonus, we allow them to write actions on the fly.

This could be a little bit more understandable using async/await.

```js
actions: {
  addUsers: state => users => ({ users: state.users.concat(users) })
  async fetchUser(state, actions) {
    const res = await fetch("/newusers")
    const data = await res.json()
    
    actions.addUsers(data.users)
  }
}
```

Now you will be allowed to return an anonymous actions that will be called once the async actions is resolved

```js
actions: {
  async fetchUser(state, actions) {
    const res = await fetch("/newusers)
    const data = await res.json()

    return state => ({ users: state.users.concat(data.users) })
  }
}
```

And with Promise resloving data as partial

```js
actions: {
  setUser: () => (user) => ({ user: user }),
  async fetchUser(state, actions) {
    const res = await fetch("/user")
    const user = await res.json()
    
     actions.setUser(user)
  }
}
```

Will be

```js
actions: {
  async fetchUser(state, actions) {
    const res = await fetch("/user")
    const user = await res.json()
    
     return ({ user: user })
  }
}
```

olstenlarck:
> But we do not give update to users anymore

yea, just wanted to note it and clarify that for someone.

JorgeBucaran:
@olstenlarck I don't know what freactal is, but I loved this from their docs.

- Beautiful is better than ugly.
- **Explicit is better than implicit.**
- Simple is better than complex.
- Complex is better than complicated.
- **Flat is better than nested.**
- Sparse is better than dense.
- Readability counts.

JorgeBucaran:
@Swizz What's the purpose of an anonymous action and why do I need to make them? Currently, I prefer writing my actions explicitly.

lukejacksonn:
So without getting into too much detail.. these patterns LGTM

```js
actions: {
  fetchNewUsers: (state, actions) => 
    fetch("/newusers")
      .then(res => res.json())
      .then(data => state => ({ users: state.users.concat(data.users) })),
  fetchUser: (state, actions) => 
    fetch("/user")
      .then(res => res.json())
      .then(user => ({ user })),
  reset: ({ counter: 0 })
}
```

dmitry-vakhnenko:
:+1:

Looks nice, but i prefer 

```
actions: {
  reset: () => ({ counter: 0 })
}
```

over
```
actions: {
  reset: ({ counter: 0 })
}
```

Functions that return new state is more convient than just object (looks strange).

JorgeBucaran:
@Swizz @lukejacksonn What about a compromise? Allow actions to return a promise that resolves to the new state, but that's it. Similar to what we used to have very early in the game.

Allowing you to do something like this.

```js
actions: {
  download: state => ({ url }) => {
    fetch(url).then(({ newData }) => ({
      isFetching: false,
      data: newData
    }))

    return { isFetching: true }
  }
}
```

But if your async reducer depends on the state you still need to write this:

```js
actions: {
  incrementAsync: (state, actions) =>
    fetch("/random").then(({ newCount }) => {
      actions.incrementBy(newCount)
    })
}
```

You can always create your own `getState` action to access the latest state from anywhere if you want, but I think the pattern above is more explicit.

Interestingly, the following would also be possible, which I consider even better.

```js
actions = {
  incrementAsync: (state, actions) =>
    fetch("/random").then(({ newCount }) => actions.incrementBy(newCount))
}
```

olstenlarck:
>  Let's allow actions to return a promise that resolves to the new state, but nothing more.

Definitely not a good thing. Pretty much things returns an object - apis and etc. I don't believe you can just assume that it is the new (partial) state. That's why reducers must return a function, so you can easily distinguish which is effect and which is reducer. Effects are collectors / requesters of data, reducers are just updaters.

I still don't understand why there should have multiple ways of doing things, it's enough mess that both type of actions are merged into one concept called "actions" ;d Let effects do and return whatever they want and force reducers to return function that returns the new state.

Swizz:
@olstenlarck This is the main purpose behind "anonymous action".
To be honest, I had the "resolving data as partial" for convenience and to add a positive opinion to my proposal. But I will add only the `State => Partial<State>` to avoid this behavior.

The only purpose of "anonymous action" is to allow to define action on the go, without the need of a named action to do only one job one time.

Anonymous actions are actions with the same signature of a regular action but returned from a promise, an observable or another action.

````js
async yourAction(state, actions) {
  const data = await getDataAsync()
  return state => ({ users: state.users.concat(data.users) })
}
```

If you want to make this explicit an update function returning an anonymous actions could be easily done : 

```js
export function update(action) {
  return typeof action === "function" ? action : () => action
}

// ...

async yourAction(state, actions) {
  const data = await getDataAsync()
  return update({ users: data.user })
}
```

olstenlarck:
Yea i'm agree and understand the whole thing.

> But I will add only the State => Partial<State> to avoid this behavior.

Yup. Btw, talking about "full state" and "partial state"... have someone here ever noticed that it's impossible to override the state? Currently it is only possible to pass partial state, but you can't "hard update" it - it is designed that way in the core and can't be workaround. As about if it is the opposite you can implement `update` (meaning partial state) yourself. That's why i believe in the core the coming object should directly override the whole (here "whole" mean the whole slice) state - currently it merge them both.

For what is "hard update" look over `freactal`. Last week i experimented and had some codepen/jsbin to proof it, but i'll see if i can get some more time to write it again.

JorgeBucaran:
@Swizz @lukejacksonn What if we give this one a shot instead?

```js
    handleKeyup: ({ target: { value } }) => state => async actions => {
      if (state.isFetching || value === "") {
        actions.setUrl({ url: "" })
      } else {
        actions.toggleFetching()

        const url = await fetch(
          `//api.giphy.com/v1/gifs/search?q=${value}&api_key=${GIPHY_API_KEY}`
        )
          .then(data => data.json())
          .then(({ data }) => (data[0] ? data[0].images.original.url : ""))

        actions.setUrl({ url })
        actions.toggleFetching()
      }
    }
```

JorgeBucaran:
More experimentation led to this too.

```js
handleKeyup: ({ target: { value: url } }) => state =>
  state.isFetching || url === ""
    ? { url }
    : [
        { isFetching: true },
        async actions => ({
          url: await fetch(
            `//api.giphy.com/v1/gifs/search?q=${url}&api_key=${
              GIPHY_API_KEY
            }`
          )
            .then(data => data.json())
            .then(
              ({ data }) => (data[0] ? data[0].images.original.url : "")
            ),
          isFetching: false
        })
      ]
```

It's interesting, as it removes the need for setUrl and toggleFetching, but you know this doesn't scale well, because what happens when you wish to set the URL or toggle isFetching from a different UI transaction? In that case, it's better to have `setUrl` and `toggleFetching`, making the snippet above less impressive.

SkaterDad:
> @JorgeBucaran It's interesting, as it removes the need for setUrl and toggleFetching, but you know this doesn't scale well, because what happens when you wish to set the URL or toggle isFetching from a different UI transaction? In that case, it's better to have setUrl and toggleFetching, making the snippet above less impressive.

These anonymous actions do seem nice at first, and if you're on a project with a single developer, it's probably not a problem.  The downsides may be long-term scalability & maintainability.  If you're allowed to create actions on the fly (potentially complicated ones), you could end up with state modifications that are harder to debug since dynamic code created the modification.

It's much easier to enforce "blessed" state changes when you restrict yourself to pre-defined actions.  The code would be more obvious also, relying less on hidden framework features.

@JorgeBucaran Is there a benefit of the `(data) => (state) => (actions)` function signature if this anonymous actions proposal doesn't go through?

JorgeBucaran:
@SkaterDad I left this PR open to gather feedback and toy with some ideas, but I think we've already established that we want a single way to update the state, and while the proposal is tempting, I am going to close as wontfix.


> @SkaterDad Is there a benefit of the (data) => (state) => (actions) function signature if this anonymous actions proposal doesn't go through?

Yes, and I think that is the API I want to move forward with. It's the most transparent to the user. I was a bit skeptical at first, because I didn't want to break the perfect (state, actions) signature, but that is nonsense, considering how that signature makes actions closer to WYSIWYG.

```js
app({
  state: {
    count: 0
  },
  view: state => actions => (
    <main>
      <h1>{state.count}</h1>
    </main>
  ),
  actions: {
    down: () => state => null,
  }
})
```

It's much clear now.

SkaterDad:
Which variations of that signature will be supported?

```js
app({
  state: { count: 0 },
  actions: {
    // only taking payload
    minimal: data => ({ count: 0 })
    // taking both payload and state
    medium: data => state => ({ count: state.count + data})
    // taking all 3 params
    full: data => state => actions => {
      actions.doSomething() // not defined, but not important.
      return ({ count: data })
    }
  }
})
```

In my opinion, the current API of `(state, actions) => (data)` is the most intuitive of those options.  It's very closely mirroring what hyperapp does to your actions before you call them.

JorgeBucaran:
@SkaterDad https://github.com/hyperapp/hyperapp/pull/458

PR #453, opened 2017-11-09
Remove modules.

JorgeBucaran:
Hello everyone! 👋

Modules are no doubt a convenient mechanism to "safely" inject state and action slices into your model. The problem is now we have two ways to define our state and actions.

This means it is very hard to wrap your state and actions outside the app() call, e.g, to implement your own module assembly functionality.

```js
import Global from "./global"
import Nav from "./nav"
import Layout from "./layout"
import module from "./module"

app(
  module({
    Global, Nav, Layout
  })
)
```

Without modules, we can assemble the state outside the app call if we need to. This way it's up to you how you mix the state with other sources.

```js
const state = JSON.parse(localStorage.getItem("state"))

app({
  state,
  view,
  actions
})
```

Instead of removing modules, one could fix the problem above if we bring back `init`. Init would be called with the state and actions after they have been assembled/wired by Hyperapp. I want to avoid introducing more complexity, so I really want to move forward with this PR instead.

Removing modules means you will need to **explicitly** assemble your state and actions (and possibly subscriptions in the future). Explicit is good, don't be afraid of it.

```js
import foo from "./foo"
import bar from "./bar"

app({
  state: {
    foo: foo.state,
    bar: bar.state
  },
  actions: {
    foo: foo.actions,
    bar: bar.actions
  }
})
```

### Inspiration

I want to emulate the [Zen of Python](https://www.python.org/dev/peps/pep-0020) :snake: 

I :heart: these principles:
- Explicit is better than implicit.
- Flat is better than nested.
- _There should be one-- and preferably only one --obvious way to do it._
- If the implementation is hard to explain, it's a bad idea.


JorgeBucaran:
@Swizz I know, but see #454! 🎉

Swizz:
Are you going to remove DTS in the nearest future ?

JorgeBucaran:
@Swizz No, I removed it here, because I can't maintain it, but @Mytrill has created a PR with up-to-date typings.

Issue #397, opened 2017-10-02
Support for i18n


kenota:
I've been looking at hyperapp for my project and I really like this library, although I am only getting used to pure functional approach.

I think the doc is missing important point of what would be 'the framework' way to add support for internationalisation?

okwolf:
@kenota I've dealt with supporting internationalization (i18n) in large apps before, and there's not really "one right way" to do it.

Part of the problem is localizing your dates and numbers per national conventions. For this we have the [`Intl` API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) (And [polyfills](https://github.com/andyearnshaw/Intl.js) thereof).

Another part of i18n is translating user-visible text into different languages. This one comes with lots of options depending on your use case and design constraints. If you are going to have multiple deployments for supporting different locales then you could pull in the correct content during your build/bundle step, assuming you have one. If, however, you have requirement to support dynamically changing the language at runtime then the next question is where does your content live? Are you requesting each new language from some API/CDN or does your bundle have all the languages already baked into it? For some folks it makes sense to keep their language strings static, but in other cases they need to be more dynamic.

Once you have answers the previous questions, we get to what I think is the heart of your original question: how do we render different language strings in our UI? If we want to support using a component to do this (a la [React Intl](https://github.com/yahoo/react-intl)) then the next problem to solve is the need to pass the available language strings around as `props` to each component that needs them for rendering localized output (React Intl uses `context` for this purpose, see [here](https://github.com/hyperapp/hyperapp/issues/365#issuecomment-329995132) for a discussion about how you might pre-wire components in Hyperapp, and some of the drawbacks).

Another approach might be deciding to write a (or use an existing) text wrapper function that will handle this formatting for you as an external dependency to your components. [FormatJS](https://formatjs.io/github) has some building blocks that would help with just such a thing. And if you want to minimize the boilerplate there could be work added to your build step to automate any of the tedious parts of this.

kenota:
@okwolf thank you for your reply. Very good decomposition of my question 👍 

You are totally right, I was thinking about problem: how to render different language strings in the UI in hyperapp way. Since my app is very simple, I was thinking about having a global key-value map which is populated once on app load based on user locale, something like: 

```jsx
function Greeting (state, actions) => {
    return (<div> {GlobalLang['WELCOME_MSG']}, {state.username}!</div>);
}
```

to avoid passing it in state. 

[...]

PR #455, opened 2017-11-10
Remove modules from docs

okwolf:
This part of the docs references functionality that has been removed.


willin:
@JorgeBucaran  why removed?

in router plugin

```js
const state = {
  location: location.state
}

const actions = {
  location: location.actions
}
```

still use so

JorgeBucaran:
Would be nice to have docs about modules in general.

okwolf:
@JorgeBucaran I realized that since we moved from an object to positional arguments for `app` it's now tougher to add this kind of functionality with an HOA. Any options would really have to be passed to the HOA:
```
withModules({
  // ...
})(app)(state, actions, ...)
```

Instead of being available in the props like we used to have long ago:

```
withModules(app)({
  state,
  actions,
  modules: {
    // ...
  },
  // ...
})
```

JorgeBucaran:
@okwolf 🤔 I am pretty sure I can write a function that accepts `app` as an argument and returns a function that takes whatever object or number of objects, e.g., { state, actions, ... }.

okwolf:
> @JorgeBucaran 🤔 I am pretty sure I can write a function that accepts app as an argument and returns a function that takes whatever object or number of objects, e.g., { state, actions, ... }.

You could, but it wouldn't play nicely with any other HOAs that rely on a different signature for `app`. You'd have fragmentation of tools that worked for each different style. I personally prefer the object form because it has maximum flexibility and futureproofing.

JorgeBucaran:
@okwolf I see what you are saying, but IMO making HOAs easier to write does not warrant changing the core. You could also make it `app(state, actions, view, modules, container)`.

JorgeBucaran:
Then make modules come last? I know, not so nice. Then, maybe HOAs will need to be programme defensively? TBH I don't think I'd want to use an HOA for modules, but instead something like [combineModels](https://gist.github.com/JorgeBucaran/4dfa567e38ca222c25c46184b3a5df99).

okwolf:
@JorgeBucaran that object `combineModels` returns would spread very nicely into `app` if only it took an object instead 😜

JorgeBucaran:
> is writing higher-order apps a hack or a supported extension mechanism?

I wouldn't call it a hack, but I don't want to say it's a supported extension mechanism either. The term "support" is slightly odd since Hyperapp is doing nothing to prevent or promote creating HOAs. I think HOAs are not frowned upon, but they should be used sparingly.

> If the former, then we need a new way to support extending Hyperapp like we've previously had with mixins, events, hooks, etc...

This could be discussed.

willin:
i'm trying to make a i18n plugin, and now wondering about how to create a module named i18n?

okwolf:
@willin on the subject of i18n, make sure you read our previous issue: #397.

willin:
yep but it's not so graceful. and i want to write my own.

JorgeBucaran:
@willin Just create it? :)

willin:
the point is not i18n,

is how to create a sub module

willin:
if i figure it out... perhaps it can be i18n, ui-framework, visualization or so on...

willin:
In source:

```js
// example
export const state = {
   lang: '',
   langs: []
}

export const actions = {
   setLocale: (lang) => () => ({lang})
}
```

When use:

```js
const {state} = require('hyperapp-i18n');
// state.js
export default {
   // blabla
   i18n: state
}
```

```js
const {actions} = require('hyperapp-i18n');
// actions.js
export default {
   // blabla
   i18n: actions
}
```

then how to use state/actions?

```js
const main = app(state,actions,view,document.body);

// is this right?
main.i18n.setLocal('en-US');
```

```js
// is this right?
   <div>{stat.i18n.lang}</div>
);
```

PR #456, opened 2017-11-10
Functions called before being defined

okwolf:
[Hoisting](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting) is awesome and everything, but for me code is easier to read when functions are defined before being called. I also have this weird thing about putting `export`s at the end of my files.

Feel free to reject this if it's too nitpicky.

### DIFF
--- a/test/dom.test.js
+++ b/test/dom.test.js
@@ -1,5 +1,38 @@
 import { h, app } from "../src"
 
+function testTreeSegue(name, trees) {
+  test(name, done => {
+    app({
+      state: {
+        index: 0
+      },
+      view: (state, actions) =>
+        h(
+          "main",
+          {
+            oncreate: actions.next,
+            onupdate: actions.next
+          },
+          [trees[state.index].tree]
+        ),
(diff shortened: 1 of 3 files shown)


JorgeBucaran:
@okwolf +3 bytes.

JorgeBucaran:
@okwolf I'll merge this and rearrange and tweak a bit later to my taste.

test/dom.test.js
```diff
@@ -1,5 +1,38 @@
 import { h, app } from "../src"
 
+function testTreeSegue(name, trees) {
```
JorgeBucaran: @okwolf What's happening here?

okwolf: @JorgeBucaran I'm not familiar with this function, all I did was move it from the bottom of the file up toward to top.

okwolf:
> @JorgeBucaran +3 bytes.

I'm a bit surprised the size changed from reordering the existing code. Interesting 🤔

PR #457, opened 2017-11-11
Virtual Node Docs

okwolf:
I was looking for a writeup on all the `props` that are allowed on vnodes, and after some archeology I remembered that this was already written up all the way back in the days of 0.12.1: https://github.com/hyperapp/hyperapp/blob/0.12.1/docs/vnodes.md

If we decide not to keep all the gory details from this, I'd like to at least include this useful tidbit somewhere:

> Virtual node props may include any valid [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes) or [SVG](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute) attributes, [DOM](https://developer.mozilla.org/en-US/docs/Web/Events), [lifecycle events](lifecycle.md), and [keys](keys.md).

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -106,7 +106,7 @@ actions: {
 The view describes your user interface as a function of the [state](#state).
 Bind user events and [actions](#actions) together to create interactive applications. The view function is called every time we need to re-render the application due to state changes.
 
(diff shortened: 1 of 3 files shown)


JorgeBucaran:
I'll tweak it a bit later. Thanks! 👍 👍

okwolf:
@JorgeBucaran cheers! 🍻

PR #458, opened 2017-11-15
Point-free style actions API: D=>S=>A 🍛

JorgeBucaran:
### Change actions signature
```js
actions: {
  setFoo: value => ({ foo: value }),
  multiplyFoo: value => state => ({ foo: state.foo * value }),
  calculateFoobar: value => state => actions => {
    // ...
  }
}
```

### Change view signature

```jsx
view: state => <h1>Hello</h1> 
```

```jsx
view: state => actions => <button onclick={actions.hello}>Hello</button>
```

## Background

When this idea was first suggested by @lukejacksonn, I was hesitant because it meant parting ways with what I felt was the "perfect" signature `(state, actions)`, also seen in the view. 

I was also unsure of the hanging `()` parens, like in the following example.

```jsx
actions: {
  toggle: () => state => ({ paused: !state.paused }),
  reset: () => ({ count: SECONDS }),
  drop: () => state => ({ count: state.count - 1 }),
  tick: () => state => actions => {
    if (state.count === 0) {
      actions.reset()
      actions.toggle()
    } else if (!state.paused) {
      actions.drop()
    }
  }
}
```

But then I asked myself: if `toggle: () => state => ({ paused: !state.paused })` bothers you, how do you deal with `actions.toggle()`? 

To someone looking at this code for the first time, it will be self-explanatory that toggle, reset and drop do not receive a payload, as they are called **without** arguments.

On the other hand, how do you explain calling `toggle: state => ({ paused: !state.paused })` as `actions.toggle()` in the current API? It's nice not having to write the parens when the action does not receive a payload, but it's completely magical why it works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,15 +23,15 @@ app({
   state: {
     count: 0
   },
-  view: (state, actions) =>
+  view: state => actions =>
     <main>
       <h1>{state.count}</h1>
       <button onclick={actions.down}>–</button>
       <button onclick={actions.up}>+</button>
     </main>,
   actions: {
-    down: state => ({ count: state.count - 1 }),
-    up: state => ({ count: state.count + 1 })
+    down: () => state => ({ count: state.count - 1 }),
+    up: () => state => ({ count: state.count + 1 })
   }
 })
 ```
--- a/docs/README.md
+++ b/docs/README.md
@@ -33,21 +33,25 @@ app({
   state: {
     count: 0
   },
-  view: (state, actions) => (
+  view: state => actions => (
     h("main", {}, [
       h("h1", {}, state.count),
       h("button", {
-        onclick: actions.down,
+        onclick() {
+          actions.down(1)
+        },
         disabled: state.count <= 0
       }, "–"),
       h("button", {
-        onclick: actions.up
+        onclick() {
+          actions.up(1)
+        },
       }, "+")
     ])
   ),
   actions: {
-    down: state => ({ count: state.count - 1 }),
-    up: state => ({ count: state.count + 1 })
+    down: n => state => ({ count: state.count - n }),
+    up: n => state => ({ count: state.count + n })
   }
 })
 
@@ -86,20 +90,12 @@ Actions are used to manipulate the [state](#state). If your application consumes
 
 ```jsx
 actions: {
-  down: state => ({ count: state.count - 1 }),
-  up: state => ({ count: state.count + 1 })
+  down: n => state => ({ count: state.count - n }),
+  up: n => state => ({ count: state.count + n })
 }
 ```
 
 Actions must never mutate the state directly. Returning a new state from an action updates the current state and schedules a re-render.
-
-You can also pass arguments to actions by returning a function.
-
-```jsx
-actions: {
-  upWithValue: state => value => ({ count: state.count + value })
-}
-```
 
 ### View
 
@@ -123,21 +119,25 @@ app({
   state: {
     count: 0
   },
-  view: (state, actions) => (
+  view: state => actions => (
     main([
       h1(state.count),
       button({
-        onclick: actions.down,
+        onclick() {
+          actions.down(1)
+        },
         disabled: state.count <= 0
       }, "–"),
       button({
-        onclick: actions.up
+        onclick() {
+          actions.up(1)
+        },
       }, "+")
     ])
   ),
   actions: {
-    down: state => ({ count: state.count - 1 }),
-    up: state => ({ count: state.count + 1 })
+    down: n => state => ({ count: state.count - n }),
+    up: n => state => ({ count: state.count + n })
   }
 })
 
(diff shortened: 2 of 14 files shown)


SkaterDad:
This one didn't take long to warm up to.   I can already imagine some nice benefits :+1:

kingoftheknoll:
I really like this. The ability to use async functions feels very natural.

dmitry-vakhnenko:
thats awesome 👍 

( 1.0 ready? ;-) )

zaceno:
Any particular reason why not: `data => (state, actions) => {...}` ?

If you just want state, it'll be the exact same

if you want state and actions it'll be:

```js
data => state => actions => {...}
//vs
data => (state, actions) => {...}
```

If you just use actions, it'll be:

```js
data => _ => actions => {}
//vs
data => (_, actions) => {}
```

sure, the extra `=>` is fancy and all, but it's not that much better DX, and I believe `D => (S, A) => {...}` would be easier to implement. But perhaps there are reasons I'm missing?

olstenlarck:
@zaceno totally agree, i vote for it! :+1: :100:

SkaterDad:
Also, what about keeping the signature for views as it is?  Not sure I see any benefit for that to be curried. Can anyone come up with an example where that gives us more power?

olstenlarck:
@SkaterDad, mostly for convenience. But also async (loading of) views.

SkaterDad:
Are you suggesting that this PR enables async views in a way the current API cannot?  Or that you want him to also implement that?  I'm not really seeing how this PR changes that behavior.

olstenlarck:
No, not suggesting anything. And i don't believe it is possible currently. Also not sure what i exactly have in mind, didn't played with such things.

SkaterDad:
Okay, just making sure I didn't miss anything!  That's definitely a good idea, though I'm sure it's already possible in user land.   The view code is certainly a great candidate for code splitting.

JorgeBucaran:
@zaceno API symmetry, which is important to me. That little `=>` is a visual aid telling me what kind of action is my action. I agree it's pointless in the case of the view function, but again, symmetry.

okwolf:
I really want to like this, but personally I feel like there's just as much magic either way. Since users won't be calling actions this way: `actions(toggle)(state)(actions)` they will still need to read the docs (or code) to understand where the non sequitur `state` and `actions` parameters come from, since they're not passed when the action is called.

JorgeBucaran:
@okwolf 

> The lesser of two evils principle (or lesser evil principle, lesser-evilism) is the principle that when faced with selecting from two immoral options, the one which is least immoral should be chosen.

A choice needs to be made.

This is the current API.

```jsx
function multiplyFoo(state, actions) {
  return number => {
    return { foo: state.foo * number }
  }
}

multiplyFoo(5) 
```

Call `multiplyFoo(5)` and define it as a function expecting the the state that returns a function expecting the number. 

---

This new API is **less** magical. Call `multiplyFoo(5)` passing some data/payload and define it as a function expecting some data/payload. 

```jsx
function multiplyFoo(number) {
  return state => {
    return { foo: state.foo * number }
  }
}

multiplyFoo(5)
```

okwolf:
@JorgeBucaran IMHO the only difference is with one you write a state/updater function (reducer) that returns an action function and with the other you write an action function that returns a reducer. I must be missing the fundamental difference 🤔

JorgeBucaran:
@okwolf The difference is you call:

```js
multiplyFoo(5)
```

and  define it as:

```js
function multiplyFoo(number) {
  // ...
}
```

That's a considerable difference, specially when the reducer doesn't use the state at all. In that case is a huge difference.

```js
setValue: value => ({ foo: value })
```

olstenlarck:
@okwolf D=>S=>A is also a lot better for testing and "reason about". Testing really fully pure reducer action is just calling it and expecting to return an object or return a function that returns an object. Effects in any way are a bit harder to test than reducers, because they need to be passed with `actions`.

depoulo:
> Change view signature
> ```   
> view: state => <h1>Hello</h1> 
> view: state => actions => <button onclick={actions.hello}>Hello</button>
> ```

Oops, I missed the part where it says that, and why, this was not done.

depoulo:
> Change view signature
> ```   
> view: state => <h1>Hello</h1> 
> view: state => actions => <button onclick={actions.hello}>Hello</button>
> ```

Oops, I missed the part where it says that, and why, this was not done.

**Update**: Is any of this in 0.18.0? Quite confused...

JorgeBucaran:
@depoulo The docs are up to date with the currently released version, this PR is already old business.

The current API is as follows (from the README):

```js

const state = {
  count: 0
}

const actions = {
  down: () => (state, actions) => ({ count: state.count - 1 }),
  up: () => (state, actions) => ({ count: state.count + 1 })
}

  <main>
    <h1>{state.count}</h1>
  </main>
)

export const main = app(state, actions, view, document.body)
```

PR #459, opened 2017-11-16
Reflect Actions API changes in types

Swizz:
Pretty easy.

Only the format stuff is weird but Prettier hold all the truth, right ?

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -57,9 +57,12 @@ export function h<Props>(
 export type ActionResult<State> = Partial<State> | Promise<any> | null | void
 
 export type MyAction<State, Actions> = (
-  state: State,
-  actions: Actions
-) => ((data: any) => ActionResult<State>) | ActionResult<State>
+  data: any
+) =>
+  | ((
+      state: State
+    ) => ((actions: Actions) => ActionResult<State>) | ActionResult<State>)
+  | ActionResult<State>
 
 /** The interface for actions (exposed when implementing actions).
  *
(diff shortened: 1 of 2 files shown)


Mytrill:
LGTM

JorgeBucaran:
@Swizz Is there any way to split up MyAction type so that is more readable? I don't mean the format. 🤔

Swizz:
By splitting you mean using side types ?

```ts
type Reducer<State> = (state: State) => ActionResult<State>
type Effect<State, Actions> = (
  state: State
) => (actions: Actions) => ActionResult<State>

export type MyAction<State, Actions> = (
  data: any
) => ActionResult<State> | Reducer<State> | Effect<State, Actions>
```

But "Reducer" is a wrong name as an effect could do not use Actions :thinking: 

```ts
type NeedActions<State, Actions> = (actions: Actions) => ActionResult<State>
type NeedState<State, Actions> = (
  state: State
) => ActionResult<State> | NeedActions<State, Actions>

export type MyAction<State, Actions> = (
  data: any
) => ActionResult<State> | NeedState<State, Actions>
```

Just a bunch of things helping you to think more a bout what you are requesting.

Both way, using the following `state` will have the right type assertions ! :tada:

```ts
app<State, Actions>({
  state: {
    value: 0
  },
  actions: {
    addNumber(num: number) {
      return state => ({ value: state.value + num })
    }
  }
})
```

JorgeBucaran:
@Swizz Yes, something like that. It would help with readability IMO.

Swizz:
Which one to use ?

JorgeBucaran:
@Swizz The one that says Reducer and Effects is what I meant. Is that the good one?

PR #467, opened 2017-11-29
Adding failing test for the onremove lifecycle event

4lejandrito:
When the parent element is removed, the child element's onremove
lifecycle event is not fired. It looks like only the top level element's
lifecycle method is being run.


4lejandrito:
Sorry, wrong test. I will resend after work :)

PR #469, opened 2017-11-30
Update the new API with respect to the View type

Alber70g:
The view is converted from `(state, actions) => VNode` to `state => actions => VNode`

I believe it should be like this. Please don't pin me down on it, as I'm not that good with Type definitions ;)

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -76,9 +76,11 @@ export type MyActions<State, Actions> = {
  *
  * @memberOf [App]
  */
-export interface View<State, Actions> {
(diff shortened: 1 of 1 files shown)


SkaterDad:
@JorgeBucaran Can we get a quick patch version release for this? :pray:

Swizz:
Is there any other API changes ?

JorgeBucaran:
@SkaterDad →0.16.2 🎉

JorgeBucaran:
@Swizz This was about the view type change, that was changed for symmetry with the new actions api that you typed for us. 😄

Issue #468, opened 2017-11-29
onremove won't fire for children elements


4lejandrito:
Please see https://github.com/hyperapp/hyperapp/pull/467 with a failing test for this. 

Basically when an element its removed only its `onremove` method seems to be called. None of the `onremove` lifecycle events are called for its children. 

My specific use case is:

Modal Dialog Component
|__ Camera custom component (`oncreate` calls [getUserMedia](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia) and `onremove` closes the stream).

When I close the modal, the `onremove` method of my camera is not called.

Thanks for all your work!

4lejandrito:
Looks like I need to dig into this a little bit more, I will reopen if necessary.

JorgeBucaran:
Maybe related. https://github.com/picodom/picodom/issues/61

SkaterDad:
Here's a Codesandbox where a parent element waits for children `onremove`:
https://codesandbox.io/s/r1yzxl3kwo

I implemented all of the `onremove` functions as Promises so you can wait for async things like animation, but the same idea applies for sync things which just teardown objects & listeners.

It's not much code to implement this, but since it would imply a perf hit on every `onremove`, I would suggest it either goes to a new (optional) lifecycle event, or is left up to devs.

```js
// recursive, so can be slow on large trees
function collectChildrenOnremove(result, child) {
  if (child.onremove) {
    result.push(
       // call child onremove before pushing it to the array
      child.onremove(child, function() {
        // Fake done() function
      })
    )
  }
  if (child.children != null) {
    const arr = Array.prototype.slice.call(child.children)
    return arr.reduce(collectChildrenOnremove, result)
  }
  return result
}

// put this on the parent element
function waitForChildrenOnremove(element, done) {
  let promises = Array.prototype.slice
    .call(element.children)
    .reduce(collectChildrenOnremove, [])

  return Promise.all(promises).then(done)
}

...

// use it later
 h('div', { onremove: waitForChildrenOnremove } , [ ... ])
```

4lejandrito:
Hi @SkaterDad,

I appreciate you spending the time to provide a workaround that would work without patching Hyperapp (I might actually use it for the time being). However I think this is such a common use case that it should somehow be added to the core (see https://github.com/hyperapp/hyperapp/pull/470). 

Thanks!

JorgeBucaran:
[...]

PR #470, opened 2017-11-30
Nested onremove

4lejandrito:
A possible solution to https://github.com/hyperapp/hyperapp/issues/468.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -159,13 +159,20 @@ export function app(props, container) {
     }
   }
 
-  function removeElement(parent, element, props) {
+  function removeElement(parent, element, node) {
     function done() {
       parent.removeChild(element)
+      if (node.children) {
+        for (var i = 0; i < node.children.length; i++) {
+          if (typeof node.children[i] !== "string") {
+            removeElement(element, element.children[i], node.children[i])
+          }
+        }
+      }
     }
 
-    if (props && props.onremove) {
-      props.onremove(element, done)
+    if (node.props && node.props.onremove) {
+      node.props.onremove(element, done)
     } else {
       done()
     }
@@ -239,7 +246,7 @@ export function app(props, container) {
         var oldChild = oldNode.children[i]
         var oldKey = getKey(oldChild)
         if (null == oldKey) {
-          removeElement(element, oldElements[i], oldChild.props)
+          removeElement(element, oldElements[i], oldChild)
         }
         i++
       }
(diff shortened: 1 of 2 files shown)


dmitry-vakhnenko:
Awesome!

I like it more than introducing destroy event.

JorgeBucaran:
Ping @mindplay-dk.

src/index.js
```diff
@@ -159,13 +159,20 @@ export function app(props, container) {
...
   }
 
-  function removeElement(parent, element, props) {
+  function removeElement(parent, element, node) {
     function done() {
       parent.removeChild(element)
+      if (node.children) {
+        for (var i = 0; i < node.children.length; i++) {
+          if (typeof node.children[i] !== "string") {
+            removeElement(element, element.children[i], node.children[i])
```
mindplay-dk: Also, with the [changes I attempted here](https://github.com/picodom/picodom/compare/master...mindplay-dk:destroy-event), I was specifically trying to avoid recursive removal of every individual element of a removed a subtree - mainly performance reasons. That's a lot of unncessary DOM operations. A single removal of the parent node would be preferable - only the destroy-notifications need to account for child element removals.

SkaterDad:
> @mindplay-dk for example, if an entire table was removed, would you expect to see an animation of every individual row getting removed before the table itself gets removed?

I agree that this implementation is probably not going to allow animation sequencing, since it has no way of waiting for children `done()` calls.

The way `onremove` currently works is very limiting, and I don't really use it.  It would be far more intuitive if it always got called, even if a parent was removed.   I'll try to find time to review this implementation more deeply, but I like the idea.

JorgeBucaran:
Do you have any comments? @zaceno @lukejacksonn

4lejandrito:
Hi! 

I do agree with @mindplay-dk. 

I was oversimplifying just to start this conversation. What I implemented works for my use case but I am aware of the drawbacks.

With my solution, every child is going to run its animations (if any) when its parent is no longer in the DOM. This is definitely a waste of resources but (without having tried this specific use case) I think it would work anyway.

There are indeed 2 uses cases that we are (or I am) trying to cover with the same event: **animations** and **subscriptions**. My use case was all about the latter (I _subscribe_ to the camera through _getUserMedia()_).

Basically I was trying to port my app written in React to Hyperapp and I wanted `onremove` to behave the same way as [componentWillUnmount](https://reactjs.org/docs/react-component.html#componentwillunmount). React does not support animations through the lifecycle methods (you use [react-transition-group](https://github.com/reactjs/react-transition-group) for this).

If we want to be accurate and avoid wasting resources, I am on @mindplay-dk's side. I think we need 2 different lifecycle events, one for **animations** and a different one for **subscriptions**.

I am happy to extend my solution to support both if that's what you guys decide.

Regards

JorgeBucaran:
> @4lejandrito I am happy to extend my solution to support both if that's what you guys decide.

Great, please do so.

4lejandrito:
Closing in favour of https://github.com/hyperapp/hyperapp/pull/478.

PR #474, opened 2017-12-03
Update rollup and uglify versions

okwolf:
We're a whole major release behind on our UglifyJS version.

Minified size decreased from 3136 -> 3102 bytes however gzipped size increased from 1427 -> 1443 bytes? 😕

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "test": "jest --coverage --no-cache && tsc -p test/ts",
     "build": "npm run bundle && npm run minify",
     "bundle": "rollup -i src/index.js -o dist/hyperapp.js -m -f umd -n hyperapp",
-    "minify": "uglifyjs dist/hyperapp.js -o dist/hyperapp.js --mangle --compress warnings=false --pure-funcs=Object.defineProperty -p relative --in-source-map dist/hyperapp.js.map --source-map dist/hyperapp.js.map",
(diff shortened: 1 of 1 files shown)


JorgeBucaran:
@okwolf Thanks, I'll have a look.

JorgeBucaran:
@okwolf There is no good reason to upgrade uglifyjs that I know. Am I missing something? As soon as we decide to drop support for IE11, I'll drop uglifyjs altogether.

JorgeBucaran:
@okwolf I can accept the other changes, including the simplified CLI flags.

okwolf:
@JorgeBucaran I can accept the other changes, including the simplified CLI flags. If you can only change the uglify version back to how it was, then I'll merge this.

The changes to the CLI flags isn't compatible with our older version. Should we just merge the rollup update?

JorgeBucaran:
Hmm I see.
> Should we just merge the rollup update?

Yeah.

JorgeBucaran:
Nevermind. Let's upgrade. 🤷

JorgeBucaran:
@okwolf Hopefully this will not break sourcemaps.

whaaaley:
uglify3 has es6 stuff and a better cli interface

okwolf:
> @JorgeBucaran Hopefully this will not break sourcemaps.

The output looked equivalent, but I didn't test it out in a real browser yet.

JorgeBucaran:
@okwolf Well, could you?

okwolf:
@JorgeBucaran verified sourcemaps are still working with the latest build from master 👍

Issue #473, opened 2017-12-02
Unsupported Node.js version for CI


okwolf:
Currently, [Travis is building with Node.js versions 7 and 8](https://github.com/hyperapp/hyperapp/blob/master/.travis.yml#L3-L4). Node.js 7 is no longer supported (last release 2017-07-11), and we dropped Node.js 6 even though it is still under LTS (last release 2017-11-06).

My suggestion would be to stick with the actively supported versions of Node.js: 6, 8, and 9.

JorgeBucaran:
Sure, no problem.

okwolf:
FYI here is the schedule for Node support:
![](https://raw.githubusercontent.com/nodejs/Release/master/schedule.png)

selfup:
Thanks for bringing it up @okwolf!

PR has been merged: #475

Closing 😄

PR #475, opened 2017-12-04
actively supported versions of Node.js: 6, 8, and 9

selfup:
References #473

Update Travis Node Versions to Supported Versions 🎉

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: node_js
 node_js:
+  - "6"
   - "8"
-  - "7"
+  - "9"
 
 env:
   - NODE_ENV=development


selfup:
Looks like we are good to go!

screen shot 2017-12-04 at 7 08 45 am

JorgeBucaran:
Thanks @selfup! :9

.travis.yml
```diff
@@ -1,7 +1,8 @@
 language: node_js
 node_js:
+  - "6"
   - "8"
-  - "7"
+  - "9"
```
frenzzy: maybe `"stable"` instead of `"9"`? ..and only LTS versions define explicitly

selfup: Good question. @JorgeBucaran what do you think?

JorgeBucaran: What do you folks make out of this?

screen shot 2017-12-04 at 23 49 27

JorgeBucaran: https://github.com/travis-ci/docs-travis-ci-com/issues/362

selfup: lots of options 😂

PR #476, opened 2017-12-06
Typing update for View & VNodes

SkaterDad:
I've started trying out Hyperapp with Typescript, and noticed a couple issues.

The current typing for the app's view parameter (`export type View`) did not allow `null` for the props, so this was giving compiler errors:
```js
app({
  view: () => h('div', null, 'Test')
})
```

I also noticed you were allowed to put weird things like numbers in the props of vnodes.  This wasn't giving any errors at compilation time. 
```js
app({
  view: () => h('div', 1, 'Test')
})
```
Since JS considers nearly everything an object, that means that `{}` is not a very specific type.  Typescript 2.2 they added `object`, which is more specific, referring to non-primitive types.  It still won't prevent you from passing an array, but it's a little better.  This article explains it well:  https://blog.mariusschulz.com/2017/02/24/typescript-2-2-the-object-type

Anyone have any `object`ions to using `object`? :laughing: 

Codesandbox which uses these updated types so you can try it.  It's based on the React Typescript template, and compiler errors show up in the editor:  https://codesandbox.io/s/7z8v2nqv56

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -9,7 +9,7 @@ export as namespace Hyperapp
 export interface VNode<Props> {
   type: string
   props: Props
-  children: VNodeChild<{} | null>[]
+  children: VNodeChild<object | null>[]
 }
 
 /** In the VDOM a Child could be either a VNode or a string
@@ -23,7 +23,7 @@ export type VNodeChild<Props> = VNode<Props> | string
  * @memberOf [VDOM]
  */
 export interface Component<Props> {
-  (props: Props, children: VNodeChild<{} | null>[]): VNode<{}>
+  (props: Props, children: VNodeChild<object | null>[]): VNode<{}>
 }
 
 /**The type for the children parameter accepted by h().
@@ -31,8 +31,8 @@ export interface Component<Props> {
  * @memberOf [VDOM]
  */
 export type VNodeChildren =
-  | Array<VNodeChild<{} | null> | number>
-  | VNodeChild<{} | null>
+  | Array<VNodeChild<object | null> | number>
+  | VNodeChild<object | null>
   | number
 
 /** The soft way to create a VNode
@@ -78,9 +78,7 @@ export type MyActions<State, Actions> = {
  */
 export type View<State, Actions> = (
   state: State
-) =>
-  | ((actions: Actions) => VNode<{}>)
-  | VNode<{}>
+) => ((actions: Actions) => VNode<object | null>) | VNode<object | null>
 
 /** The props object that serves as an input to app().
  *


JorgeBucaran:
LGTM. Ping @Swizz @Mytrill.

@mindplay-dk Maybe picodom should learn this trick too?

JorgeBucaran:
@SkaterDad Instead of `object`, shouldn't we define a props type? Like `Props` or `Attributes`?

That way we could say that attributes is a typeof:
```
HTMLAttributes | SVGAttributes | DOMEvents | VDOMEvents
```

where VDOMEvents refer to `oncreate`, `onupdate` and `onremove`.

SkaterDad:
I think the difficulty in defining a props type is that users may want to attach arbitrary things.   For example, if someone wanted to use a 3rd party library that watched for certain non-standard attributes.

JorgeBucaran:
@SkaterDad In that case:

```
HTMLAttributes | SVGAttributes | DOMEvents | VDOMEvents | any
```

SkaterDad:
Sounds like a good idea.  I think 'any' can be replaced with something more specific.  I'll try to implement that tomorrow unless someone else chimes in.

JorgeBucaran:
@SkaterDad I edited my post, yes you are right.

mindplay-dk:
I've been doing something similar in my own private fork - I use the term `VChildNode` rather than `VNodeChild` for consistency with DOM terminology though, where they're commonly known as "child nodes" rather than "node children".

JorgeBucaran:
@mindplay-dk I would love to integrate those improvements to Hyperapp! Any reason you want to keep this private? 🤔

mindplay-dk:
@JorgeBucaran it's part of a larger experiment - I'm basically rewriting picodom in Typescript, while also attempting some very aggressive optimizations: no recursion, no function calls. Doesn't work yet ;-)

JorgeBucaran:
PR welcome in picodom 😉

mindplay-dk:
@JorgeBucaran of course, if/when I complete this, you will hear about it :-) But this is really a rewrite, tests rewritten in TS too (using Tape) and it's factored quite differently, placing more responsibility on the `h` function, in order to get a `VNode` with fewer permutations and simpler types. It'll be a while before I'm ready to show this ;-)

JorgeBucaran:
@mindplay-dk So `h()` is not recursive in your rewrite?

mindplay-dk:
@JorgeBucaran yes, `h` is recursive - I can't do anything about that, since this is how Typescript compiles JSX expressions. But the call to `updateElement` internally (the rendering step) will be non-recursive.

SkaterDad:
I've been tinkering with adding more specific VNodeProps types, and it seems challenging to get right.  

The main TS lib has type defs for HTMLElement & SVGElement, but nothing specifically just for attributes/properties.   Additionally, since every HTML element has its own set of props/attributes, I'm not sure how we could model that with just the `h()` function.  You could define more specific types in `hyperapp/html`?

**Looking at other libraries for reference:**

Snabbdom just defines their attributes and props as simple records.
Their props type for example: `export type Props = Record<string, any>;`

`ivi` went above & beyond the call of duty, and manually defined all of the relevant attributes for each type of DOM node.  It's truly impressive how much effort was put in:  https://github.com/ivijs/ivi/tree/master/packages/ivi-core/src

**Path forward:**
For simplicity, how about we include the VDOM lifecylce hooks, and a few of the most helpful HTML attributes?  `class`, `style`, `onclick`, and `onsubmit` are pretty handy, so having them pop up in autocompletes would be nice.

Here's how that would look, with all of the generics `<Props>` removed.   Can anyone explain why `<Props>` was used (cc: @Swizz)?  It seems simpler to remove it, but I wouldn't want to hinder any use cases.

```typescript
interface VDOMEvents {
  oncreate?(el: HTMLElement): any
  onupdate?(el: HTMLElement, oldProps: VDOMProps): any
  onremove?(el: HTMLElement, done: () => any): any
}

interface VDOMPropsInterface extends VDOMEvents {
  class?: string
  style?: Record<string, string>
  onclick?(e: any): any
  onsubmit?(e: any): any
  [key: string]: any
}

type VDOMProps = VDOMPropsInterface | null
```
The Codesandbox has these updates so you can try it:
https://codesandbox.io/s/7z8v2nqv56

SkaterDad:
I'm going to close this for now.  There will need to be other typing updates for the recent API changes.

PR #478, opened 2017-12-07
Ondestroy

4lejandrito:
Adds ondestroy event to fix https://github.com/hyperapp/hyperapp/issues/468.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -159,13 +159,27 @@ export function app(props, container) {
     }
   }
 
-  function removeElement(parent, element, props) {
+  function destroyNode(element, node) {
+    if (node.children) {
+      for (var i = 0; i < node.children.length; i++) {
+        if (typeof node.children[i] !== "string") {
+          destroyNode(element.childNodes[i], node.children[i])
+        }
+      }
+    }
+    if (node.props && node.props.ondestroy) {
+      node.props.ondestroy(element)
+    }
+  }
+
+  function removeElement(parent, element, node) {
     function done() {
       parent.removeChild(element)
+      destroyNode(element, node)
     }
 
-    if (props && props.onremove) {
-      props.onremove(element, done)
+    if (node.props && node.props.onremove) {
+      node.props.onremove(element, done)
     } else {
       done()
     }
@@ -239,7 +253,7 @@ export function app(props, container) {
         var oldChild = oldNode.children[i]
         var oldKey = getKey(oldChild)
         if (null == oldKey) {
-          removeElement(element, oldElements[i], oldChild.props)
+          removeElement(element, oldElements[i], oldChild)
         }
         i++
       }
@@ -248,7 +262,7 @@ export function app(props, container) {
         var keyedNode = oldKeyed[i]
         var reusableNode = keyedNode[1]
         if (!keyed[reusableNode.props.key]) {
-          removeElement(element, keyedNode[0], reusableNode.props)
+          removeElement(element, keyedNode[0], reusableNode)
         }
       }
     } else if (element && node !== element.nodeValue) {
@@ -259,7 +273,7 @@ export function app(props, container) {
           createElement(node, isSVG),
           (nextSibling = element)
         )
-        removeElement(parent, nextSibling, oldNode.props)
+        removeElement(parent, nextSibling, oldNode)
       }
     }
 
--- a/src/index.js
+++ b/src/index.js
@@ -174,8 +174,8 @@ export function app(props, container) {
 
   function removeElement(parent, element, node) {
     function done() {
-      parent.removeChild(element)
       destroyNode(element, node)
+      parent.removeChild(element)
     }
 
     if (node.props && node.props.onremove) {
--- a/test/lifecycle.test.js
+++ b/test/lifecycle.test.js
@@ -82,6 +82,77 @@ test("onremove", done => {
   })
 })
 
+test("ondestroy", done => {
+  let destroyed = false
+  app({
+    state: {
+      value: true
+    },
+    view: state => actions =>
+      state.value
+        ? h(
+            "ul",
+            {
+              oncreate: () => actions.toggle()
+            },
+            [
+              h("li"),
+              h("li", {
+                onremove(element, remove) {
+                  expect(destroyed).toBe(false)
+                  remove()
+                  expect(destroyed).toBe(true)
+                  done()
+                },
+                ondestroy() {
+                  destroyed = true
+                }
+              })
+            ]
+          )
+        : h("ul", {}, [h("li")]),
+    actions: {
+      toggle: () => state => ({ value: !state.value })
+    }
+  })
+})
+
+test("nested ondestroy", done => {
+  let removed = false
+  app({
+    state: {
+      value: true
+    },
+    view: state => actions =>
+      state.value
+        ? h(
+            "ul",
+            {
+              oncreate: () => actions.toggle()
+            },
+            [
+              h("li"),
+              h("li", {}, [
+                h("span", {
+                  onremove(element, remove) {
+                    removed = true
+                    remove()
+                  },
+                  ondestroy() {
+                    expect(removed).toBe(false)
+                    done()
+                  }
+                })
+              ])
+            ]
+          )
+        : h("ul", {}, [h("li")]),
+    actions: {
+      toggle: () => state => ({ value: !state.value })
+    }
+  })
+})
+
 test("event bubling", done => {
   let count = 0
   app({
--- a/test/lifecycle.test.js
+++ b/test/lifecycle.test.js
@@ -153,6 +153,44 @@ test("nested ondestroy", done => {
   })
 })
 
+test("several nested ondestroy", done => {
+  let childDestroyed = false
+  app({
+    state: {
+      value: true
+    },
+    view: state => actions =>
+      state.value
+        ? h(
+            "ul",
+            {
+              oncreate: () => actions.toggle()
+            },
+            [
+              h("li"),
+              h("li", {}, [
+                h("span", {
+                  ondestroy() {
+                    expect(childDestroyed).toBe(true)
+                    done()
+                  }
+                }, [
+                  h("strong", {
+                    ondestroy() {
+                      childDestroyed = true
+                    }
+                  })
+                ])
+              ])
+            ]
+          )
+        : h("ul", {}, [h("li")]),
+    actions: {
+      toggle: () => state => ({ value: !state.value })
+    }
+  })
+})
+
 test("event bubling", done => {
   let count = 0
   app({
--- a/docs/lifecycle.md
+++ b/docs/lifecycle.md
@@ -40,7 +40,7 @@ function Textbox({ placeholder }) {
 
 ## `onremove`
 
-This event is fired before the element is removed from the DOM. Use it for cleaning up resources, creating slide out animations, etc. Use `done` inside the function to remove the element.
+This event is fired before the element is removed from the DOM. Use it for creating slide out animations, etc. Use `done` inside the function to remove the element.
 
 ```jsx
 function MessageWithFadeout({ title }) {
@@ -49,6 +49,31 @@ function MessageWithFadeout({ title }) {
       onremove={(element, done) => fadeout(element).then(done)}>
       <h1>{title}</h1>
     </div>
+  )
+}
+```
+
+## `ondestroy`
+
+This event is fired after the element has been removed from the DOM, either directly or as a result of a parent being removed. Use it for cleaning up resources.
+
+```jsx
+function Camera({onerror}) {
+  return (
+    <video
+      poster="loading.png"
+      oncreate={
+        element => {
+          navigator.mediaDevices
+            .getUserMedia({video: true})
+            .then(stream => (element.srcObject = stream))
+            .catch(onerror)
+        }
+      }
+      ondestroy={
+        element => element.srcObject.getTracks()[0].stop()
+      }
+    />
   )
 }
 ```


test/lifecycle.test.js
```diff
@@ -82,6 +82,77 @@ test("onremove", done => {
...
+            ]
+          )
+        : h("ul", {}, [h("li")]),
+    actions: {
+      toggle: () => state => ({ value: !state.value })
+    }
+  })
+})
+
+test("nested ondestroy", done => {
```
SkaterDad: Would you mind adding another test with multiple nested `ondestroy`s?  The code looks fine, but it would be a nice test to have.

Could just add a child to the `h('span', { ondestroy: ... })`, perhaps a `h('strong', {ondestroy: ...} )`?

JorgeBucaran:
@4lejandrito @SkaterDad @zaceno @lukejacksonn What if we rename the events to:

```
~ondestroy~ → onremove
~onremove~ → onunmount
```

🤔

SkaterDad:
If anything should be renamed, the recursive `destroyNode` could be `seekAndDestroy`. 🤘

JorgeBucaran:
@4lejandrito @SkaterDad If ondestroy is only used for cleaning resources, unsubscribing from global events, clearing timers, etc., then it might make sense to use the same stack we use for oncreate and flush it after patch.

mindplay-dk:
May be somewhat unrelated, but while reading through this code, one thing that occurs to me is, how can you know when an update is complete?

Because `onremove` can asynchronously trigger `ondestroy`, there's nothing telling the outside world (e.g. code calling `patch`) when the entire patch operation is done - which means there's no way to defer the next update until the current one is done.

The asynchronous nature of `onremove`, as it relates to this PR, could mean that multiple updates get invoked before the callback to an `onremove` handler triggers the `ondestroy` notification - at which time, this could lead to notifications that, at the time, would have been correct, but when they get invoked, they could actually be wrong. That is, maybe we're broadcasting a remove notification too late, after it's parent has been been destroyed in a subsequent update.

Like I said, maybe this is somewhat unrelated to this PR (which looks pretty good otherwise) but I wonder if the synchronous `patch` API makes these life-cycle features potentially error-prone? Perhaps the `patch` function ought to return a `Promise` that doesn't resolve before `onremove` handlers have invoked their `remove` functions and triggered the destroy notifications?

JorgeBucaran:
@mindplay-dk What is an update?

JorgeBucaran:
@mindplay-dk What do you mean by update? And what kind of problems? I don't understand very well what you are saying. Also, you are familiar with snabbdom's code base yes? What do they do here if at all?

4lejandrito:
Hi @JorgeBucaran,

Regarding:

> @4lejandrito @SkaterDad If ondestroy is only used for cleaning resources, unsubscribing from global events, clearing timers, etc., then it might make sense to use the same stack we use for oncreate and flush it after patch.

I think we can't use the lifecycle stack since the `ondestroy` event can be potentially called asynchronously (after ' onremove's `done` function is called). In this case those lifecycle events would be ignored by [the consumer](https://github.com/hyperapp/hyperapp/blob/master/src/index.js#L309) which seems to work synchronously.

Does that make sense?

mindplay-dk:
@JorgeBucaran I think Snabbdom likely ignores this problem as well, or they're not aware of it. Not sure. Maybe just ignore my comment for now... in my Typescript port/rewrite, the `patch()` function returns a `Promise` to make it possible to deal with this problem, but it requires the update loop etc. to be structured in a radically different way - it's not a small change, and maybe this is meeting a requirement that you/others don't currently care about.

zaceno:
I suspect that if ondestroy is only meant to be used for freeing up resources, it shouldn't matter that it's asynchronous and the rest of the app shouldn't need to wait for it to finish?  @mindplay-dk What's your use case for having a promise that resolves when all the ondestroy are done being called?

src/index.js
```diff
@@ -159,13 +159,27 @@ export function app(props, container) {
...
+    }
+    if (node.props && node.props.ondestroy) {
+      node.props.ondestroy(element)
+    }
+  }
+
+  function removeElement(parent, element, node) {
     function done() {
       parent.removeChild(element)
+      destroyNode(element, node)
```
zaceno: Could we call `destroyNode` before the element is removed instead?

I'm not sure, but I suspect the child elements may be in a somehow less usable state when their tree is detached from the DOM, potentially causing issues for the ondestroy handlers.

mindplay-dk:
> What's your use case for having a promise that resolves when all the ondestroy are done being called?

It's not so much a use-case, more a concern that if a second `patch()` call is made while the first one is still completing, this could cause problems.

For example, if the first call causes an `onremove` handler to start an animation that takes 200 milliseconds to complete, and the next call to `patch()` happens before the animation has completed, e.g. `parent.children` indices will be out of sync with indices in `oldNodes`, because the element that's still animating won't be removed from `children` before the call to `removeChild` which hasn't happened yet.

I'm pretty sure you could write a simple test-case to demonstrate this problem - I don't believe we have a test-case demonstrating the effects of asynchronous removals in the first place? e.g. a test-case where the first cycle creates two adjacent elements:

    <div>
        <div id="1" onremove={ (remove) => setTimeout(remove, 1000) }></div>
        <div id="2" oncreate={ log.push('created') }></div>
    </div>

Then, immediately after, a second call to patch removes the first element, which will trigger `onremove`, but the removal of the element from DOM won't happen until a second later: `oldNodes` will contain one entry, but the DOM will still contain two elements, and you'd expect to get the second one, but you're going to get the first one instead - you're going to patch that, and then a second later, the element you patched is going to get removed.

You can't safely do a premature call to patch, because, even though it's basing the diff logic on a set of old/new nodes, it isn't entirely free of DOM dependency: it does look up existing elements, and so it does depend on child indices - if these change uncontrollably, you can't possibly know if the DOM child indices have caught up with the indices of the old list of VNodes.

I don't see how there's any way around using promises for this.

src/index.js
JorgeBucaran: What issues do you imagine those would be?

JorgeBucaran:
> @mindplay-dk I'm pretty sure you could write a simple test-case to demonstrate this problem.

You said it's simple so can you just write the test? I still don't understand what you are concerned with and I suspect you may be misusing keys. You've gave us a lot to read and I am grateful for you leaving a well documented explanation of the issue, but it would be easier if there's code to look at. 

Maybe if we use a stack as I suggested the problem go away, as there would be no handlers to call left by the time the inner patch (the one you are worried about?) returns. Ignore that if it's something else what you mean.

zaceno:
@mindplay-dk Ah, ok I see the problem now.

I don't exactly have a solution, but perhaps more of an explanation why I haven't run into this issue myself.

Generally, when using lifecycle events like oncreate and onremove, it's a good idea to key your vnodes, so that creation / removal of vnodes is always conencted with the same actual dom element. 

For keyed vnodes, after patch has run through, the elements corresponding to the removed vnodes will always end up at the bottom of the list of siblings -- meaning that the patch function (which doesn't loop that far down) won't encounter them again.

So those elements can just hang out there and wait to get removed properly.

(this leads to a secondary problem for animated transitions, which is that elements in the process of being removed, are suddenly laid out differently. Which is why my hyperapp-transitions library is way more complex than one would think it should be)

But again it all hinges on using keys, or things will probably get very confusing like you described...

JorgeBucaran:
I was suspecting it was @mindplay-dk missing something about keys and now I think that is actually the issue. 😅 

> @zaceno But again it all hinges on using keys, or things will probably get very confusing like you described...

The only issue I see here is VDOM authors failing to document keys usage in more depth or making it a more prominent section in the docs.

src/index.js
zaceno: I can't say off-hand -- but I know I've run into trouble before with elements not being able to do everything I expect, when they're not mounted in the dom. But maybe those things wouldn't be relevant here. 

Anyhow, is there a reason for the ordering or is it arbitrary? If it's arbitrary I'd feel safer putting destroyNode first. That's all

zaceno:
> @JorgeBucaran The only issue I see here is VDOM authors failing to document keys usage in more depth or making it a more prominent section in the docs.

True. *Especially* for delayed removal. If you try to do a delayed removal without keys, it will totally mess up the next patch (like @mindplay-dk was describing).

mindplay-dk:
As far as I can figure, keys or no keys, this has no bearing on the issue I'm highlighting.

The list of elements gets initialized as follows:

```js
      for (var i = 0; i < oldLen; i++) {
        var oldElement = (oldElements[i] = element.childNodes[i])
        ...
      }
```

And the `while` loop uses the same index for `oldElements` as for `oldNode.children`:

```js
      while (j < len) {
        var oldElement = oldElements[i]
        var oldChild = oldNode.children[i]
        ...
        if (keyed[oldKey]) {
          i++
          continue
        }
        ...
      }
```

While the `if`-statement skips elements that *were* keyed, for indices *following* any indices with keyed nodes, the remaining indexes will still be off, as far as I can figure.

Anyhow, yes, the best assure this is not the case, would be to write a test. I'm pressed for time right now and don't know when I'll be able to do that, sorry.

JorgeBucaran:
@mindplay-dk If you use keys you won't be having this problem.

zaceno:
@mindplay-dk If you don't have time to write a test, perhaps you could play with this example on codepen (from my hyperapp-transitions examples) to demonstrate the problem (or lack thereof) 

Perhaps try adjusting the time it takes to remove a toast, so you have time to add a new one while the old is still dissapearing. I haven't tried that (I'm pressed for time too at the moment) but that would demonstrate your problem I think (but actually I think it will work fine since the example is using keys)

https://codepen.io/zaceno/pen/QOYOZd

mindplay-dk:
Watch closely and notice how clicking the two first toasts erases the correct ones - then, clicking two toasts in rapid succession, it ends up erasing the wrong element, the one that's still animating:

bug

As said, I'm fairly certain keys have no bearing on this issue - the problem is that the indices in DOM `children` are updated too late... for the duration of the animation, the indices are going to be wrong, because the removal from the old VNodes list was immediate, and removal from DOM is not.

mindplay-dk:
(you can test this yourself more accurately by changing the animation time to, say, 3 seconds.)

JorgeBucaran:
@mindplay-dk 🔥😮🔥

zaceno:
@mindplay-dk  ... that's very wierd.

I tried my best to reproduce it (even tried different browsers, on the chance that it's a browser-related bug in hyperapp-transitions, and not a vdom issue) -- but no where could i make it happen.  I tried setting the time of the exit transition long too (if anyone else wants to try, change line 5 to ` <transitions.exit time={3000} easing="ease-in-out" ...` to make it exit over 3 seconds) ... and generally mistreat the vdom as much I could, shuffling and adding and while stuff is being removed. But I couldn't make anything wierd happen.

That's not to say there isn't a problem though. You obviously saw (and screengrabbed!) something. But more investigations are needed before I'd be comfortable saying exactly what the problem is. This needs more eyes on it I'd say.

JorgeBucaran:
@zaceno Thanks, I didn't know how to set the time! And I wish I could, but couldn't repro @mindplay-dk's bug either. Can you explain what you did? Perhaps something else we can't see from the gif or code changes you forgot to mention? 🤔 

Also, what gif recording software did you use? (Just curious)

zaceno:
@mindplay-dk Also, just to be on the safe side: what browser did you use?

JorgeBucaran:
FWIW I tested it on both Chrome/Firefox.

src/index.js
frenzzy: > What issues do you imagine those would be?

For example unable to save last element dimensions because offsetHeight will return 0 for detached node.

mindplay-dk:
> Can you explain what you did? Perhaps something else we can't see from the gif or code changes you forgot to mention?

I didn't change anything.

Clicked @zaceno's link, started clicking around, and was able to repeat this issue.

Still am, though now it's behaving a bit differently... it may be a bug in the animation, as it appears I'm still seeing four elements in the inspector although the top-most element has vanished.

bug

> what browser did you use?

Current version of Chrome, Win10/64.

    Version 63.0.3239.84 (Official Build) (64-bit)

zaceno:
@mindplay-dk 
> it may be a bug in the animation

... quite possible. And bugs in hyperapp-transitions are likely browser (and even possibly OS based) unfortunately. 

So, considering the fact that your demonstrated odd behavior is not universally reproducible, I'd say it's likely due to browser inconsistencies that hyperapp-transitions isn't able to handler.

That's not to say you don't have a point with your original concern necessarily. But it seems just "clicking around" in the toasts-demo is not going to demonstrate it. The question is how else 🤔

JorgeBucaran:
@zaceno It's incredible though, I can see how in @mindplay-dk's demo the toast immediately disappears after the click, but I was unable to reproduce this no matter what I did.

zaceno:
@JorgeBucaran I agree. But what OS are you on? The thing is, I had to mess around quite a bit with timeouts and rAFs in hyperapp-transitions to get somewhat consistent behavior on the versions of chrome, safari and firefox that I had on my mac. So I wouldn't be surprised if this is about some subtle difference in different versions of chrome

JorgeBucaran:
Mac like you.

zaceno:
Aha... I'll see if I can test it on a windows machine here at work...

SkaterDad:
I was able to repeat @mindplay-dk 's issue in that Codepen without changing settings.  The 2nd element clicked in rapid succession disappears immediately without animation.   With dev tools open, the invisible element is still in the `Elements` tab.

Win 7, Chrome 63,

I tried testing in IE11 also, but it does really weird stuff.  :laughing:

JorgeBucaran:
Let's merge this! 🎉

JorgeBucaran:
@4lejandrito You did an amazing job! Thanks! 🙏

PR #480, opened 2017-12-09
Fix type definiton formatting

okwolf:
Noticed that our latest `hyperapp.d.ts` was changing contents after `npm run format`.


JorgeBucaran:
Great thanks! @okwolf 🎉

PR #482, opened 2017-12-09
Fix test formatting

okwolf:
Updating format command to include test js files and caught one that wasn't formatted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "bundle": "rollup -i src/index.js -o dist/hyperapp.js -m -f umd -n hyperapp",
     "minify": "uglifyjs dist/hyperapp.js -o dist/hyperapp.js -mc pure_funcs=Object.defineProperty --source-map includeSources,url=hyperapp.js.map",
     "prepare": "npm run build",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -25,7 +25,7 @@
     "bundle": "rollup -i src/index.js -o dist/hyperapp.js -m -f umd -n hyperapp",
     "minify": "uglifyjs dist/hyperapp.js -o dist/hyperapp.js -mc pure_funcs=Object.defineProperty --source-map includeSources,url=hyperapp.js.map",
     "prepare": "npm run build",
-    "format": "prettier --semi false --write \"src/**/*.js\" \"{,test/ts/}*.{ts,tsx}\"",
+    "format": "prettier --semi false --write {src,test}/**/*.js {,test/ts/}*.{ts,tsx}",
```
JorgeBucaran: @okwolf Didn't you made the previous commit yourself to support windows or something?

okwolf: The previous change was from single quotes to double quotes, but it doesn't look like quotes are even needed at all.

JorgeBucaran: @okwolf Sure, not needed in unix.

okwolf: @JorgeBucaran verified still working on Windows too! 🎉

Issue #477, opened 2017-12-07
Combine the concept of state and actions into models 🤯


JorgeBucaran:
> EDIT: @Alber70g suggested an even more extreme simplification:
>
> Instead of `app({ model, view }, root)` → `app(model, view, root)`, but we can discuss that on another issue.

We did it [once](https://github.com/hyperapp/hyperapp/issues/126) and here we go again. Like everything, there are trade-offs involved, not to mention the pain that comes with the churn. There's also "done is better than perfect", but Hyperapp is not done yet, and we still have some time before 1.0. I need your help to evaluate if this idea holds or not, and identify possible drawbacks or disadvantages.

The proposal (via @Alber70g) is to combine the concepts of state and actions into a new abstraction: _models_. A model is a regular object that holds your state props and actions.

The typical counter would look like this:

```jsx
app({
  model: {
    count: 0,
    down: () => model => ({ count: model.count - 1 }),
    up: () => model => ({ count: model.count + 1 })
  },
  view: model => (
    <main>
      <h1>{model.count}</h1>
      <button onclick={model.down}>–</button>
      <button onclick={model.up}>+</button>
    </main>
  )
})
```

An application that is wired to state and actions from multiple sources, could look like this:

```jsx
app({
  model: {
    A,
    B,
    C,
    D
  }
})
```

Instead of the current:

```jsx
app({
  state: {
    A: state.A,
    B: state.B,
    C: state.C,
    D: state.D
  },
  actions: {
    A: actions.A,
    B: actions.B,
    C: actions.C,
    D: actions.D
  }
})
```

And an app that subscribes to external events.

```jsx
const { tick } = app({
  model: {
    time: Date.now(),
    tick: () => ({ time: Date.now() })
  },
  view: model => <Clock time={model.time} />
})

setInterval(tick, 1000)
```

You can see that the app() now returns the wired `model`, so you can access not only actions, but also state props.

Now, you can think of models like a "module" that contains your state and actions under the same namespace.

```jsx
// A counter model
const counter = {
  count: 0,
  add: n => model => ({ count: model.count + n })
}

app({
  model: { counter },
  view: ({ counter }) => (
    <main>
      <h1>{counter.count}</h1>
      <button onclick={e => counter.add(1)}>Add</button>
    </main>
  )
})
```

Notice how usage of the model inside the view function mirrors how you define the model above.

### Pros
- Simplify your code (save bytes!)
- Simplify core (save bytes!)

### Cons
- Breaking change
[...]

Issue #439, opened 2017-10-27
Need to dynamic actions in order to support code splitting


Pyrolistical:
#435 showed its possible to write an HoA to encapsulate state+action+view into module and https://github.com/hyperapp/hyperapp/issues/435#issuecomment-340067619 showed its possible to load lazy said module, **except** for the module actions.

# Motivation
Code splitting.

# Problem
There is currently no way to add actions dynamically.

# Possible Solutions
## 1. Include `addAction` action in app
- solves the problem in the most direct way
## 2. Include `loadModule` action in app
- solves the bigger issue at the same time
## 3. Merge the concepts of `state` and `actions`
- state is expected to change, so what if we just put state and actions in the same place?
- replace `state` and `actions` with `root`, keep `view`
- allow `root` to hold both data and functions. functions would take in root and slice the tree as it does now
- functions return and object which will merge into the subtree of where it lives, it can return both data and more functions (aka actions!)
- since actions don't need to be there at init time, then modules don't need to exist either! modules can be put directly on the root or added during init or add lazily later

Option 3 is most interesting to me since it removes even more concepts from Hyperapp.

JorgeBucaran:
@Pyrolistical This is currently impossible given we don't have stateful components. It could work in userland if you have different apps talking to each other, which is not unheard of, but not what you want probably. 

I would prefer not having code splitting than implementing it in any of the proposed ways, sorry. But there's still hope, we just need to think more about it.

Dynamically loading views, when using e.g., a router, is feasible and I can see how this could be useful, given that views and components can make up a very large chunk of your app bundle too.

Pyrolistical:
I don't understand your stateful components comment. What does stateful components have to do with this?

JorgeBucaran:
AFIK, in React you can dynamically load components and its child components with all its moving parts, and setup codesplitting for that. In Hyperapp you can't because there is only one app and a single entry point.

Honorable mention: Have you looked into HOAs?

JorgeBucaran:
@Pyrolistical How about this?

```js
app(load => ({
  actions: {
    reactorStart() {
      import("/remote-module").then(load)
    }
  }
}))
```

Pyrolistical:
???? isn't `app(function)` the old HoA style that has been removed?

Pyrolistical:
we don't need stateful components for code splitting, we already have modules, state slice, its just dynamic actions that is the blocker

JorgeBucaran:
[...]

PR #484, opened 2017-12-09
Introduce models: state+actions 🤯

JorgeBucaran:
- Read the discussion here: https://github.com/hyperapp/hyperapp/issues/477
- You can try this out if you install hyperapp with `npm i hyperapp@hot`

---

## What changed?

This PR combines the concepts of state and actions into a single one, models.

A model is as if you literally merged current Hyperapp state and actions props into a single object. This change does not affect how slices work, so if you were wondering about that, now you know the answer!

How the app() function is called changed too. Instead of passing an object with named props, now you pass each prop as an argument like this:

```js
app(model, view)
```

or

```js
app(model, view, container)
```

Like before, if the view is undefined, nothing will be rendered. I found this order made the most logical sense.

### Counter

```js

const model = {
  count: 0,
  down: () => model => ({ count: model.count - 1 }),
  up: () => model => ({ count: model.count + 1 })
}

const view = model => (
  <main>
    <h1>{model.count}</h1>
    <button onclick={model.down}>-</button>
    <button onclick={model.up}>+</button>
  </main>
)

app(model, view)
```

### Clock
```jsx

const angle = t => 2 * Math.PI * t / 60000

const model = {
  time: Date.now(),
  tick: () => ({
    time: Date.now()
  })
}

const view = model => (
  <svg viewBox="0 0 100 100" width="200">
    <circle cx="50" cy="50" r="45" fill="blue" />
    <line
      x1="50"
      y1="50"
      x2={50 + 40 * Math.cos(angle(model.time))}
      y2={50 + 40 * Math.sin(angle(model.time))}
      stroke="white"
    />
  </svg>
)

const model = app(model, view)

setInterval(model.tick, 1000)
```

### Improve the nomenclature
Stores, models, modules, slices it's all the same. Models **organize** state and actions that belong together. They are not "mixed", as that simply suggests we are conflating things that should be apart. There is no violation of separation of concerns here. 

#### Before
```js
// ABC.js
import { B } from "./B"
import { C } from "./C"

export const state = {
  value: 0,
  B: B.state,
  C: C.state
}
export const actions = {
  add: value => state => ({ value: state.value + value }),
  B: B.actions,
  C: C.actions
}
```
#### After
```js
// ABC.js
import { B } from "./B"
import { C } from "./C"

export const model = {
  value: 0,
  add: value => state => ({ value: state.value + value }),
  B, 
  C
}
```

### Simplify slices
Because the action tree must mimic the state tree in order for slices to work, you need to duplicate the same structure twice for every namespace in the state/actions tree. You also need to keep this in mind at all times. 

#### Before
```js
import  { A } from "./A"
import  { B } from "./B"
import  { C } from "./C"

app({
  state: {
    A: A.state,
    B: B.state,
    C: C.state
  },
  actions: {
    A: A.actions,
    B: B.actions,
    C: C.actions
  }
})
```

#### After
```js
import  { A } from "./A"
import  { B } from "./B"
import  { C } from "./C"

app({ A, B, C })
```

#### Before

```jsx
const view = state => actions => (
  <App>
    <Feed state={state.feed} actions={actions.feed} />
    <Trend state={state.trend} actions={actions.trend} />
  </App>
)

// OR

const view = state => actions => (
  <App>
    <Feed
      model={{
        ...state.feed,
        ...actions.feed
      }}
    />
    <Trend
      model={{
        ...state.trend,
        ...actions.trend
      }}
    />
  </App>
)

// OR

import { location, Route, Link } from "@hyperapp/router"
import { auth } from "./simple-auth"

const otherState = { ... }
const otherActions = { ... }

const actions = app({
  state: {
    ...otherState,
    auth: auth.state,
    location: location.state
  },
  actions: {
    ...otherActions,
    location: location.actions,
    auth: auth.actions
  },
  view: state => actions => (
    <AuthExample
      location={{
        go: actions.location.go,
        previous: state.location.previous
      }}
      login={actions.auth.login}
      userId={state.auth.userId}
      isLoggedIn={state.auth.userId !== null}
    />
  )
})

const unsubscribe = location.subscribe(actions.location)
```

#### After
```jsx
const view = model => (
  <App>
    <Feed model={model.feed} />
    <Trend model={model.trend} />
  </App>
)

//  OR

import { location, Route, Link } from "@hyperapp/router"
import { auth } from "./simple-auth"

const otherModel = { ... }

const model = app(
  {
    ...otherModel,
    auth,
    location
  },
  model => (
    <AuthExample
      location={model.location}
      auth={model.auth}
      isLoggedIn={model.auth.userId !== null}
    />
  )
)

const unsubscribe = location.subscribe(model.location)
```

### Add actions dynamically

No example, but this basically means the ability to add new actions on runtime and pave the way for code splitting #439.

### Other

- Easier refactoring
- Less bugs because of typos
- This change encourages you to setup your model and view separately in a variable and then pass it to the app. This makes it easier envision how to import/export your model and view properly. 🎉 

## Cons

- When creating a model from [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage), you need to deep merge it with the original model source (because the actions are removed from the object when it is persisted).

--

Bundle size: 1442→**1399**


lukejacksonn:
I just had a play around with this and everything seems to check out ⛳️

Here is a codepen with the `src/index.js` of this PR imported already if anyone wants to experiment https://codepen.io/lukejacksonn/pen/mqNqjM

src/index.js
```diff
@@ -1,4 +1,4 @@
-export function h(type, props) {
+export function h(tag, props) {
```
selfup: With argument name changes like `tag` and `from`/`source` will we merge this first and have others update their current PRs? Just wondering!

JorgeBucaran:
@lukejacksonn I published this as `npm i hyperapp@hot`! 🔥 Let's play!

src/index.js
JorgeBucaran: @selfup Good question. I want to get some feedback here first, but what you say sounds good to me. It's just a tiny change anyway.

yuku-t:
In my understanding, actions are wrapped during `init()` for slice feature. Currently nothing happens to functions returned by actions, they are just set to model. Should we do the same thing to these returned functions? In other words, does hyperapp support defining *dynamic* actions? 🤔

JorgeBucaran:
@yuku-t Yes, it seems to be possible by changing init to this:

```jsx
function init(source, path) {
  for (var key in source) {
    typeof source[key] === "function"
      ? (function(key, action) {
          source[key] = function(data) {
            source = get(path, model)

            if (typeof (data = action(data)) === "function") {
              data = data(source)
            }

            if (data && data !== source && !data.then) {
              repaint(
                (model = setDeep(
                  path,
                  merge(source, init(data, path)),
                  model
                ))
              )
            }

            return data
          }
        })(key, source[key])
      : typeof source[key] === "object" &&
        !Array.isArray(source[key]) &&
        init(source[key], path.concat(key))
  }

  return source
}
```

Using that I made this contrived example:

```jsx
const model = {
  count: 0,
  new: () => store => ({
    count: store.count - 1,
    mult: () => store => {
      return {
        count: store.count * 10
      }
    }
  }),
  up: () => store => ({ count: store.count + 1 })
}

const view = store => (
  <main>
    <h1>{store.count}</h1>
    <button onclick={store.up}>Increment</button>
    <button onclick={store.mult}>Multiply ({store.mult ? "Ready" : "Not Ready"})</button>
    <button onclick={store.new}>Add Dynamic Action</button>
  </main>
)

app(model, view)
```

33804359-6629b620-dde6-11e7-8d86-5e6ba819a060

I don't know if this is a great idea or not, but it seems interesting! 

This means we need to call init for every action that schedules a render, but it's only called on the data returned by the action, not the entire model, so I don't think this is a significant performance hit.

zaceno:
@JorgeBucaran @yuku-t 

There are a couple problems I see with calling init on every returned store-slice by default.

One is of course the performance hit (which is probably small, but still).

The other is if you're doing some sort of integration with a 3d party component you may find yourself having to put things things like instance objects, in the store. Automatically calling init on that will break the instance's methods.

What if we, instead, passed init as a second argument, along with the model. So the signature for actions would become: `data => (store, init) => { .... }` ?  That would let the user *chose* when to apply init and not.

Additionally (unless I'm mistaken), it would make it possible to put store-slices in arrays, which would be very useful: 

```jsx
const TodoItem = text => ({
  text,
  done: false,
  toggle: _ => me => ({done: !me.done})
})

const TodoList = {
   list: [],
   add: text => ({list}, init) => ({list: list.concat(init(TodoItem(text)))})
}

app(TodoList, store => (
  <div>
    <input type="text" onchange={ev => {
      model.add(ev.currentTarget.value)
      ev.currentTarget.value = ''
    }} />
  <ul>
    {model.list.map(todo => <li class={todo.done ? 'done' : ''}>
      <input type="checkbox" checked={todo.done} onclick={todo.toggle} />
      {todo.text}
    </li>)}
  </ul>
)) 
```

src/index.js
```diff
@@ -9,113 +9,98 @@ export function h(type, props) {
...
-      children.push(typeof node === "number" ? (node = node + "") : node)
+      children.push(typeof node === "number" ? (node += "") : node)
     }
   }
 
-  return typeof type === "string"
+  return typeof tag === "string"
     ? {
-        type: type,
+        tag: tag,
```
zaceno: Why are we naming `type` back to `tag` again?

(I never knew why we changed it to type in the first place...)

lukejacksonn: I noticed this.. and @selfup noted that it might be easier for everyone if we don't bundle these changes in with this proposal. Perhaps @JorgeBucaran could revert these bits and make a separate PR with `type -> tag` if it is something that we want to do. It would make for a cleaner diff here.

JorgeBucaran: 👍

JorgeBucaran:
@zaceno Just going to paraphrase here from slack. Slices in arrays don't play well with serialization/deserialization so I am not keen on that idea. It would also require a complete rewrite of init and who knows what other changes. I prefer to think of those items in the array as "primitive/basic types" rather than models themselves.

Passing init to the action along with the model is one possibility. Another is to do it out of the box. Another is to put the `init` function in the object returned by `app()`, the `store` and let you do whatever you want. The good news is we can and there is not one, but several ways to accomplish this. We just need to pick the best. — easy right? 🙄

I am going to leave this (dynamic models) for a future release and proceed with the current PR and after I am done rewriting tests, publish 0.17.0 🎉

src/index.js
```diff
@@ -27,95 +27,80 @@ export function h(type, props) {
...
+
+              if (typeof (data = action(data)) === "function") {
+                data = data(source)
               }
 
-              if (data && data !== state && !data.then) {
-                repaint(
-                  (appState = setDeep(path, merge(state, data), appState))
-                )
+              if (data && data !== source && !data.then) {
```
Mytrill: I just thought of something here...
If you also check `&& !(data.type && data.props && data.children)` you can allow actions to return VNodes, i.e. wired components for cheap.

JorgeBucaran: @Mytrill Let me see.

lukejacksonn: This felt awfully fitting with respect to the API changes proposed here. Awesome that it could be done with such a small change to core.

JorgeBucaran:
@Mytrill Reminded me of [`JSON.parse(text[, reviver])`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse). Maybe we can use it to avoid a deep merge (since it's already traversing the object tree) and generate a complete model.

src/index.js
JorgeBucaran: @lukejacksonn @Mytrill It works, but also complicates the API. I need to think more about it. 😄

okwolf:
The pragmatic side of me thinks this is an improvement.

My ideological side says this is more complex since it very literally complects `state` and `actions`.

Right now the pragmatic side is winning.

okwolf:
I'm also struggling with what names make sense to choose for each of these:

```js
const model = {
  value: 0,
  inc: () => thing1 => ({ value: thing1.value + 1 })
}

const view = thing2 => h("main", {}, [
  thing2.value,
  h("button", { onclick: thing2.inc }, "+")
])

const thing3 = app(model, view)
```

In the examples `thing1`, `thing2`, and `thing3` were all called `store`, which feels like a naming collision to me. Things were easier to name when they weren't trying to be a mix of different things. Dr. Seuss would be pleased though.

src/index.js
```diff
@@ -27,98 +27,101 @@ export function h(type, props) {
     : type(props || {}, children)
 }
 
-export function app(props, container) {
+export function app(model, view, container) {
```
okwolf: @JorgeBucaran we should add a trigger warning for folks who have PTSD from their MVC experiences

src/index.js
```diff
@@ -27,98 +27,101 @@ export function h(type, props) {
...
+    next = view(model)
+
+    if (!lock) {
+      root = patch(container, root, node, (node = next))
+    }
+
+    while ((next = stack.pop())) next()
+  }
+
+  function schedule() {
```
okwolf: Why `schedule` instead of `repaint`? The former is less clear to me what it's doing than the latter.

src/index.js
JorgeBucaran: @okwolf Haha!

src/index.js
JorgeBucaran: Probably, I'm not really done though.

JorgeBucaran:
@okwolf Same. I am not sure what to call them either. This is a problem.

lukejacksonn:
If you move your call to `thing3.*` into the view and call it from a component there.. then you don't need to return or name anything returned from app 😉

src/index.js
```diff
@@ -33,7 +33,7 @@ export function app(model, view, container) {
   var node = vnode(root, [].map)
   var stack = []
 
-  schedule()
+  repaint()
```
okwolf: 👍

src/index.js
```diff
@@ -102,7 +102,7 @@ export function app(model, view, container) {
     while ((next = stack.pop())) next()
   }
 
-  function schedule() {
+  function repaint() {
```
okwolf: 👍

okwolf:
> @lukejacksonn If you move any call to thing3 into the view and call it from a component there, just using model.* ..then you don't need to return or name anything returned from app 😉

I'm not sure I'm following. I thought the whole point of `thing3` was to call actions from outside of Hyperapp, including outside of the view.

JorgeBucaran:
@okwolf It is and that's still an important key of the API. @lukejacksonn prefers doing everything inside the view, which is possible, but not as pleasant IMO.

lukejacksonn:
Why would you want to call actions from outside of Hyperapp, including outside of the view? How does it make your app _more pleasant_?

JorgeBucaran:
To subscribe to global events, etc. That's how the router works too.

lukejacksonn:
Subscribing to global events can be done inside the view, whats more is that you can unsubscribe from them from the view as well. It was also not an answer to the question either.. which was, what makes it _more pleasant_?

I am trying to find out what pushed you outside of the app, when everything you are doing outside of the app can be done inside (and more). I'm not saying you are wrong, just curious. What advantages I'm missing out on by keeping my apps simply `app(model, view)` (other than interop; if you were putting many apps on a page or using your hyperapp as part of a larger app).

zaceno:
@lukejacksonn The way I see it, while it's possible to "hoist" actions out of the view for interop purposes, it means I'm *not* making my app agnostic of the interop environment. If app returns the actions (or even the model with current state which would be 💯 ) then I don't need to think about interop *inside* my app, I just know it will work. That makes it more pleasing. Hoisting stuff out of view feels like a hack, which is unpleasant.

JorgeBucaran:
@lukejacksonn Subscribing and unsubscribing to global events using oncreate and the upcoming ondestroy (onremove for now on the top-level vnode) makes sense when you have a vnode, otherwise we don't have a better solution but to call actions outside the app call.

JorgeBucaran:
https://github.com/hyperapp/hyperapp/issues/489

Zhami:
> @lukejacksonn Why would you want to call actions from outside of Hyperapp, including outside of the view? How does it make your app more pleasant?

I am working on an application that connects to a websocketsc server to present a real-time updating view of quickly updating data streams. A variety of messages arrive. Presently, the piece of code that handles the incoming messages exists as a distinct module (i.e., it is not in a view). It is initialized with the `actions` object returned from invoking `app()`. This message handler invokes specific update actions, which make changes to different parts of the `state` (which is deeply nested). I don't want this aspect (message receipt handling) of my app in any view because there may be multiple views that are updated as a consequence to changes made by the receipt of a message. 

This may be a naïve understanding of how to implement an app with hyperapp since I am very new to hyperapp. Were I not using hyperapp with its top-down single state object, I would have wanted to have different view components "subscribe" to various objects in a store, and they would update themselves when the data changed. Now, it feels to me that the entire view is computed top-down (whether this is for each of these machine-gun-fire rate of state updates, or some accumulation of state updates related to animation frame request timing, I don't know). Am I approaching the architecture of my app appropriate for hyperapp?

btw: by posing this question, I am add fuel to the "framework" side of the discussion about library vs. framework.

JorgeBucaran:
@Zhami You could also create a node and use their oncreate/onremove lifecycle! :)

> machine-gun-fire rate of state updates

Hyperapp renders the view efficiently avoiding multiple re-renders when the state changes rapidly / in succession, you'll be fine.

SteveALee:
I see this was closed not merged but not why. Perhaps I missed something?

JorgeBucaran:
@SteveALee And also labeled as `wontfix`, meaning it will not be implemented. The tl;dr was this brings more problems than it solves, so I decided against it. I don't think there is one good issue explaining the rationale that way against this, but a lot of was spoken and talked about on Slack. We discussed this a lot.

SteveALee:
Thank you.

PR #487, opened 2017-12-10
Improve namespacing

w-mcilhagga:
These changes replace the ```isSVG``` flag with a namespace string. Namespacing is automatically triggered by using ```svg``` or ```math``` tags, and inherited by children. Namespacing can be explicitly triggered by adding a ```namespace``` to the ```props``` object passed to ```h```. This is useful for e.g. embedding HTML inside SVG via the ```foreignObject``` tag.

Defined namespace URIs are added as a property of the ```h``` function, for convenience - saves having to look up and type them when doing explicit namespacing.

The other change is to always ```createElementNS``` even for ```html``` tags. There is no performance advantage to ```createElement```

NB: my ```package.json``` and typescript definitions file are out of sync with the master - ignore please.


src/index.js
```diff
@@ -27,6 +27,12 @@ export function h(type, props) {
     : type(props || {}, children)
 }
 
+h.ns = {
```
JorgeBucaran: @w-mcilhagga What's this?

w-mcilhagga: This is the bit where I said:

Defined namespace URIs are added as a property of the h function, for convenience - saves having to look up and type them when doing explicit namespacing.

So if you're doing a foreign object inclusion in svg you can use a canned namespace e.g.

```javascript
h('svg',{},
    h('foreignobject',{width:100, height:100},
        h('p',{namespace:h.ns.html}, 'html included')
    )
)
```

Pure convenience for the end-user to expose the namespace strings via a property attached to h. They have to be defined anyway, so why not give them to the programmer as well.

Up to you if you like this or not. The only advantage to ***not*** exposing them is that you don't need to define the ```html``` namespace in the code.

JorgeBucaran: @w-mcilhagga Thanks! Yes, I would probably not expose anything to the user.

JorgeBucaran:
@w-mcilhagga Hyperapp currently picks up SVG automatically, how does this change affect that? 🤔

src/index.js
JorgeBucaran: @w-mcilhagga Ping

JorgeBucaran:
@w-mcilhagga Can you give it another shot at explaining what's happening here? This is going to add several more bytes to the dist bundle, so why do we want this?

w-mcilhagga:
Made some changes that reduce the build size by 0.05KB (the discussions around model=state+actions helped me understand the hyperapp mindset a little better, and I think the changes I've made make the pull request fit better with that)

But as to why hyperapp should do this?

1. It's the standard. hyperapp should handle the namespace property correctly. Users who ***want*** to give a namespace to an element shouldn't be told to go away, even if there are perhaps only a few of them.
2. The standard also says that the namespace can't be changed once element is created, so the update element code is changed to read the namespace rather than attempt to write it, and default to xhtml if null is returned. Returning null is only an issue with older browsers, according to MDN.
3. Since svg is the most common alternative to the xhtml namespace, it is automatically namespaced for convenience. (Dropped mathml automatic namespacing, since not that common outside of scientific/technical, which is my interest). This duplicates existing functionality.
4. Currently with hyperapp the svg namespace is a black hole that you can't escape once you are in it. The standard says you should be able to embed html, canvas, or video in svg with appropriate namespacing. This change lets you do that too.

JorgeBucaran:
@w-mcilhagga What's the total bundle size? This branch has unresolved conflicts, so I can't easily check myself without resolving the conflicts first. 😄

w-mcilhagga:
Yeah, I'm a bit out of sync. Windows says it is 3.07 KB (3,145 bytes). Used to be 3.12KB.

JorgeBucaran:
I mean, what is the bundle increase?

JorgeBucaran:
@w-mcilhagga Also can you explain how to use this? For example a mathml example or html embedded in svg.

w-mcilhagga:
Good questions - I have to catch the train right now but I'll get back to you when I'm at work

JorgeBucaran:
@w-mcilhagga Catch you later then! 😄

infinnie:
#678

PR #490, opened 2017-12-13
Make view an argument to app as app({state,actions}, view).

JorgeBucaran:
@whaaaley and I came up with an improvement to the current API without needing to merge state & actions, but making it easier to reduce boilerplate.

It consists of extracting the view from the props argument to app.

```jsx
app(props, view, container)
```

Props is the state and actions, same API as 0.16.2. While we are at it, I find it more logical to call it `model`, instead of props, so I'll do just that.

```jsx
const model = {
  state: {
    count: 0
  },
  actions: {
  }
}

const view = state => actions => (
  <main>
    <h1>{state.count}</h1>
  </main>
)

app(model, view)
```

So almost nothing changes, but the advantage of this approach is that it makes it now easier to merge multiple models.

```jsx
app(
  combine({ A, B, C, D }),
  view,
  document.getElementById("root")
)
```

or

```jsx
app(
  {
    ...model,
    ...combine({ A, B, C, D })
  },
  view,
  document.getElementById("root")
)
```

Of course you need to write that custom `combine` function, but surely this can be pushed to userland.

---

## `combine(models)`
Creates a single model from multiple models. A model is an object of `{state,actions}` pairs. This function creates a single model from multiple models organized by their keys.

```jsx
const A = {
  state: { ... },
  actions: { ... }
}
const B = {
  state: { ... },
  actions: { ... }
}
const C = {
  state: { ... },
  actions: { ... }
}

combine({ A, B, C }) /*
{
  state: {
    A: A.state,
    B: B.state,
    B: B.state,
  },
  actions: {
    A: A.actions,
    B: B.actions,
    B: B.actions,
  },
}
*/
```

### Imperative

```jsx
const combine = models => {
  var out = {state:{},actions:{}}
  for (var key in models) {
    out.state[key] = models[key].state
    out.actions[key] = models[key].actions
  }
  return out
}
```

### Functional

```jsx
const combine = models =>
  Object.keys(models).reduce(
    (obj, key) => ({
      ...obj,
      ...{
        state: {
          ...obj.state,
          ...{ [key]: models[key].state }
        },
        actions: {
          ...obj.actions,
          ...{ [key]: models[key].actions }
        }
      }
    }),
    {}
  )
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,9 +34,9 @@
   "devDependencies": {
     "babel-preset-env": "^1.6.1",
     "jest": "^21.2.1",
-    "prettier": "^1.8.2",
-    "rollup": "^0.52.0",
-    "typescript": "^2.6.1",
-    "uglify-js": "^3.2.1"
+    "prettier": "^1.9.2",
+    "rollup": "^0.52.1",
+    "typescript": "^2.6.2",
+    "uglify-js": "^3.2.2"
   }
 }
--- a/src/index.js
+++ b/src/index.js
@@ -27,15 +27,15 @@ export function h(type, props) {
     : type(props || {}, children)
 }
 
-export function app(props, container) {
+export function app(model, view, container) {
   var lock
   var root = (container = container || document.body).children[0]
   var node = vnode(root, [].map)
   var lifecycle = []
-  var appState = props.state || {}
+  var appState = model.state || {}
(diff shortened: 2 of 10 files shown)


okwolf:
This is a much smaller change, but I feel like it's a step in the right direction.

PR #497, opened 2017-12-16
1.0.0

JorgeBucaran:
## Summary
This PR contains probably my last major changes to core before 1.0. After I merge this, I'll proceed to merge #498 and release 0.18.0 so those using hyperapp right now can update to the latest and not need to wait until 1.0, which should happen in the coming weeks.

## Boilerplate

You can reduce setup boilerplate with a function that maps your props to a model.

```jsx
const { state, actions } = mapPropsToModel({
  counter, mouse, keyboard, location
})
```

**OR**

```jsx
const { state, actions } = mapPropsToModel({
  count: 0,
})
```

### Counter.js

```jsx

const state = {
  count: 0
}

const actions = {
}

  <main>
    <h1>{state.count}</h1>
  </main>
)

```

### Counter.ts

```tsx
import { h, app, WiredActions } from "hyperapp"

namespace Counter {
  export interface State {
    count: number
  }

  export interface Actions {
    sub(): State
    add(value: number): State
  }

  export const state: State = {
    count: 0
  }

  export const actions: WiredActions<State, Actions> = {
    sub: () => state => ({ count: state.count - 1 }),
    add: (value: number) => state => ({
      count: state.count + value
    })
  }
}

const main = app<Counter.State, Counter.Actions>(
  Counter.state,
  Counter.actions,
  (state, actions) => <div>{state.count}</div>,
  document.getElementById("app")
)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,15 +17,16 @@ Hyperapp is a JavaScript framework for building frontend applications.
 ```jsx
 import { h, app } from "hyperapp"
 
-const model = {
-  state: { count: 0 },
-  actions: {
-    down: () => state => ({ count: state.count - 1 }),
-    up: () => state => ({ count: state.count + 1 })
-  }
+const state = {
+  count: 0
 }
 
-const view = ({ state, actions }) => (
+const actions = {
+  down: () => state => ({ count: state.count - 1 }),
+  up: () => state => ({ count: state.count + 1 })
+}
+
+const view = (state, actions) => (
   <main>
     <h1>{state.count}</h1>
     <button onclick={actions.down}>-</button>
@@ -33,7 +34,7 @@ const view = ({ state, actions }) => (
   </main>
 )
 
-const { actions } = app(model, view, document.body)
+export const main = app(state, actions, view, document.body)
 ```
 
 ## Community
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -1,27 +1,27 @@
 # Contributing to Hyperapp
 
-Thank you for taking the time to read our contribution guidelines. You can contribute in many ways like writing tutorials, improving the documentation, and filing bug reports.
+Thank you for taking the time to read our contribution guidelines. You can start contributing in many ways like filing bug reports, improving the documentation, and helping others.
 
 ## Style
 
-- We write in ES5.
-- Format your code before creating a new commit using `npm run format`.
-- We prefer keeping all the moving parts inside as few files as possible. We don't have plans to break up the library into smaller modules.
+* We write in ES5.
+* Format your code before creating a new commit using `npm run format`.
+* We prefer keeping all the moving parts inside as few files as possible. We don't have plans to break up the library into smaller modules.
 
 ## Bugs
 
-- Before submitting a bug report, search the issues for similar tickets. Your issue may have already been discussed and resolved. Feel free to add a comment to an existing ticket, even if it's closed.
-- Determine which repository the problem should be reported in. If you have an issue with the Router, you'll be better served in [hyperapp/router](https://github.com/hyperapp/router), etc.
-- If you have a question or need help with something you are building, we recommend joining the [Hyperapp Slack Team](https://hyperappjs.herokuapp.com).
-- Be thorough in your title and report, don't leave out important details, describe your setup and include any relevant code with your issue.
-- Use GitHub [fenced code blocks](https://help.github.com/articles/creating-and-highlighting-code-blocks/) when sharing code. If your code has JSX in it, please use <code>```jsx</code> for best syntax highlighting.
+* Before submitting a bug report, search the issues for similar tickets. Your issue may have already been discussed and resolved. Feel free to add a comment to an existing ticket, even if it's closed.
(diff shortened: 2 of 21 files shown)


okwolf:
@JorgeBucaran isn't your `Counter.js` example missing the `app` call?

```js
app(state, actions, view, document.getElementById("app"))
```

Also, I'd like to mention that one of the things I liked most about Hyperapp when I first came across it was passing the parameters to `app` as a single object. Then we moved the `root`/`container` to a separate argument.

Now it looks like we've swung to the other extreme of spreading the parameters out to their own arguments. I've probably been biased by [Rich Hickey](https://youtu.be/2V1FtfBDsLU?t=46m51s), but I much prefer named arguments à la:

```js
fn({
  state: a,
  actions: b,
  view: c
})
```

versus positional semantics:

```js
fn(a, b, c)
```

Yes, you can always look it up in the docs, but with the object approach, a dev build of Hyperapp could actually know that a user got the name of something wrong, instead of crashing and burning in weird ways by trying to call your `actions` as the `view` because you got the order of the arguments wrong. Not to mention the names would be preserved even in a minified build you were unlucky enough to have to debug in production without source maps.

selfup:
Are we ditching `model` already from `0.17.1`?

Just wondering

zaceno:
What does app return now? Anything?

okwolf:
> @zaceno What does app return now? Anything?

Pretty sure we are back to `app` returning `actions` again: https://github.com/hyperapp/hyperapp/blob/1.0.0/src/index.js#L40

zaceno:
@okwolf thanks, good on you for checking the code! Wonder why I didn't do that?! 😆 Anyway, so my warning holds true: no async access to current state, except via actions (which may not always be desirable)

okwolf:
@zaceno state getters are dead, long live state getters.

zaceno:
For example: no access to *current* state in `onremove` handlers. Or in `setTimeouts` called from event handlers.

JorgeBucaran:
@okwolf 

```js
```

okwolf:
> @zaceno Oh I totally forgot about the `{ getState: _ => state => state, ...}` hack. That probably works still.

That was what I meant by state getters.

JorgeBucaran:
@selfup Ditching models as a built-in concept, because you can easily do it yourself.

```jsx
const model = {
  state: { ... },
  actions: { ... }
}

const { state, actions } = model

```

JorgeBucaran:
> @okwolf ...because you got the order of the arguments wrong.

I guess one could get the order wrong, specially when using Nodepad or TextEdit.

Remember you call the app function only once, so if you make a mistake, you fix it only once. And if you prefer named props, then write an HOA that takes your props and calls the app function for you! 🎉

JorgeBucaran:
@zaceno `getState` is not a hack. And:

> For example: no access to current state in onremove handlers. Or in setTimeouts called from event handlers.

`getState` or create a vnode wrapper in the view function.

selfup:
Ok thanks for clearing that up!

okwolf:
> @JorgeBucaran And if you prefer named props, then write an HOA that takes your props and calls the app function for you!

That is true. This was my only concern about the API, and I don't feel strongly enough about it to fight passionately for it.

I think we just need docs updates and this will be good to merge, right?

JorgeBucaran:
And types!  💪🔥

okwolf:
> @JorgeBucaran And types! 💪🔥

So should #498 merge into this branch instead of master then? 🤔

JorgeBucaran:
I'll merge here first, then merge @Mytrill's PR and finally work a bit more on the types myself as I need some changes, some of which are not so technical, but "artistic" and don't want to bother @Mytrill! 😅💦

JorgeBucaran:
@zaceno I updated the example to illustrate the app returns an object I'm going to call `main`. This object represents your app and it's how you communicate with it from the "outside". It's nothing but the good old `actions`, **wired** to the state-update—view-render mechanism. tl;dr app returns your actions.

```js
```

zaceno:
@JorgeBucaran 
>  create a vnode wrapper in the view function and use the lifecycle.

If you're talking about defining getters in the view, that's precisely what *doesn't* work for `onbeforeremove`,  `onremove`, and inside `setTimeout` called from event handlers (for example) -- because by the time they're called, any state they need may be outdated. 

One could solve it by maintaining an up to date synced copy of the state you might need to access async (maybe that's what you were talking about?) but that's obviously far from ideal.

As you mentioned, `getState` works too, but that's *also* not ideal, because that causes a run through merge, and a repaint for no good reason (all you wanted was to read the state)

So it's not an unsolvable problem. But I think that whoever else encounters this class of problems (including @JorgeBucaran, when and if it happens), will feel that the available solutions are unsatisfactory.

JorgeBucaran:
@zaceno Sorry, what are you talking about? I am interested in this problem, but here seems like an odd place to discuss this, that's maybe why I don't understand what you are saying.

zaceno:
@JorgeBucaran Yes this is an odd place, I agree. 😄 - It's the same thing I've been talking about before on slack, but it probably deserves a proper issue, so I will open one.

PR #507, opened 2017-12-26
Merry Xmas! 🎁

JorgeBucaran:
# 1.0

- Slices ✅ 
- Automatic SSR DOM re-hydration ✅ 
- Tiny size: 1398 bytes
- Simple API ✅ 💁‍♀️
  ```jsx
  import { state, actions, view } from "./app"
  export const main = app(state, actions, view, document.body)
  ```
- TypeScript ✅ 
- ~vnode.type~ → vnode.name
- Lifecycle events (oncreate, onupdate, onremove, ondestroy)
  - ~onremove~ → ondestroy
  - ~onbeforeremove~ → onremove
    - onremove(element, **done**)
- Combining multiple {state,actions} objects
  - See [combineModels](https://gist.github.com/JorgeBucaran/4dfa567e38ca222c25c46184b3a5df99)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Travis CI](https://img.shields.io/travis/hyperapp/hyperapp/master.svg)](https://travis-ci.org/hyperapp/hyperapp) [![Codecov](https://img.shields.io/codecov/c/github/hyperapp/hyperapp/master.svg)](https://codecov.io/gh/hyperapp/hyperapp) [![npm](https://img.shields.io/npm/v/hyperapp.svg)](https://www.npmjs.org/package/hyperapp) [![Slack](https://hyperappjs.herokuapp.com/badge.svg)](https://hyperappjs.herokuapp.com "Join us")
 
-Hyperapp is a JavaScript framework for building frontend applications.
+Hyperapp is a JavaScript library for building frontend applications.
 
 * **Minimal**: Hyperapp was born out of the attempt to do more with less. We have aggressively minimized the concepts you need to understand while remaining on par with what other frameworks can do.
 * **Functional**: Hyperapp's design is inspired by [The Elm Architecture](https://guide.elm-lang.org/architecture). Create scalable browser-based applications using a functional paradigm. The twist is you don't have to learn a new language.
--- a/src/index.js
+++ b/src/index.js
@@ -33,7 +33,7 @@ export function app(state, actions, view, container) {
   var node = vnode(root, [].map)
   var stack = []
 
-  repaint(init([], (state = assign(state)), (actions = assign(actions))))
+  repaint(init([], (state = copy(state)), (actions = copy(actions))))
 
   return actions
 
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -2,7 +2,7 @@
 
 [![Travis CI](https://img.shields.io/travis/hyperapp/hyperapp/master.svg)](https://travis-ci.org/hyperapp/hyperapp) [![Codecov](https://img.shields.io/codecov/c/github/hyperapp/hyperapp/master.svg)](https://codecov.io/gh/hyperapp/hyperapp) [![npm](https://img.shields.io/npm/v/hyperapp.svg)](https://www.npmjs.org/package/hyperapp) [![Slack](https://hyperappjs.herokuapp.com/badge.svg)](https://hyperappjs.herokuapp.com "Join us")
 
-Hyperapp is a JavaScript framework for building frontend applications.
+Hyperapp is a JavaScript library for building frontend applications.
```
okwolf: 👍  for consistency!

infinnie:
That 1KB size limit is gone.

Swizz:
Hyperapp is 1.4 KB since a while.

Take a look at the bundlephobia history
https://bundlephobia.com/result?p=hyperapp@0.13.0

README.md
selfup: Either or 😄 But yea consistency is key 🎉

src/index.js
```diff
@@ -40,7 +40,7 @@ export function app(state, actions, view, container) {
   function vnode(element, map) {
     return (
       element && {
-        type: element.tagName.toLowerCase(),
+        name: element.nodeName.toLowerCase(),
```
selfup: I much prefer `name` / `nodeName` as well 😄

src/index.js
```diff
@@ -130,20 +127,22 @@ export function app(state, actions, view, container) {
   function setElementProp(element, name, value, oldValue) {
     if (name === "key") {
     } else if (name === "style") {
-      for (var i in copy(oldValue, (value = value || {}))) {
-        element.style[i] = null == value[i] ? "" : value[i]
+      for (var i in copy(oldValue, value)) {
```
selfup: Do we default to `{}`?

Seems we do?

src/index.js
JorgeBucaran: Exactly!

src/index.js
JorgeBucaran: @selfup `copy` will return at least {}, the trick is in the assignment in the line below:

```js
element[name][i] = null == value || null == value[i] ? "" : value[i]
```

...here we make sure to access `value[i]` only if `value` exists. I found this way reduced a few bytes and kept the same functionality (tests passed).

JorgeBucaran:
@infinnie FYI the project's hard limit is 1530 bytes (1530/1024=> 1.49KB), currently we are at 1.36KB.

calderaro:
esto es un gran regalo! 👍 👍

okwolf:
> @JorgeBucaran FYI the project's hard limit is 1530 bytes (1530/1024=> 1.49KB), currently we are at 1.36KB.

Not to be that guy or anything, but technically shouldn't the limit be 1535 bytes? 🤓 

I do think that since we're hovering around 1400 bytes, we should try and remove something less important whenever we decide to add an important new feature.

JorgeBucaran:
@okwolf That's going to be tough! 😉

src/index.js
selfup: Ok awesome! Just making sure 🎉

okwolf:
:shipit: 🚢  🚀

selfup:
This is it! 🌲

src/index.js
```diff
@@ -63,31 +63,28 @@ export function app(state, actions, view, container) {
   }
 
   function repaint() {
-    if (view && !lock) {
```
okwolf: Is view a required parameter now? This can lead to `TypeError: view is not a function` later when using Hyperapp without a view. 😕

JorgeBucaran: @okwolf Yes, it's required now! So, don't use hyperapp without a view.

Btw, you don't need to think about this function as solely serving the purpose of a view. You can think of it as an event that is subscribing to state changes.

okwolf: @JorgeBucaran thanks for clearing that up! Is there a reason we are [checking if `container` is defined](https://github.com/hyperapp/hyperapp/blob/c7741845556c3f101e931ed7e29ab183c56317ff/src/index.js#L58) then? I think we might want to skip calling the `view` function too and not just `patch`? 🤔 

Either way, should we make this clearer in the docs regarding what's required?

JorgeBucaran: Ping @Swizz

JorgeBucaran: @okwolf The idea is to avoid patching `null` or `undefined` if no container is given. But when would you do that? I think @Swizz could answer that.

Swizz: We made this choice to allow users to use hyperapp as a State manager without the Vdom engine. 
Basically you can now use the view hook to call you own render engine (canvas, native, worker events, etc..).

We removed the optional container capabilities because of if you attempt to use a HTML element that doesn't exist as container the whole body will be used as container and this can lead to an unexpected behavior. 

Finally, we can said you would use hyperapp without a container but why would you use hyperapp without a view/render ?

PR #510, opened 2017-12-28
Added docs for actions and substate

Alber70g:


selfup:
This is much needed! Thank you 🎉

docs/concepts/README.md
```diff
@@ -1,5 +1,6 @@
 ## Concepts
 
+* [Actions](actions.md)
```
selfup: 🙏

docs/concepts/actions.md
```diff
@@ -0,0 +1,64 @@
...
+Actions are _the_ way to change the state. Actions have this signature: `payload => (state, actions) => newState`. Actions are passed to the `app` as the second parameter ([app](tbd)).
+
+## Usage
+
+Actions get the payload passed as parameter of the first function. This can return an `object` (the new state) or a function. When returning an `object` it'll be (shallow) merged with the state. Otherwise it'll be called and passed the `state` and `actions` which in turn returns a new state object that will be (shallow) merged.
+
+- `() => newState`
+- `payload => newState`
+- `payload => (state, actions) => newState`
+
```
selfup: Clear as day! 🌞

docs/README.md
```diff
@@ -4,7 +4,7 @@
   * [Hello World](introduction/hello-world.md)
   * [Installation](introduction/installation.md)
 * [Concepts](concepts/README.md)
-  * [Actions](concept/actions.md)
+  * [Actions](concepts/actions.md)
```
selfup: good catch! 😂

docs/concepts/actions.md
```diff
@@ -0,0 +1,64 @@
...
+### Example
+
```
JorgeBucaran: These examples are a bit shoddy and don't match the spirit of the rest of the docs and Hyperapp in general.

docs/concepts/actions.md
```diff
@@ -0,0 +1,64 @@
...
+The `state` parameter that's passed to an `action` can also be a `substate`. A `substate` is a part of the state defined by the `path` on which the action is defined on the actions object. This allows us to have actions and components being reused in different places.
+
+### Example 
+In the following example you see that the `inputActions` can be 'reused' across parts of the actions tree because it get's the part that is relevant for that particular `substate`.
+
+The property `state.view` is not touched because the result of `inputActions.setValue` which is 'mounted' on `actions.loginForm.username` and `actions.loginForm.password` is being (shallow) merged with the paths `state.loginForm.username` and respectively `state.loginForm.password`.
+
+```js
+const state = {
```
JorgeBucaran: None of these examples use Hyperapp style conventions.

docs/concepts/actions.md
```diff
@@ -0,0 +1,64 @@
+# Actions
+
+Actions are _the_ way to change the state. Actions have this signature: `payload => (state, actions) => newState`. Actions are passed to the `app` as the second parameter ([app](tbd)).
```
JorgeBucaran: I would prefer to put the signature on a separate line.

docs/concepts/actions.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Usage
+
```
JorgeBucaran: These explanation is too elaborate.

docs/concepts/actions.md
Alber70g: What is the Hyperapp style convention and where can I read about them?

docs/concepts/actions.md
```diff
@@ -0,0 +1,64 @@
...
+Actions get the payload passed as parameter of the first function. This can return an `object` (the new state) or a function. When returning an `object` it'll be (shallow) merged with the state. Otherwise it'll be called and passed the `state` and `actions` which in turn returns a new state object that will be (shallow) merged.
+
+- `() => newState`
+- `payload => newState`
+- `payload => (state, actions) => newState`
```
JorgeBucaran: I don't really like using `code = like > this` unless it is for a single variable. It doesn't look very pretty.

docs/concepts/actions.md
JorgeBucaran: `npm run format`

docs/concepts/actions.md
Alber70g: What should be the example?

docs/concepts/README.md
JorgeBucaran: I feel this is needed, but it's weird we are talking only about actions and not about state or the view. 

If actions get their own section on the docs, then the explanation in hello-world.md needs to be simplified, since it is currently trying to explain everything about actions.

docs/concepts/actions.md
JorgeBucaran: No idea! 😕

docs/concepts/actions.md
selfup: Actions are _the_ way to change the state and have this signature: `payload => (state, actions) => newState`.

***

That's nice and short!

docs/concepts/actions.md
selfup: I like it... but what would you suggest?

PR #513, opened 2017-12-28
Bit of a refactoring of the docs to make it better findable.

Alber70g:
See also issue #514


JorgeBucaran:
@Alber70g Why?

Alber70g:
As I find it struggling to not have separate docs for each of the concepts/functions in Hyperapp (and [others](https://www.reddit.com/r/programming/comments/7mphvg/introducing_hyperapp_10_1_kb_javascript_library/drwf6jf/) as well), I thought it'd be good to create them.

JorgeBucaran:
> I'm still struggling with how to format the documents so they're not too overwhelming, but still describe what can be done with each part.

I know. This is really hard, that's why I published 1.0 with the current docs. 🤔

JorgeBucaran:
Does this PR depend on your other PR?

PR #516, opened 2017-12-29
Added a badge for npm download

rishabh09:


Alber70g:
While you're at it, can you put them on separate lines? Markup doesn't break when they're on separate lines.

JorgeBucaran:
Thanks @rishabh09! I don't know if I want to add more badges, it's already pretty loaded. Please don't close as I'd like to think about it, but I wouldn't hold my breath to it! 😅

JorgeBucaran:
@Alber70g Prettier should do that for you! 😄

Alber70g:
Oh wow that's annoying ;)

JorgeBucaran:
@Alber70g I think I fixed that with `"prettier.proseWrap": false`.

JorgeBucaran:
Sorry folks! I messed up my GPG and had to push --force to master. I didn't know it was going to mess up the PR like this. 🙇🙇  

With regards to adding the downloads badge, I feel there are already too many badges on the README.

PR #520, opened 2017-12-31
Numeric Text Nodes

okwolf:
Handle text nodes as part of `patch` instead of converting to  a string in `h`.

Existing tests make sure this logic is covered.

-4 bytes

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -14,7 +14,7 @@ export function h(name, props) {
       }
     } else if (null == node || true === node || false === node) {
     } else {
-      children.push(typeof node === "number" ? node + "" : node)
+      children.push(node)
     }
   }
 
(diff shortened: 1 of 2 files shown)


JorgeBucaran:
@okwolf I think this will break something else, so like your other PR we'll need a test for that. The problem is I am not sure what "that" is yet.

okwolf:
@JorgeBucaran I've tested this out in some pens and haven't found anywhere this would affect other than the creation of nodes so far. It looked funny that this logic should be happening where it did to me. I understand why the `h` is responsible for filtering out certain falsey values, but I would think you'd want to pass on through the rest.

okwolf:
Would this also yield a _very slight_ performance improvement from not having to create string copies for each of the numbers in your view? 🤔

JorgeBucaran:
@okwolf I think this looks good, let me test a few things.

PR #521, opened 2017-12-31
Fix typo in docs

avarun42:
Should say "PAUSE" instead of "PAUSED"


JorgeBucaran:
Sounds right. Thanks @avarun42! 😄👍

jozsefsallai:
@JorgeBucaran I believe this should be updated in the CodePen example as well! 😃

PR #525, opened 2018-01-01
Move hydrate from core to userland 🚰

okwolf:
Why not start 2018 off right by losing some weight in Hyperapp? 🎉 

Most new apps won't need SSR/hydration initially, and those that will are the more sophisticated users on larger projects that probably won't mind adding a small function to do this in userland:
```js
const hydrate = element => {
  const hydrateNode = node =>
    node && {
      name: node.nodeName.toLowerCase(),
      props: {},
      children: [...node.childNodes].map(
        child =>
          child.nodeType === Node.TEXT_NODE
            ? child.nodeValue
            : hydrateNode(child)
      )
    }
  return hydrateNode(element.children[0])
}
...
container = document.getElementById("app")

app(state, actions, view, container, hydrate(container))
...
```

Adding a new optional argument isn't a breaking change (although this would be easier with the old object props instead of spread out), and I considering removing hydration from core more of a performance change since the initial `patch` will handle a nullish `oldNode` perfectly fine by creating all new elements.

This saves 57 bytes in the process.


selfup:
Interesting idea and valid points. Doing `import hydrate from '@hyperapp/hydrate';` seems reasonable!

Either way I am down (keep it or leave it).

Let's see what others have to say 😄

okwolf:
Considering this one feature (that not many seem to actually use) makes up over 4% of our small library footprint, I'd like to at least consider moving this out of our core responsibilities.

JorgeBucaran:
@okwolf Let's gather more feedback. I find small just works things like these seal the deal more often than not.

zaceno:
I'm for removing hydration from core. Mainly from the perspective that I think routing is a more popular feature than hydration, and since routing doesn't get to be in core, why should hydration? In fact, if it were possible to replace routing with hydration I think that would be a win. Maybe we could reduce the router code size by putting it in core (allow shortcuts other modules don't get) too.

I know a lot of people would be against that, and I'm not 100% sure it's a good idea either, but it makes more sense than hydration in core imo.

JorgeBucaran:
Bringing back the router into core is an interesting idea. I am not sure about hydration. I think there was an issue about this already and it was closed as wontfix.

zaceno:
> @JorgeBucaran ...but then we would not be a 1 KB library anymore...
... unless we take out hydration? Or is the size difference still too big?

Anyway, just to clarify my preference here:

1) *neither* hydration or routing in core
2) routing in core, hydration out of core
3) hydration in core, routing out of core
4) hydration and routing in core

daynox:
In preference order :
1. neither hydration or routing in core
2. routing in core, hydration out of core

SkaterDad:
I'm good with neither routing or hydration being in core, so long as we can still add hydration easily when needed.

If routing is put into core again, I'd like it tree-shakable and not super baked in, since I have my own routing implementation.

lukejacksonn:
I'd be happy to see hydrate abstracted to it's own repo but don't see much to be gained by pulling the router into core.

dmitry-vakhnenko:
Why you want routing in core?

is next step will be drop routing from core and append something like `effects` or introducing new v-dom algo in core?

Please be consistent.

I propose drop hydration from core and save routing outside because @JorgeBucaran want improve v-dom mechanism that will grow hyperapp size.

and guys, routing inside core will be bad idea because someone want other (own) routing solution that will be better (for they project) or will have more features.

Dont forget some special cases: hyperapp usable not only in browsers.

@JorgeBucaran we can drop hydration from core and just wait new v-dom algo without introducing router in core.

zaceno:
So far I don't think anyone has said they actually want routing in core. I only said it would make more sense than hydration in core since routing is a more popular feature. So, I'm basically just saying: taking hydration out makes sense. Nothing more.

okwolf:
What do you folks think of using an object instead of multiple args for additional options we might add?

```js
app(state, actions, view, container, { ...optionsHere })
```

@vdsabev already brought this up: https://github.com/hyperapp/hyperapp/issues/532#issuecomment-355371483

I already think having 4 positional arguments is pushing it, and would like to consider moving back to a single props argument for `app` in 2.0.

JorgeBucaran:
An options object is something we can always add **someday** if there is the need for it! 

> ...and would like to consider moving back to a single props argument for app in 2.0.

Sorry, no.

okwolf:
@JorgeBucaran that's fine, just one more HOA to write 😂

JorgeBucaran:
@okwolf It seems rewriting the history messes up active PRs real bad, just look at the mess I caused on this PR. 😓 

As for whether we should keep hydration in core or not, I'd prefer to keep it for now.

okwolf:
@JorgeBucaran if that's the case I won't recreate the PR right away.

PR #527, opened 2018-01-01
Installation process

sadick254:
Having an installation process stated will help devs get up an running quickly with hyper app. The current README is not clear on how to setup hyperapp in your dev environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,46 @@ Hyperapp is a JavaScript library for building frontend applications.
 
 [Read the Docs](/docs/README.md#documentation)
 
+## Installation
+
+```
+yarn add hyperapp
+or 
+npm install --save hyperapp
+```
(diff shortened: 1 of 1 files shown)


Swizz:
Did you see the https://github.com/hyperapp/hyperapp/blob/master/docs/introduction/installation.md ?

What about adding a link from the README to this file ?

sadick254:
My bad. I didn't see that. Adding a link would do but i think it hyperapp would benefit from having https://github.com/hyperapp/hyperapp/blob/master/docs/introduction/installation.md  on the root README file.

JorgeBucaran:
@sadick254 It's a matter of style. I prefer to keep that information on the docs, but thanks for the suggestion and nice you showed how to get up and running with parcel! 💯

PR #535, opened 2018-01-03
Added typings for wired actions

Alber70g:
Please have a look at the added typings. This shouldn't be breaking. It only adds more security to the typescript project.

[edit: a bit more context]
With this change the state and actions object needs to adhere to the StateType and ActionsType. 
This protects you from passing a wrong type as generic into `app<State, Actions>`. 

```ts
export function app<
  State extends StateType,
  Actions extends ActionsType<State, Actions>
>
```


JorgeBucaran:
I thought we were calling those (the wired actions) simply `Actions`. 

Ping @Mytrill.

Alber70g:
That would be fine, but it cannot be Actions in the generic of `app<State, Actions>` because the `Actions` generic defines what can be passed into `app(state, actions: Actions, ....)`

JorgeBucaran:
@Alber70g My understand is that there are two types of actions in the typings:

- Actions: The actions returned by app.
- ActionsType: The actions implementation.

Alber70g:
Thats fine as well. Either way we cannot use Actions for both the generic
in app and the return value from app

JorgeBucaran:
> @Alber70g This protects you from passing a wrong type as generic into app<State, Actions>.

Can you provide two examples? One using the current typings and one without where it shows how this PR adds more security?

Alber70g:
Examples with state:

More security
```ts
  export interface State extends StateType {
    count: number,
    view: () => {} // this is not allowed when using extends StateType
  }
```

Less security
```ts
export interface State {
    count: number,
    view: () => {} // this is now allowed even though you cannot put functions in the state.
  }
```

JorgeBucaran:
@Alber70g  Why is there a view on the State? 😕

Alber70g:
Fixed it. please its an example...

JorgeBucaran:
@Alber70g I know it's an example, but I need to understand why you want me to merge this. Thanks for fixing the example, but now I see an function in the state, which doesn't make sense either.

Alber70g:
My bad, using 'view' could be confusing, I admit.

JorgeBucaran:
Haha `foo` is okay. Why is there a function though?

Alber70g:
Because this is the example where the new typings add more security.

JorgeBucaran:
@Alber70g Oh, okay, so one of the purposes of this PR is to prescribe what types are valid in the State. That sounds good.

What about the other stuff? What's the difference between `Actions`, `ActionsType` and `WiredActions`?

Alber70g:
@JorgeBucaran would you like me to change the WiredActions to Actions and refactor the wrong uses? Since I wasn't aware of the ActionType and Actions since they're used ambiguous

JorgeBucaran:
Yes, please.

Alber70g:
@JorgeBucaran Although I still feel we should use a more defined term for the Actions that are returned by `app()`.

JorgeBucaran:
@Alber70g Should we update hte TS example as well? 🤔😉 

> Although I still feel we should use a more defined term for the Actions that are returned by app().

I get that, but AFAIK the user almost never needs to touch those types.

hyperapp.d.ts
```diff
@@ -85,6 +85,28 @@ export interface View<State, Actions> {
   (state: State, actions: Actions): VNode<object>
 }
 
+/** The action wired to the app which are in the object 
+ * that's being returned from the App()
```
JorgeBucaran: `App()`?

Alber70g: Fixed :)

hyperapp.d.ts
```diff
@@ -94,12 +116,15 @@ export interface View<State, Actions> {
  * @returns The actions wired to the application.
  * @memberOf [App]
  */
-export function app<State, Actions>(
+export function app<
+  State extends StateType,
+  AppActions extends ActionsType<State, AppActions>
```
JorgeBucaran: @Alber70g Why `AppActions`?

hyperapp.d.ts
```diff
@@ -94,12 +116,15 @@ export interface View<State, Actions> {
...
+>(
   state: State,
-  actions: ActionsType<State, Actions>,
-  view: View<State, Actions>,
+  actions: ActionsType<State, AppActions>,
```
JorgeBucaran: Also `AppActions` here.

hyperapp.d.ts
Alber70g: Because otherwise it's ambiguous and can Typescript not detect which one to use. This is the generic where 'Actions' is also a type itself

hyperapp.d.ts
Alber70g: Same reason

JorgeBucaran: I don't understand.

JorgeBucaran:
I like how `StateType` appears to be symmetric with `ActionsType`, but I am confused about `AppActions`. Maybe it would be best to remove that.

hyperapp.d.ts
Alber70g: You need to understand how Generics work. A generic can be named with a single letter, which is usually done, but also as a name. In this case we could even prefix generics with a G so it isn't ambiguous. The Typescript compiler needs to know the difference between a `Type` and a `Generic`. Read up more here: https://www.typescriptlang.org/docs/handbook/generics.html

Alber70g:
> I like how StateType appears to be symmetric with ActionsType, but I am confused about AppActions. Maybe it would be best to remove that.

See https://github.com/hyperapp/hyperapp/pull/535#discussion_r159433724

hyperapp.d.ts
JorgeBucaran: I will try. BTW you gave me a good example with StateType, we made progress there, but there was none for actions, hence we are still stuck.

Alber70g: Have a look here, we want the generic type for `Actions` in `app()` (ie `app<MyStateType, MyActionsType>`) to be constraint to having only functions of the following signature:
- `(data?) => state | ((state?, actions?) => state)`. Have a look here at 'Generic Constraints' https://www.typescriptlang.org/docs/handbook/generics.html#generic-constraints

Alber70g:
@JorgeBucaran Do you need explanation about why I call it `AppActions` or why the generic constraint is there: `AppActions extends ActionsType<State, AppActions>`?

JorgeBucaran:
@Alber70g In order to give you a proper review, I need an example that demonstrates the shortcomings of the current typings. The example should try to explain why using the new typings makes TS apps more type-safe and ideally cover as much as possible, `StateType`, `AppActions`, `Actions`, etc. We can work on more reader-friendly names once we have that example. 🙏

lassecapel:
can you add an interface for app too? so you can extend from that function if you create a HOA

Alber70g:
I've changed the typings considerably and I need to polish them. In the mean time it would be nice to review them for what there is now. 

A list of the changes:

- all generics are prefixed with T to make the difference between types/interfaces and generics clear
- I've added a generic NestedMap<T> as helper type to form the other types (specifically State and Actions, and their derivatives)
- I changed the exported 'app()' signature. Right now it's an implementation of an interface that's being exported (this allows @lassecapel to type the app that's passed into his HOA)
- I've added WiredAction and UnwiredAction and their object wrappers as aliases so it's more convenient for implementing objects for respectively State/SubState and Actions/SubActions.

JorgeBucaran:
@Alber70g Sorry for the confusion, could you please create a new PR with the typings! 🙏

Issue #533, opened 2018-01-03
Dynamic actions: How to add new actions at runtime?


jorgebucaran:
## Summary

There is currently no way to add actions dynamically. 

* Why do I want the ability to add actions at runtime? (Code splitting, etc.)
 * Why do I want code splitting?
   * But why do you even need code splitting? Why not instead consider making a simpler app? If you've reached this extreme situation, why not creating _multiple_ apps? Those should definitely be "code-splittable" without intervention from the core.
   *  Why not be aware that an SPA is **no different** from an actual app you'd download like from an "app store". What about implementing a one-time install screen?
* What about route-based code splitting? I think this is what Elm plans to do for 0.19.
* What would be the simplest, non-breaking way to add this functionality? How would it look like on a simple app?

## Related
https://github.com/hyperapp/hyperapp/issues/439

timjacobi:
Since actions operate on state we probably also need to think about how to dynamically add state.
Could we not just return a reference of the state together with the action from `app` and reinitialize the app when new actions and state are loaded? We could also mark returned actions as already initialised and then skip them during reinitialization.

```js
const initState = { ... }
const initActions = { ... }

let { state, actions } = app(initState, initActions, ...)
// each action has the property action.initialized = true

// later load new actions and state
load('./mymodule.js').then(({newState, newActions}) => {
  { state, actions } = app({...state, ...newState}, {...actions, ...newActions})
});
```

JorgeBucaran:
@timjacobi Running `app` twice?

zaceno:
@JorgeBucaran yeah multiple app calls - or at least multiple init calls makes sense to me. It's along the lines of "multiple stores" I was thinking about before. If you're initing new actions, it's always going to be for new initial state. So there won't be any crossover possible wether or not you splice the new state on to the original state tree or just keep it in a different one.

The only problem, of course, is how to do full state persistence/time travel.

But I think the dynamically initialized actions use cases rule out persistent/time traveling state anyway.

selfup:
I think this is a very important question not just for code splitting but for importing reusable "components".

Going to grab my laptop and edit this comment brb

zaceno:
@selfup have a look at hyperapp-nestable if you haven't already. Just at the source I mean. It's basically the "magic" you're talking about, which works by calling app inside lifecycle methods.

Now wether that's good enough I don't know, but it's fully possible to make shareable components today using that method.

[...]

PR #542, opened 2018-01-06
Dynamic Actions: expose updateApp function _per_ hyperapp

selfup:
Ok so as per #533 I have made some changes.

This is **purely** a POC and we can try to figure out what needs to be done.

Here is the code that users would need to do on their end (slight breaking change unfortunately):

**After the latest commit I was able to not break the current API but leaving the example here so we can decide which direction we go in**

```js
const {
  actions: exposed,
  updateApp,
} = app(
  state,
  actions,
  view,
  document.body,
);

// ...

setTimeout(() => {
  const newState = {
    hello: 'hello world',
    num: 42,
  };

  const newActions = {
    sayHello: () => s => ({ num: s.num * 2 }),
  };

  Object.assign(
    exposed,
    updateApp(
      newState,
      newActions,
    ),
  );
}, 3000);
```

As you can see the exposed actions are nested one more level deep. But we now get to expose a very interesting feature!

This comment is what got me to this conclusion: https://github.com/hyperapp/hyperapp/issues/533#issuecomment-355596451

Which was heavily inspired by this: https://github.com/hyperapp/hyperapp/issues/533#issuecomment-355596451

Thank you @timjacobi for talking with us! Let me know if you have any other ideas?

We will need to add tests and whatnot later, but this is just so some people can play around locally when they check out this branch and link hyperapp with a file path in their package.json!
  
### Workaround to not break the current API:

We can expose the `updateApp` in the exposed actions! Kind of like:

```
return copy(actions, { updateApp: updateApp })
```

_Let me try that real quick_


selfup:
Ok I got it to work without breaking the current API 🎉

selfup:
I think having it built in provides a clean way for others to use something. An official API that requires very little code. Also in the discussion of the issue it seems that exposing it from Hyperapp is the direction we were going

okwolf:
@zaceno as @selfup said I've already shown this is absolutely doable in an HOA: https://github.com/hyperapp/hyperapp/issues/533#issuecomment-355226795

JorgeBucaran:
@selfup Whoops! It seems my push --force to master blew up this PR devastatingly! 🙇🍅 

Please create a new PR so we can continue discussing an actual implementation of dynamic actions. 🙏

PR #543, opened 2018-01-06
convert var to let

emersonlaurentino:


okwolf:
As much as I agree with this change, we would need to agree on dropping IE 10 support.

JorgeBucaran:
Thanks @emersonlaurentino and welcome aboard, but I prefer to write and maintain Hyperapp in ES3/5! See our contribution guidelines [style section here](https://github.com/hyperapp/hyperapp/blob/master/docs/CONTRIBUTING.md#style). 

I like to think that if I went back in time to 2005, I could most likely use Hyperapp as it is today.

PR #548, opened 2018-01-07
expose updateApp function per hyperapp

selfup:
🎉 #533 #542 

#### Pros

1. Simple
  
#### Cons

1. Memory Leaks (for now until we solve/if we decide to solve)
2. `oncreate` will be fired (each one)
3. `ondestory` needs to be fired (each one)
4. other unknowns

### Please consider using a HOA (Higher Order App)

https://github.com/hyperapp/hyperapp/issues/533#issuecomment-355765388

Brief example of a HOA being used:

```js
const main = injectUpdate(app)(state, actions, view, document.body);

main.updateApp({
  newActions: {
    mult: () => ({ count }) => ({ count: count * 2 }),
  },
  newState: {
    count: 90,
  },
});

// OR

const main = injectUpdate(app)(state, actions, view, document.body);

import('./thing.js')
  .then(({ state: newState, actions: newActions }) => {
    main.updateApp({ newState, newActions });
  });
```

If we decide to delegate to a HOA, we should add a file in the docs `dynamic_loading.md` and explain the reason and point to the issue/PR etc..!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -34,7 +34,7 @@ export function app(state, actions, view, container) {
 
   repaint(init([], (state = copy(state)), (actions = copy(actions))))
 
-  return actions
+  return copy(actions, { updateApp: updateApp })
 
   function vnode(element, map) {
     return (
@@ -66,6 +66,12 @@ export function app(state, actions, view, container) {
       patchLock = !patchLock
       setTimeout(render)
     }
+  }
+
+  function updateApp(newState, newActions) {
+    actions = copy(actions, newActions)
+    repaint(init([], (state = copy(state, newState)), actions))
+    return actions
   }
 
   function copy(a, b) {


JorgeBucaran:
@selfup I like:

> other unknowns

😆

selfup:
Heh 😆 You never know!

selfup:
https://github.com/hyperapp/hyperapp/issues/533#issuecomment-355830938

> If we pick the HOA we have two options:
> 
> Extract to userland (under @hyperapp/update)
> Expose from hyperapp const { h, app, updateApp } = hyperapp but that would add a lot of bytes

JorgeBucaran:
I'll experiment with adding core support to calling app multiple times, but as @okwolf commented [here](https://github.com/hyperapp/hyperapp/issues/533#issuecomment-355856008), `updateApp` is a pattern that can be achieved using an HOA. It not even needs to be an "official" package.

selfup:
Yep agreed!

PR #551, opened 2018-01-08
Refactor typings to allow for better intellisense and support in Typescript

Alber70g:
As a continuation of this PR (https://github.com/hyperapp/hyperapp/pull/535 that might be gone...) I've created this PR.

**!! Please have a look at this, anyone who has interest in Typings !!**

I've changed the typings considerably and I need to polish them. In the mean time it would be nice to review them for what there is now. 

A list of the changes:

- all generics are prefixed with T to make the difference between types/interfaces and generics clear
- I've added a generic NestedMap<T> as helper type to form the other types (specifically State and Actions, and their derivatives)
- I changed the exported 'app()' signature. Right now it's an implementation of an interface that's being exported (this allows @lassecapel to type the app that's passed into his HOA)
- I've added WiredAction and UnwiredAction and their object wrappers as aliases so it's more convenient for implementing objects for respectively State/SubState and Actions/SubActions
- I changed the signature of the generics so that the TPayload is always last and optional

Things to do:

- [x] fix the build (test/index.tsx in particular)
- [x] fix formatting according to Hyperapp standards
- [x] allow for nested types and actions to be in the same way accessible as the return value of `App` interface
- [x] change and move documentation in the proper place
- [ ] allow for nested actions to get the nested state based on the path where the action is mounted
- [ ] _In progress_: implement recommendations from review (please take a critical look at the current types and the uses of `any` as defaults for generics)
- [ ] [discussion] there are now the concepts of `UnwiredAction`, `WiredAction`. I'm wondering if we should keep it this way. I feel that it gives good distinction between the two, but others think otherwise. 
- [x] Not applicable anymore ~~[discussion] a lot of generics are now typed optional with `any`, it might be better to have them non-optional~~

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -63,7 +63,7 @@ export type NestedMap<T> = {
 /**
  * Type to describe anything that's not a function or object
  */
-export type Primitive = string | number | boolean | null | any[]
+export type Primitive = string | number | boolean | null | undefined | any[]
 
 /** The result of an action.
  *
@@ -83,46 +83,7 @@ export interface View<
   TState extends Primitive | NestedMap<Primitive>,
   TActions extends NestedMap<WiredAction<any, TState>>
 > {
-  (state: TState, actions: TActions): VNode<object>
-}
-
-/**
- * The action that's being passed to actions and returned from app()
- */
-type WiredAction<
-  TPayload extends Primitive | NestedMap<Primitive>,
-  TState extends Primitive | NestedMap<Primitive>
-> = (data?: TPayload) => TState
-
-/**
- * The action that's being passed to the actions object in app. 
- * Use this for implementing actions.
- */
-export type ActionType<
-  TState extends Primitive | NestedMap<Primitive> = any,
-  TActions extends
-    | NestedMap<WiredActions<TState, any>>
-    | NestedMap<UnwiredActions<TState, any>> = any,
-  TPayload extends Primitive | NestedMap<Primitive> = any
-> = (
-  data?: TPayload
-) =>
-  | ((state: TState, actions: TActions) => ActionResult<TState>)
-  | ActionResult<TState>
-
-/**
- * The interface for app(). Use this for implementing Higher Order App's
- */
-export interface App {
-  <
-    TState extends Primitive | NestedMap<Primitive>,
-    TActions extends UnwiredActions<TState, TActions>
-  >(
-    state: TState,
-    actions: TActions,
-    view: View<TState, TActions>,
-    container: HTMLElement | null
-  ): { [P in keyof TActions]: TActions[P] }
+  (state: TState, actions: TActions): VNode<any>
 }
 
 /** The app() call creates and renders a new application.
@@ -135,27 +96,71 @@ export interface App {
  */
 export declare const app: App
 
-/** 
- * Convenience type for State
+/**
+ * The interface for app(). Use this for implementing Higher Order App's
  */
-export interface StateType extends NestedMap<Primitive> {}
+export interface App {
+  <TState extends State, TActions extends UnwiredActions<TState, TActions>>(
+    state: TState,
+    actions: TActions,
+    view: View<TState, { [P in keyof TActions]: WiredAction<TState, any> }>,
+    container: HTMLElement | null
+  ): { [P in keyof TActions]: WiredAction<TState, any> }
+}
 
 /**
- * Convenience alias to extend your actions object from, when implementing actions.
+ * The action that's being passed to actions and returned from app()
  */
+type WiredAction<
+  TState extends Primitive | NestedMap<Primitive>,
+  TPayload extends Primitive | NestedMap<Primitive> = any
+> = (data?: TPayload) => TState
+
+/**
+ * The action that's being passed to app(). Use when you implement your actions.
+ */
+export type UnwiredAction<
+  TState extends State,
+  TActions extends UnwiredActions<TState, TActions>,
+  TPayload extends Primitive | NestedMap<Primitive> = any
+> = (
+  data?: TPayload
+) =>
+  | ((
+      state: TState,
+      actions: { [P in keyof TActions]: WiredAction<State, any> }
+    ) => ActionResult<TState>)
+  | ActionResult<TState>
+
+/** 
+ * Convenience type for defining initial State
+ */
+export interface State extends NestedMap<Primitive> {}
+
+/**
+ * Convenience alias from which you extend your actions object.
+ */
+// interface UnwiredActions<
+//   TState extends State,
+//   TActions extends { [P in keyof TActions]: WiredAction<TState> } = any
+// > extends NestedMap<TActions> {}
 interface UnwiredActions<
-  TState extends Primitive | NestedMap<Primitive> = any,
-  TActions extends NestedMap<
-    WiredActions<TState, TActions> | UnwiredActions<TState>
-  > = any
-> extends NestedMap<any> {}
+  TState extends State,
+  TActions extends UnwiredActions<
+    TState,
+    { [P in keyof TActions]: WiredAction<TState> }
+  >
+>
+  extends NestedMap<
+      UnwiredAction<TState, { [P in keyof TActions]: WiredAction<TState> }>
+    > {}
 
 /**
  * Convenience alias for NestedMap<WiredAction>
  */
 interface WiredActions<
-  TState extends Primitive | NestedMap<Primitive>,
-  TActions extends UnwiredActions
+  TState extends State,
+  TActions extends UnwiredActions<TState, TActions>
 > extends NestedMap<WiredAction<any, TState>> {}
 
 /** @namespace [JSX] */
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -2,7 +2,8 @@ export as namespace Hyperapp
 
 /** @namespace [VDOM] */
 
-/** The VDOM representation of an Element.
+/** 
+ * The VDOM representation of an Element.
  *
  * @memberOf [VDOM]
  */
@@ -12,13 +13,15 @@ export interface VNode<Props> {
   children: VNodeChild<object | null>[]
 }
 
-/** In the VDOM a Child can be either a VNode or a string.
+/** 
+ * In the VDOM a Child can be either a VNode or a string.
  *
  * @memberOf [VDOM]
  */
 export type VNodeChild<Props = object> = VNode<Props> | string
 
-/** A Component is a function that returns a custom VNode.
+/** 
+ * A Component is a function that returns a custom VNode.
  *
  * @memberOf [VDOM]
  */
@@ -26,7 +29,8 @@ export interface Component<Props> {
   (props: Props, children: VNodeChild<object | null>[]): VNode<object>
 }
 
-/** The type of the children argument passed to h().
+/** 
+ * The type of the children argument passed to h().
  *
  * @memberOf [VDOM]
  */
@@ -35,7 +39,8 @@ export type VNodeChildren =
   | VNodeChild<object | null>
   | number
 
-/** The soft way to create a VNode.
+/** 
+ * The soft way to create a VNode.
  * @param name      An element name or a Component function
  * @param props     Any valid HTML atributes, events, styles, and meta data
  * @param children  The children of the VNode
@@ -51,21 +56,30 @@ export function h<Props>(
 
 /** @namespace [App] */
 
-/**
+/** 
  * Type to describe a nested map of type T
  * 
  * example: { 'foo' : { 'bar': 1 }, 'baz': '' }
  */
-export type NestedMap<T> = {
+type NestedMap<T> = {
   [key: string]: NestedMap<T> | T
 }
 
-/**
+/** 
+ * Type alias to proxy UnwiredActions to WiredActions.
+ */
+type ActionProxy<
+  TState extends StateType,
+  TActions extends UnwiredActions<TState, TActions>
+> = { [P in keyof TActions]: WiredAction<TState, any> }
+
+/** 
  * Type to describe anything that's not a function or object
  */
 export type Primitive = string | number | boolean | null | undefined | any[]
 
-/** The result of an action.
+/** 
+ * The result of an action.
  *
  * @memberOf [App]
  */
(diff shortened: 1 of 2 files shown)


test/ts/index.tsx
```diff
@@ -1,38 +1,49 @@
...
     <div>{state.count}</div>
     <button onclick={actions.down}>-</button>
-    <button onclick={actions.up}>+</button>
+    <button onclick={() => actions.up()}>+</button>
+    <button onclick={() => actions.up(5)}>+5</button>
   </main>
 )
 
-app<Counter.State, Counter.Actions>(
+const appActions = app<Counter.State, Counter.Actions>(
```
JorgeBucaran: `main`

test/ts/index.tsx
```diff
@@ -1,38 +1,49 @@
...
   Counter.state,
   Counter.actions,
   view,
   document.body
 )
+
+appActions.up()
```
JorgeBucaran: `main.up()`

test/ts/index.tsx
```diff
@@ -1,38 +1,49 @@
-import { h, app, ActionsType, View } from "hyperapp"
+import { h, app, View, StateType, UnwiredActions, UnwiredAction } from 'hyperapp'
```
JorgeBucaran: `"`

test/ts/index.tsx
```diff
@@ -1,38 +1,49 @@
...
 
 namespace Counter {
-  export interface State {
+  export interface State extends StateType {
     count: number
+    clickCount: number
```
JorgeBucaran: If you are going to use `clickCount`, then maybe change the counter example as well, because `count` and `clickCount` is confusing.

test/ts/index.tsx
```diff
@@ -1,38 +1,49 @@
...
   }
 
-  export interface Actions {
-    down(): State
-    up(value: number): State
+  export interface Actions extends UnwiredActions<State, Actions> {
+    down: UnwiredAction<State, Actions>
```
JorgeBucaran: UnwiredActions is rather long. What's wrong with ActionsType or ActionsTemplate?

test/ts/index.tsx
Alber70g: What should I change it to?

hyperapp.d.ts
```diff
@@ -51,58 +56,126 @@ export function h<Props>(
...
+/** 
+ * The action that's being passed as second param in an object to app(). 
+ * Use this when implementing your actions.
+ */
+export type UnwiredAction<
+  TState extends StateType,
+  TActions extends UnwiredActions<TState, TActions>,
+  TPayload extends Primitive | NestedMap<Primitive> = any
+> = (
+  data?: TPayload
```
JorgeBucaran: What is TPayload?

test/ts/index.tsx
Alber70g: Nothing, just wanted to put something that's not even remotely confusing

hyperapp.d.ts
Alber70g: The generic to type the payload

JorgeBucaran: What is the payload?

test/ts/index.tsx
Alber70g: I'm for changing UnwiredAction to ActionsTemplate. But what about WiredActions? It's an internal type that no-one should be using anyway.

test/ts/index.tsx
JorgeBucaran: I don't know, maybe come up with a different example. @Mytrill had a todo list originally, but I thought it was too long, so I replaced it for the current counter.

hyperapp.d.ts
Alber70g: The payload passed to an action when calling from the view or inside an action.

JorgeBucaran:
Do you have any links that I can start reading to help me understand this PR? I don't expect to understand everything about TS in one sit, but at least something that can help figure out what's going on here.

hyperapp.d.ts
Alber70g: Redux also calls it payload and it's called payload in the Flux architecture. I believe it's wise to keep it the same as the rest of the community calls it.

Alber70g:
Start reading [advanced types](https://www.typescriptlang.org/docs/handbook/advanced-types.html)

test/ts/index.tsx
Alber70g: Fixed

test/ts/index.tsx
Alber70g: I put it there so we can call this action in the `up` and `down` action to demonstrate that the types are working properly.

JorgeBucaran:
@Alber70g Any way we can simplify this?

Alber70g:
@JorgeBucaran Why and what would need to be simplified? The reason it's complex is because we want to pass an type that defines the actions as UnwiredActions but want to use them as WiredActions.

```typescript
const actions = {
  up: () => (state, actions) => ({})
}
```
I'm talking about `(state, actions) => ({})`. Here the actions we pass to this function aren't the original actions.

JorgeBucaran:
@Alber70g What would makes the types simpler? It would have been nice we had this convo before the 1.0.

hyperapp.d.ts
JorgeBucaran: @Alber70g Ah, the data? 

```js
actions.foo(payload)
```
?

test/ts/index.tsx
JorgeBucaran: I see. Don't change it yet please.

test/ts/index.tsx
JorgeBucaran: If it's not a new example, then maybe just refactor the names.

Alber70g:
@JorgeBucaran I tried to change stuff before the 1.0 release but there was no attention for the typings at all. No one except me uses Hyperapp with TypeScript right now. 

Ping @lassecapel: have a look at the typings, see if they're good for your redux devtools implementation

hyperapp.d.ts
lassecapel: Yes the Data, but I think payload makes sense since it is used as a name in Redux too.

Alber70g: Yes that's correct.

hyperapp.d.ts
```diff
@@ -51,58 +56,126 @@ export function h<Props>(
...
+ */
+interface UnwiredActions<
+  TState extends StateType,
+  TActions extends UnwiredActions<TState, ActionProxy<TState, TActions>>
+> extends NestedMap<UnwiredAction<TState, ActionProxy<TState, TActions>>> {}
+
+/**
+ * Convenience type for NestedMap<WiredAction>
+ */
+interface WiredActions<
```
lassecapel: Since you implement this. It would be nice to use `Actions` here. You can then use `ActionsTemplate` then for the `UnwiredActions`

test/ts/index.tsx
```diff
@@ -1,38 +1,49 @@
...
+    },
+    up: (value = 1) => (state, actions) => {
+      actions.addClickCount()
+      return { count: state.count + value }
+    },
+    addClickCount: () => state => ({ clickCount: state.clickCount + 1 }),
   }
 }
 
 const view: View<Counter.State, Counter.Actions> = (state, actions) => (
```
Mytrill: The `View<...>`type accepts `WiredActions<...>`, but here you pass it `Actions` which are of type `UnwiredActions<...>`, so I have 2 questions:
 - does this actually compiles?
 - inside the view, the user will call an unwiredActions as if it was a wired action? Shouldn't the type of View be:
```typescript
export interface View<
  TState extends StateType,
  TActions extends WiredActions<TState, TActions>
> {
  // Use the proxy here
  (state: TState, actions: ActionProxy<TState, TActions>): VNode<any>
}
```

Other than this, neat changes! With the new API it makes sense now for users to type the unwired actions and get the wired action type auto generated!

hyperapp.d.ts
lassecapel: Another proposal: `Actions` and `AppActions`. Where `Actions` are the unwired ones and `AppActions` the wired ones

hyperapp.d.ts
```diff
@@ -51,58 +56,126 @@ export function h<Props>(
...
-  view: View<State, Actions>,
-  container: Element | null
-): Actions
+export declare const app: App
 
-/** @namespace [JSX] */
+/**
+ * The interface for app(). Use this for implementing Higher Order App's
+ */
+export interface App {
```
lassecapel: This would work for the redux-devtools HOA

hyperapp.d.ts
```diff
@@ -51,58 +56,126 @@ export function h<Props>(
...
- * @memberOf [App]
+/** 
+ * Type alias to proxy UnwiredActions to WiredActions.
  */
-export type ActionType<State, Actions> = (
-  data?: any
-) =>
-  | ((state: State, actions: Actions) => ActionResult<State>)
-  | ActionResult<State>
+type ActionProxy<
```
lassecapel: Nice way to define the wiring. This was causing me some trouble understanding this out of the documentation.

test/ts/index.tsx
Alber70g: You're right, it's a bit strange. But it does compile somehow... I'll investigate more :) Thanks for your review 👍

hyperapp.d.ts
JorgeBucaran: I agree with this! Thanks for explaining.

hyperapp.d.ts
JorgeBucaran: It would be nice to use `Actions` for the most common one.

test/ts/index.tsx
Mytrill: IMHO, the test should have an example of asynchronous action (i.e. returning a promise) and an action where the return type is actually used by the view, since these could potentially be shortcomings of these types, which were not there with the previous version.

test/ts/index.tsx
Alber70g: You're right. I've changed the signature of the View. But in your snippet the `TActions extends WiredActions<..., ...>` should be `TActions extends UnwiredActions<..., ...>` because you can only proxy UnwiredActions -> WiredActions. To see that it's correct now, you can look at the index.tsx and see what signature the view's `onclick={() => actions.up(5)}`: `(property) up: (data?: any) => Counter.State`. This is correct.

The reason why it compiles is because the UnwiredActions' TActions generic is ambiguous because it's being proxied to WiredActions :')

Alber70g:
@JorgeBucaran (& @Mytrill) I've found something that can be simplified in a way that you don't need to type the object containing the UnwiredActions, like so:

```ts
type UnwiredActions<
  TState extends StateType,
  TActions extends UnwiredActions<TState, TActions>
> = { [P in keyof TActions]: NestedMap<UnwiredAction<TState, TActions>>
```

test/ts/index.tsx
Mytrill: Ah, nice!

Alber70g:
It's still not perfect. You cannot add nested actions in the actions object. This disallows the use of slices. Also we need to figure out a way to declare nested types and infer their type. Not sure if that's possible right now, but we can provide a 'helper' that allows for a certain level of nesting: https://github.com/Microsoft/TypeScript/issues/10693

hyperapp.d.ts
Alber70g: What is 'most common' where you implement actions, or where you use them (e.g. in actions)?

Alber70g:
I'm a bit stuck right now. The problem I face is that whenever you create a nested state and nested action, the type of the nested UnwiredAction has a state and actions passed living on the same path as the UnwiredAction. I don't know how to define that state... somehow, in the types, there needs to be a connection between state and actions. This isn't possible with two generics. 

If anyone would like to show me some suggestions, I might be able to integrate it.

Mytrill:
I used something like this before, and it seemed to work:
```typescript
type ActionProxy<TState extends Partial<Record<keyof TActions, any>>, TActions> = {
  [P in keyof TActions]: WiredAction<TState[P], TActions[P]>
}
```
The `Partial<Record<keyof TActions, any>` makes sure that `TState` and `TActions` have some keys in common, and allows to write `TState[P]`.

Hope that helps :)

Alber70g:
Sorry for the radio silence the past few weeks. I've been working on fixing this problem on and off, and in the mean time I've been ill.

This issue of the nested types can be fixed for good once conditional types (https://github.com/Microsoft/TypeScript/pull/21316) are landing in TypeScript.

This solution (https://github.com/Microsoft/TypeScript/issues/12424#issuecomment-356685955) can be used in the mean time. I haven't started experimenting with it yet, but I'll try that this weekend. 

The final way to solve this issue is to not type it (actions) at all, and just use any. This means however, that we cannot have inferred typings for the view and the actions parameter in the actions, unless you pass them explicitly as a generic.

JorgeBucaran:
@Alber70g Any plans to continue this PR. 🤔

Alber70g:
Right now it can be used as is, but it'll not be flawless with slices. Until conditional types land in Typescript 2.8 which should be released soon.

Alber70g:
I've done some digging with typescript@2.8 alpha, and its not working in a way I want it to be.

There's no way to infer types and use the children of a *different* nested type (i.e. how slices work in HA) in another type. 
In other words: you cannot make the *nested* type of the `Actions` refer to the *nested* type of `State` in order to type the `State` that's being passed to a particular `Action` living on the same 'slice'. 

Example:
```ts
app({ counter: 0, nested: { counter: 0} }, 
{
  inc: () => (state) => ({ counter: state.counter + 1 }),
  nested: {
    inc: () => (nestedState) => ({ counter: nestedState.counter + 1 }) 
    // currently the typings do not support Typescript in figuring out 
    // that this "nestedState" is the "state.nested".
  }
}
```

I will finalise this pull request in the coming week.

jorgebucaran:
> @Alber70g ...currently the typings do not support Typescript in figuring out that this "nestedState" is the "state.nested".

Is this also true in TS 2.8?

Alber70g:
> typescript@2.8 alpha

jorgebucaran:
😢

jorgebucaran:
With 2.0 we won't need most of this work, so I am going to close here. We'll need types though! But I imagine you'll prefer to create a new PR for that then? 😅 

Closing.

Alber70g:
That's fine 👌 Once the API of 2.0 is clear I can start working on that one

PR #554, opened 2018-01-10
Add unpkg entry for bare URLs

mjackson:
This allows people to use src/index.js using a "bare" unpkg URL like
https://unpkg.com/hyperapp instead of requiring people to use the
?module flag, which is a little overkill for hyperapp since everything
is in a single file anyway.

It does, however, require that src/index.js always be runnable in a
browser environment, which may be something that you don't want. In that
case, it'd be better to point unpkg at dist/hyperapp.js to make sure people
get your built files instead. Up to you!

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "dist/hyperapp.js",
   "jsnext:main": "src/index.js",
   "module": "src/index.js",
+  "unpkg": "dist/hyperapp.js",
   "typings": "hyperapp.d.ts",
   "license": "MIT",
   "repository": "hyperapp/hyperapp",


package.json
```diff
@@ -5,6 +5,7 @@
   "main": "dist/hyperapp.js",
   "jsnext:main": "src/index.js",
   "module": "src/index.js",
+  "unpkg": "src/index.js",
```
Swizz: Using Hyperapp in the browser along to unpkg I would prefer to get the for-browser umd builded file `dist/hyperapp.js`.
The current behavior.

mjackson: I think I agree. But @JorgeBucaran recommended the `module` file [on Twitter the other day](https://twitter.com/JorgeBucaran/status/951009930609278976) so I was just following suit. It all depends on which file you'd expect to load when you use `https://unpkg.com/hyperapp`.

JorgeBucaran:
@mjackson Would this also make `https://unpkg.com/hyperapp@1.0.1` output `src/index.js`?

mjackson:
@JorgeBucaran Yes. `https://unpkg.com/hyperapp` will always redirect to the latest version. If you want to serve the minified bundle at that URL, use `{ "unpkg": "dist/hyperapp.js" }` instead.

It's a trade-off, really. It all depends on whether or not you're comfortable serving your `src` file to browsers, or whether you are doing stuff in `src` that browsers may not support (some ES6 stuff). In that case, the code in `dist` may be safer.

selfup:
Yea currently our non mangled code has an export statement in it, so unclear how that’ll do 😂

mjackson:
Ah, I missed that `export` statement. Thanks for the heads up, @selfup. I updated the PR to point to `dist/hyperapp.js`. 🍻

JorgeBucaran:
Yep, that's why serving the raw index.js works as an ES module.

@mjackson Can you please clarify, if `https://unpkg.com/hyperapp` loads `https://unpkg.com/hyperapp@1.0.1/dist/hyperapp.js`, then what difference does it make adding `unpkg` to pkgjson? 🤔

selfup:
@JorgeBucaran so for Chrome only? Or what other browsers support straight up modules?

mjackson:
Oh, duh. Ya, you're right @JorgeBucaran. I wasn't thinking straight. If your `main` file is a pre-built file then this won't make any difference. Closing.

mjackson:
(I should note however that people *usually* put CommonJS in their `main` config so that modules can be used inside node, but if you're putting a UMD build there then that should work too)

JorgeBucaran:
@mjackson You are right, but Node doesn't have ES modules enabled by default (yet), and even if it did, we would have to provide a `.mjs` file, but we are only distributing a `.js` file at the moment. 

@selfup I just checked and it works in Safari, Chrome and FF (behind the `dom.moduleScripts.enabled` setting).

PR #562, opened 2018-01-20
Size badge

okwolf:
I noticed [Preact has a gzip size badge on their repo readme](https://github.com/developit/preact#preact) and was wondering if we should have something similar since we frequently mention our small footprint.


JorgeBucaran:
Actually, I was thinking of removing the coverage badge!

So, no thanks! 😆

Issue #566, opened 2018-01-22
Animations, Transitions & Experiential Sites


mattdesl:
Hey there! I'm really enjoying Hyperapp so far. I'm trying to use it for building a rich/experiential/animated site (imagine lots of animations/transitions, WebGL/Canvas, etc) and it seems to have a couple of unique issues.

1. When using `onremove` with a delayed callback (e.g. fade out), if the user quickly adds/removes the component from the tree they will end up with multiple of the same component. I am hoping for something closer to "React Transition Group" so that my view looks like this. But in this case I don't want multiple error messages to stack up.

```js
<div>
  <TextField />
  <TransitionGroup>
    {state.isError ? <AnimatedErrorMessage /> : undefined}
  </TransitionGroup>
</div>
```

2. Right now, using `onremove` sometimes leads to an error:

```Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.```

Full code to reproduce. Try clicking +/- really quickly.

```js
import { h, app } from 'hyperapp';

const noop = () => {};

const state = {
  count: 0
};

const actions = {
  down: () => state => ({ count: state.count - 1 }),
  up: () => state => ({ count: state.count + 1 })
};

const fadeIn = (el) => {
  console.log('fade in', el);
};

const fadeOut = (el, done = noop) => {
  console.log('fade out', el);
  setTimeout(() => done(), 1000);
};

const animations = {
  oncreate: fadeIn,
  onremove: fadeOut
};

const Message = (state, children) => {
  return <div {...state} {...animations}>Above one!</div>;
};

const view = (state, actions) => {
  return <main>
    <h1>{state.count}</h1>
    <button onclick={actions.down}>-</button>
    <button onclick={actions.up}>+</button>
    {state.count >= 1 && <Message key='message' />}
  </main>;
};

app(state, actions, view, document.body);
```

3. This is more a question... I often end up needing local state in a component. I have seen a few issues mentioning this but they all point to different things: components, modules, mixins, `embed` utility functions that call `app()`, etc. I'm wondering what the official/suggested solution is for this?

For example: storing Canvas2D context for a component, or storing the millisecond time that the component was _first_ created (but not replacing it on subsequent renders). I do not want to move all of this information into my global app state/actions.

zaceno:
1) What you're asking for sounds close to https://github.com/zaceno/hyperapp-transitions

2) Try:

```
const Message = (state, children) => {
[...]

Issue #530, opened 2018-01-02
Fragment node type to support <></> JSX syntax?


frenzzy:
```js
const view = <>
  <h1>{state.count}</h1>
  <button onclick={actions.down}>-</button>
  <button onclick={actions.up}>+</button>
</>
```
to avoid writing commas in jsx:
```js
const view = [
  <h1>{state.count}</h1>,
  <button onclick={actions.down}>-</button>,
  <button onclick={actions.up}>+</button>,
];
```

refs:
- [babel-plugin-transform-react-jsx#pragmaFrag](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-react-jsx#pragmafrag)
- [Fragments in React.js](https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html)

by the way, returning an array of elements from root view leads to `<undefined></undefined>` in DOM but nested views are rendered as expected

lukejacksonn:
AFAIK.. hyperapp does not support an array of elements as the root view, this is by design and don't think there is any plan to change this behaviour. As you point out though, it does support fragment like subviews by simply returning an array of elements. For this reason, the need for a fragment syntax was redundant.

Out of curiosity.. what does this transpile to?
```
<>
  <h1>{state.count}</h1>
  <button onclick={actions.down}>-</button>
  <button onclick={actions.up}>+</button>
</>
```

JorgeBucaran:
@lukejacksonn 

> In React, this desugars to a <React.Fragment/> element, as in the example from the previous section. (Non-React frameworks that use JSX may compile to something different.)

So, maybe it compiles to `h(Fragment, {}, [the, elemen, ts])` ? 🤔 

---

I find mapping over a list/array like so:

```jsx
const ImageGallery = ({ images }) =>
  images.map(({ hash, url, description }) => (
    <li key={hash}>
      <img src={url} alt={description} />
    </li>
  ))
```

...more common than returning an array like in your example, but I am curious to know much code this would add.

lukejacksonn:
I tend to return an array of elements that might include a nested `map` to fill my `main` container. Useful for `Routes` I have found but it is all personal pref. Either way you are returning an array which is much more natural (wysiwyg) than `<></>` imo.

This is somewhat a non-issue for me in particular, as you know, I am more of a _JSnoX_ kinda guy. So happy to hand this one over to someone who knows more on the topic.

olstenlarck:
[...]

PR #568, opened 2018-01-23
app() should not remove root element

mattdesl:
Right now, if you have a body like this: `<body><script><script></body>` and use the demo code in the [Readme intro](https://github.com/hyperapp/hyperapp/blob/master/README.md), the script tag will be replaced with the `<main>` element. This is probably not desirable.

This PR fixes that and adds a test. However, please take a close look at the PR since I am not 100% sure if it is a strong fix — I don't have enough familiarity with the codebase to know what sort of situations this might break in. In what cases is this last `else` actually triggered, other than initially on the first patch?

Cheers.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -37,6 +37,11 @@ export function app(state, actions, view, container) {
 
   return wiredActions
 
+  function insert (element, newNode, referenceNode) {
+    if (newNode === referenceNode) element.prepend(newNode)
+    else element.insertBefore(newNode, referenceNode)
+  }
+
   function toVNode(element, map) {
     return {
       name: element.nodeName.toLowerCase(),
@@ -308,7 +313,9 @@ export function app(state, actions, view, container) {
         createElement(node, isSVG),
         (nextSibling = element)
       )
-      removeElement(parent, nextSibling, oldNode)
+      if (nextSibling !== rootElement) {
+        removeElement(parent, nextSibling, oldNode)
+      }
     }
     return element
   }
--- a/src/index.js
+++ b/src/index.js
@@ -37,11 +37,6 @@ export function app(state, actions, view, container) {
 
   return wiredActions
 
-  function insert (element, newNode, referenceNode) {
-    if (newNode === referenceNode) element.prepend(newNode)
-    else element.insertBefore(newNode, referenceNode)
-  }
-
   function toVNode(element, map) {
     return {
       name: element.nodeName.toLowerCase(),
--- a/test/dom.test.js
+++ b/test/dom.test.js
@@ -738,3 +738,21 @@ testTreeSegue("elements with falsey values", [
     html: `<div></div>`
   }
 ])
+
+test('does not remove other tags', done => {
+  const actions = {
+    next: () => () => {
+      expect(document.body.innerHTML).toBe(
+        `<div></div><script></script>`
+      )
+      return done()
+    }
+  }
+
+  const view = (state, actions) => h("div", {
+    oncreate: actions.next
+  })
+
+  document.body.innerHTML = `<script></script>`
+  const main = app({}, actions, view, document.body)
+})


JorgeBucaran:
@mattdesl I think this is not related to #566, correct?

mattdesl:
Not related at all. :smile:

JorgeBucaran:
> @mattdesl Right now, if you have a body like this: <body><script><script></body> and use the demo code in the Readme intro, the script tag will be replaced with the <main> element. This is probably not desirable.

Yes, correct, but why is it not desirable?

mattdesl:
To me it seems like an unexpected side-effect: instead of prepending or appending to the parent container, it replaces the first element. This is especially odd if you have to mount your app into a container that already has other elements.

In preact (and presumably react), it appends to the end of body (which is different than this PR). To me, appending seems like a sane default.

```js
import { render, h } from 'preact';
render(<div />, document.body);
```

JorgeBucaran:
> @mattdesl This is especially odd if you have to mount your app into a container that already has other elements.

This is not supported.

frenzzy:
@mattdesl If you launch the app on a specific container, it means the app now cares about replacing its content with actual view:
```html
<div id="app">anything</div>
```
```js
const state = {}
const actions = {}
const view = (state, actions) => 'use this instead of anything above'
const container = document.getElementById('app')

app(state, actions, view, container) // by this line you say:
// "Hyperapp please replace all the content of the container with my view
// and take care about its timely update"
```
so, if you don't want that Hyperapp touch any other tags just create a separate container for it.

mattdesl:
Ok, makes sense. I think you can close this, perhaps it's just a matter of documenting it, since the docs/guides use `document.body`. Maybe mounting to a child element would be a better suggestion for new users, especially those coming from react/etc.

JorgeBucaran:
@mattdesl What do you think would be the best compromise here and how would you document it? Let's ignore the current behavior for a moment and think of what the ideal behavior would be.

frenzzy:
The ideal behavior would be if Hyperapp take care about all children of a container. Patch algorithm must remove everything what is not match the view. It will become obvious when Hyperapp will start to support an array as a root element of a view. Adding a [JSX Fragment support to the patch algorithm](https://github.com/hyperapp/hyperapp/issues/530#issuecomment-359794718) may solve this problem IMO.

frenzzy:
Also I agree with @mattdesl that all official examples should use `<div id="app"></div>` as a container instead of `document.body`. It should help new users.

mattdesl:
IMHO it would be to append the view to the end of whatever parent element you provide (like preact).

e.g. 1 - Let's say for some reason you are building a split screen experience, on the left side of the page is one app, and on the right is another. Ideally, you might want those under the same parent container.

e.g. 2 - Let's say you need to add a `<header>` element in your `document.body` that, for some reason, is not part of your hyperapp UI. Maybe it's some weird analytics requirement, or you are integrating some components from another agency, or whatever. With the current implementation, you would have to re-structure your DOM (and potentially your CSS) to include a _new_ and empty parent container in order to support hyperapp without losing your `header` element.

e.g. 3 - You are using hyperapp to generate HTML markup, maybe server-side rendering or just offline as a CLI tool. If you have a `<body><script src="bundle.js"></script></body>` HTML, and you mount to document.body, your HTML will end up without a script tag. I can imagine a lot of developers would be confused by this.

These examples may seem contrived, but I've seen some much weirder stuff when building real-world apps on a deadline... 😂

JorgeBucaran:
@mattdesl And how do you communicate between the two apps? There should be only one app.

So, we are going back to appending now? 😄

frenzzy:
1. The same container for two different views sounds strange
2. And you probably want footer too, you just need to create a container in a necessary place to render the app to
3. JFYI, once script executed it doesn't matter that `<script>` is removed from the DOM

JorgeBucaran:
@frenzzy About (3), yeah, I remember I was mindblown when I found out, but it makes sense.

JorgeBucaran:
BTW @andyrj what do you think about all this?

mattdesl:
@frenzzy —
1. Yes, strange but not unheard of. Imagine each view is a widget in a dashboard, each widget is developed and published on npm as a completely isolated app.
2. Sure, you may need a footer, or you may not. I've built and seen many sites with just a header nav and no footer. And yes, you "only" need to re-structure your DOM and CSS to support the newly added `<header>`, but that's my point...
3. Not if you are using hyperapp in Node.js to generate markup files. e.g. A CLI tool that writes HTML files.

FWIW I'm fine with the current implementation, I think perhaps just the docs could be updated.

JorgeBucaran:
@mattdesl There's a reason why we don't append, but instead take over the first element. It has to do with Hyperapp's default (can't opt out) re-hydration of existing markup. So, back in the day, I had to decide (1) more flexibility, but more burden on the user or (2) less flexibility, but works out of the box. 

I'd still like to hear what others think about this.

andyrj:
@jorgebucaran I didn't read all of the above thread, but this was indeed what we needed to have the simple default hydration.  Personally the fact that it overwrites the script tag is neat as long as you aren't trying to hmr.

Swizz:
If you are using `document.body` as container, your app should really contain only the markup for Hydratation. And consider to use defer, that is compatible will all versions of all browsers.

I dont know why the practice of adding scripts at the end of the HTML file became so much popular, as `defer` is here since the begining and `async` since HTML5.

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <script src="./index.js" defer></script>
</head>
<body>
  <main>
    <h1>?</h1>
    <button>-</button>
    <button>+</button>
  </main>
</body>
</html>
```

JorgeBucaran:
> @andyrj Personally the fact that it overwrites the script tag is neat as long as you aren't trying to hmr.

What are you talking about?

andyrj:
@mattdesl The main issue I have with default append is that it directly contradicts default hydration with the current 1.0 setup.  We had both options available back in 0.12.1-ish and forced hydration to always explicitly provide a container while the default of not providing a container appended.  So to do this we would need to add a fifth parameters to app(), switch back to an argument object, or make container optional again...  But I still don't see anything in your example list to indicate that right now hyperapp users are limited by this choice.

andyrj:
@jorgebucaran I am talking about the first example with body having just a script tag...

JorgeBucaran:
@andyrj Sorry, I am not following. You were very supportive of the current behavior if IIRC.

andyrj:
@JorgeBucaran I am 100% behind what we have now, and not going back to append.  I was just trying to clarify what we had earlier compared to now.  Also, I am wrong about hmr being a problem so ignore that above, sorry.

JorgeBucaran:
What should we do about this? Is the current behavior an issue to anyone else?

lukejacksonn:
I have had this cause an issue before when the only element in my `body` was an `svg` containing sprites so that I could reference them using `h('use', { href: '#play' })` for example. The issue was, my icons weren't displaying because the containing node was being removed.

I find it weird that is is destructive in this sense.. but it it _has_ to be or is a pain to do otherwise then it is not the end of the world. It could just be clearly documented somewhere.

shawnsandy:
I have been playing around with a ways to implement slots in my hyperapps, is there any way to catch the foot content and convert it to slot or something similar? https://vuejs.org/v2/guide/components.html#Single-Slot

Swizz:
@shawnsandy I guess you are on the wrong issue. But.
I personally use child to achieve this

```jsx
const MyComponent = (props, children) => (
  <div>
    <h2>I'm the child title</h2>
    {!children.length ? (
      <slot>
        This will only be displayed if there is no content
        to be distributed.
      </slot>
     ) : children}
  </div>
)

const view = (state) => (
  <div>
    <h1>I'm the parent title</h1>
    <MyComponent>
      <p>This is some original content</p>
      <p>This is some more original content</p>
    </MyComponent>
  </div>
)
```

zaceno:
@shawnsandy Yeah, better to open a separate issue. But there's been some discussion on the topic here: https://github.com/picodom/picodom/issues/11 Conclusion: probably possible to implement by using a custom `h` function that wraps the original `h`... but haven't tried (yet)

shawnsandy:
@Swizz @zaceno maybe thanks maybe I will open a feature request

shawnsandy:
Just had a look at the PR and maybe it can help example below

* View
``` js
export const view = (state, actions) => (
  <div
    class=""
    oncreate={ e => {

    }}
  >
    <h1 class="title is-1 is-uppercase">{u(".hyper").data("element")}  {state.count}</h1>
    <Button click={actions.down}> Min </Button>   <Button click={actions.up} classes="elm-blue"> Add </Button>

  </div>
);
```

* HTML
``` html
 <div class="section">
    <div class="container">
    <h1 class="title">Component</h1>
      <p class="hyper " data-element="app"></p>
    </div>
  </div>
```

* Output
``` html
<div class="container">

      <p class="hyper " data-element="app">
       <div class=""><h1 class="title is-1 is-uppercase">app  0</h1><button type="button" class="button-elm null"> Min </button>   <button type="button" class="button-elm elm-blue"> Add </button></div></p>
<h1 class="title">Component</h1>
      <p></p>
    </div>
```

If I'm reading this tight this PR removes the `<h1 class="title">Component</h1>` from the rendered component?

JorgeBucaran:
@shawnsandy Sorry, is this comment related to this PR? Otherwise, please post it on a different issue or join the hyperapp slack and discuss it there. Thanks!

JorgeBucaran:
@mattdesl My bad, but I still don't understand this issue. Does app() remove the root element? I believe it does not.

```js
// Consider this HTML: <div id="my-app"></div> 

app(state, actions, view, document.getElementById("my-app")

// Produces this HTML: <div id="my-app">...view...</div>
```

mattdesl:
> My bad, but I still don't understand this issue. Does app() remove the root element? I believe it does not.

It's not that it removes the root element, it's that using `document.body` as the root element leads to a surprising and seemingly arbitrary result of having the first-most element in your DOM tree replaced by the app element.

Surprising in the sense that it is unintuitive, destructive and can cause developer confusion (even [in this thread](https://github.com/hyperapp/hyperapp/pull/568#issuecomment-362034055)).

That's really the only problem. If it's intentional and well documented (like telling users to "only insert app into empty elements" and to discourage using document.body as root) then there is no need to change this behaviour.

JorgeBucaran:
> @mattdesl It's not that it removes the root element, it's that using document.body as the root element leads to a surprising and seemingly arbitrary result of having the first child element in your body replaced by the app element.

Ah, this is true. In Preact the default behavior is to append to the container, instead of attempting SSR hydration, but if you specify `render`'s third argument `replaceElement`, then it will behave as Hyperapp's default.

I opened #590 to change our behavior to match that, but not everyone likes it. We'll see! Thanks.

jorgebucaran:
Thanks for bringing this up, but we are keeping the current behavior.

Append to the provided container element. If the container is not empty, we'll attempt to recycle existing content. 🎉

PR #570, opened 2018-01-23
Improve documentation a bit.

JorgeBucaran:
I tried to improve the Hello World documentation and the section that explains virtual DOM. Thanks to @zaceno for helping me write a lot of it. 

Anyone feel free to comment on documentation areas that are not directly related to the content added by this PR. I want to take the chance to improve as much as possible.

Missing is a section that explains slices and actions in more depth. I have ideas for actions, but I am clueless about slices. Maybe it's because I've been doing this for a year that I find slices to be an extremely simple concept, but I find it hard to document.

Your feedback and comments, please! 🙏


docs/introduction/hello-world.md
```diff
@@ -74,7 +64,9 @@ const actions = {
 }
 ```
 
-Returning a new state from an action updates the current state and schedules a re-render. Actions are not required to have a return value at all. You can use them to call other actions, for example after an asynchronous operation has ended.
+Actions don't mutate the state directly but return a new fragment of the state. If you try to mutate the state inside an action and then return it, the view will not be re-rendered as you might expect. Immutable state updates allow cheap memoization of the view and components using only a strict `===` check.
```
selfup: Finally! This is a very important part of the docs. Glad it has been updated 🎉

docs/introduction/hello-world.md
```diff
@@ -85,7 +77,7 @@ const actions = {
 }
 ```
 
-An action may also return a promise, enabling you to use async functions.
+If an action returns a Promise, we'll pass it to the caller allowing you to create [async](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) actions.
```
selfup: This is a good clarification as well 🙏

docs/introduction/hello-world.md
```diff
@@ -97,68 +89,52 @@ const actions = {
...
-  }
-}
-```
-
-Every state update is immutable and produces a new object, allowing cheap memoization of the view and components using a strict `===` check.
-
 ### View
 
-The view describes your application user interface as a function of the state and actions. This function is called every time we need to re-render because the state has changed.
+Every time your application state changes, the view function is called so that you can specify how you want the DOM to look based on the new state. The view returns your specification in the form of a [virtual DOM](../concepts/vdom.md) and Hyperapp takes care of updating the actual DOM to match it.
```
selfup: 👍

PR #579, opened 2018-01-30
One readme to rule them docs! Ready or not, here we go!

### LINKED ISSUE #531: What documentation would you like to see here?
The goal of this issue is to collect feedback about what stuff should be covered in the official documentation.

- What do you think about the current documentation? 
- What's missing in the documentation?
  - What topics would you like to see covered? 
- If it was up to you, how would you structure/layout the documentation?
- What makes great documentation?
- How can we keep the documentation minimal and still useful?

JorgeBucaran:
Collapse the entire documentation into a single page. One readme to rule them docs.

The documentation about nested actions (slices) is still a work in progress, but everything else is looking more or less how I want them done. I am considering merging this right away and work on docs for slices as we go along.

Close #531.


README.md
```diff
@@ -1,85 +1,224 @@
...
+    name: "h1",
+    props: {},
+    children: ["Hi."]
+  }]
+}
+```
+
+At any point in time, you describe how you want your UI to look like. It is important to understand that the result of render is not an actual DOM node. Those are just lightweight JavaScript objects. We call them the virtual DOM.
+
+React is going to use this representation to try to find the minimum number of steps to go from the previous render to the next.
```
SkaterDad: React? :trollface:

JorgeBucaran: Whoops. 😉

selfup: _hehe_

brodybits:
-1 for removing CONTRIBUTING.md

+1 for keeping all important info in one place

Another suggestion: contributor covenant / code of conduct

JorgeBucaran:
@brodybits I am not removing CONTRIBUTING from _the final_ PR. It got removed when I deleted the docs directory, that's all.

README.md
```diff
@@ -1,85 +1,224 @@
...
+
+The state is a plain JavaScript object that describes your entire program. It consists of all the dynamic data that is moved around in the application during its execution. The state cannot be mutated after it is created. To update our program, we generate a new state object using actions.
+
+```js
+const state = {
+  count: 0
+}
+```
+
+Like any JavaScript object, the state can be a deeply nested tree of objects. We refer to nested objects in the state as partial state. A single state tree does not conflict with modularity — see [Slices] to learn how to update nested objects and split your state and actions.
```
okwolf: Is [Slices] going to link to content yet to be written?

README.md
```diff
@@ -1,85 +1,224 @@
...
+  {
+    id: "app"
+  },
+  [h("h1", null, "Hi.")]
+)
+```
+
+A virtual node props may include any valid [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes) or [SVG](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute) attributes, [DOM events](https://developer.mozilla.org/en-US/docs/Web/Events), [lifecycle events](lifecycle-events.md), and [keys](#keys).
+
+A function that returns a virtual node is also known as a [component](#components).
```
okwolf: Missing docs/example of a component.

PR #582, opened 2018-02-02
Add vNode structure mapper

sergey-shpak:
#### Allows vNode compatibility with external frameworks and libraries

Now you can, define any vNode structure without affecting hyperapp

```javascript

// before any `h` call
h.vNodeMap => ({ name, props, children }) => ({
    name,
    props,
    children,
    nodeName: name,
    attributes: props
})
```
as result virtual nodes will get the new node structure

* please notice, `name`, `props` and `children` shouldn't be overwritten

### PR Changes:
- adds vNode mapper as `h` constructor property
- adds tests

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,11 +18,15 @@ export function h(name, props) {
 
   return typeof name === "function"
     ? name(props || {}, children)
-    : {
+    : h.vNodeMap({
         name: name,
         props: props || {},
         children: children
-      }
+      })
+}
+
+h.vNodeMap = function(node) {
+  return node
 }
 
 export function app(state, actions, view, container) {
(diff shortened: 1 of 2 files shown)


JorgeBucaran:
@sergey-shpak Thanks! But I don't think this should be in the core when you could simply export vNodeMap externally and wrap <samp>h</samp> with it. What do you think?

sergey-shpak:
@JorgeBucaran hm, yes you're right

I think something like is possible:
```json
    ["transform-react-jsx", { "pragma": "h" }]
```

And then import from the following module instead of `hyperapp`

```javascript
import { h as _h } from 'hyperapp';

export const h = (...args) => {
  const { name, props, children } = _h(...args);
  return { name, props, children, nodeName: name, attributes: props }
}
```

What do you think?

sergey-shpak:
closing this as solved

JorgeBucaran:
Thanks, @sergey-shpak! 👍

PR #589, opened 2018-02-12
Mount app by replacing a given element instead of rendering into a container.

JorgeBucaran:
Without removing automatic rehydration, this PR changes Hyperapp's mount strategy. Currently, it renders the app into a given container. If the container **is not** empty, Hyperapp tries to reuse the existing markup (rehydration).

The new behavior, proposed by this PR, is to replace the given DOM element and attempt to rehydrate anyway. 

This PR also adds a `recycled` prop to the VNode to tell the updateElement that the old node was recycled. With this information, we are able to call `oncreate`, instead of `onupdate` on the first render. 

---

I would prefer to keep the default behavior instead of adding all this complexity, but if this is what everyone wants, then I am down for it.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -26,11 +26,12 @@ export function h(name, attributes /*, ...rest*/) {
       }
 }
 
-export function app(state, actions, view, container) {
+export function app(state, actions, view, element) {
   var renderLock
   var invokeLaterStack = []
-  var rootElement = (container && container.children[0]) || null
-  var oldNode = rootElement && toVNode(rootElement, [].map)
+  var container = element ? element.parentNode : document.body
+
+  var oldNode = element && toVNode(element, [].map)
   var globalState = clone(state)
   var wiredActions = clone(actions)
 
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -26,11 +26,12 @@ export function h(name, attributes /*, ...rest*/) {
       }
 }
 
-export function app(state, actions, view, container) {
+export function app(state, actions, view, element) {
   var renderLock
   var invokeLaterStack = []
-  var rootElement = (container && container.children[0]) || null
-  var oldNode = rootElement && toVNode(rootElement, [].map)
+  var container = element ? element.parentNode : document.body
```
frenzzy: Why do you want to stop support of SSR or apps with non-DOM view?

frenzzy:
Definitely `oncreate` must be called instead of `onupdate` on first render even after hydration but container element should not be modified or deleted IMO.

andyrj:
I believe you can correct the onupdate/oncreate issue without needing to add flag into vnode.  Just needs to be set within hyperapp and checked as is in this...  Hyperapp.isRecycling instead of VNode.recycled.

PR #590, opened 2018-02-12
Enable re-hydration via recycleElement passed to app.

jorgebucaran:
This new behavior is more back-portable than #589 , but still breaking, since we will not rehydrate unless a <samp>recycleElement</samp> is provided to the app function.

Now the app function accepts an additional optional argument: <samp>recycleElement</samp> that will be used to update a specific element. This means you can have multiple apps rendered into `document.body`, without interference.

The new default behavior is to append to the provided <samp>container</samp> element, without rehydration. We will not attempt to recycle any existing markup unless you also pass a <samp>recycleElement</samp> to the app function.

I still need to fix tests and we need to debate how we want lifecycle events to be invoked during rehydration (that is do we want to keep the current behavior in Hyperapp ~1.0.0 or do we want to call <samp>oncreate</samp> even if the element **already** exists). 

I added a new section "mounting" as suggested by @zaceno, but I am not happy at all with the documentation of this part. If you could help, I would appreciate it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,8 @@ const actions = {
 
 #### Nested Actions
 
+> **Work In Progress**: This section is not perfect and may be incomplete. Please help.
+
 Actions can be nested inside namespaces. Updating deeply nested state is as easy as declaring actions inside an object in the same path as the part of the state you want to update.
 
 ```jsx
@@ -195,6 +197,43 @@ const view = (state, actions) =>
 ```
 
 It is important to understand that the result of the view is a new Virtual DOM. It may seem wasteful to throw away the old Virtual DOM and re-create it entirely on every update — not to mention the fact that at any one time, Hyperapp is keeping two Virtual DOM trees in memory, but as it turns out, browsers can create hundreds of thousands of objects very quickly. On the other hand, modifying the DOM is several orders of magnitude more expensive.
+
+## Mounting
+
+> **Work In Progress**: This section is not perfect and may be incomplete. Please help.
+
+Hyperapp will mount your rendered view into the given <samp>container</samp> Element. The default behavior is non-destructive. We will not take over any existing elements inside the container.
+
+```js
+app(state, actions, view, container)
+```
+
+### SSR
+
+Hyperapp works transparently with SSR and pre-rendered HTML, enabling SEO optimization and improving your sites time-to-interactive. The process consists of serving a fully pre-rendered page together with your application and providing a <samp>recycleElement</samp> DOM element to the <samp>app</samp> function.
+
+```js
+app(state, actions, view, container, recycleElement)
+```
+
+```html
+<html>
+<head>
+  <meta charset="utf-8">
+  <script defer src="bundle.js"></script>
+</head>
+
+<body>
+  <div>
+    <h1>0</h1>
+    <button>-</button>
+    <button>+</button>
+  </div>
+</body>
+</html>
+```
+
+Then instead of throwing away the pre-rendered markdown, we'll turn your DOM element into an interactive application out of the box.
 
 ## Virtual DOM
 
@@ -468,29 +507,6 @@ const PlayerList = ({ players }) =>
       </li>
     ))
 ```
-
-## Hydration
-
-Hyperapp works transparently with SSR and pre-rendered HTML, enabling SEO optimization and improving your sites time-to-interactive. The process consists of serving a fully pre-rendered page together with your application.
-
-```html
-<html>
-<head>
-  <meta charset="utf-8">
-  <script defer src="bundle.js"></script>
-</head>
-
-<body>
-  <div>
-    <h1>0</h1>
-    <button>-</button>
-    <button>+</button>
-  </div>
-</body>
-</html>
-```
-
-Then instead of throwing away the server-rendered markdown, we'll turn your DOM nodes into an interactive application out of the box.
 
 ## Community
 
--- a/src/index.js
+++ b/src/index.js
@@ -26,11 +26,10 @@ export function h(name, attributes /*, ...rest*/) {
       }
 }
 
-export function app(state, actions, view, container) {
+export function app(state, actions, view, container, element) {
   var renderLock
   var invokeLaterStack = []
-  var rootElement = (container && container.children[0]) || null
-  var oldNode = rootElement && toVNode(rootElement, [].map)
+  var oldNode = element && toVNode(element, [].map)
   var globalState = clone(state)
   var wiredActions = clone(actions)
 
@@ -55,7 +54,7 @@ export function app(state, actions, view, container) {
 
     var next = view(globalState, wiredActions)
     if (container && !renderLock) {
-      rootElement = patch(container, rootElement, oldNode, (oldNode = next))
+      element = patch(container, element, oldNode, (oldNode = next))
     }
 
     while ((next = invokeLaterStack.pop())) next()


selfup:
I would consider this as MAJOR.MINOR and bump 1.1.2 to 1.2.

However feel free to share your own thoughts 🎉

JorgeBucaran:
@selfup Yeah, this is a teeny-weeny major breaking change. What to do though.

andyrj:
I like this change and think it makes the most sense to have "hydrated" components trigger oncreate during hydration.

okwolf:
This is the point at which positional arguments start to break down and a single "props" object would really shine 👹

zaceno:
@okwolf I agree partially. IMO it should be: `app({state, actions, view}, container=undefined, replaceElement=undefined)`
... but to keep things moving I say: let's decide on this PR first, and then decide wether to change the signature of `app`

zaceno:
Btw, this "old" PR is related right: https://github.com/hyperapp/hyperapp/pull/568 ?
Solved/invalidated by this PR?

README.md
```diff
@@ -196,6 +198,43 @@ const view = (state, actions) =>
 
 It is important to understand that the result of the view is a new Virtual DOM. It may seem wasteful to throw away the old Virtual DOM and re-create it entirely on every update — not to mention the fact that at any one time, Hyperapp is keeping two Virtual DOM trees in memory, but as it turns out, browsers can create hundreds of thousands of objects very quickly. On the other hand, modifying the DOM is several orders of magnitude more expensive.
 
+## Mounting
+
+> **Work In Progress**: This section is not perfect and may be incomplete. Please help.
```
zaceno: Here's some alternative wording:

In order to start your app, call

```js
```

where `container` is an element in your DOM. Hyperapp will produce a DOM element from the vnode returned by your `view`, and add it to the children of `container`.

If there are already children in `container`, the app's root-element will be inserted above them, but they will be otherwise left untouched.

README.md
```diff
@@ -196,6 +198,43 @@ const view = (state, actions) =>
...
+
+Hyperapp will mount your rendered view into the given <samp>container</samp> Element. The default behavior is non-destructive. We will not take over any existing elements inside the container.
+
+```js
+app(state, actions, view, container)
+```
+
+### SSR
```
zaceno: I'd like to think more about this section, but I don't think we need to hold up this PR on account of that. 
 
The thing that's confusing to me, is that even if hyperapp *didn't* do hypdration proper (the `toVnode` thing), but *only* the replaced the 5th-argument-element with itself, we'd still have the "time-to-interactive" benefit described here.

So for one, the `replaceElement`-argument is a feature in itself. You could have simply `<div id="myapp">App not started yet</div>` as the replace-element.

For two, what benefit does the fact that hyperapp calls `toVnode` behind the scenes provide (over just replacing the dom that was already there)? Something with lifecycle events? Anyway, whatever it is, we should clarify it.

And depending on what the difference is, maybe we want to split the `### SSR`  section into two.

zaceno: Oh never mind. I get it. `toVnode`  just makes the first render less expensive *if* the existing DOM matches the app's first VDOM well enough.  Right? 

In that case I don't think we need to mention the word "hydration" or "SSR" at all (to me those are scary words that I associate with complex and framework-specific server-side setup). "Pre-existing elements" or even "pre-existing html" sound more approachable, and are afaik equally correct.

andyrj: Hydration doesn't fix the "uncanny valley" situation where before it has completed the app is visible but not functional.  It does at least leave whatever the user may have input and not cause possible reflow/repaint/flicker/scroll resets etc.  Others will have to defend why they want it as is since I haven't been actively using hyperapp myself atm.

zaceno: @andyrj To clarify: I wasn't saying we should remove the hydration feature. Rather that we should not *call it* "hydration", because it makes it seem more magical than it really is.

It took me until today to realize that what we're really talking about is simply replacing static html that is already there, with whatever the app renders.  The fact that we use `toVnode` is an internal optimization of that (less dom-operations to do the replacing) that the user doesn't need to know about. From the outside it simply looks like "replacing existing html".

andyrj: Hydration is about adopting current nodes it isn't "replacing" anything other than event handlers.

andyrj: Maybe call it adopt()?

andyrj: Also I think that time to interactive was from me when @JorgeBucaran was asking me to help him clarify the point of it back some months ago now.  If someone has a less buzzwordy description I definitely think that bit could be improved upon...  Have I told you all I am terrible at documentation yet? 😉

frenzzy:
Sorry guys but I think this is wrong direction of evolution:

1. We should not make things more complex for user. After this change he will need to read more docs and understand why he need one more argument. But now it just works!
2. It is so intuitive when the patch DOM library cares about all the content in a container (see #568 comments)
  ```js
  <div id="container"><img src="spinner.gif"/> Loading... (this content will be replaced)</div>
  ```
3. There are so many use cases for the `hyperapp`, for example: make a widget dynamic on existent page, modal dialogs or tooltips rendered to a special container, nested stateful components and so on. In this PR you are focused on making container unnecessary in case of `document.body`, but it is much simpler to use a container DON node always!
4. It will be much more comfortable to users if `hyperapp` supports returning an array of nodes for the root element (see how many users had issues with it in our slack and github). But after current PR it will be much more complicated to implement this feature. Btw, this feature is also focused on reducing the number of necessary DOM elements.

JorgeBucaran:
@frenzzy What alternative do you suggest?

JorgeBucaran:
@frenzzy AFAIK this is how preact works as well. https://preactjs.com/guide/api-reference#-preact-render-

frenzzy:
If users need want to keep something untouched in `document.body` they just need to create a container for `hyperapp`:
```html
<body>
  <!-- any stuff here, untouched by hyperapp -->
  <div id="container"></div>
  <!-- any stuff here, untouched by hyperapp -->
</body>
```
Very simple!

frenzzy:
But `hyperapp` is not a `preact`, it is much simpler, lighter and "optionless" :D

JorgeBucaran:
@frenzzy 

```html
<body>
  <!-- any stuff here, untouched by hyperapp -->
  <div id="container"></div>
  <!-- any stuff here, untouched by hyperapp -->
</body>
```

Using `id=container` as app's <samp>container</samp> argument, will generate this code:

```html
<body>
  <div id="container">
    <div>
      Your App
    </div>
  </div>
</body>
```

frenzzy:
What the problem with one more extra `div` tag? Simplicity is more important, isn't it?

JorgeBucaran:
Folks, @frenzzy has a point. 🤔

SkaterDad:
I find the current behavior a tiny bit annoying, to be honest.  By specifying a container, I normally expect that container to be completely taken over, like in other frameworks.

Its a minor issue to me, so I'm good with either way this goes.

README.md
zaceno: @andyrj If your preexisting dom is `<div id="app-mount-point">App not started, yet...</div>` then when your app starts, it will replace that with the app UI. But there will be no reusing of elements since the existing DOM does not match the app's vdom closely enough.

From a users perspective, it's replacing what was there, regardless of wether hyperapp is able to repurpose those elements or not.

JorgeBucaran:
@SkaterDad When you say take over the container you mean "replace" the container with your tree?

JorgeBucaran:
@SkaterDad When you say other frameworks, to which frameworks are you referring?

This PR makes Hyperapp's mounting behave like Preact.

JorgeBucaran:
@frenzzy Support for arrays from the root seems tough and full of complications I still can't anticipate. Lifecycle events, how to diff when oldNode is an array, but node is not and viceversa, etc.

Siilwyn:
I have no strong opinion on the actual change, seems to me it's about keeping it simple vs. making it a bit nicer and flexible... Regarding the arguments form I completely agree with @okwolf, using an object would make this clearer.
Regarding the version change, this is definitely a major version bump since it changes default behavior that people can depend on in several ways.

README.md
JorgeBucaran: > @andyrj Have I told you all I am terrible at documentation yet? 😉

Here we go again. Just kidding. 🤣

logankaser:
I prefer the current behavior, I think its very intuitive and elegant.
Potential extra div tags are not a problem worth adding an argument for.
Just my two cents.

frenzzy:
With support for arrays from the root #591 this option become useless, isn't it?

JorgeBucaran:
@frenzzy It is. But I don't think I want to go for that either.

frenzzy:
Guys please vote:

1. Hyperapp is taking control over the entire contents of the container which is always matched view. No API change. As a bonus you will able to return multiple nodes from view and use placeholders like `<container>Loading...</container>`.

vs

2. Hyperapp expects empty container or requires container AND element specified. Braking API change. If you want to use a placeholder, you need to take care of removing it yourself or use `<container><element>placeholder</element></container>`. One more extra argument for `app` which you need to understand how it works and explain to your friend who doesn't like to read docs.

1 or 2 ?

SkaterDad:
@frenzzy  #1

JorgeBucaran:
It's not a simple poll. (1) is difficult to implement. Then we'll need to implement fragments, properly, as well as all their gotchas and whatnot. I don't think it's worth it.

frenzzy:
But it is already implemented in #591, hyperapp there removes everything from the container what is not matched view. Fragments are just syntax sugar which [flatten](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatten) children arrays for you.

JorgeBucaran:
> But it is already implemented in #591, hyperapp there removes everything from the container what is not matched view. Fragments are just syntax sugar which flatten children arrays for you.

As discussed on Slack (posting here for the record), this is not quite correct. It could be hacked like that to fake it out, but it's not really what fragments are about. Fragments are essentially about treating tuples/arrays of nodes as a single node when reordering the keyed DOM.

JorgeBucaran:
@okwolf Re-hydration can't be removed.

okwolf:
> @JorgeBucaran Re-hydration can't be removed.

Then my vote is keep things the way they are.

Are we keeping this open because we're still considering changing?

jorgebucaran:
I am going to keep the current behavior for the time being but might reconsider in the future. 

To summarize the current behavior is to append your rendered view to the provided container element. If the container is not empty, we'll attempt to recycle existing content. 🎉

PR #591, opened 2018-02-13
An array as a root node

frenzzy:
An alternative solution for #590 and #568 POC

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -30,8 +30,7 @@ export function app(state, actions, view, container) {
   var renderLock
   var firstRender = true
   var lifecycleStack = []
-  var rootElement = (container && container.children[0]) || null
-  var oldNode = rootElement && toVNode(rootElement, [].map)
+  var oldNode = container && toVNode(container, [].map)
   var globalState = clone(state)
   var wiredActions = clone(actions)
 
@@ -56,7 +55,8 @@ export function app(state, actions, view, container) {
 
     var next = view(globalState, wiredActions)
     if (container && !renderLock) {
-      rootElement = patch(container, rootElement, oldNode, (oldNode = next))
+      next = h(container.nodeName.toLowerCase(), {}, next)
+      container = patch(container, container, oldNode, (oldNode = next))
       firstRender = false
     }
 
--- a/test/dom.test.js
+++ b/test/dom.test.js
@@ -758,3 +758,24 @@ testTreeSegue("elements with falsey values", [
     html: `<div></div>`
   }
 ])
+
+test("an array as a root node", done => {
+  const view = () => [
+    h("div", {}, "A"),
+    h("div", {}, "B"),
+    h(
+      "div",
+      {
+        oncreate() {
+          expect(document.body.innerHTML).toBe(
+            `<div>A</div><div>B</div><div>C</div>`
+          )
+          done()
+        }
+      },
+      "C"
+    )
+  ]
+
+  app({}, {}, view, document.body)
+})


src/index.js
```diff
@@ -55,6 +55,11 @@ export function app(state, actions, view, container) {
 
     var next = view(globalState, wiredActions)
     if (container && !renderLock) {
+      next = {
+        nodeName: container.nodeName.toLowerCase(),
+        attributes: {},
+        children: [].concat(next)
```
frenzzy: or `next.pop /* Array */ ? next : [next]`

chicoxyzzy: Why not `Array.isArray(next)` (IE 9 compatible)?

JorgeBucaran:
Thanks, @frenzzy. This is an interesting proposal. What about adding Fragments to this PR as well?

src/index.js
frenzzy: because of bytes and perfomance? see 4930b5ecd06d9d5c5d6fe7311769020d313fcd6c

frenzzy:
Fragments is a separate feature which is not quite related to the current single root node problem, I would prefer to create a separate commits (PRs) for each one. And for Features I see three different possible implementations:
1. Support VNode with empty name in patch function
2. Support for nested arrays inside `children` arrays
3. Converting fragments (nested arrays) to plain `children` arrays inside `h` function

Maybe need to create three PRs and compare..

src/index.js
chicoxyzzy: Size over correctness... is bad thing

lukejacksonn: Then you might not like this repo...

JorgeBucaran: @chicoxyzzy That's a bit too dogmatic. We take a lighter approach. If we can get away with it, I suggest we do it. Not always, but if we can why not? :)

frenzzy:
Lol, JSX Fragments already supported, see #596
And this PR makes the life easier to use them :)

JorgeBucaran:
I would prefer not. Put everything in a single PR. Thanks.

JorgeBucaran:
This is clever, but if we introduce Fragments someday in the future, I'd prefer this to be properly implemented using DocumentFragment and going all the way as to allow Components to return keyed fragments, not simply an array.

frenzzy:
The PR is not about fragments directly, but about an ability to render more than one element into the `document.body`. I see your point.

JorgeBucaran:
Yes, you are right. My apologies for asking you to mix two unrelated things.

frenzzy:
+4 bytes^^

JorgeBucaran:
The only way we'll support returning an array from the root is if we add fragment support someday, but not like this. Thanks.

frenzzy:
The main purpose why I did this, I want that users do not try to put various elements in the container and look for a workaround for "which element Hyperapp should revive" like here #590 and #568

JorgeBucaran:
@frenzzy Because we may add fragment support in the future, I think the best we can do now is nothing. So, both #590 and #598 may have to wait as well. Although, I like #598 and it's also what I want to do for Picodom 2.0.

- https://github.com/picodom/picodom/pull/96

jorgebucaran:
@frenzzy Do you think we could implement this in a different way?

infinnie:
But btw, we’ve been longing for this feature since the times of 0.1x.
Are keyed elements supported now in root-level array elements?

jorgebucaran:
> Are keyed elements supported now in root-level array elements?

No, how would you implement that?

frenzzy:
@jorgebucaran what exactly you dislike in the current implementation?

Since we have subviews/lazy components now, I think it make sense to change `patch` signature to accept a parent element and already resolved an array of children VNodes. And patch function will just modify all children elements without touching the parent element. What do you think?

jorgebucaran:
> And patch function will just modify all children elements without touching the parent element. What do you think?

Do you have some code or pseudo code to explain what you mean?

frenzzy:
Something like `patch(container, nextVDomTree, prevVDomTree)` and inside patch do not touch `container` itself, only `container.childNodes`

PR #592, opened 2018-02-13
Make boolean "false" a valid attribute value

frenzzy:
before: `<div data-attr={false} />` => `<div></div>`
after: `<div data-attr={false} />` => `<div data-attr="false"></div>`

Useful for attributes such as [draggable](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable) and for custom data-attributes.

The does not affect props like `checked` and they will continue work as previously.
Hope this change also decrease library size on a few bytes :)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -137,11 +137,11 @@ export function app(state, actions, view, container) {
     } else {
       if (typeof value === "function" || (name in element && !isSVG)) {
         element[name] = value == null ? "" : value
-      } else if (value != null && value !== false) {
+      } else if (value != null) {
         element.setAttribute(name, value)
       }
 
-      if (value == null || value === false) {
+      if (value == null) {
         element.removeAttribute(name)
       }
     }
--- a/test/dom.test.js
+++ b/test/dom.test.js
@@ -749,7 +749,7 @@ testTreeSegue("elements with falsey values", [
     tree: h("div", {
       "data-test": false
     }),
-    html: `<div></div>`
+    html: `<div data-test="false"></div>`
   },
   {
     tree: h("div", {


JorgeBucaran:
I need to do a bit more testing to make sure this does not break other stuff.

Good job @frenzzy!

frenzzy:
Oh.. looks like this PR introduces inconsistency between client and server side rendering because on server side we can't check if attribute belongs to element without whitelisting. It is much easier to just not render falsey attributes. See [react-dom/src/shared/DOMProperty.js](https://github.com/facebook/react/blob/1fd205ad2d101f441622c9bb7e2ccb480281e0a0/packages/react-dom/src/shared/DOMProperty.js#L242-L334)

frenzzy:
After this PR only non-standard attributes will render `false` as a string:
- `<div data-custom={false} />` => `<div data-custom="false"></div>`
- `<button disabled={false} />` => `<button></button>`

JorgeBucaran:
@SkaterDad I have a feeling this has been discussed before and we agreed with each other.

> For some attributes, like disabled, it's preferable (to me) to remove it completely when false...

Why was that again? 🤔

JorgeBucaran:
@frenzzy This needs more tests and also more cowbell. 🐮🔔

SkaterDad:
> @frenzzy After this PR only non-standard attributes will render false as a string:

Can you put tests to verify this? 

When I code this:
```js
    h('input', { oncreate: (el) => { el.setAttribute('disabled', false)} })
```

Chrome renders this, which is still a disabled element functionally:
```html
<input disabled="false">
```

frenzzy:
Yes, I'm scared about this change too. But do not have a real app to test this yet.

frenzzy:
@SkaterDad there are check in code:
```js
if (attributeName in element) {
  element[name] = value
} else {
  // use element.setAttribute
}
```
https://github.com/hyperapp/hyperapp/blob/c1a22ae45817f681a43059db27730d82d2704632/src/index.js#L140-L141

SkaterDad:
@frenzzy Thanks for pointing that out.   I must have missed that `hyperapp` distinguished between props and attributes again! :+1: 

My use cases shouldn't be affected by this PR, then, since I'm not doing any SSR currently.

frenzzy:
I gonna close this PR because it introduces mismatching between client and server side rendering and does not help in any case to render boolean `false` as a string to reserved DOM attributes like `draggable`, you need to pass `"false"` as a string anyway.

PR #594, opened 2018-02-14
oldProps→oldAttributes in README.md

frenzzy:
continues 25b85ef848c40b7030ecb7241ee55b49e4ef84fd

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,15 +393,15 @@ const Textbox = ({ placeholder }) => (
 
 ### onupdate
 
-This event is fired every time we update the element attributes. Use <samp>oldProps</samp> inside the event handler to check if any attributes changed or not.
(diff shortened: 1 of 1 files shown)


frenzzy:
maybe `prevAttributes` is better name?

JorgeBucaran:
@frenzzy What if we just call them old <samp>attributes</samp>?

frenzzy:
Second argument of `onupdate` is actually previous VNode attributes. If we call it just `attributes` it may confuse users. Also `old` prefix in my mind is some version in the past, not exactly previous. And `prevAttributes` explains what it really is.

JorgeBucaran:
Ah, well you are right. About `old` vs `previous`. We are already using `old` internally and in the docs, so I prefer we use `old` here too.

JorgeBucaran:
Thanks, @frenzzy!

PR #595, opened 2018-02-14
Invoke oncreate instead of onupdate on hydration

### LINKED ISSUE #588: Hyperapp do not call oncreate handler when view mounted to document.body
Code:
```js

function view() {
  return (<div oncreate={() => console.log('log oncreate')} />);
}

app({}, {}, view, document.body);
```

In this case we do not see a message `log oncreate`. It take some time to understand that app mount view to body and this element is not re-created. Maybe I miss something, but I have assumption that this handler should be called in any case when view mounted first time.

I solve a "problem" in this way: just add element with ID and mount view to it.

frenzzy:
Allows to init third-party libs if `app` was started on non empty container.

Fixes #588 
Refs #589

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -27,6 +27,7 @@ export function h(name, attributes /*, ...rest*/) {
 }
 
 export function app(state, actions, view, container) {
+  var init = true
   var renderLock
   var invokeLaterStack = []
   var rootElement = (container && container.children[0]) || null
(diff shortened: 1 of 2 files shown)


JorgeBucaran:
@frenzzy Thanks. I'll likely implement this together with #590.

frenzzy:
But #590 does not fix the issue yet and more focused on a introducing new api

JorgeBucaran:
🎉🎉🎉

frenzzy:
Interesting side effect here:
```js
<div
  oncreate={(element, attributes) => {
    if (attributes !== undefined) {
      console.log('node was hydrated!')
    } else {
      console.log('node was created from scratch')
    }
  }}
/>
```

JorgeBucaran:
@frenzzy Haha, true, danger zone. 💀

PR #597, opened 2018-02-15
name→nodeName

frenzzy:
What do you think about more descriptive argument name? :)


JorgeBucaran:
@frenzzy Yeah, this is fine, but <samp>name</samp> can also be a function. If no one has a better suggestion, let's go with <samp>nodeName</samp>.

jorgebucaran:
I don't have a good reason and you are making a good case, but I'm going to keep it like now for time being.

PR #598, opened 2018-02-15
Breaking! Mount view by replacing given DOM element.

jorgebucaran:
## Summary

In <samp>app(state, actions, view, **element**)</samp>, replace <samp>element</samp> with your rendered view _instead_ of appending to it.

### (I)

```jsx
// <body>
//   <div id="app"></div>
// </body>

app(state, actions, () => <h1>Hello</h1>, document.getElementById("app"))

// <body>
//   <h1>Hello</h1>
// </body>
```

### (II)

```jsx
// <body>
//   <div id="app"></div>
// </body>

app(state, actions, () => (
  <div>
    <h1>Hello</h1>
  </div>
), document.getElementById("app"))

// <body>
//   <div>
//     <h1>Hello</h1>
//   </div>
// </body>
```

### (III)

```jsx
// <body>
//   <div id="app"></div>
// </body>

app(state, actions, () => (
  <body>
    <h1>Hello</h1>
  </body>
), document.body)

// <body>
//   <h1>Hello</h1>
// </body>
```

## Why?

Because it favors the most common use case: rendering into some container instead of <samp>document.body</samp> and it's consistent with Hyperapp not allowing returning arrays from the top view. It's also nice not having a dummy container enclosing the app.

## Notes

This PR is competing with #590. 

## See Also

- https://medium.com/@dan_abramov/two-weird-tricks-that-fix-react-7cf9bbdef375

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -26,27 +26,27 @@ export function h(name, attributes /*, ...rest*/) {
       }
 }
 
-export function app(state, actions, view, container) {
+export function app(state, actions, view, element) {
+  var map = [].map
+  var oldNode
   var renderLock
   var firstRender = true
-  var lifecycleStack = []
-  var rootElement = (container && container.children[0]) || null
-  var oldNode = rootElement && toVNode(rootElement, [].map)
   var globalState = clone(state)
   var wiredActions = clone(actions)
+  var lifecycleStack = []
 
   scheduleRender(wireStateToActions([], globalState, wiredActions))
 
   return wiredActions
 
-  function toVNode(element, map) {
+  function elementToVNode(element) {
     return {
       nodeName: element.nodeName.toLowerCase(),
       attributes: {},
       children: map.call(element.childNodes, function(element) {
         return element.nodeType === 3 // Node.TEXT_NODE
           ? element.nodeValue
-          : toVNode(element, map)
+          : elementToVNode(element)
       })
     }
   }
@@ -55,8 +55,13 @@ export function app(state, actions, view, container) {
     renderLock = !renderLock
 
     var next = view(globalState, wiredActions)
-    if (container && !renderLock) {
-      rootElement = patch(container, rootElement, oldNode, (oldNode = next))
+    if (element && !renderLock) {
+      element = patch(
+        element.parentNode,
+        element,
+        oldNode || elementToVNode(element),
+        (oldNode = next)
+      )
       firstRender = false
     }
 


frenzzy:
I think replacing a container is a bad idea, because the ref to it could be used by third party code and when you need to unmount hyperapp, you don't know know which element to remove.
I still prefer #591 over this.

JorgeBucaran:
> ... because the ref to it could be used by third party code and when you need to unmount hyperapp, you don't know know which element to remove.

This is only true if you replace the given element with an element of a different type, which doesn't make much sense, as it would for example defeat the purpose of hydration.

```js
const myRoot = document.getElementById("app")
const state = ...
const actions = ...
  <div>
    <h1>{state.count}</h1>
    <button onclick={() => actions.down(1)}>-</button>
    <button onclick={() => actions.up(1)}>+</button>
  </div>
)

app(state, actions, view, myRoot)
```

screen shot 2018-02-16 at 0 55 16

src/index.js
```diff
@@ -26,12 +26,12 @@ export function h(name, attributes /*, ...rest*/) {
...
 }
 
-export function app(state, actions, view, container) {
+export function app(state, actions, view, element) {
   var renderLock
   var firstRender = true
   var lifecycleStack = []
-  var rootElement = (container && container.children[0]) || null
-  var oldNode = rootElement && toVNode(rootElement, [].map)
+  var container = element.parentNode
```
frenzzy: `element && element.parentNode` ?

frenzzy:
Another disadvantage is you need to duplicate the same node in the DOM and in a view.

JorgeBucaran:
Why is that a disadvantage?

frenzzy:
Because of DRY principe, why do you need to do it twice? Why not just create a container in the DOM and mount your view into it..

Swizz:
I dont see this as a disadvantage. It make a lot more sense to me :  Your app container is your view root. 

It make a lot more sens that : Your app container will be whiped out and the view root will be added to it.

zaceno:
I think the main reason I like this change is that it gives absolute control to the user over where the top-element of the app gets placed in the rest of the dom.

The old (=current) solution only lets you specify the container. And if you want your app mounted at a specific place among siblings (without an extra wrapper div)  you're out of luck -- unless that place is "first".

This control makes mounting a lot easier to reason about, and much easier to explain in the docs as well.

src/index.js
```diff
@@ -55,8 +55,13 @@ export function app(state, actions, view, container) {
     renderLock = !renderLock
 
     var next = view(globalState, wiredActions)
-    if (container && !renderLock) {
-      rootElement = patch(container, rootElement, oldNode, (oldNode = next))
+    if (element && !renderLock) {
+      element = patch(
+        element.parentNode,
```
frenzzy: `parentNode` could be `null` for unmounted node, is it ok?
```js
const element = document.createElement('div')
app(state, actions, view, element)
```

JorgeBucaran: Why?

frenzzy: You won't able to mount `app` on detached node because it will throw an exception (example updated)

frenzzy:
Also I think this change may cause a lot of headache to users in case of misunderstanding of the concept and misuse. Immutable container is much more robust in this sense IMO. I like when users do not have a way to misuse things.

JorgeBucaran:
Well, it was nice to dream about this.

mrozbarry:
I'm on board with this. I really like that It allows the developer to control the container in code rather than externally define it (ie in the dom/html page) and do some magic to maintain it.

I think it needs to be clear in the documentation that this is the behaviour to avoid confusion, but I think this is the most correct approach.

swandir:
Has this option been discussed?

`app(state, actions, view, container, elementToReplace)`

`elementToReplace` argument is optional, and if it's a child of `container`, then it'll be replaced.
That is how [preact does it](https://preactjs.com/guide/api-reference#-preact-render-).

frenzzy:
@swandir see #590 and #591

JorgeBucaran:
@swandir I don't mind that option either, but some people think it's too ugly, which I can't disagree with. My favorite is #590.

jorgebucaran:
I am going to keep the current behavior for the time being, but might reconsider in the future. 

To summarize the current behavior is to append your rendered view to the provided container element. If the container is not empty, we'll attempt to recycle existing content. 🎉

PR #599, opened 2018-02-15
[fix] setTimeout example

rma4ok:
`setTimeout` returns `timeoutId` which will be saved as state

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,8 @@ Actions used for side effects (writing to databases, sending a request to a serv
 ```js
 const actions = {
   upLater: value => (state, actions) => {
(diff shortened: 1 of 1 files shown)


Swizz:
When you use arrow function with body `(a, b) => { ... }` nothing will be returned by the function, to return something you will need to use an explicit return statement.

So, this action return nothing, that will not result at the merge issue you are referring.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Function_body

rma4ok:
@Swizz that's right. Thank you

PR #604, opened 2018-02-17
Add connected components

etalisoft:
### Why?
Currently all components are dumb components.  Given a component hierarchy A->B->C->D->E.  If Component `E` needed access to `state.name` then components A, B, C, and D would each need to be updated to pass along state.name.  This causes components to become tightly coupled, or written generically so that all state and actions are handed down to every child component, because of the "you never know when you'll need it" mentality.

### The Solution
Natively support both dumb components and connected components. 

Normal dumb component signature:
```jsx
const Component = (props, children) => (
  <div>Hello</div>
);
```

Connected component signature:
```jsx
const ConnectedComponent = (props, children) => (state, actions) => (
  <div>Hello, {state.name}</div>
);
```

This solution does not require use of either a Higher Order App or a Higher Order Component. 

### How does it work?
In hyperapp's `render` function, whenever the view is processed, the resulting object tree is traversed.  Any child node that is a function is called with the `globalState` and `wiredActions`.

This change raised the gzip size by ~30 bytes.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -150,8 +150,8 @@ export function app(state, actions, view, container) {
   }
 
   function connect(children, i) {
-    typeof children[i] === "function" &&
-      (children[i] = children[i](globalState, wiredActions))
+    if (typeof children[i] === "function")
+      children[i] = children[i](globalState, wiredActions)
   }
 
   function createElement(node, isSVG) {
--- a/src/index.js
+++ b/src/index.js
@@ -51,7 +51,7 @@ export function app(state, actions, view, container) {
     }
   }
 
-  function inflate(children) {
+  function connect(children) {
     for (var i in children) {
       if (typeof children[i] === "function")
         children[i] = children[i](globalState, wiredActions)
@@ -175,7 +175,7 @@ export function app(state, actions, view, container) {
         })
       }
 
-      var c = inflate(node.children)
+      var c = connect(node.children)
       for (var i in c) {
         element.appendChild(createElement(c[i], isSVG))
       }
@@ -270,7 +270,7 @@ export function app(state, actions, view, container) {
       var i = 0
       var j = 0
 
-      var c = inflate(node.children)
+      var c = connect(node.children)
       while (j < c.length) {
         var oldChild = oldNode.children[i]
         var newChild = node.children[j]
--- a/src/index.js
+++ b/src/index.js
@@ -51,14 +51,6 @@ export function app(state, actions, view, container) {
     }
   }
 
-  function connect(children) {
-    for (var i in children) {
-      if (typeof children[i] === "function")
-        children[i] = children[i](globalState, wiredActions)
-    }
-    return children
-  }
-
   function render() {
     renderLock = !renderLock
 
@@ -157,6 +149,11 @@ export function app(state, actions, view, container) {
     }
   }
 
+  function connect(children, i) {
+    typeof children[i] === "function" &&
+      (children[i] = children[i](globalState, wiredActions))
+  }
+
   function createElement(node, isSVG) {
     var element =
       typeof node === "string" || typeof node === "number"
@@ -175,8 +172,9 @@ export function app(state, actions, view, container) {
         })
       }
 
-      var c = connect(node.children)
+      var c = node.children
       for (var i in c) {
+        connect(c, i)
         element.appendChild(createElement(c[i], isSVG))
       }
 
@@ -270,8 +268,9 @@ export function app(state, actions, view, container) {
       var i = 0
       var j = 0
 
-      var c = connect(node.children)
+      var c = node.children
       while (j < c.length) {
+        connect(c, i)
         var oldChild = oldNode.children[i]
         var newChild = node.children[j]
 
--- a/src/index.js
+++ b/src/index.js
@@ -172,10 +172,9 @@ export function app(state, actions, view, container) {
         })
       }
 
-      var c = node.children
-      for (var i in c) {
-        connect(c, i)
-        element.appendChild(createElement(c[i], isSVG))
+      for (var i = 0; i < node.children.length; i++) {
+        connect(node.children, i)
+        element.appendChild(createElement(node.children[i], isSVG))
       }
 
       for (var name in node.attributes) {
@@ -267,9 +266,8 @@ export function app(state, actions, view, container) {
       var i = 0
       var j = 0
 
-      var c = node.children
-      while (j < c.length) {
-        connect(c, i)
+      while (j < node.children.length) {
+        connect(node.children, i)
         var oldChild = oldNode.children[i]
         var newChild = node.children[j]
 
--- a/src/index.js
+++ b/src/index.js
@@ -53,9 +53,10 @@ export function app(state, actions, view, container) {
 
   function inflate(children) {
     for (var i in children) {
-      typeof children[i] === "function" && (children[i] = children[i](globalState, wiredActions));
+      typeof children[i] === "function" &&
+        (children[i] = children[i](globalState, wiredActions))
     }
-    return children;
+    return children
   }
 
   function render() {
@@ -174,7 +175,7 @@ export function app(state, actions, view, container) {
         })
       }
 
-      var c = inflate(node.children);
+      var c = inflate(node.children)
       for (var i = 0; i < c.length; i++) {
         element.appendChild(createElement(c[i], isSVG))
       }
@@ -268,7 +269,7 @@ export function app(state, actions, view, container) {
       var i = 0
       var j = 0
 
-      var c = inflate(node.children);
+      var c = inflate(node.children)
       while (j < c.length) {
         var oldChild = oldNode.children[i]
         var newChild = node.children[j]
--- a/src/index.js
+++ b/src/index.js
@@ -90,7 +90,7 @@ export function app(state, actions, view, container) {
   }
 
   function get(path, source) {
-    for (var i in path) {
+    for (var i = 0; i < path.length; i++) {
       source = source[path[i]]
     }
     return source
@@ -214,9 +214,8 @@ export function app(state, actions, view, container) {
 
   function removeChildren(element, node, attributes) {
     if ((attributes = node.attributes)) {
-      var c = node.children
-      for (var i in c) {
-        removeChildren(element.childNodes[i], c[i])
+      for (var i = 0; i < node.children.length; i++) {
+        removeChildren(element.childNodes[i], node.children[i])
       }
 
       if (attributes.ondestroy) {
@@ -254,7 +253,7 @@ export function app(state, actions, view, container) {
       var oldKeyed = {}
       var newKeyed = {}
 
-      for (var i in oldNode.children) {
+      for (var i = 0; i < oldNode.children.length; i++) {
         oldElements[i] = element.childNodes[i]
 
         var oldChild = oldNode.children[i]
@@ -320,8 +319,9 @@ export function app(state, actions, view, container) {
       }
 
       for (var i in oldKeyed) {
-        if (!newKeyed[oldKeyed[i][1].key])
+        if (!newKeyed[oldKeyed[i][1].key]) {
           removeElement(element, oldKeyed[i][0], oldKeyed[i][1])
+        }
       }
     } else if (node.nodeName === oldNode.nodeName) {
       element.nodeValue = node
--- a/src/index.js
+++ b/src/index.js
@@ -52,7 +52,7 @@ export function app(state, actions, view, container) {
   }
 
   function inflate(children) {
-    for (let i in children) {
+    for (var i in children) {
       typeof children[i] === "function" && (children[i] = children[i](globalState, wiredActions));
     }
     return children;
--- a/src/index.js
+++ b/src/index.js
@@ -53,8 +53,8 @@ export function app(state, actions, view, container) {
 
   function inflate(children) {
     for (var i in children) {
-      typeof children[i] === "function" &&
-        (children[i] = children[i](globalState, wiredActions))
+      if (typeof children[i] === "function")
+        children[i] = children[i](globalState, wiredActions)
     }
     return children
   }
@@ -98,7 +98,7 @@ export function app(state, actions, view, container) {
   }
 
   function get(path, source) {
-    for (var i = 0; i < path.length; i++) {
+    for (var i in path) {
       source = source[path[i]]
     }
     return source
@@ -176,7 +176,7 @@ export function app(state, actions, view, container) {
       }
 
       var c = inflate(node.children)
-      for (var i = 0; i < c.length; i++) {
+      for (var i in c) {
         element.appendChild(createElement(c[i], isSVG))
       }
 
@@ -216,8 +216,9 @@ export function app(state, actions, view, container) {
 
   function removeChildren(element, node, attributes) {
     if ((attributes = node.attributes)) {
-      for (var i = 0; i < node.children.length; i++) {
-        removeChildren(element.childNodes[i], node.children[i])
+      var c = node.children
+      for (var i in c) {
+        removeChildren(element.childNodes[i], c[i])
       }
 
       if (attributes.ondestroy) {
@@ -255,7 +256,7 @@ export function app(state, actions, view, container) {
       var oldKeyed = {}
       var newKeyed = {}
 
-      for (var i = 0; i < oldNode.children.length; i++) {
+      for (var i in oldNode.children) {
         oldElements[i] = element.childNodes[i]
 
         var oldChild = oldNode.children[i]
@@ -320,9 +321,8 @@ export function app(state, actions, view, container) {
       }
 
       for (var i in oldKeyed) {
-        if (!newKeyed[oldKeyed[i][1].key]) {
+        if (!newKeyed[oldKeyed[i][1].key])
           removeElement(element, oldKeyed[i][0], oldKeyed[i][1])
-        }
       }
     } else if (node.nodeName === oldNode.nodeName) {
       element.nodeValue = node
--- a/src/index.js
+++ b/src/index.js
@@ -51,10 +51,21 @@ export function app(state, actions, view, container) {
     }
   }
 
+  function inflate(node) {
+    if (node) {
+      var c = node.children || []
+      for (var i in c) {
+        typeof c[i] === "function" &&
+          (c[i] = inflate(c[i](globalState, wiredActions)))
+      }
+    }
+    return node
+  }
+
   function render() {
     renderLock = !renderLock
 
-    var next = view(globalState, wiredActions)
+    var next = inflate(view(globalState, wiredActions))
     if (container && !renderLock) {
       rootElement = patch(container, rootElement, oldNode, (oldNode = next))
       firstRender = false
--- a/src/index.js
+++ b/src/index.js
@@ -51,21 +51,17 @@ export function app(state, actions, view, container) {
     }
   }
 
-  function inflate(node) {
-    if (node) {
-      var c = node.children || []
-      for (var i in c) {
-        typeof c[i] === "function" &&
-          (c[i] = inflate(c[i](globalState, wiredActions)))
-      }
+  function inflate(children) {
+    for (let i in children) {
+      typeof children[i] === "function" && (children[i] = children[i](globalState, wiredActions));
     }
-    return node
+    return children;
   }
 
   function render() {
     renderLock = !renderLock
 
-    var next = inflate(view(globalState, wiredActions))
+    var next = view(globalState, wiredActions)
     if (container && !renderLock) {
       rootElement = patch(container, rootElement, oldNode, (oldNode = next))
       firstRender = false
@@ -178,8 +174,9 @@ export function app(state, actions, view, container) {
         })
       }
 
-      for (var i = 0; i < node.children.length; i++) {
-        element.appendChild(createElement(node.children[i], isSVG))
+      var c = inflate(node.children);
+      for (var i = 0; i < c.length; i++) {
+        element.appendChild(createElement(c[i], isSVG))
       }
 
       for (var name in node.attributes) {
@@ -271,7 +268,8 @@ export function app(state, actions, view, container) {
       var i = 0
       var j = 0
 
-      while (j < node.children.length) {
+      var c = inflate(node.children);
+      while (j < c.length) {
         var oldChild = oldNode.children[i]
         var newChild = node.children[j]
 
--- a/README.md
+++ b/README.md
@@ -373,6 +373,20 @@ const HelloBox = ({ name }) => (
 )
 ```
 
+### Connected Components
+
+Normal components are dumb (not wired to your application state or actions), since they only receive their props and children from the parent component.  To create a connected component simply change the component signature.
+
+```jsx
+const DumbComponent = (props, children) => (
+  <div>Hello</div>
+);
+
+const ConnectedComponent = (props, children) => (state, actions) => (
+  <div>Hello, {state.name}!</div>
+);
+```
+
 ## Lifecycle Events
 
 You can be notified when elements managed by the Virtual DOM are created, updated or removed via lifecycle events. Use them for animation, data fetching, wrapping third party libraries, cleaning up resources, etc.
--- a/test/inflation.test.js
+++ b/test/inflation.test.js
@@ -1,0 +1,33 @@
+import { h, app } from "../src"
+
+beforeEach(() => {
+  document.body.innerHTML = ""
+})
+
+test("inflation of connected components", done => {
+  const state = {
+    value: 1
+  }
+
+  const actions = {}
+
+  const view = state =>
+    h("a", {}, (state, actions) =>
+      h("b", {}, [
+        state.value,
+        (state, actions) =>
+          h(
+            "c",
+            {
+              oncreate() {
+                expect(document.body.innerHTML).toBe("<a><b>1<c>1</c></b></a>")
+                done()
+              }
+            },
+            state.value
+          )
+      ])
+    )
+
+  app(state, actions, view, document.body)
+})


SkaterDad:
Very interesting idea!  Will take a closer look at this later. 

While the framework size increases, I wonder if this would lower the actual application bundles in some cases.

Swizz:
My question at this stage is more pragmatic. What is the benefits of Connected Components regarding of the explicit use of state and actions properties ?
```jsx
const Component = (props, children) => (
  <div>Hello, {props.state.name}</div>
)

  <div>
    <Component state={state} actions={actions} />
  </div>
)
```

Yes, you will need to add `state={state} actions={actions}` to each tree step. 
Why is this a problem for you ? Only about DRY ?

JorgeBucaran:
@jason-whitted I would prefer if people created userland solutions for this like you did with hyperapp-connect.

etalisoft:
> Hyperapp applications consist of three interconnected parts: the State, View, and Actions.

From the moment of instantiation an app binds these pieces together.  The view is composed of a hierarchy of components.  Every instance of each component in the app is directly tied to the state and actions; however, currently they do not have native access to them.

Due to this limitation it forces the parent component to be concerned with the business of the child component.  This causes tight coupling and a violation of separation of concerns.  To bypass this, component developers will be forced to relay action and state down through every component they create to every child.

@SkaterDad 
> My big concern with this approach is the performance hit. We're already traversing the entire view before this new inflate function is called, so you're effectively doubling the work on each render. If this idea has traction, maybe there's a way to accomplish this within the 1st iteration?

I would love to do handle this in the 1st iteration, but my unfamiliarity with the codebase is slowing down progress.  Could you point me in the right direction?  It appears to me that the view does one of two things:

1. Simply returns an object map (without the use of `h`), in which case the `inflate` call is the first iteration.
2. Invokes a nested chain of `h` functions, which returns an object map.  Since the `h` functions are not bound to the `app` instance it cannot carry the state/actions down to the object it creates.

I would assume the `patch` function then could be considered the 1st iteration.  Would this be a better place to put the inflation logic?

@JorgeBucaran 
> I would prefer if people created userland solutions for this like you did with hyperapp-connect.

My "userland solution" is a hack at best.  You even had questions about why the key was needed.  Requiring a HOA and an HOC and the cognitive load required to do a common task goes against one of the main points of Hyperapp.
> We have aggressively minimized the concepts you need to understand
> Out of the box...all with no dependencies

etalisoft:
I have updated the code to no longer prematurely recursively inflate before the patch process.  It now happens during the patch process.

okwolf:
Why the name `inflate` exactly? 🤔

etalisoft:
A node has children.  Some of the children may be functions instead of other nodes.  The inflate function is converting them.  Hydrate was already used.  Inflate sounded good.  Haha, have a better name?

okwolf:
> @etalisoft A node has children. Some of the children may be functions instead of other nodes. The inflate function is converting them. Hydrate was already used. Inflate sounded good. Haha, have a better name?

Not sure, maybe `renderChildren`? 🤷‍♂️

etalisoft:
Or `connect`?  I don't know why I didn't just use that begin with.

JorgeBucaran:
I have nothing against the name. It reminded me of this from one of my favorite childhood games:



@etalisoft Can you try implementing this touching only `h`?

etalisoft:
I renamed `inflate` to `connect`.  I also found that converted the `for (var i = 0; i < array.length; i++)` to `for (var i in array)` I could cut out 20 bytes.  Which means this patch only has a net gain of 10 bytes.  ;)

etalisoft:
@JorgeBucaran 
> Can you try implementing this touching only `h`?

I thought long and hard on trying to do that.  It was the first logical place I tried.  I touched on the difficulties previously:

> Since the `h` functions are not bound to the app instance it cannot carry the state/actions down to the object it creates.

I'm not sure how I could accomplish this, since the `h` function isn't derived from the `app` instance.

src/index.js
```diff
@@ -90,7 +98,7 @@ export function app(state, actions, view, container) {
   }
 
   function get(path, source) {
-    for (var i = 0; i < path.length; i++) {
+    for (var i in path) {
```
JorgeBucaran: Why are you touching this? @etalisoft

JorgeBucaran:
@etalisoft It's a tough one. What else can we do to mitigate the performance hit?

src/index.js
etalisoft: @JorgeBucaran 
Given the focus on the end product file size, I was looking for ways to make up for some of the bytes that I added.  Obviously this can get reverted as it has no impact on the Connected Component code.

```js
var arr1 = ['a'];
arr1[100] = 'b';
for (var i = 0; i < arr1.length; i++) 
	console.log('arr1', i);
// outputs 100 times
  
var arr2 = ['a'];
arr2[100] = 'b';
for (var i in arr2)
	console.log('arr2', i);
// outputs twice
```

Based on the above logic, in addition to savings file size, it seemed it could have the added benefit of possibly reducing iterations.

JorgeBucaran: Let's implement it first, then worry about file size. Please revert it.

JorgeBucaran:
```js
arr2[100] = 'b';
```
Haha yeah, but that **can't** happen and regular for loops are faster than slow foreach loops.

etalisoft:
@JorgeBucaran Is there some kind of performance load test that the team has been using?  Maybe something that renders thousands of children and fires state changes.  If not, maybe it's the next thing I build :)

I'd like to compare the before and after performance.  I think the performance impact will be extremely minimal.

JorgeBucaran:
@etalisoft You are iterating over children twice every time.

etalisoft:
Not any more! 😉

JorgeBucaran:
@etalisoft You forgot to revert the <samp>for..in</samp> to <samp>for</samp>.

etalisoft:
@JorgeBucaran Oh yeah, sorry about that.  Done.

src/index.js
```diff
@@ -167,8 +172,10 @@ export function app(state, actions, view, container) {
         })
       }
 
-      for (var i = 0; i < node.children.length; i++) {
-        element.appendChild(createElement(node.children[i], isSVG))
+      var c = node.children
+      for (var i in c) {
```
JorgeBucaran: @etalisoft <samp>for..in</samp>

src/index.js
```diff
@@ -167,8 +172,10 @@ export function app(state, actions, view, container) {
         })
       }
 
-      for (var i = 0; i < node.children.length; i++) {
-        element.appendChild(createElement(node.children[i], isSVG))
+      var c = node.children
```
JorgeBucaran: <samp>var children = node.children</samp> 🙏

src/index.js
```diff
@@ -149,6 +149,11 @@ export function app(state, actions, view, container) {
     }
   }
 
+  function connect(children, i) {
+    typeof children[i] === "function" &&
```
JorgeBucaran: @etalisoft Can you please use a regular <samp>if</samp> clause?

src/index.js
```diff
@@ -260,7 +267,9 @@ export function app(state, actions, view, container) {
       var i = 0
       var j = 0
 
-      while (j < node.children.length) {
+      var c = node.children
```
JorgeBucaran: @etalisoft <samp>var children = node.children</samp>

src/index.js
```diff
@@ -149,6 +149,11 @@ export function app(state, actions, view, container) {
     }
   }
 
+  function connect(children, i) {
+    if (typeof children[i] === "function")
+      children[i] = children[i](globalState, wiredActions)
```
JorgeBucaran: @etalisoft This made me wonder, what if we make all components lazy by default?

JorgeBucaran:
@etalisoft What if we make all components lazy? Just wondering. 😏

etalisoft:
@JorgeBucaran Oh man, let me see if I can wrap my noodle around this.  😰

etalisoft:
@JorgeBucaran Here's what I've come up with so far: etalisoft/hyperapp@e1f378da8dd61b3ecb61ca7051350ea6105abc9e
I put this experiment in another branch, since these changes obviously break a few of the test cases.

zaceno:
I think this PR is a step in a good direction!

But personally I'd prefer something more flexible, along the lines of `hyperapp-context` (...which no longer requires a special `Context` component, btw). 

If we agree that the word "context" means: "a branch of your component tree, in which certain data is available to components without needing to pass it to them explicitly as props"  ... then this PR basically makes a context for your entire app tree with the state/actions in it.

But it *doesn't* allow you to add anything else to this context, and doesn't allow defining more specific sub-contexts for particular branches of your component tree. 

One reason for wanting to enrich the context would be for values computed from the state. If there is some type of computed value that depends on multiple values in the state, and many components need that value, then they all have to call the computation function in them. Having the value computed once and made available through context would be great in that case.

One reason for wanting to define sub-contexts, might be, because an entire branch of your component tree depends on a particular slice/namespace of your state & actions. Then, rather than have to use the namespace (`state.foo.prop`, `actions.foo.doThing`, ...) everywhere in the branch, it would be nice to be able to just write `state.prop`, `actions.doThing`. Since everything in those components is meant to be in the *context* of `foo`.

Basically, having the ability to define & enrich subcontexts can decouple your view or branches of it from how you chose to structure your state and actions. You may not want to make your components completely generic and reusable, but perhaps you'd like to make certain sets of components generic for use together. 

So what I'm suggesting is to also allow a sort of middle-ground between this PR and the current behavior of hyperapp's components.

etalisoft:
@zaceno
> this PR basically makes a context for your entire app tree with the state/actions in it.

Absolutely!

> But it doesn't allow you to add anything else to this context

It's your state, you can add anything you want to it.

> One reason for wanting to enrich the context would be for values computed from the state. If there is some type of computed value that depends on multiple values in the state, and many components need that value, then they all have to call the computation function in them.

A memoized selector library, like reselect, already adds this functionality.  They are memoized so long as the state doesn't change, which means they are reusable, composable, and only calculate the value once.  The selectors can even return components.

Here's an example I created:  https://codepen.io/jason-whitted/pen/WMMeWL

zaceno:
@etalisoft 
> > But it doesn't allow you to add anything else to this context

> It's your state, you can add anything you want to it.

I think we're talking about different things. What I was talking about is more about line 48 in this codepen: https://codepen.io/zaceno/pen/gvGgQP?editors=0010 

I set the index of a todo-item in the main Item component, so the index doesn't have to be passed along as a prop to all the sub-components where it's eventually used. The index is a value that changes during the buildup of the virtual dom tree -- not something that belongs (or could reasonably be put in) the state.

I know, one can question the point of using context at all for such a trivial thing. But it is, after all a technique meant for big complex apps, demonstrated using a very simple app.

zaceno:
@etalisoft 👍 For the reselect example.

etalisoft:
@zaceno I definitely get what you are describing.  It's very similar to how block scope works. 
```js
{ // app
   var index = 123;
    { // List1
        console.log(index);  // 123
    }
    { // List2
        var index = 321;  // this does not change app's index
        { // List 2 item
            console.log(index); // 321
        }
    }
}
```
Scoped context will require the generation and maintenance of scope while traversing the child nodes.  Additionally context is not state or actions, so it sounds like what you are requesting is an additional object that gets passed around.

Something like:
```js
const Component = (props, children, context) => ...
```

Which is similar how React handled context (at least until they re-engineered it).  Parent components make context available with `getChildContext()` and even the stateless functional components had the context sent as `(props, context) => ...`.

These concepts are unrelated to this PR.  The app has a single instance of state and a single source of actions per render.  This change is only to allow child nodes to have access to the app's state / actions.

etalisoft:
@zaceno It seems like you could update:
```js
// hyperapp-connect/src/index.js, line 6
node(context, function (props) { context = Object.assign({}, context, props)}),
```
If you pass `state`, `actions`, and then `context` it would still work wouldn't it?  Then your components would be connected and have context.

lukejacksonn:
I see this proposal reducing a lot of boilerplate in hyperapps. We know one of the biggest flaws with the current arch is that passing state and actions down the view can become very verbose.

If I understand this implementation correctly, it would allow any view at any any level in the vdom tree to access the global state and actions. This is similar to the notion of "pre-wired views" that have been discussed. Except that this is much more general (no involvement of slices). Presumably _connected_ views are expected to destructure what they require from the global state/actions objects like the root view. This encourages a flatter state structure and a better separation between view and model structure (you won't try appropriate the shape of either one to fit the other).

So what are the downsides? Besides +10 bytes or so.

Also, lazy views.. what do we mean by this? What doors does it open?

JorgeBucaran:
> @zaceno Having the global state & actions available everywhere is certainly useful. It does, however, make your views dependent on the shape of your entire state/action tree. You can't make views that depend only on a particular slice of state/actions. So if you want to decompose/modularize your app in that way, this doesn't help.

Exactly. This and the fact we went with slices (instead of no slices as we used to be) makes me doubt this will work for Hyperapp.

jason-whitted:
The view **_IS_** bound to state and actions.  And since the view is composed of components those components are inherently also bound to the entire state and actions -- even if you do not explicitly pass them down.

Providing only slices down to components and actions is extremely limiting.  The choices are then to either:
- Hack together a userland solution (which would be inferior to having integration built into the library)
- Explicitly pass down the entire state/actions to every single component (which is just avoiding slices altogether)

Redux, in a limited fashion, works as slices.  A rootReducer is composed of individual reducers.  The reducers only have access to their own 'slice' in the state.  However when you create a Redux store you receive an object with two functions:
- `getState()` - returns the global state
- `dispatch()` - allows you to execute actions globally
You can connect any component to the state globally.

Maybe the solution is to have the `app` function return `{ getState, getActions }` or something similar.  At least this way a component could get access to these functions.

Sure the users could individually create their own HOA function that stores the `state` and `actions` into an object for use elsewhere.  However hyperapp touts itself as being `Minimal` and `Batteries Included`, yet this exclusion forces the developers to: learn how to write their own Higher-Order Application, include an external library to do it for them, or to avoid slices altogether by explicitly passing down props to every component they create.

JorgeBucaran:
> @jason-whitted Maybe the solution is to have the app function return { getState, getActions } or something similar. At least this way a component could get access to these functions.

Are you aware that the <samp>app</samp> returns your wired actions?

jason-whitted:
Yes, of course I am.  Which is half of the needed response from `app`.  

The `getActions` call currently could be replaced with just `actions`, which is what the `app` call already returns.  However, #533 would suggest that there is a possibly that the actions could be mutated within the lifetime of the app, which would mean you need a way to get an updated list of actions.

JorgeBucaran:
@jason-whitted If the comparison you draw with Redux it's 100% accurate, then I don't understand how they deal with implicit dependencies, which is why I am "opposed" to sharing the global state & actions with a slice of the actions tree.

jason-whitted:
> @JorgeBucaran I don't understand how they deal with implicit dependencies

React-Redux uses a `Provider` component, which uses React's context to share state to all of its child components.  The `connect` HOC reads from the `Provider`'s context and binds a component to the state/actions.

Since Hyerpapp doesn't have context, some other way is needed to pass around the state/actions.  The app only provides the state to the initial view.  Which means it becomes the responsibility of EVERY child component to pass that state around if/when a component needs access to parts of the state that it's direct parent did not have access to.

Mytrill:
In redux, the connect() function accept a pure function to create the state transmitted from the global state to the component. This function can be reused and allow components to operate on the relevant state slices. Maybe we could think of a similar system? Maybe a path (or an array of paths) could be used when connecting the component or something?

jason-whitted:
@Mytrill Technically `connect` takes two arguments, a pure function used to map state and also on object containing actions to bind.

One could technically do:
```js
const * as actions from './actions'; // all unbound actions
const state = o => o;
const ConnectedComponent = connect(state, actions)(Component);
```

This would give the entire state and all bound actions to the component.

zaceno:
Since context was brought up again (not my fault this time! 😅 ): If Redux's connect function is based on React's context, wouldn't it make sense to have context in hyperapp, and solve the connected-components goal through that?

PS
The only reason I didn't PR my HOA yet is because I'm like 99.9% sure it will be rejected 😄: 
> @JorgeBucaran I would prefer if people created userland solutions for this like you did with hyperapp-connect.

jason-whitted:
@zaceno @JorgeBucaran 
I was just playing around with the entire context idea, just to be able to compare and contrast.

I created a modified version of `app` that provides context as a third parameter to components.
Components then have the signature:
```jsx
const Component = (props, children, context) => <div>{children}</div>;
```

The context is just a plain object.  It is cloned from the parent object.  If a component wants to add something to the context for it's children they just do it in the function body.

```jsx
const Parent = (props, children, context) => {
    context.apiKey = "abc123";
    return <div>{children}</div>;
}
const NestedChild = (props, children, context) => <div>{context.apiKey}</div>;
```

With this ability I then created a `Provider` and `connect` that works almost identical to `react-redux`.

If you want to take a look, I put it on [CodePen](https://codepen.io/jason-whitted/pen/vdrEgR).

This is obviously a completely different solution.
```jsx
// `current` solution
const Component = (props, children, { getState, getActions }) => ...

// `context` solution
const Component = (props, children, context) => ...
```

The `current` solution should have 0 performance impact.  Since the `current` object is generated and frozen once during the initial app mounting and then just past around.

The `context` solution is extremely flexible, and it pretty much works the same way as React.  Unfortunately, this does not come for free since the context has to be cloned from the parent.

JorgeBucaran:
If in order for this to exist in userland without severely patching the <samp>app</samp> function we need a similar concept like context in React, that's a course I would be interested in exploring too. We can't just add _all the_ features, but we can add something new that opens the door to new, exciting and generalized possibilities.

zaceno:
@etalisoft 
> The context solution is extremely flexible, and it pretty much works the same way as React.  Unfortunately, this does not come for free since the context has to be cloned from the parent.

Nice solution for context! And you're right -- there is a penalty to it.

Not sure if you're deep-cloning or shallow, but I think shallow should be enough and the impact should be negligible. Still: it's there. ..and would annoy anyone who doesn't want/use it.

Your connected-components approach is essentially free, and as such "harmless" to those who don't care about it. On the other hand: not as flexible as context.

frenzzy:
I agree that hyperapp must have a way to pass things (global state/actions/context/etc) down to components without the need to do it manually (using props through the whole tree). I also agree with @JorgeBucaran that core should not implement all the features but just allow you to extend it in userland.

In my mind we have two ways how to achieve this (an ability to communicate parent and child components without modifying components tree in the middle):
1. Context object
  For example via component signature:
  ```js
  const Component = (attributes, children, context) => <div />
  // or
  const Component = (attributes, children) => context => <div />
  ```
  But I do not see a way how to make it explicit and simple to understand/use for users. Example above with `context.apiKey = "abc123"` looks very ugly/untestable/etc.
2. Global actions
  For example with one of the following component signatures:
  ```js
  const Component = (attributes, children, actions) => <div />
  const Component = (attributes, children) => (actions) => <div />
  const Component = (attributes, children, state, actions) => <div />
  const Component = (attributes, children) => (state, actions) => <div />
  ```
  By using global actions you can solve the same problems which context solves. But actions already described by hyperapp and you know how to use them. And you also can implement your own mutable context based on actions or anything else.

So, IMO we should approve only something like #604 to be in the core.

zaceno:
@frenzzy And you also can implement your own mutable context based on actions or anything else.

This is not exactly true. Context isn't necessarily about the state, but about *where in the component tree a component is being rendered* -- not something you can keep in state, or use actions to update. Because the second you use an action... boop! The tree is being rendered again.

Mytrill:
@frenzzy Have you checked out [hyperapp-context](https://github.com/HyperappCommunity/hyperapp-context)? Would you agree that such an API would be an elegant solution for 1. that has much more possibilities than 2.?

frenzzy:
@zaceno under the hood context is just an object which is available in all components. You can add any features to it to behave as you want. For example on a component level you can detect where you are and add this into to context.

Context implementation example using global action:
```js
const context = { key: 'value' }

const state = {}
const actions = { getContext: () => context }

const ChildComponent = () => (state, actions) => {
  const context = actions.getContext()
  return <div>context: {context.key}</div>
}
const ParentComponent = () => <div><ChildComponent /></div>
const view = (state, actions) => <main><ParentComponent /></main>

```
or you may use higher order component:
```js
const withContext = (component) => (attributes, children) => (state, actions) =>
  component(attributes, children, actions.getContext())

const YourComponent = withContext((attributes, children, context) => <div />)
```

Mytrill:
@frenzzy As far as I understand, in hypperapp-context's case, any value set on the context is only available to the children of this component, not to the entire tree, as such I don't think you can (easily) implement it with an action/state slice.

frenzzy:
@Mytrill it is solvable with a knowledge about component executing order. For example how implementation of [new React.js context API](https://medium.com/dailyjs/reacts-%EF%B8%8F-new-context-api-70c9fe01596b) may look in Hyperapp:
```js
function createContext(defaultValue) {
  let value = defaultValue
  return {
    Provider(attributes, children) {
      value = attributes.value
      return children
    },
    Consumer(attributes, children) {
      return attributes.render(value)
    }
  }
}

const ThemeContext = createContext('light')

const view = (state, actions) =>
  <main>
    <ThemeContext.Provider value="dark">
      <div>
        <ThemeContext.Consumer render={val => <div>{val}</div>} />
      </div>
    </ThemeContext.Provider>
  </main>

app({}, {}, view, document.body)
```
Demo: https://codepen.io/frenzzy/pen/oEyGeJ?editors=0010

It does not work right now just because component execution order must be top to bottom, but currently components are executed in the opposite order. Lazy components PR #606 will solve this problem.

zaceno:
@frenzzy Yes, that implementation looks more workable. It works because it does not use state/actions to read & create context. That's the only point I was making (and I think @Mytrill ).

Like you say, it requires "lazy components". Actually, your implementation of context, when you include the lazy-components aspect, is technically pretty much how my `hyperapp-context` hoa works -- except the API is totally different of course 😄

jason-whitted:
@JorgeBucaran @frenzzy @Mytrill @zaceno 

I took @frenzzy's `createContext` idea:
```jsx
// Could be the output of a hyperapp-context package
const createContext = defaultValue => {
  let value = defaultValue;
  return {
    Provider(attributes, children) {
      value = attributes;
      return <div>{children}</div>;
    },
    Consumer(attributes, children) {
      return attributes.render(value)
    },
  };
};
```

Created `Provider` and `connect` using the context:
```jsx
// Could be the output of a hyperapp-connect package
const { Provider, Consumer } = createContext({});
const connect = (mapStateToProps, mapActionsToProps) => BaseComponent =>
  (props, children) => {
    const render = ({ state, actions }) =>
      <BaseComponent {...props} {...mapStateToProps(state)} {...mapActionsToProps(actions)} />;
    return <Consumer render={render} />;
  };
```

The end result is context AND connected components -- and the only thing that needs to be added to the hyperapp core library is #606 (@frenzzy's Lazy Components) and #611 (@titouancreach's fix minify).

```jsx
import { Provider, connect } from 'hyerpapp-connect'; // uses hyperapp-context internally
const state = { value: 0 };
const actions = {
  value: value => state => ({ value }),
};

const Component = (props, children) => <div />;
const mapStateToProps = state => ({ value: state.value });
const mapActionsToProps = actions => ({ setValue: actions.value });
const ConnectedComponent = connect(mapStateToProps, mapActionsToProps)(Component);
// The ConnectedComponent would receive the following props:
//   name: string
//   value: number
//   setValue: function

  <Provider state={state} actions={actions}>
    <ConnectedComponent name="bob" />
  </Provider>
);
```

Here's a working [CodePen](https://codepen.io/jason-whitted/pen/xYzQYK).

JorgeBucaran:
> @etalisoft The end result is context AND connected components -- and the only thing that needs to be added to the hyperapp core library is #606 (@frenzzy's Lazy Components) and #611 (@titouancreach's fix minify).

Meaning this PR is no longer relevant? Or did I misunderstand what you said?

And why #611? What difference did it make?

etalisoft:
> @JorgeBucaran this PR is no longer relevant?
Correct, if lazy comments gets added then this PR is not necessary. 

> why #611?
The minified code was not properly relaying props down to child nodes. The unminified code did not have this problem. When I incorporated #611 the minified code worked properly.

JorgeBucaran:
> Correct, if lazy comments gets added then this PR is not necessary.

I know someone who loves lazy comments.

okwolf:
> @etalisoft Correct, if lazy comments gets added then this PR is not necessary.

comments === components?

etalisoft:
> @JorgeBucaran lazy comments

Ha 😃 That’s what I get for typing on my phone. 

Additionally, lazy components makes #615 irrelevant also

okwolf:
I think we are fast approaching a major decision point for this project. Right now we have a pretty equal compromise between the simple/functional/verbose and complex/imperative/performance/pragmatic sides. If we choose to support connected components that would be a shift toward the complex side in order to make certain things easier and more like React with somewhat stateful components.

My vote is to keep things simple in core, unless we can come up with another approach that is simple made easy 👌

JorgeBucaran:
@okwolf What about lazy **components**? That closes #604 and #615.

okwolf:
> @JorgeBucaran What about lazy components? That closes #604 and #615.

That approach is more flexible and allows for the community to decide how to wire their components. Maybe some awesome approach (à la selectors) will emerge. Probably it's something no one has even thought of yet.

etalisoft:
@okwolf Selectors, like reselect?  If so, I’ve already codepenned that in one of these threads.  They already work and pair perfectly with Provider/connect model listed earlier.

okwolf:
> @etalisoft Selectors, like reselect? If so, I’ve already codepenned that in one of these threads. They already work and pair perfectly with Provider/connect model listed earlier.

Conceptually, yes. I'd love to see an approach that improves on the approach of Reselect by integrating more tightly with the VDOM engine. This would give some of the same ergonomic improvements as the shorthand syntax for the state updates has over reducers.

titouancreach:
> @etalisoft  The minified code was not properly relaying props down to child nodes. The unminified code did not have this problem. When I incorporated #611 the minified code worked properly.

Wow, just in time :+1:

JorgeBucaran:
@etalisoft I think we can close here then?

JorgeBucaran:
Closed in favor of #606.

Issue #564, opened 2018-01-22
Compile JSX right into virtual node objects avoiding `h` function


frenzzy:
There are an idea to avoid using [`h`](https://github.com/hyperapp/hyperapp/blob/1.0.2/docs/concepts/vnodes.md) in favor of compilation step which will do the same work.
_Compile JSX:_
```js
const view = (state, actions) => (
  <main>
    <h1>{state.count}</h1>
    <button onclick={actions.down}>-</button>
    <button onclick={actions.up}>+</button>
  </main>
)
```
_to:_
```js
const view = (state, actions) => ({
  name: 'main',
  props: {},
  children: [
    { name: 'h1', props: {}, children: [state.count] },
    { name: 'button', props: { onclick: actions.down }, children: ['-'] },
    { name: 'button', props: { onclick: actions.up }, children: ['+'] }
  ]
})
```
Potentially the app should work faster this way because each function call requires additional computing resources. Also `h` which is still necessary for JSX haters could be tree shakied from your build.

Possible future improvements: adopt [babel-plugin-transform-react-constant-elements](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-react-constant-elements) to speedup the app even more:
```js
const h1 = { name: 'h1', props: {}, children: ['Static content'] }
const view = (state, actions) => h1
```

Can anyone foresee potential pitfalls?

vdsabev:
I started some work on a `jsx-to-object` compiler in https://github.com/vdsabev/jsx-to-object/pull/1/files#diff-1dd241c4cd3fd1dd89c570cee98b79dd but got held up with an Esprima issue parsing the `...` spread operator.

If anyone has ideas or suggestions on an alternative parser that supports JSX, I'd be happy to drop Esprima and use something else instead 😅

frenzzy:
@vdsabev have you tried [babylon](https://github.com/babel/babel/tree/master/packages/babylon) parser? Or could it be implemented as a babel plugin?

vdsabev:
Great suggestions, I'll try when I get the time!

I started with acorn first (on which babylon is based), but decided to switch to esprima (don't remember why anymore). From what I've read recently, babylon supports all the use cases I need, and has even had contributions from the TypeScript team.

I might write a babel plugin on top of `jsx-to-object`, the goal is to have it as a standalone utility that you can use from a CLI, gulp, whatever, so as to not strictly depend on babel. I don't have experience doing anything like that, so it'll be a learning opportunity.

[...]

Issue #499, opened 2017-12-17
Rewrite patch algo / improve diffing performance


jorgebucaran:
I want to rewrite patch to incorporate some prefix/suffix trimming techniques as described in the first link posted below.

Hyperapp unkeyed and keyed fares relatively well according to [js-framework-benchmarks](https://github.com/krausest/js-framework-benchmark) (arguably the best benchmark out there and by far), but we could do a bit better and still get away with our 1 KB proposition.

- https://github.com/hyperapp/hyperapp/issues/216#issuecomment-314051357
- https://github.com/yelouafi/petit-dom/blob/master/src/vdom.js

leeoniya:
there's a non-zero chance that we'll absorb uibench's 4 "worst case" scenarios [1]. prefix/suffix checks will help some of these, but not all.

all of uibench is actually a great testing/perf regression tool.

[1] https://github.com/localvoid/uibench-base/blob/efacae672bbf4133360a928305876ee1de643e64/lib/uibench.ts#L302-L319

JorgeBucaran:
@leeoniya Can you elaborate? What am I looking at? 😄

leeoniya:
the public page is here: https://localvoid.github.io/uibench/

what the linked code does should be fairly obvious from the function names.

JorgeBucaran:
@leeoniya Thanks. I've merged your comments into the same post. I don't want to derail this issue and if possible keep the convo on point.

okwolf:
The core VDOM (Picodom) looks pretty close to vanilla in the [latest js-framework-benchmark](http://www.stefankrause.net/js-frameworks-benchmark7/table.html).

What I find really interesting is where Hyperapp underperforms Picodom:

![screen shot 2017-12-16 at 23 11 13](https://user-images.githubusercontent.com/3735164/34077339-94742cde-e2b6-11e7-9336-d12ec1d4a8af.png)

By a factor of 3x in some cases! 😮

JorgeBucaran:
@okwolf That's a different library: https://github.com/hville/pico-dom, the picodom I wrote is not on that list.

okwolf:
@JorgeBucaran oh. I guess there can never be too many libraries with similar names 😕 

Well, I think comparing the performance of the _real_ Picodom library with Hyperapp might still be a worthwhile experiment.

okwolf:
@JorgeBucaran have you done any more research on this topic yet?

[...]

PR #606, opened 2018-02-18
Lazy components

frenzzy:
- Lazy execute components at runtime, probably it may save computation in some cases
- It will be possible to easily extend component with `connect` proposal, ref #604 
- It will be possible to compile JSX right into virtual node objects avoiding `h` function with components support, ref #564

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -16,14 +16,12 @@ export function h(name, attributes /*, ...rest*/) {
     }
   }
 
-  return typeof name === "function"
-    ? name(attributes || {}, children)
-    : {
-        nodeName: name,
-        attributes: attributes || {},
-        children: children,
-        key: attributes && attributes.key
-      }
+  return {
+    nodeName: name,
+    attributes: attributes || {},
+    children: children,
+    key: attributes && attributes.key
+  }
 }
 
 export function app(state, actions, view, container) {
@@ -51,12 +49,23 @@ export function app(state, actions, view, container) {
     }
   }
 
+  function getVNode(node) {
+    return typeof node.nodeName === "function"
+      ? getVNode(node.nodeName(node.attributes, node.children))
+      : node
+  }
+
   function render() {
     renderLock = !renderLock
 
     var next = view(globalState, wiredActions)
     if (container && !renderLock) {
-      rootElement = patch(container, rootElement, oldNode, (oldNode = next))
+      rootElement = patch(
+        container,
+        rootElement,
+        oldNode,
+        (oldNode = getVNode(next))
+      )
       firstRender = false
     }
 
@@ -168,7 +177,9 @@ export function app(state, actions, view, container) {
       }
 
       for (var i = 0; i < node.children.length; i++) {
-        element.appendChild(createElement(node.children[i], isSVG))
+        element.appendChild(
+          createElement((node.children[i] = getVNode(node.children[i])), isSVG)
+        )
       }
 
       for (var name in node.attributes) {
@@ -262,7 +273,7 @@ export function app(state, actions, view, container) {
 
       while (j < node.children.length) {
         var oldChild = oldNode.children[i]
-        var newChild = node.children[j]
+        var newChild = (node.children[j] = getVNode(node.children[j]))
 
         var oldKey = getKey(oldChild)
         var newKey = getKey(newChild)
--- a/test/h.test.js
+++ b/test/h.test.js
@@ -79,35 +79,12 @@ test("skip null and Boolean children", () => {
   expect(h("div", {}, null)).toEqual(expected)
 })
 
-test("components", () => {
-  const Component = (props, children) => h("div", props, children)
+test("component", () => {
+  const Component = () => h("div", {}, ["baz"])
 
   expect(h(Component, { id: "foo" }, "bar")).toEqual({
-    nodeName: "div",
+    nodeName: Component,
     attributes: { id: "foo" },
     children: ["bar"]
-  })
-
-  expect(h(Component, { id: "foo" }, [h(Component, { id: "bar" })])).toEqual({
-    nodeName: "div",
-    attributes: { id: "foo" },
-    children: [
-      {
-        nodeName: "div",
-        attributes: { id: "bar" },
-        children: []
-      }
-    ]
-  })
-})
-
-test("component with no props adds default props", () => {
-  const Component = ({ name = "world" }, children) =>
-    h("div", {}, "Hello " + name)
-
-  expect(h(Component)).toEqual({
-    nodeName: "div",
-    attributes: {},
-    children: ["Hello world"]
   })
 })
--- a/test/components.test.js
+++ b/test/components.test.js
@@ -1,0 +1,60 @@
+import { h, app } from "../src"
+
+beforeEach(() => {
+  document.body.innerHTML = ""
+})
+
+test("components", done => {
+  const Component = (props, children) => h("div", props, children)
+
+  const view = () =>
+    h(
+      Component,
+      {
+        oncreate() {
+          expect(document.body.innerHTML).toBe("<div>foo<div>bar</div></div>")
+          done()
+        }
+      },
+      ["foo", h(Component, {}, "bar")]
+    )
+
+  app({}, {}, view, document.body)
+})
+
+test("nested components", done => {
+  const Child = (props, children) => h("div", props, children)
+  const Parent = (props, children) => h(Child, props, children)
+
+  const view = () =>
+    h(
+      Parent,
+      {
+        oncreate() {
+          expect(document.body.innerHTML).toBe("<div>foo</div>")
+          done()
+        }
+      },
+      ["foo"]
+    )
+
+  app({}, {}, view, document.body)
+})
+
+test("component with no props adds default props", done => {
+  const Component = ({ name = "world" }, children) =>
+    h(
+      "div",
+      {
+        oncreate() {
+          expect(document.body.innerHTML).toBe("<div>Hello world</div>")
+          done()
+        }
+      },
+      "Hello " + name
+    )
+
+  const view = () => h(Component)
+
+  app({}, {}, view, document.body)
+})


okwolf:
I like the idea, but wouldn't this have a negative impact on testing your app? You wouldn't be able to assert on any VDOM subtree that used a component since it would just be the function and props instead of expanding into the resulting VDOM tree?

frenzzy:
For tests we can use something similar to [react's test-renderer](https://reactjs.org/docs/test-renderer.html)
```js
import TestRenderer from 'hyperapp-test-renderer'

const testRenderer = TestRenderer.create(<Component />)
expect(testRenderer.toTree()).toBe(theWholeExpectedTreeHere)
```

test/h.test.js
```diff
@@ -67,47 +67,12 @@ test("vnode with attributes", () => {
   })
 })
 
-test("skip null and Boolean children", () => {
```
JorgeBucaran: @frenzzy Why did you remove this test? Please add it back.

frenzzy: This test was moved from `h.test.js` to `components.test.js` because the check for `null`, `true` and `false` was moved from `h` to runtime.

Swizz:
Interesting idea, I see on this feature an entry point to allow lifecycle event on Components too.

Do we know the performance impact ?

src/index.js
```diff
@@ -51,12 +49,27 @@ export function app(state, actions, view, container) {
     }
   }
 
+  function getVNode(node) {
+    if (node == null || typeof node === "boolean") {
+      return ""
+    }
+    if (typeof node.nodeName === "function") {
+      return getVNode(node.nodeName(node.attributes, node.children))
```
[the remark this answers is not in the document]
etalisoft: `(attributes, children, state, actions) => ...` was my first inclination on the component signature -- it's actually the signature that hyperapp-connect used.  This signature also makes it a little bit easier for userland HOA/HOC solution development.  That signature would obviate the need for #604.

JorgeBucaran:
Before I worry about the perf impact, I worry about how this changes the way we build apps, etc. I need to think about this a lot. 😅

SkaterDad:
This looks pretty neat.  Can someone post some pseudo-code of the new possibilities this creates?

Mytrill:
Is it me or would this PR allows to implement @zaceno 's [hyperapp-context](https://github.com/HyperappCommunity/hyperapp-context) right into core for no performance penalty?

zaceno:
@Mytrill 

There's an inescapable (but small, I think) performance hit from maintaining the context boundaries. 

Aside from that, yes! This PR should be more efficient than the "postprocessing" of the vtree I do in my HOA, because I basically have to walk through the entire tree one extra time before every render -- whereas here, it happens during the walkthrough we do anyway.

JorgeBucaran:
The way this is implemented conflicts with a relatively recent feature (it wasn't always like this, though), which is the "ability" to call actions within the view function (/cc @lukejacksonn). A good example of this can be seen in the router [here](https://github.com/HyperappCommunity/router/blob/48d81a0cca9ce0545905021b1cd78d9d4840a33d/src/Redirect.js#L3) where <samp>replaceState</samp> is called, which then calls an action thanks to this other code [here](https://github.com/HyperappCommunity/router/blob/48d81a0cca9ce0545905021b1cd78d9d4840a33d/src/location.js#L40).

zaceno:
> @JorgeBucaran The way this is implemented means that an action can be called before patch returns, which will definitely have unwanted side effects, as well as break the lifecycle callback stack.

Aha! Good point. That's definitely a good argument to process lazy nodes *before* patch, despite the penalty. (Actions in views is a much more important feature IMO). If only there were a way to have both... 🤔

JorgeBucaran:
@zaceno I think that's one of the basic points of React Fiber.

zaceno:
@JorgeBucaran interesting. I haven't looked into React Fiber yet (never looked at React's code at all actually). What was the trick that let them solve the problem (if you know) ?

JorgeBucaran:
@zaceno It's described more or less [here](https://github.com/acdlite/react-fiber-architecture#what-is-a-fiber). I haven't looked much at React's code either, that's why I made Hyperapp right? 😉 

---

If we want to introduce this idea of "lazy" components, we'll need to go back to not allow actions in the view. Or call this off.

zaceno:
@JorgeBucaran Thanks for the link! It gave me an idea: 

I wonder if we could get around the conflict between lazy components and calling actions in views, if we separate the process like how `virtual-dom` (and React Fiber, from the looks of it) does things, by first computing the diff to  apply (reconciliation), and then in a separate step applying the diff (the rendering). 🤔 

Even if that works... I realize it's quite a long way off. So I think yeah, for now, lets call off lazy components. Actions in views i way more important I think.

frenzzy:
Why do you need to call actions inside a view? Can you provide an example? I can't imagine an example when you need to use actions not from handlers or life cycle events.

Also locking action calls during patching is just a one boolean variable.

JorgeBucaran:
I am not a big fan of the idea either, but that's how we implemented the router. I would be happy to remove this feature and think of other ways to move forward. See [my comment here](https://github.com/hyperapp/hyperapp/pull/606#issuecomment-367609333).

zaceno:
@frenzzy Also fetching data on route, like here: https://github.com/lukejacksonn/hyperapp-fetch 

Correct me if I'm wrong, @JorgeBucaran, but this conflict extends beyond just calling actions *directly* in the view, but also to calling actions in lifecycle events, right?

JorgeBucaran:
> ...calling actions in lifecycle events, right?

No, why? 🤔

SkaterDad:
> A solution is going back to not allowing calling actions inside the view function.

That's a good thing :smile: .  I know this debate was had in the past, and I still 100% believe that keeping the view function "pure" without side effects has great benefits.   It's part of the reason I don't use the new router.

All of these PR's lately are a bit hard to follow.  Great to see so much enthusiasm!

zaceno:
@JorgeBucaran Sorry you're right ... disregard! 😅 

@SkaterDad I agree actions in views is inelegant. I too wish there was a better way. Now that I think more about it, I think I've only needed it for the classic: "fetch data on route" problem. If we can think of an alternative, that should be good.

One "solution" I won't accept, though, is to say "call the action in `oncreate` or `onupdate` instead". Because at that point we're already rendering a state I know I don't want. 

A related mechanism that I hope we don't lose in the future (for this or any reason) is the ability to call multiple actions in sequence without having to render after each.

SkaterDad:
> @zaceno A related mechanism that I hope we don't lose in the future (for this or any reason) is the ability to call multiple actions in sequence without having to render after each.

Agree!  I often call a few different actions within another one.   Particularly when doing async stuff.

> @zaceno One "solution" I won't accept, though, is to say "call the action in oncreate or onupdate instead". Because at that point we're already rendering a state I know I don't want.

I agree, those aren't great either.  In my mind, every action should be triggered by user input or browser events.

`<off-topic>`
<details>
My own routing implementation is similar to [redux-little-router](https://github.com/FormidableLabs/redux-little-router) and [redux-first-router](https://github.com/faceyspacey/redux-first-router).   There are [some articles](https://formidable.com/blog/2016/07/11/let-the-url-do-the-talking-part-1-the-pain-of-react-router-in-redux/) written by those library authors which go into great detail on why the `react-router` style isn't a good fit for a global state tree architecture or when you need SSR.

```
user click
-> router action begins
-> calls functions you provide for guarding/data fetching, providing the "new route" params/query/etc
-> update state
-> let the view figure out the rest
```

`</off-topic>`

JorgeBucaran:
@SkaterDad Can we see your router's source code? :)

SkaterDad:
@JorgeBucaran You bet.  When I get a chance I'll extract the current version into a Codesandbox.

src/index.js
```diff
@@ -11,19 +11,17 @@ export function h(name, attributes /*, ...rest*/) {
       for (length = node.length; length--; ) {
         rest.push(node[length])
       }
-    } else if (node != null && node !== true && node !== false) {
```
JorgeBucaran: @frenzzy Why would you remove this line here?

frenzzy: Component execution was moved away from `h` function but they also may return `null` and boolean values. We have to check the result of component call, so I added this check to `app` (`getVNode)`. And just to check the same thing twice I removed this line.

JorgeBucaran: @frenzzy But calling a component function will call `h`, so you could leave this here too.

frenzzy:
```js
// no h() call here
const Component = (attributes, children) => null
const Component = ({ condition }) => condition && <div />
const Component = ({ condition }) => condition ? <div /> : null
```
ref: [React: Preventing Component from Rendering](https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering)

JorgeBucaran: @frenzzy 

```js
h(Component)
```

frenzzy: `h(Component)` in this PR returns `{ nodeName: Component, attributes: {}, children: [] }`. Inside patch function I execute `Component` and replace VNode with the result.
```js
const Component = () => result

<div><Component /></div>
// =>
h('div', {}, [h(Component)])
// =>
{ nodeName: 'div', attributes: {}, children: [
  { nodeName: Component, attributes: {}, children: [] }
] }
// =>
{ nodeName: 'div', attributes: {}, children: [
  result
] }
```

okwolf:
> @frenzzy For tests we can use something similar to react's test-renderer

Is someone signing up for that support? I have the `hyperapp-test` package that's out of date, but I'd like to bring new life to some point soon.

JorgeBucaran:
@frenzzy What about returning arrays from components? 😉

JorgeBucaran:
I am benchmarking this code with js-framework-benchmarks and there is definitely a bug when removing elements somewhere.



😱

frenzzy:
> @frenzzy What about returning arrays from components? While this is currently possible, it seems this
PR removes that ability. 😉

Maybe introduce JSX Fragments for it?) In React an ability to return arrays from components appeared after introducing fragments...

frenzzy:
By the way returning arrays from components is just a possibility which is not officially supported because we don not have docs and tests for it. I am right?

JorgeBucaran:
@frenzzy @etalisoft @zaceno 

### Pros and cons of lazy components:

#### PROS
- ~~Better performance~~: I run this branch through js-framework-benchmarks and noticed actually slightly worse performance. It's not terrible, maybe it was a quantum fluctuation. It was disappointing we gained nothing. Now, this is, to some extent, to be expected, the bottleneck **is in the diff**, and rewriting the algo is the only way to score better in the benchmarks. #499 
- Connected components / context / etc. #604 . @etalisoft Could you explain how is this possible?

#### CONS
- **Can't** call actions within the view / component. You can use `oncreate` or `onupdate`, etc. as usual though.
- **Lose** the ability to return arrays from components. Solution: implement fragments. See https://github.com/hyperapp/hyperapp/pull/606#issuecomment-367933729.

selfup:
> Lose ability to call actions within the view / component. You can use oncreate or onupdate, etc. as usual though.

Yikes..

SkaterDad:
> Lose ability to call actions within the view / component. You can use oncreate or onupdate, etc. as usual though.

And user events like clicks and inputs.  Mentioning just in case people are scared by this comment. 

As I've said before, I think losing that ability  is a good thing. :+1

JorgeBucaran:
Sure, let me edit my post.

etalisoft:
> @JorgeBucaran Could you explain how is this possible?

Sorry in advance, this won't be a short post.  🙁 

# The Example App
Let's start off by declaring a simple closure that allows us `read`/`write` functions.  This function in itself doesn't have anything to do with hyperapp.
```js
const memory = () => {
  const values = {};
  return {
    write(key, value) { values[key] = value; },
    read(){ return values; }
  };
}
const { read, write } = memory();
// read() returns {}
// write('name', 'Bob');
// read() returns { name: 'Bob' }
```

Now let's create a simple app that uses the `read`/`write` functions internally.
```jsx
const DisplayMemory = () => <pre>{JSON.stringify(read(), null, 2)}</pre>;
const StoreMemory = ({ key, value }, children) => {
  write(key, value);
  return <div>{children}</div>;
};
const view = () => (
  <StoreMemory key="animal" value="dog">
    <DisplayMemory />
  </StoreMemory>
);
app({}, {}, view, document.body);
```

### Observing the `h` function lifecycle
In order to observe execution, I added some simple logging to the `h` function.
```js
function h(name, attributes /*, ...rest*/) {
  console.group('h', name.name || name);
  ...
  console.log(result);
  console.groupEnd();
  return result;
}
```

# Eager `h` function (current behavior)
The current `h` function returns:
```js
  ? name(attributes || {}, children)
  : {
      nodeName: name,
      attributes: attributes || {},
      children: children,
      key: attributes && attributes.key
    }
```
If `name` is a component function, the `h` function is eagerly invoking this function.  This means the child components are rendered before the parent components.

### Resulting DOM
```html
<div>
  <pre>{}</pre>
</div>
```

### Console Output
image

### Analysis
As you can see, the `h` function runs components from the inside out.  Since `DisplayMemory` was executed before `StoreMemory`, there was nothing to display.

Want to see it in action?  Here's the [CodePen](https://codepen.io/jason-whitted/pen/xYarpM).

# Lazy `h` function
The Lazy `h` function returns:
```js
return {
    nodeName: name,
    attributes: attributes || {},
    children: children,
    key: attributes && attributes.key
  }
```
If `name` is a component, the component is not executed.  Instead it defers this to the `app`, which can invoke the component within the context of the app.  This allows parent components to render the child components JIT.

### Resulting DOM
```html
<div>
  <pre>
    {
      "animal": "dog"
    }
  </pre>
</div>
```

### Console Output
image

### Analysis
Moving the component execution from the `h` function to the `app` function allows child components the ability to be executing within context of their parent.

Want to see it in action?  Here's the [CodePen](https://codepen.io/jason-whitted/pen/ZrMJxg).

etalisoft:
> @JorgeBucaran how does the current PR enable your "connected components" since there is no passing of global state or actions anywhere in this PR.

Lazy Components enable data to be shared between components via context.  In my example above the `DisplayMemory` and `StoreMemory` were sharing context managed by the `memory` closure.  

[#604 comment](https://github.com/hyperapp/hyperapp/pull/604#issuecomment-367755974) demonstrates how the `state` and `actions` can be passed around with context.

### Hypothetical hyperapp-context package
```js
import { createContext } from 'hyperapp-context';
const { ContextProvider, ContextConsumer } = createContext();
```
A `createContext` function could generate a `ContextProvider`/`ContextConsumer` pair that share data using closure.

### Hypothetical hyperapp-connect package
```js
import { Provider, connect } from 'hyperapp-connect';
```
Uses the ContextProvider/ContextConsumer in order to provide a `Provider` component and `connect` HOC.  The Provider component stores the `state`/`actions` in context using `ContextProvider`.  The `connect` HOC uses the `ContextConsumer` to read the `state` /`actions` and pass specified props to the connected component.

This [CodePen](https://codepen.io/jason-whitted/pen/xYzQYK) shows it all in action.

etalisoft:
@JorgeBucaran @frenzzy @zaceno 

In response to some Slack conversation that I missed:

> @zaceno It depends only on lazy components. And it whould work for multiple apps, wouldn’t it?
> @frenzzy Nope, because internal “value” variable must be different for each instance
> @zaceno You have `createContext` you could call for each instance, no?
> Oh… you’re right. How would you
> And technically, you’d need to call `createContext` for each render

I don't understand the limitation you two were referring to with regards to context.  I created a [CodePen](https://codepen.io/anon/pen/LQJgbz) with 3 apps.  App1 and App2 share context.  App 3 does not share.  All 3 apps exist on one page and are working as intended.

The only issue I ran into was getting the two apps to render when the other updated context.  In the end I created a `syncUpdate` function that causes a render on both apps.

zaceno:
@etalisoft Nice codepen! Cool use of the `render` attribute to get around the current lack of lazy components!

After some thinking, I believe you're right about multiple apps. If there is a `Provider`/`Consumer` pair of singletons, two app's should be able to import them without writing over eachother's context. 

However it's important that you only use a `Provider` once in the tree of your app. The problem I still see is maintaining "context boundaries".  If you need to create a more specialized context with more/refined data for a part of your app, you'd need to use a `Provider2`/`Consumer2` pair. 

That's a line of thinking I haven't been down, because I was always thinking in terms of having a single mechanism for setting/adding to the context, but only make it apply to the descendants of where the context is set. (That's how hyperapp-context works)

Simply having multiple providers gets around that problem. At the cost of having to export a pair of Provider/Consumer components for every branch of context-specialization you need.

frenzzy:
@etalisoft if you will use createContext pattern as it is in your example then it will introduce the most difficult to catch bugs that you have ever seen because of sharing memory.

Imagine node.js app, where each user request processed independently and for each request `Provider` component will be executed. Data will be written to the same `value` variable shared between all requests. And you can't control the order of write/read to this variable, its value become inconsistent. It may work fine for 999 of 1000 requests which makes it impossible to debug and catch the bug.

In reactjs they additionally use internal context object which is unique for each app instance to write/read data produced by `Provider`/`Consumer` type of components. See https://github.com/facebook/react/pull/11818

zaceno:
@JorgeBucaran  Well, as far as I understand the idea of a Provider, is that only Consumers that are *descendants* of the provider, should have access to what it provides. That is not the case in this implementation (or any implementation I've seen).

JorgeBucaran:
@zaceno Do you mean [this](https://github.com/hyperapp/hyperapp/pull/604#issuecomment-367755974)?

src/index.js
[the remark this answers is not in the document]
frenzzy: A crazy idea: what if we will create an internal context object:
```js
export function app(state, actions, view, container) {
  // ...
  var context = { state: globalState, actions: wiredActions }
  // ...
}
```
and will provide it to all components as `this`
```js
return getVNode(node.nodeName.call(context, node.attributes, node.children))
```
so, component signature will stay the same but for higher-order components we provide access to internal context which we can use for implementing `connect` or `Provider`/`Consumer` patterns..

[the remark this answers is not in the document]
JorgeBucaran: Nope.

Mytrill:
@SkaterDad @JorgeBucaran I have created a small snippet for the router where there is no action call in the view for state change: https://github.com/hyperstart/hyperapp-recipes/blob/2788ffa476d0dfc3dfe6644b8ebb7f94525a0a17/router/index.js#L31

Is this the kind of implementation you are using, SkaterDad?

zaceno:
> @JorgeBucaran We want lazy components, don't we? I am really excited about all the new room for activities once this is merged.

Not sure...

I'm not opposed. I've verified that hyperapp-context is still possible with lazy components, but at an additional small performance penalty.

Yet I don't really see how it benefits me either.

jorgebucaran:
@zaceno Possible without lazy components, but you need a `render` attribute in your component. It doesn't work out of the box.

zaceno:
@jorgebucaran ? - I think you misunderstood me.

I meant that https://github.com/zaceno/hyperapp-context is can be updated to keep working if we merge lazy components. I was worried it would be permanently broken with no way of implementing context properly on top of hyperapp. I have now verified that is not the case.  I'm not opposed to merging this. I'm just not sure I see the gains.

> but you need a render attribute in your component. It doesn't work out of the box.

Correct. It is possible to make components lazy "in userland" by using a render prop. I think that would be preferable, since it gives userland control over what we pass into the lazy components -- so hyperapp doesn't have to have an opinon. And it makes laziness opt-in.

jorgebucaran:
@zaceno If we don't make components lazy by default, then it's impossible to share the global state and actions with components. That's the other part to this feature.

zaceno:
@jorgebucaran @etalisoft demonstrated above how it *is* possible to share global state and actions with all components -- without lazy components --, using a render prop like you just mentioned

jorgebucaran:
@zaceno Alright, but I mean by Hyperapp, default and out of the box.

```jsx
const Subview = state => props => <h1>{state.title}</h1>

const state = { title: "Hello" }
const actions = ...
const view = () => <Subview />

app(state, actions, view, ...)
```

jorgebucaran:
@zaceno Sorry, I could've avoided this confusion by saying that I meant so from the beginning. 😅

zaceno:
@jorgebucaran Ah I see. Ok well my *personal* opinion on that (I know it's contentious) is:

- it is not useful to me
- it is too opinionated of hyperapp to do that particular thing
- it doesn't break anything for me, so: go ahead if you want 😄 
- if you do go ahead, it would be a lot nicer if regular generic components could keep their current signature. I e, both of these should be valid ways of defining a component:

```
const GenericComponent = (props, children) => <div class={props.foo}>{children}</div>

const ConnectedComponent = (props, children) => (state, actions) => <button onclick={actions.do} disabled={state.possible}>Yo!</button>

//alternatively

const ConnectedComponent = (props, children, state, actions) => <button onclick={actions.do} disabled={state.possible}>Yo!</button>
```

jorgebucaran:
@zaceno I think you can have all the signatures without interference from this feature. I'll give you a proper answer when a I get a chance.

jorgebucaran:
@frenzzy @zaceno Okay, this is what I have in mind.

```js
const Subview = props => state => <h1>{state}</h1>

const view = () => {
  return (
    <div>
      <Subview />
      <h1>Title</h1>
    </div>
  )
}
```

This signature `(props => (state, actions) => VNode` is more flexible because then you can do this too:

```jsx
const Subview = props => <h1>{props.title}</h1>

const view = () => {
  return (
    <div>
      {state => <Subview title={state} />}
      <h1>Title</h1>
    </div>
  )
}
```

SkaterDad:
> @jorgebucaran This signature: (props => (state, actions) => VNode is backward compatible and lets you do this too:

I'm 100% liking this API for components.  I don't see any reason to have React-style `context` since we don't have stateful components, but reducing boilerplate by passing the full state/actions to every component is nice and simple.  Nice bonus that it's backwards compatible, too :+1:

jorgebucaran:
@frenzzy Does it make sense to support vnodes as a function for the root? It seems pointless since you already have the state and actions provided by the view.

```jsx
// What's the point of this?

const App = props => (state,actions) => <...>
const view = (state, actions) => <App />
```

frenzzy:
@jorgebucaran 
```js
const view = () => <Layout>...</Layout>
const view = () => <Router>...</Router>
const view = () => <Provider>...</Provider>
// etc.
```

jorgebucaran:
Thanks, @frenzzy! It's merged. I'll be making some changes (components will **not** be lazy by default) and push up in a few hours. 🎉

PR #608, opened 2018-02-20
element[name] -> element.style

38elements:
Since `name` is `style`, I think `element.style` is more explicit.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -134,7 +134,7 @@ export function app(state, actions, view, container) {
     if (name === "key") {
     } else if (name === "style") {
       for (var i in clone(oldValue, value)) {
-        element[name][i] = value == null || value[i] == null ? "" : value[i]
+        element.style[i] = value == null || value[i] == null ? "" : value[i]
       }
     } else {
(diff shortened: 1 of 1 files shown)


Swizz:
I didn't have authored this line, but I see many points to let it as it is.

1. **Bundle size** : Once minimifed `e[n]` will be more light than `e.style` an object keys will never be minified for some reasons.
2. **Gzip size** : Gzip algorithm love repetition, more repetitions you have better the compression will be. We use `element[name]` elsewhere.
3. **In case of changes** : If we need to change the `name === "style"` condition, it will not be necessary to change the used key.

38elements:
Thank you for feedback.
I think your pointing out is reasonable.

JorgeBucaran:
Thanks, @38elements. The reason it's like this is just for saving bytes. This does not reduce readability in my opinion, so I prefer to keep it this way. BTW I will be making some updates to the core that will result in a few more bytes to the bundle (+10 or so) but will make the code more readable overall. Wish me luck!

PR #609, opened 2018-02-20
Support array in state or partial state root

viljami:
Currently state root or action divided root supports only objects, due to the clone method. With small changes following could be possible.

```js
var state = [];
var actions = {
    add: n => state => state.concat(n),
    getState: () => state => state
};
var main = app(state, actions, () => {})

main.add(1); // => [1]
main.getState(); // => [1]
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -71,6 +71,17 @@ export function app(state, actions, view, container) {
   }
 
   function clone(target, source) {
+    if (Array.isArray(target)) {
+      if (!source) {
+        return target.concat();
+      }
+
+      return target.map(function(a, i) {
+        return source[i] || a;
+      })
+      .concat(source.slice(target.length));
+    }
+
     var obj = {}
 
     for (var i in target) obj[i] = target[i]
--- a/test/slices.test.js
+++ b/test/slices.test.js
@@ -51,6 +51,56 @@ test("slices", done => {
   done()
 })
 
+test("slices and array", done => {
+  const bar = {
+    state: [],
+    actions: {
+      add: () => state => state.concat('b')
+    }
+  }
+
+  const foo = {
+    state: {
+      value: [],
+      bar: bar.state
+    },
+    actions: {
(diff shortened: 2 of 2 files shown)


frenzzy:
This looks similar to #512

viljami:
@frenzzy, It does look similar. It think it's focus is slightly different. My only purpose is to support arrays in store as same as objects. I run into this while writing a todo app earlier. https://github.com/viljami/hyperapp-todolist/blob/master/src/app/store/todos.js

And I would like to write the todo.js file like this:
https://gist.github.com/viljami/fde50888c2346576e99d19cf16f3d1db

Here's a bit more about the issue I had.
```js
var state = [];
var actions = {
    add: n => state => state.concat(n),
    getState: () => state => state
};
var main = app(state, actions, () => {})

// What happens now
main.add(1); // => {'0': 1}
main.getState(); // => {'0': 1}

// What I would expect is
main.add(1); // => [1]
main.getState(); // => [1]
```
So, I don't want my array to change from array to object. Secondly, I can't iterate over object. 

```js
// For a list of items I want the following to be always possible
main.getState().map(a => <Item {...a} />)
```
I could workaround the problem by writing the above code like this.

```js
var state = {items: []};
var actions = {
    add: n => state => ({items: state.items.concat(n)}),
    getState: () => state => state
};
var main = app(state, actions, () => {})

// What happens now
main.add(1); // => {items: [1]}
main.getState(); // => {items: [1]}
```

And now I wrote slightly more code. The action add got a bit more complex and there is an object wrapping the array which was not originally needed. 

Now lets imagine a larger application which state tree is way deeper. The wrapping objects could start to add to the complexity of the tree. I'd like to keep it as simple as possible without added workarounds.

Swizz:
The goal is the same as I had imaginated in #512

JorgeBucaran:
Thanks, @viljami, but I don't want to support arrays as slices.

PR #611, opened 2018-02-20
fix minify

titouancreach:
Use the correct uglifyJs syntax:

see https://github.com/HyperappCommunity/hyperapp-effects/pull/23 and https://github.com/HyperappCommunity/hyperapp-effects/pull/21

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "test": "jest --coverage --no-cache && tsc -p test/ts",
     "build": "npm run bundle && npm run minify",
     "bundle": "rollup -i src/index.js -o dist/hyperapp.js -m -f umd -n hyperapp",
-    "minify": "uglifyjs dist/hyperapp.js -o dist/hyperapp.js -mc pure_funcs=Object.defineProperty --source-map includeSources,url=hyperapp.js.map",
+    "minify": "uglifyjs dist/hyperapp.js -o dist/hyperapp.js -mc pure_funcs=['Object.defineProperty'] --source-map includeSources,url=hyperapp.js.map",
     "prepare": "npm run build",
     "format": "prettier --semi false --write {src,test}/**/*.js {,test/ts/}*.{ts,tsx}",
     "release": "npm run build && npm test && git commit -am $npm_package_version && git tag $npm_package_version && git push && git push --tags && npm publish"


JorgeBucaran:
@titouancreach I want to remove the <samp>Object.defineProperty(__esmodule, {...})</samp> part, and I think that's already working as I can't find it here: https://unpkg.com/hyperapp.

frenzzy:
Have you read [webpack 4: import() and CommonJs](https://medium.com/webpack/webpack-4-import-and-commonjs-d619d626b655) article? 

https://unpkg.com/hyperapp@latest/dist/hyperapp.js is a non-ESM module, but it could be transpiled-ESM if use [named exports](https://rollupjs.org/guide/en#named-exports) in rollup.

titouancreach:
@JorgeBucaran It removes the `Object.defineProperties` but it compiles other part of the code differently! 
I had exactly the same behavior here https://github.com/HyperappCommunity/hyperapp-effects/pull/21#issuecomment-367139859. 

In both case, `Object.defineProperties` was removed. 
I think the bad syntax will lead to weird behaviors later (very weird). 

I should probably read the uglify JS to know what it does with the bad syntax.

Even if its work for now, it's better to follow the official documentation :)

JorgeBucaran:
@titouancreach Sounds good, but do you mean uglify's documentation? and do you have a link, please?

titouancreach:
@JorgeBucaran It is what I used:

https://github.com/mishoo/UglifyJS2#cli-compress-options
https://github.com/mishoo/UglifyJS2#compress-options

> You can pass an array of names and UglifyJS will assume that those functions do not produce side effects.

JorgeBucaran:
@titouancreach So, to clarify, the problem is that the current uglify syntax/instruction is invalid, therefore it generates different (unexpected) code that luckily works?

titouancreach:
Exactly !! :+1:  Sorry to be unclear

okwolf:
@titouancreach so does this change make any change to the minified bundle at all? Is it just accidentally working correctly?

titouancreach:
I didn't read the minified code this time! (it's not that much fun). I just run the tests and they all pass.

okwolf:
> @titouancreach I didn't read the minified code this time! (it's not that much fun). I just run the tests and they all pass.

I wouldn't expect you to try and read that mess, but I'm curious if there's a diff between the output with and without this change.

titouancreach:
Yes I can do that today I guess :)

jorgebucaran:
Thank you, @titouancreach! 🎉

titouancreach:
I'm proud to appear in hyperapp the contributors (even for 1 line) :)

PR #612, opened 2018-02-21
CSS Variables (Custom Properties) 💅

frenzzy:
Usage:
```js
<main style={{ '--foo': 'red' }}>
  <h1 style={{ color: 'var(--foo)' }}>Hello</h1>
  <button style={{ backgroundColor: 'var(--foo)' }}>+</button>
</main>
```
Refs:
- MDN: [Using CSS custom properties (variables)](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables)
- Can I use: https://caniuse.com/#feat=css-variables
- PR to React.js: https://github.com/facebook/react/pull/9302

### DIFF
--- a/test/dom.test.js
+++ b/test/dom.test.js
@@ -481,11 +481,11 @@ testVdomToHtml("styles", [
     html: `<div></div>`
   },
   {
-    vdom: <div style={{ color: "red", fontSize: "1em" }} />,
+    vdom: <div style={{ color: "red", fontSize: "1em", "--foo": "red" }} />,
     html: `<div style="color: red; font-size: 1em;"></div>`
   },
   {
-    vdom: <div style={{ color: "blue", float: "left" }} />,
+    vdom: <div style={{ color: "blue", float: "left", "--foo": "blue" }} />,
     html: `<div style="color: blue; float: left;"></div>`
   },
   {
--- a/src/index.js
+++ b/src/index.js
@@ -145,7 +145,12 @@ export function app(state, actions, view, container) {
     if (name === "key") {
     } else if (name === "style") {
       for (var i in clone(oldValue, value)) {
-        element[name][i] = value == null || value[i] == null ? "" : value[i]
+        var style = value == null || value[i] == null ? "" : value[i]
+        if (i[0] === "-") {
+          element[name].setProperty(i, style)
+        } else {
+          element[name][i] = style
+        }
       }
     } else {
       if (name[0] === "o" && name[1] === "n") {


frenzzy:
FYI, you also can use `cssText` as a workaround for current version of `hyperapp`:
```js
<main style={{ cssText: '--foo:red' }}>
  <h1 style={{ color: 'var(--foo)' }}>Hello</h1>
  <button style={{ backgroundColor: 'var(--foo)' }}>+</button>
</main>
```
Demo: https://codepen.io/frenzzy/pen/RQygyL?editors=0010

JorgeBucaran:
Thanks, @frenzzy. Amazing work. It's not as much code as I actually imagined, but I'll not merge this before I can do some research of my own. Exploring alternatives like using <samp>cssText</samp> to implement inline styles, or maybe even something more wacko like [Picostyle](https://github.com/picostyle/picostyle).

test/dom.test.js
```diff
@@ -605,11 +605,15 @@ testTreeSegue("styles", [
...
+      style: { color: "red", fontSize: "1em", "--foo": "red" }
+    }),
     html: `<div style="color: red; font-size: 1em;"></div>`
   },
   {
-    tree: h("div", { style: { color: "blue", float: "left" } }),
+    tree: h("div", {
+      style: { color: "blue", float: "left", "--foo": "blue" }
+    }),
     html: `<div style="color: blue; float: left;"></div>`
```
frenzzy: looks like jsdom does not support custom properties yet =/
```js
test('example', () => {
  const div = document.createElement('div')
  div.style.setProperty('--foo', 'red')
  document.body.appendChild(div)

  console.log(document.body.innerHTML) // => '<div></div>'
  console.log(div.style.getPropertyValue('--foo')) // => undefined

  expect(document.body.innerHTML).toBe('<div style="--foo:red"></div>') // => fail
  expect(div.style.getPropertyValue('--foo')).toBe('red') // => fail
})
```

frenzzy: Google Search crawler is based on Chrome 41, which does NOT support CSS Custom Properties: https://twitter.com/ebidel/status/968989651888295936

jorgebucaran:
@frenzzy What's the total byte increase?

frenzzy:
file | master | branch | diff
--- | --- | --- | ---
hyperapp.js | 3465 | 3516 | +51
hyperapp.js.gz | 1545 | 1578 | +33

jorgebucaran:
@frenzzy Sounds like a reasonable compromise. Is there anything else to add? Does this cover everything? 

What about `--mainColor` or `$mainColor` instead of `--main-color`? Like:

```
<div style={{
  $cssVariable: "red"
}} />
```

Probably a lot more code, but it would be cool haha.

/cc @lukejacksonn @zaceno @SkaterDad @okwolf

frenzzy:
According to [custom properties specification](https://www.w3.org/TR/css-variables/#custom-property) the only requirement is any custom property name must starts with two dashes.

So, what notation to use (`--mainColor` or `--main-color`) is up to you.

Symbol `$` more about preprocessors like Sass, Less an so on but not about native CSS.

jorgebucaran:
@frenzzy 

> So, what notation to use (--mainColor or --main-color) is up to you.

You are right.

> Symbol $ is more about preprocessors like Sass, Less an so on, but not about native CSS.

Right, but if we (just an idea) treated names starting with a `$` as a CSS variable it would be a lot easier to use/type in the style object.

```js
{ $cssVariable: "red" }
```
...instead of:
```js
{ "--cssVariable": "red" }
```
But yeah, I probably be crucified if we introduced this convenient oddball.

lukejacksonn:
For once, I am not totally against breaking the standard here as you point out `$var` is way way better than `"--var"` but it is probably still a bad idea 😅

frenzzy:
How about this?
```js
const brandColor = '--brand-color'
const $brandColor = `var(${brandColor})` 

const view = () => (
  <main style={{ [brandColor]: 'red' }}>
    <h1 style={{ color: $brandColor }}>Hello</h1>
    <button style={{ backgroundColor: $brandColor }}>+</button>
  </main>
)
```

jorgebucaran:
@frenzzy What about rewriting the code like this?

```js
for (var i in clone(oldValue, value)) {
  if (i[0] === "-") {
    element[name].setProperty(i, value)
  } else {
    element[name][i] = value == null || value[i] == null ? "" : value[i]
  }
}
```

jorgebucaran:
Test `#ff0000` `#ff00ff` `#ffff00` `#ffffff`

jorgebucaran:
@frenzzy I think we can use `setProperty` https://github.com/facebook/react/issues/6411#issuecomment-290643973

SteveALee:
> So, what notation to use (--mainColor or --main-color) is up to you.

But  I suggest stick to the current convention of snale case in CSS and camel case in JS

PS I only just found out you can set CSS vars easily in JS. no need to mess about manipulating stylesheets in JS.

jorgebucaran:
@SteveALee We went with the --no-surprise version of CSS variables! 😉 

```jsx
<div
  style={{
    "--color": "red",
    color: "var(--color)"
  }}
>
  OK
</div>
```

SteveALee:
Totally Agree. Just making a note for users of HA :)

Issue #532, opened 2018-01-02
Allow a custom merge method


titouancreach:
Currently, when a action return a new state, this one is merged with the previous state. That make the object "state" to grow but never shrink. I would imagine this behavior to be configurable with a pure function. where the default would be:

```javascript
const mergeState = (previousState, newState) => ({...previousState, ...newState});
```

To have the same behavior as redux for example, we can use this function
```javascript
const mergeState = (previousState, newState) => newState;
```

JorgeBucaran:
@titouancreach How would that look in a simple app and why would I want my state to shrink?

titouancreach:
The first example that would come to my mind would be: I want to create a simple router, I would be able to do:

```javascript
import defaultStatePageFooBar from '...';

actions: {
  goToPageFooBar: (state) => defaultStatePageFooBar
}
```

I find the second `mergeState` more functional and more natural because you don't hide a "merge" to the user.

JorgeBucaran:
@tiaanduplessis And how would that look like on your app?

titouancreach:
You specify the merge method in the constructor as an optional parameter:

```javascript
app(state, actions, view, document.body, mergeState);
```

whaaaley:
I like the idea of non-merged state. I know merged state has it's own pros, but merging doesn't really help me with my own apps.

If I have some state like this and and I want to update the state.

```
// initial
{
  foo: 'bar',
  bar: 'baz'
}

// now i want this
{
  foo: 'ooga booga'
}
```

I won't be able to. I'd have to set `bar` to `null` myself to denote that it's "removed".

```
// initial
{
  foo: 'bar',
  bar: 'baz'
}

// now i want this
{
  foo: 'ooga booga',
  bar: null
}```

It's not a "con" per se. It's just a way I would prefer to build my apps. Imo, when being able to replace a state object "wholesale" like this, is easier to read and less to code.

JorgeBucaran:
Why would anyone want this `{ foo: 'ooga booga' }`?

okwolf:
The current state update approach in Hyperapp reminds me of [updating records in Elm](http://elm-lang.org/docs/records#updating-records) or `Object.assign(state, { key: value })`/`{ ...state, key: value }` from ES 6, but without the boilerplate of bringing in the previous state values.

This is by far the most common use case, and the only one I've used in [reducers for Redux](https://redux.js.org/docs/recipes/UsingObjectSpreadOperator.html).

[...]

PR #613, opened 2018-02-21
Support for non-object slices

frenzzy:
An alternative solution for #609 and #512

- It allows to write more simple code as described in this comment: https://github.com/hyperapp/hyperapp/pull/609#issuecomment-367253627
- It allows to change data type of a slice or easily override/remove it
- It makes custom merge method unnecessary #532 (I hope)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -96,6 +96,10 @@ export function app(state, actions, view, container) {
     return source
   }
 
+  function isObject(val) {
+    return Object.prototype.toString.call(val) === "[object Object]"
+  }
+
   function wireStateToActions(path, state, actions) {
     for (var key in actions) {
       typeof actions[key] === "function"
@@ -106,12 +110,16 @@ export function app(state, actions, view, container) {
               }
 
               if (
-                data &&
+                data != null &&
                 data !== (state = get(path, globalState)) &&
                 !data.then // Promise
               ) {
                 scheduleRender(
-                  (globalState = set(path, clone(state, data), globalState))
+                  (globalState = set(
+                    path,
+                    isObject(data) ? clone(state, data) : data,
+                    globalState
+                  ))
                 )
               }
 
@@ -120,7 +128,9 @@ export function app(state, actions, view, container) {
           })(key, actions[key])
         : wireStateToActions(
             path.concat(key),
-            (state[key] = clone(state[key])),
+            state[key] == null || isObject(state[key])
+              ? (state[key] = clone(state[key]))
+              : state[key],
             (actions[key] = clone(actions[key]))
           )
     }
--- a/test/slices.test.js
+++ b/test/slices.test.js
@@ -83,3 +83,29 @@ test("state/actions tree", done => {
   main.foo.bar.baz.foobarbaz()
   expect(state).toEqual({ foo: {} })
 })
+
+test("non-object slices", () => {
+  const state = {
+    number: 1,
+    string: "foo",
+    array: ["a"]
+  }
+
+  const actions = {
+    number: { add: value => state => state + value },
+    string: { add: value => state => state + value },
+    array: { add: value => state => state.concat(value) },
+    getState: () => state => state
+  }
+
+  const main = app(state, actions, () => {})
+
+  expect(main.number.add(-1)).toBe(0)
+  expect(main.string.add("bar")).toBe("foobar")
+  expect(main.array.add("b")).toEqual(["a", "b"])
+  expect(main.getState()).toEqual({
+    number: 0,
+    string: "foobar",
+    array: ["a", "b"]
+  })
+})


JorgeBucaran:
@frenzzy Can I also do something like this?

```jsx
const state  = 0
const actions = { up: () => state => state  + 1 }

**Comment by Swizz** [comment]
Only curiosity, not required : 
The #512 side effect is still working ? 

```jsx
const state = {
  counts: [{
    count: 0
  }]
}

const actions = {
  counts: [{
  }]
}

  <div>
   {state.counts[0].count}
  </div>
)

main.counts[0].up()
```

frenzzy:
@JorgeBucaran this is possible only if change this line:
https://github.com/hyperapp/hyperapp/blob/8b501cea26179e223c0fadf188e8bcad5219b442/src/index.js#L35
because `clone` converts everything to an object

JorgeBucaran:
@frenzzy Gotcha, not worth it.

frenzzy:
@Swizz no, actions nested to arrays are still not supported.
But you can rewrite the example:
```js
const state = {
  counts: [
    { count: 0 }
  ]
}

const actions = {
  counts: {
    up: (index, value) => state => state.map(
      (slice, idx) => idx === index ? { count: slice.count + value } : slice
    )
  }
}

  <div>
    {state.counts[0].count}
  </div>
)

const counterIndex = 0
const upValue = 100
main.counts.up(counterIndex, upValue)
```

jason-whitted:
@frenzzy Your example won't actually work as illustrated in [this CodePen](https://codepen.io/jason-whitted/pen/wyjRgO?editors=0010).

Currently the actions only accept one value, because of this block of code:
```js
  function wireStateToActions(path, state, actions) {
    for (var key in actions) {
      typeof actions[key] === "function"
        ? (function(key, action) {
            // The function is created only supporting the single parameter 'data'
            actions[key] = function(data) {
              // action is then called with only one argument
              if (typeof (data = action(data)) === "function") {
                data = data(get(path, globalState), actions)
              }
              ...
```

frenzzy:
@jason-whitted the issue you found is not related to the current PR, see #619

okwolf:
I don't really understand how the primitive type support is an improvement since it lets you treat the values as objects for nesting actions, which means inconsistent state/actions shape.

For arrays, I have the same concerns as I did before: https://github.com/hyperapp/hyperapp/pull/512#issuecomment-354497710

frenzzy:
@okwolf this is not about nested actions, but about an ability to write more simple actions for primitive types. See https://github.com/hyperapp/hyperapp/pull/609#issuecomment-367253627 how it simplifies todo example.
```js
const state = {
  name: "",
  age: 0
}
const actions = {
  setName: value => { name: value }
  setAge: value => { age: value }
}
actions.setName("Jhon")
actions.setAge(18)
```
vs
```js
const state = {
  name: "",
  age: 0
}
const actions = {
  name: { set: value => value }
  age: { set: value => value }
}
actions.name.set("Jhon")
actions.age.set(18)
```

JorgeBucaran:
I find this feature rather confusing. It's easier if we just deal with objects.

okwolf:
> @frenzzy this is not about nested actions, but about an ability to write more simple actions for primitive types

Perhaps we are in agreement that nested actions shouldn't be used for primitive types, but that's exactly what I see in this proposal. The actions for `name` and `age` are nested in an object, which is not the shape of the state they update. I think we might be using two different definitions of "simple" though.

How exactly is someone to tell if `actions.name` is for a primitive type or a state slice? I understand that returning a non-object from an action today can break things, but at least there's only one valid way to return the new state once you learn how to use it everywhere. What explosions can we expect in a world where there are multiple valid options and someone accidentally returns a partial state object where a primitive was supposed to be? 💣 

And what exactly are we gaining from this? IMHO `actions.setName` is _not_ a bigger burden to use than `actions.name.set`.

Swizz:
So the complementary of this PR should be to remove partials ?

@zaceno and I were discussing about this feature some weeks ago.
One of the benefits of this feature is to make the user life a litle bit easier. And every one here know I am on this side since the beggining.

But, I am not on the same page regarding `actions.setName`. I am using a `set` actions in each slice, and I use objects to tell the key to set, with bonus to be able to set multiple once `actions.set({ name: "Jhon })`

JorgeBucaran:
@Swizz Sorry, but what are partials? 😄

Swizz:
> To update the state, an action must return **a partial state object**

I took the bad habit (maybe ?) to call them partials.

JorgeBucaran:
No problem. If the complement of this PR is to remove partials on another one, then I don't see myself merging this ever. I think for the looks of it this is probably going to be another wontfix.

frenzzy:
When I opened this PR the idea was to allow to easily reduce slices with primitive types [like in Redux](https://redux.js.org/basics/reducers#source-code) and mostly to make work with arrays simpler.

I absolutely agree with @okwolf and @JorgeBucaran points. This feature may introduce confusion and misuse and this feature is not worth it.

So, should we just close this PR or will wait for more opinions?

Swizz:
Or open a RFC as an issue to gather feedbacks and wish about this feature ?

JorgeBucaran:
I say let's leave it open for a bit more to collect more feedback.

Issue #615, opened 2018-02-21
Global access to state/actions in actions/components


jason-whitted:
Slices are convenient when they are helping you out, but they very frequently impede the developer from being able to accomplish a goal.  This is a recurring theme as seen in:

- #601 - @spapas was requesting global access to actions in an action
- #604 - connected components, which make global access to state/actions available in components
- #606 - @frenzzy commenting on possibly adding state/actions to every component's arguments

I have experimented with hyperapp over at [etalisoft/hyperapp-extended](https://github.com/etalisoft/hyperapp-extended) to get this to work.

It modifies the following signatures:
```js
// action signature
value => (state, actions, current) => ...
// component signature
(attributes, children, current) => ...
```

`current` is an object with the following methods:
- `getState()` - returns the global state
- `getActions()` returns the global actions

This allows both actions and components the flexibility to have access to current state/actions even in nested async callbacks.

The `current` object is only created once and frozen and merely passed along as an additional argument, which should have no negative performance impact.

JorgeBucaran:
Closed in favor of #606.

Issue #601, opened 2018-02-16
How to call an action in a different slice from a nested asynchronous action


spapas:
I have a (nested) login() action which is inside an `auth` object/namespace  and an `updateLoading` function that is not inside that namespace (it is called from other places in my app). Something like this:

```
const reducers = module.exports = {
    updateLoading: loading => state => ({
	loading
    }),
    auth: {
        login: () => (state, actions) => {
            actions.updateLoading(true)
	}
	// Do the login etc ...
    }, 
    // ...
}
```

Now, since login is asynchronous, I want it to call the `updateLoading()` while doing the http post request to the login API but the `actions` object that it receives contains *only* the actions that are inside the `auth` object/namespace.

Is it possible to call the `upadeLoading` action from inside `auth.login` ?

zaceno:
Short answer: no, not directly like in your example. 

There are several ways to work around it though, each with its own pros and cons. Various ways I've handled this sort of thing in the past:

- Pass a prop `onupdateloading: actions.updateLoading`  to your auth-component. Then in your `onsubmit` of the login form: `onsubmit: ev => actions.auth.login({onupdateloading})`, and finally in the `login` action: `onupdateloading(true)`

- Use a global event bus in your app. In your `login` action: `eventbus.emit('loading', true)`. And during your app setup:

```js
const main = app(state, actions, view, container)
eventbus.on('loading', x => main.updateLoading(x))
```

- Somehow leverage that the state changes when you login, in your view. Perhaps something like:

```jsx
const view = (state, actions) => {
  if (state.auth.attempting && !state.loading) actions.updateLoading(true)
  return (
    <main>
      ...
    </main>
  )
}
```

spapas:
Hello @zaceno thank you very much for the thorough answer! I have tried a modified version of your first solution (the second one adds extra complexity I don't want and the third one is really confusing to me). So, because I actually need more actions when logging in (i.e change the location from the login view) I tried pass the whole `reducers` object down to the login action (instead of only updateLoading) and it is working fine now. 

So my login action is something like this:
```
login: (global_actions) => (state, actions) => {
// I can use every action here using the global_actions object
}
```


However I'd like to also ask wouldn't it be possible for hyperapp to pass by default the `global_actions` a third argument to the async action signature ? I.e something like this:

```
login: () => (state, actions, global_actions) => {
...
}
```

This shouldn't be a big change (and it won't really change the API - you'll just define global_actions if you need it) and would make everything much easier !

[...]

Issue #569, opened 2018-01-23
Change component signature from (props, children) to (props) and props.children


frenzzy:
Current component signature:
```js
const Component = (props, children) => (
  <div {...props}>{children}</div>
)
```
Desirable component signature:
```js
const Component = (props) => (
  <div {...props}>{props.children}</div>
)
// or even
const Component = (props) => (
  <div {...props} />
)
```
**Props**
- You may omit the unused argument if your component uses only `children`, ref #567
  ```js
  const Component = (props) => (
    <div>{props.children}</div>
  )
  ```
  vs
  ```js
  const Component = (_, children) => (
    <div>{children}</div>
  )
  ```
- It is easier to use memoization because you don't need to compare children separately from other props
  ```js
  const Component = (props) => (
    <h1 title={props.title}>Hello {props.children}</h1>
  )

  const MemoizedComponent = memoize(
    Component,
    // by default (props, nextProps) => !shallowEqual(props, nextProps)
  )
  const DeepMemoizedComponent = memoize(
    Component,
    (props, nextProps) => !_.deepEqual(props, nextProps)
  )
  const CustomMemoizedComponent = memoize(
    Component,
    (props, nextProps) => props.children !== nextProps.children
  )
  ```
  vs
  ```js
  const Component = (props, children) => (
    <h1 title={props.title}>Hello {children}</h1>
  )

  const MemoizedComponent = memoize(
    Component,
    // by default (props, nextProps, children, nextChildren) => !shallowEqual(props, nextProps) || children !== nextChildren
  )
  const DeepMemoizedComponent = memoize(
    Component,
    (props, nextProps, children, nextChildren) => !_.deepEqual(props, nextProps) || children !== nextChildren
  )
  const CustomMemoizedComponent = memoize(
    Component,
    (_, _, children, nextChildren) => children !== nextChildren
  )
  ```
- Using `children` prop is valid and it will not throw an error:
  `TypeError: Cannot assign to read only property 'children' of object '#<HTMLHeadingElement>'`
- It will be easier to add something new in the future (or via HOA)
  ```js
  const ComponentWithContext = (props, context) => (
    <div>{props.children}</div>
  ) // Context for SSR aka context in React.js
  const ComponentWithConnect = (props, store, actions) => (
    <div>{props.children}</div>
  )  // aka React-Redux Connect
  // etc.
  ```

**Cons**
- It is a breaking change and requires at least Hyperapp v2 release

[...]

Issue #388, opened 2017-09-24
Error Handling


okwolf:
Please forgive me if we've already had a discussion on this topic, and I know this is subject to change with the API proposals brewing, but I wanted to capture the state of affairs as of this moment.

~~Because of the indirection with `events.action` and `events.resolve` it's not possible to catch any errors that are thrown within the execution of an action today. I believe the same will also hold true for the coming `hooks` API.~~

It is possible to catch and handle view errors with an [HOC](https://facebook.github.io/react/docs/higher-order-components.html).

React 16 is introducing [Error Boundaries](https://facebook.github.io/react/blog/2017/07/26/error-handling-in-react-16.html) to handle this problem. I'm not proposing we add anything to core, but wondering what the best practice should be.

Here's a small demo pen showing view/sync action/async action errors and how one might handle them with an HOA: https://codepen.io/okwolf/pen/rGWpgJ?editors=0010

johanalkstal:
I like the idea of having something catch errors and allowing you to react to it, without crashing the app.

I don't like the idea of it being a component.

I feel there's an already existing problem (talking about React now) with the creation of humongous component trees filled with higher order components that distort the actual features of your application. I see a real danger of error boundaries becoming littered all over, adding to that.

But again, I like the idea! The problem to solve for me would then be, how can I create such a thing that also catches errors on itself, eliminating the need for a wrapper?

JorgeBucaran:
@johanalkstal @okwolf What about a new VDOM event?

```js
app({
  view(state) {
    return (
      <main onerror={() => console.log("Look what you've done!")} /> 
    )
  }
})
```

SkaterDad:
@JorgeBucaran I assume that event would only catch errors which occur during the creation of v-dom nodes?

JorgeBucaran:
@SkaterDad Yes, should it work otherwise?

SkaterDad:
Nope, just verifying.  I think it would be a useful feature.  There are definitely times when I'm iterating on frontend/backend, and get data mismatches.

okwolf:
> @JorgeBucaran What about a new VDOM event?

Would the `onerror` function return a VNode to render in place of the node it's specified for? What would happen if you defined nested `onerror` props? 🤔

JorgeBucaran:
@okwolf Hmm not sure. TBH I still don't understand why we need this in Hyperapp.

okwolf:
[...]

PR #616, opened 2018-02-22
New function to allow `app` creation with inversion of control

etalisoft:
# Inversion of Control

## Why?
The base `app` function that Hyperapp supplies is perfect for covering the common use case.  It can be customized pretty easily with Higher-Order Apps and Higher-Order Components.  Unfortunately HOA/HOC can only take you so far.  Some features need to be implemented inside `app` itself.  Features like connected components or context would be beneficial to some, but those not using the features would have to pay the performance price.  The PR is to enable app creation with inversion of control -- empowering users to add the features themselves.

## Details
```jsx
```
With this PR these would continue to work exactly as they do today; however, a new `hyper` function would also be available.
```jsx
import { h, hyper } from 'hyperapp';
```
The `hyper` function takes a `parseNode` function and returns an `app` function.
The `parseNode` function has the following signature:
```jsx
const parseNode = (node, parent, globalState, wiredActions) => /* node */
```

✳️ **EDIT:** _The `hyper` function has been replaced with an overloaded `app` function.  Example:_
```js
app(parseNode)(state, actions, view, document.body);
```
✳️ **EDIT:** _To be clear, the `parseNode` functions are not supplied by HyperApp, but instead developed by the community.  Below are a few examples I created just to show some of the capabilities._

### Example: App supporting Context
```jsx
const { h, hyper } from 'hyperapp';
// Create the app with a custom node parser
const app = hyper(nodeParserWithContext);
// Thanks to nodeParserWithContext every component now receives context!
const Component = (props, children, context) => <div />;
app(state, actions, view, document.body);
```
Want to see it in action?  Check out this [CodePen](https://codepen.io/jason-whitted/pen/BYVjmb).

### Example: Using Context to create Provider/connect (like redux)
Building upon the previous Context example, we can easily create a `Provider` component and `connect` HOC.

Want to see it in action?  Check out this [CodePen](https://codepen.io/jason-whitted/pen/LQrGam).

### Example: App supporting Connected Components
```jsx
const { h, hyper } from 'hyperapp';
// Create the app with a custom current parser
const app = hyper(nodeParserWithCurrent);
// Thanks to nodeParserWithCurrent every component now receives a current object
// The current object has getState() and getActions() methods
const Component = (props, children, { getState, getActions }) => <div />;
app(state, actions, view, document.body);
```
Want to see it in action?  Check out this [CodePen](https://codepen.io/jason-whitted/pen/OQENPL).

### Example: App that combines children into props
```jsx
const { h, hyper } from 'hyperapp';
// Create the app with a custom parser
const app = hyper(nodeParserWithChildrenInProps);
// Thanks to nodeParserWithChildrenInProps every component now only receives one props object
const Component = ({ children, ...other }) => <div>{children}</div>;
app(state, actions, view, document.body);
```

Want to see it in action?  Check out this [CodePen](https://codepen.io/jason-whitted/pen/GQGZqV).

### Example: Error Boundaries
```jsx
const { h, hyper } from 'hyperapp';
// Create the app with a custom error boundary node parser
const app = hyper(nodeParserWithErrorBoundary);
// Thanks to nodeParserWithErrorBoundary components that throw errors can be handled
const Throw = (props, children) => { throw new Error('Boom!'); }
app(state, actions, view, document.body);
```

Want to see it in action?  Check out this [CodePen](https://codepen.io/jason-whitted/pen/GQGZyg)

## Conclusion
This PR should have no negative performance impact; however, it should be able to satisfy many of the following issues / PRs: #388, #615, #601, #569, #604, etc.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -33,21 +33,25 @@ function clone(target, source) {
   return obj
 }
 
-export var app = hyper(
-  // (node , parent, globalState, wiredActions) => node
-  function getVNode(node) {
-    if (node == null || typeof node === "boolean") {
-      return ""
-    }
-    if (typeof node.nodeName === "function") {
-      return getVNode(node.nodeName(node.attributes, node.children))
-    }
-    return node
+function getVNode(node) {
+  if (node == null || typeof node === "boolean") {
+    return ""
   }
-)
+  if (typeof node.nodeName === "function") {
+    return getVNode(node.nodeName(node.attributes, node.children))
+  }
+  return node
+}
 
-export function hyper(parseNode) {
-  return function app(state, actions, view, container) {
+export function app(state, actions, view, container) {
+  var parseNode = getVNode
+  if (typeof state === "function") {
+    parseNode = state
+    return createApp
+  }
+  return createApp(state, actions, view, container)
+
+  function createApp(state, actions, view, container) {
     var renderLock
     var firstRender = true
     var lifecycleStack = []
--- a/src/index.js
+++ b/src/index.js
@@ -24,31 +24,17 @@ export function h(name, attributes /*, ...rest*/) {
   }
 }
 
-export function app(state, actions, view, container) {
-  var renderLock
-  var firstRender = true
-  var lifecycleStack = []
-  var rootElement = (container && container.children[0]) || null
-  var oldNode = rootElement && toVNode(rootElement, [].map)
-  var globalState = clone(state)
-  var wiredActions = clone(actions)
+function clone(target, source) {
+  var obj = {}
 
-  scheduleRender(wireStateToActions([], globalState, wiredActions))
+  for (var i in target) obj[i] = target[i]
+  for (var i in source) obj[i] = source[i]
 
-  return wiredActions
+  return obj
+}
 
-  function toVNode(element, map) {
-    return {
-      nodeName: element.nodeName.toLowerCase(),
-      attributes: {},
-      children: map.call(element.childNodes, function(element) {
-        return element.nodeType === 3 // Node.TEXT_NODE
-          ? element.nodeValue
-          : toVNode(element, map)
-      })
-    }
-  }
-
+export var app = hyper(
+  // (node , parent, globalState, wiredActions) => node
   function getVNode(node) {
     if (node == null || typeof node === "boolean") {
       return ""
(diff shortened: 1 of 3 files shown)


JorgeBucaran:
@etalisoft I don't want to export a new function. What other options do we have?

JorgeBucaran:
> @etalisoft The PR is to enable app creation with inversion of control -- empowering users to add the features themselves.

Too much power leads to option paralysis, so while I like this, it's not for Hyperapp. I would be interested in exploring adding only one more feature to Hyperapp, for example, a "context" feature like React's.

- https://github.com/hyperapp/hyperapp/pull/604#issuecomment-367552428

etalisoft:
> @JorgeBucaran I don't want to export a new function. What other options do we have?

Wow, you read that wall-of-text in 2 minutes? 

A couple of options I can think of:

- An optional 5th parameter to `app`
  `app(state, actions, view, document.body);`
- If the first parameter is a function then apply inversion of control
  `app(parseNode)(state, actions, view, document.body);`

JorgeBucaran:
@etalisoft For starters, let's make a few changes. Extract all the functions from the app body, to the top level of the module/file and rename <samp>hyper</samp> to <samp>createApp</samp>.

Also, I am still not sure if I want to go with #606 or not. Perhaps consider answer the questions you were asked on that issue first.

JorgeBucaran:
> @etalisoft If the first parameter is a function then apply inversion of control...

I think this is an acceptable compromise. What do you think? @okwolf @zaceno

etalisoft:
> @JorgeBucaran Also, I am still not sure if I want to go with #606 or not. Perhaps consider answer the questions you were asked on that issue first.

#606 is @frenzzy's PR.  I don't think I have any pending questions.  I did piggyback off of his change, though, because it moved the node function invocation out of the `h` function and into the `app` function.

JorgeBucaran:
@etalisoft You are right, my bad. Those are @frenzzy's pending questions.

JorgeBucaran:
@etalisoft Sorry, but could you remove @frenzzy's lazy components from this PR?

etalisoft:
> @JorgeBucaran could you remove @frenzzy's lazy components from this PR?

Previously the `h` function invoked all components BEFORE they were passed to the `app` function.
```js
export function h(name, attributes /*, ...rest*/) {
  ...
    ? name(attributes || {}, children) // <-- this line executes outside of the app context
    : {
        nodeName: name,
        attributes: attributes || {},
        children: children,
        key: attributes && attributes.key
      }
}
```

The `h` function immediately invokes child components outside of the context of the `app`.  However the `parseNode` function needs to be ran within the scope of the `app`.  

The lazy-loaded solution instead returns an object with the nodeName remaining as a function.
```js
export function h(name, attributes /*, ...reset */){
  ...
  return {
    nodeName: name,  // <-- Components remain as functions until the app processes them
    attributes: attributes || {},
    children: children,
    key: attributes && attributes.key
  }
}
```

The lazy-loading was the gateway to allow this to work.  This is similar to how [React's](https://reactjs.org/docs/shallow-renderer.html) and [Enzyme's](https://github.com/airbnb/enzyme/blob/master/docs/api/shallow.md) shallow rendering works.

zaceno:
>  Some features need to be implemented inside app itself. 

This sentiment I agree with. Actually, more than parseNodes, I'd like the ability to inject middleware (or replace) how the actions are wired.

So regardless of parseNodes, I think this discussion is interesting. Personally I wouldn't mind exporting another function -- especially If we want to offer multiple other injection points in the future. `app()` can only be overloaded so far.

But about parseNodes:

So the idea is to use it like this:

```js

const myParseNodes = node => {
   ...
   return parsedNode
}
...
app(parseNodes)(state, actions, views, container)

```

Other than the terser code, what's the benefit over using a HOA? Like this:

```js
const myParseNodes = node => {
  ...
  return parsedNode
}

const myApp = app => (state, actions, originalView, container) => {
  const myView = (state, actions) => {
    return myParseNodes(originalView(state, actions))
  }
  return app(state, actions, myView, container)
}


myApp(app)(state, actions, view, container)
```

okwolf:
>> @etalisoft If the first parameter is a function then apply inversion of control...
>
> @JorgeBucaran I think this is an acceptable compromise. What do you think? @okwolf @zaceno

Sounds fine to me. If we refactor the existing `app` to reuse this same path it shouldn't add too much to our footprint either.

jorgebucaran:
@etalisoft Thanks to your work on this and your other work with connected components we figured out lazy components and #606 is now merged into master and will be published in the upcoming 1.2.0. I think that what you are proposing here is super interesting, but not aligned with Hyperapp extreme minimal goals. Good job anyway! 👋

Issue #617, opened 2018-02-22
About supporting string type style


huozhi:
Hey there! I'm trying to write tiny project with hyperapp recently it's really awesome!

One thing I wonder to know is that will hyperapp decided to support string type style value? I think that will be pretty much easier for people to build apps.

```html
<div style="color: blue; backgound-color: red"></div>
```

Currently only object style is supported. If you guys think making compatible work for string style is good here is the pull request https://github.com/hyperapp/hyperapp/pull/618. Otherwise it's still OK if you think supporting may bring redundance. : )

JorgeBucaran:
Hi, @huozhi! Thank you, but I don't want to support this for the reasons given [here](https://github.com/hyperapp/hyperapp/pull/618). 👋

huozhi:
Still many thanks for your quick reply! 👍

Issue #258, opened 2017-06-29
Suggestion add default units for style props


mauron85:
I'm using more and more hyperapp and must admit I like it. Good job.

What I'm describing here is not bug or direct issue with hyperapp. Just want to raise attention for others who might have similar problem.

The problem is that by default hyperapp doesn't assign any units to style props: `width`, `height`, `left` ...
As result these styles props are silently ignored by the browser. This is nothing special, that's how browsers work. It's same behaviour as assigning style props with bare DOM eg. `el.style.height=20` -
would not be applied either.

Normally it's not problem when developer is aware of this, but for people with React background, this would bit surprising, as React handles this somehow.

Example:

Component responsible for rendering usernames from keepassxc db when user click on username login form field).

```
const CredentialsMenu = ({ top, left }) => {
  const style = { position: 'absolute', top, left };
   return (
    <div style={style}>
      <div>
        <ul >
          <li>1</li>
          <li>2</li>
          <li>3</li>
        </ul>
      </div>
    </div>
  );
};
```

This component needs to be "aligned" to some html input field

```
const { bottom, left } = inputElement.getBoundingClientRect();
const menuCoords = { top: bottom + scrollY, left: left + scrollX };
...
<CredentialsMenu {...menuCoords} />
...
```

`top` `left` styles will not be applied, only `position 'absolute'` will.

Fix is quite easy. Just add `px` unit, but you have to do it for every component.

jbucaran:
Here's the implementation in Preact.

-
 https://github.com/developit/preact/blob/52bb91e17a56e6f3dba050c3ae001292b6ff2285/src/constants.js#L12
- https://github.com/developit/preact/blob/33fc697ac11762a1cb6e71e9847670d047af7ce5/src/dom/index.js#L58

leeoniya:
that regex may be somewhat fragile for future, but hyperapp is too size-focused to absorb the actual whitelist [1] :/

[1] https://github.com/leeoniya/domvm/blob/3.x-dev/src/view/addons/autoPx.js

jbucaran:
@leeoniya Ouch.

leeoniya:
yeah, i only have it in my "larger" builds 😆

lazarljubenovic:
I'm can't help but wonder if something can be hacked together by changing `valueOf` on the Number prototype to 

```
Number.prototype.valueOf = function() { return `${this}px` }
```

Or creating a new prototype, say Pixel:

```
[...]

PR #618, opened 2018-02-22
feature: support string type styles (#617)

huozhi:


JorgeBucaran:
Thanks, @huozhi! Sorry, but I don't want to support this. The current solution is an improvement over that approach. 

As [described in the docs](https://github.com/hyperapp/hyperapp#styles), Hyperapp's element's style attribute takes a JavaScript object with camelCased properties rather than a CSS string. [Same deal with React](https://reactjs.org/docs/dom-elements.html#style) by the way.

### Related
- https://github.com/JorgeBucaran/picodom/issues/25
- https://github.com/hyperapp/hyperapp/issues/258

PR #619, opened 2018-02-22
Actions with multiple arguments

frenzzy:
Looks like a bug:
```js
const actions = {
  up: (a, b, c) => state => {
    console.log(a) // => 1
    console.log(b) // => undefined (expected: 2)
    console.log(c) // => undefined (expected: 3)
  }
}

const main = app({}, actions, () => {})

main.up(1, 2, 3)
```
Demo: https://codepen.io/frenzzy/pen/KQeygR?editors=0010

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -101,7 +101,7 @@ export function app(state, actions, view, container) {
       typeof actions[key] === "function"
         ? (function(key, action) {
             actions[key] = function(data) {
-              if (typeof (data = action(data)) === "function") {
+              if (typeof (data = action.apply(null, arguments)) === "function") {
                 data = data(get(path, globalState), actions)
               }
 
--- a/test/actions.test.js
+++ b/test/actions.test.js
@@ -100,3 +100,31 @@ test("call action within action", done => {
 
   main.upAndFoo()
 })
+
+test("action with multiple arguments", done => {
+  const state = {
+    value: 0
+  }
+
+  const actions = {
+    up: (a, b, c) => state => ({
+      value: state.value + a + b + c
+    })
+  }
+
+  const view = state =>
+    h(
+      "div",
+      {
+        oncreate() {
+          expect(document.body.innerHTML).toBe(`<div>6</div>`)
+          done()
+        }
+      },
+      state.value
+    )
+
+  const main = app(state, actions, view, document.body)
+
+  main.up(1, 2, 3)
+})


JorgeBucaran:
Actions don't take multiple arguments. As described [in the docs](https://github.com/hyperapp/hyperapp#actions):

> An action is a unary function (accepts a single argument) expecting a payload. The payload can be anything you want to pass into the action.

frenzzy:
O-o
Why do we need such limitation?
Isn't it simpler to not create intermediate functions and use actions as a regular functions for everything?
```js
<Item onSort={(id, newIndex) => actions.sortItem({ id, newIndex })} />
// vs
<Item onSort={actions.sortItem} />
```

JorgeBucaran:
Which one should you use?
```js
some.action(id, value)
```
or
```js
some.action({ id, value })
```

I want to have only one way to do most things. That's why.

JorgeBucaran:
Can't  you change <samp>Item</samp>'s <samp>onSort</samp> signature so that this is possible?

```js
<Item onSort={actions.sortItem} />
```

frenzzy:
Mmm, I see your point.
Action is just a function. It should be up to user how he will use functions, dont'you think? I mean it could be not obvious why only the first argument is working. Like for me, but I read the docs, honestly :D

JorgeBucaran:
Thank you, for understanding. Now it's you and me _vs_ @Swizz! 😄

frenzzy:
If #613 will be merged, it become more weird:
```js
const state = {
  value: 0
}

const actions = {
  value: {
    sum = (a, b) => a + b
  }
}

const view = (state, actions) => <main>{state.value}</main>

actions.value.sum(1, 2) // => 3
main.value.sum(1, 2) // => NaN

// => <main>NaN</main>
```
It looks like Hyperapp breaks my functions =/

JorgeBucaran:
@frenzzy I don't get it, is this joke? :smi

frenzzy:
nope, I am serious)

Also if an action may accept only one argument, why action signature is not like this:
```js
const action = (data, state, actions) => newState
```
it would explain the current behavior.

A current signature as though it is telling you use as many arguments as you want:
```js
const action = (...args) => (state, actions) => newState
```

Sorry for the debates, hope it is still changeable.

Swizz:
This is because, at start I suggest the new API with multiple arguments #440.

JorgeBucaran:
@frenzzy I don't understand what you are saying.

The actions signature is:

```js
action: data => (state, actions) => newState
// or
actions: data => newState
```

> Sorry for the debates, hope it is still changeable.

It's not. Actions will continue to be unary functions.

PR #624, opened 2018-02-25
!newKeyed[oldKeyed[i][1].key] -> !newKeyed[i]

38elements:
Since `oldKey` is `oldChild.key`, I think this is more simpler.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -310,7 +310,7 @@ export function app(state, actions, view, container) {
       }
 
       for (var i in oldKeyed) {
-        if (!newKeyed[oldKeyed[i][1].key]) {
+        if (!newKeyed[i]) {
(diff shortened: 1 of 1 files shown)


JorgeBucaran:
Thanks, @38elements. How come this works? Super weird. 😆

38elements:
I looked this lines.
I think that `oldChild.key` is `oldKey`.
I could not find a line where `oldChild.key` changes.  


https://github.com/hyperapp/hyperapp/blob/a7cf41933f98e1474a90b66dec7c8adbb2fe93d6/src/index.js#L252-L257

jorgebucaran:
Let's merge this! 

Thanks! @38elements. This is one of the coolest and really simplest contributions in a long time that also save us some bytes! 🎉

PR #626, opened 2018-02-27
Provide global state and actions to subview function

frenzzy:
Continues #606 and implements @jorgebucaran idea from this comment: https://github.com/hyperapp/hyperapp/pull/606#issuecomment-368863726
```js
const Subview = props => state => <h1>{state}</h1>

const view = () => {
  return (
    <div>
      <Subview />
      <h1>Title</h1>
    </div>
  )
}
```
```js
const Subview = props => <h1>{props.title}</h1>

const view = () => {
  return (
    <div>
      {state => <Subview title={state} />}
      <h1>Title</h1>
    </div>
  )
}
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -16,14 +16,12 @@ export function h(name, attributes /*, ...rest*/) {
     }
   }
 
-  return typeof name === "function"
-    ? name(attributes || {}, children)
-    : {
-        nodeName: name,
-        attributes: attributes || {},
-        children: children,
-        key: attributes && attributes.key
-      }
+  return {
+    nodeName: name,
+    attributes: attributes || {},
+    children: children,
+    key: attributes && attributes.key
+  }
 }
 
 export function app(state, actions, view, container) {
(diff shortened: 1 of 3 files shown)


jorgebucaran:
@frenzzy I want to support only one use case. Either nodeName===Function or node===Function, but not both.

frenzzy:
But they are about different things, one about the subview function another one about the component.

jorgebucaran:
We need to choose one for Hyperapp.

jorgebucaran:
@frenzzy Also, go ahead and close #606 in favor fo this.

frenzzy:
This PR is more for collecting feedback.

> We need to choose one for Hyperapp.

I see these cases:
1. Components #606 `<div><Component/></div>`
2. Component + Subview #626  `<div><Component/></div>` and `<div>{()=>'subview'}</div>`
3. Subview `<div>{()=>'subview'}</div>`

You are saying we need to choose 1 or 3 but not 2, correct?

jorgebucaran:
@frenzzy If we didn't care about JSX support, this would be a lot easier to settle I'll make my own PR.

PR #629, opened 2018-03-01
Boolean attributes

### LINKED ISSUE #628: Certain attributes behave differently given the same value
__Environment:__
* `OSX 10.11.6`
* `Chrome 62.0.3202.94`
* `hyperapp@1.1.2`

Consider the following component:
```jsx
const Input = () =>
  <input type="text" spellcheck="false" aria-invalid="false" />;
```

I would expect this to be rendered as:
```html
<input type="text" spellcheck="false" aria-invalid="false" />
```

But the actual output is:
```html
<input type="text" spellcheck="true" aria-invalid="false" />
```

I'm able get the desired `spellcheck` value by setting it to an empty string in the component, but `aria-invalid` behaves differently
```jsx
const Input = () =>
  <input type="text" spellcheck="" aria-invalid="" />;
```

Now the rendered output is:
```html
<input type="text" spellcheck="false" aria-invalid />
```

Ultimately, I have to pass different values to the `spellcheck` and `aria-invalid` attributes in the component to get the same values in the output:
```jsx
const Input = () =>
  <input type="text" spellcheck="" aria-invalid="false" />;
```
```html
<input type="text" spellcheck="false" aria-invalid="false" />
```

frenzzy:
Fixes #628

With this change we may rich consistent behavior between client and server rendering for reserved DOM attributes such as `spellcheck` or `draggable`, see:

before
```js
<input spellcheck="false" /> // => <input spellcheck="true" /> <== Oh NO!
<input spellcheck={false} /> // => <input spellcheck="false" />
```

after
```js
<input spellcheck="false" /> // => <input spellcheck="false" /> <== Oh YES!
<input spellcheck={false} /> // => <input spellcheck="false" />
```

SSR
```js
<input spellcheck="false" /> // => <input spellcheck="false" />
<input spellcheck={false} /> // => <input />
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -186,7 +186,15 @@ export function app(state, actions, view, container) {
         } else {
           element.removeEventListener(name, eventListener)
         }
-      } else if (name in element && name !== "list" && name !== "type" && !isSvg) {
+      } else if (
+        name in element &&
+        name !== "list" &&
+        name !== "type" &&
+        name !== "draggable" &&
+        name !== "spellcheck" &&
+        name !== "translate" &&
+        !isSvg
+      ) {
         element[name] = value == null ? "" : value
       } else if (value != null && value !== false) {
         element.setAttribute(name, value)
--- a/test/dom.test.js
+++ b/test/dom.test.js
@@ -652,3 +652,30 @@ testVdomToHtml("events", [
     html: `<button id="clicked"></button>`
   }
 ])
+
+testVdomToHtml("boolean attributes", [
+  {
+    vdom: (
+      <main>
+        <input
+          checked={true}
+          spellcheck="true"
+          autocomplete="on"
+          translate="yes"
+        />
+        <input
+          checked={false}
+          spellcheck="false"
+          autocomplete="off"
+          translate="no"
+        />
+      </main>
+    ),
+    html: `
+        <main>
+          <input spellcheck="true" autocomplete="on" translate="yes">
+          <input spellcheck="false" autocomplete="off" translate="no">
+        </main>
+      `
+  }
+])


jorgebucaran:
What's the difference with https://github.com/hyperapp/hyperapp/pull/592?

frenzzy:
#592 behavior:

```js
// before
<input spellcheck="false" data-x="false" /> // => <input spellcheck="true" data-x="false" />
<input spellcheck={false} data-x={false} /> // => <input spellcheck="false" />

// after
<input spellcheck="false" data-x="false" /> // => <input spellcheck="true" data-x="false" />
<input spellcheck={false} data-x={false} /> // => <input spellcheck="false" data-x="false" />
```

so, this PR about reserved DOM attributes while #592 is about custom DOM attributes.

src/index.js
```diff
@@ -153,7 +153,7 @@ export function app(state, actions, view, container) {
         typeof value === "function" ||
         (name in element && name !== "list" && !isSVG)
       ) {
-        element[name] = value == null ? "" : value
+        element[name] = value == null || value === "false" ? "" : value
```
frenzzy: or we can use regex here: `/^false$/i.test(value)`
it will be more robust (users can use `False`, `FALSE` etc) but slower
```js
// as it
<input spellcheck="false" /> // => <input spellcheck="false" />
<input spellcheck="False" /> // => <input spellcheck="true" />
<input spellcheck="FALSE" /> // => <input spellcheck="true" />

// regexp
<input spellcheck="false" /> // => <input spellcheck="false" />
<input spellcheck="False" /> // => <input spellcheck="false" />
<input spellcheck="FALSE" /> // => <input spellcheck="false" />
```

Swizz: HTML spec allow different cases ?

frenzzy: Yes, HTML case insensitive

jorgebucaran:
Unrelated but related. Adding it here before I forget. https://github.com/developit/preact/pull/983

jorgebucaran:
@frenzzy But why not use `{false}`?

frenzzy:
> @frenzzy But why not use {false}?

Because `false` does't render attribute at all:
```js
<input data-example={false} /> // renders <input />
```
Also `"false"` already works correctly for most of attributes:
```js
<input data-example="false" /> // renders <input data-example="false" />
```
The problem only in bunch of DOM attributes such as `spellcheck`, `dragging` etc.

After this change **`"false"`** not longer converted to **`"true"`** for _SOME_ attributes. Which makes `false` value consistent between different attributes and server-side rendering.

jorgebucaran:
@frenzzy Why does "false" work for some attributes and not for others? I am trying to leave this information documented here for our records.

frenzzy:
> @frenzzy Why does "false" work for some attributes and not for others?

Because of the way how we modify attributes (props vs attributes):
```js
const el = document.createElement('input')

el.spellcheck = false                   // => <input spellcheck="false" />
el.spellcheck = 'false'                 // => <input spellcheck="true" /> !!! Noooo...
el.setAttribute('spellcheck', false)    // => <input spellcheck="false" />
el.setAttribute('spellcheck', 'false')  // => <input spellcheck="false" />
el.setAttribute('data-custom', false)   // => <input data-custom="false" />
el.setAttribute('data-custom', 'false') // => <input data-custom="false" />
```
see also #634

frenzzy:
Looks like hyperapp have problems with all kind of boolean attribute values `"true"`/`"false"`, `"yes"`/`"no"` and `"on"`/`"off"`, for example:
```js
// jsx                     => html
<div translate="yes" /> // => <div translate="yes"></div>
<div translate="no" />  // => <div translate="yes"></div> <= No!
```

Probably related: https://github.com/facebook/react/issues/10589

jorgebucaran:
I think we should make only `{true}` or `{false}` works.

frenzzy:
Demos:
- master branch: https://codepen.io/frenzzy/pen/Zxrrzz/left/?editors=0010
- current branch: https://codepen.io/frenzzy/pen/EEQoBP/left/?editors=0010

jorgebucaran:
Let's wait after 2.0. Then let's figure out what to do.

src/index.js
```diff
@@ -184,7 +184,12 @@ export function app(state, actions, view, container) {
         } else {
           element.removeEventListener(name, eventListener)
         }
-      } else if (name in element && name !== "list" && !isSvg) {
+      } else if (
+        name in element &&
+        name !== "list" &&
+        !(typeof value === "string" && typeof element[name] === "boolean") &&
```
jorgebucaran: @frenzzy What about only `typeof value === "string"`?

So, removing `typeof element[name] === "boolean"`.

jorgebucaran: Nevermind :D

jorgebucaran:
Thanks, @frenzzy! **1.2.8** 🎉

PR #635, opened 2018-03-04
Update README.md with DOM event info

mickeyvip:
Add a section with an example about DOM event usage in elements like <input />

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,7 +351,6 @@ export const Jumbotron = ({ text }) => (
 )
 ```
 
-### Action Handlers
 An action handler will be passed an DOM <code>event</code> object that can be used to get the current state of the HTML element.
 
 ```jsx
--- a/README.md
+++ b/README.md
@@ -351,6 +351,21 @@ export const Jumbotron = ({ text }) => (
 )
 ```
 
+### Action Handlers
+An action handler will be passed an DOM <code>event</code> object that can be used to get the current state of the HTML element.
+
+```jsx
+const view = (state, actions) => (
+  <div class="container">
+    <input 
+      type="search"
+      placeholder="Search..."
+      oninput={event => actions.search(event.target.value)}
+    />
+  </div>
+);
+```
+
 ### Lifecycle Events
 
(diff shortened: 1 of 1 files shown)


jorgebucaran:
Thanks, @mickeyvip. I think this is a good start. I don't want to add a new section "Action Handlers", however, because to a newcomer it will seem as if Hyperapp has this thing called "action handlers", when there is, in fact, no such thing. In fact, this has nothing to do with actions.

mickeyvip:
@jorgebucaran , I'll remove the section title

jorgebucaran:
@mickeyvip I am starting to think there's a missunderstanding here. Hyperapp does not pass the event object to your DOM element's event handler, the browser does. Since we're no wrapping DOM events in anyway, the current behavior is to be expected. There's nothing surprising about it.

jorgebucaran:
Thanks, @mickeyvip. I'll put this on hold until I can figure out how to explain this without bloating the README too much. Sorry, to keep you waiting and thanks again for contributing to the documentation! ❤️

mickeyvip:
Another thought - update the first example with `input` field:

```js
const { h, app } = hyperapp
// @jsx h 

const state = {
  count: 0,
  step: 1
}

const actions = {
  down: value => state => ({ count: state.count - state.step }),
  up: value => state => ({ count: state.count + state.step }),
  stepUpdated: value => state => {
    let newStep = parseInt(value, 10);
    return isNaN(newStep) ? state : {step: newStep};
  }
}

  <main>
    <h1>{state.count}</h1>
    <button onclick={() => actions.down(1)} disabled={state.count <= 0}>ー</button>
    <button onclick={() => actions.up(1)}>＋</button>
    <section>
      <label>Step</label>
      <input type="number" value={state.step} oninput={(e) => actions.stepUpdated(e.target.value)} />
    </section>
  </main>
)

```

jorgebucaran:
All of this stuff is going to change in 2.0, so we won't need this after all. Thanks, @mickeyvip.

PR #636, opened 2018-03-05
Allow Lazy components to return null as on the normal components

jonlov:
I was checking out the lazy components! I'm loving it!

I just found a little bug when you return null on a lazy component it will throw

```javascript
Uncaught TypeError: Cannot read property 'nodeName' of null on line 172
```

But this doesn't happen with normal components. So I had a look at the code and I realized that
resolveNode function will return null and so it means that node will be null
```javascript
  function resolveNode(node) {
    return typeof node === "function"
      ? resolveNode(node(globalState, wiredActions)) // will return undefined and will crash
      : node
  }
```

I fixed by just resolving normal node if the function is null like this:

```javascript
  function resolveNode(node) {
    return typeof node === "function"
      ? resolveNode(node(globalState, wiredActions)) || node
      : node
  }
```

I'm not sure if it's the best approach but I've been testing it and it works just fine.
Thanks for the lazy components it's awesome!!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,4 @@
-export function h(name, attributes /*...rest*/) {
+export function h(name, attributes) {
   var rest = []
   var children = []
   var length = arguments.length
@@ -7,10 +7,7 @@ export function h(name, attributes /*...rest*/) {
 
   while (rest.length) {
     var node = rest.pop()
-    if (
-      node &&
-      node.pop // isArray
-    ) {
+    if (node && node.pop) {
       for (length = node.length; length--; ) {
         rest.push(node[length])
       }
@@ -20,7 +17,7 @@ export function h(name, attributes /*...rest*/) {
   }
 
   return typeof name === "function"
-    ? name(attributes || {}, children) // h(Component)
+    ? name(attributes || {}, children)
     : {
         nodeName: name,
         attributes: attributes || {},
@@ -31,11 +28,11 @@ export function h(name, attributes /*...rest*/) {
 
 export function app(state, actions, view, container) {
   var map = [].map
-  var events = []
   var element = (container && container.children[0]) || null
-  var oldNode = element && recycleNode(element)
+  var oldNode = element && recycleElement(element)
+  var lifecycle = []
   var skipRender
-  var isFirstRender = true
+  var isRecycling = true
   var globalState = clone(state)
   var wiredActions = clone(actions)
 
@@ -43,14 +40,14 @@ export function app(state, actions, view, container) {
 
   return wiredActions
 
-  function recycleNode(element) {
+  function recycleElement(element) {
     return {
       nodeName: element.nodeName.toLowerCase(),
       attributes: {},
       children: map.call(element.childNodes, function(element) {
         return element.nodeType === 3 // Node.TEXT_NODE
           ? element.nodeValue
-          : recycleNode(element)
+          : recycleElement(element)
       })
     }
   }
@@ -68,9 +65,9 @@ export function app(state, actions, view, container) {
       element = patch(container, element, oldNode, (oldNode = node))
     }
 
-    skipRender = isFirstRender = false
+    skipRender = isRecycling = false
 
-    while (events.length) events.pop()()
+    while (lifecycle.length) lifecycle.pop()()
   }
 
   function scheduleRender() {
@@ -142,6 +139,10 @@ export function app(state, actions, view, container) {
     return node ? node.key : null
   }
 
+  function eventListener(event) {
+    return event.currentTarget.events[event.type](event)
+  }
+
   function updateAttribute(element, name, value, oldValue, isSVG) {
     if (name === "key") {
     } else if (name === "style") {
@@ -149,10 +150,19 @@ export function app(state, actions, view, container) {
         element[name][i] = value == null || value[i] == null ? "" : value[i]
       }
     } else {
-      if (
-        typeof value === "function" ||
-        (name in element && name !== "list" && !isSVG)
-      ) {
+      if (name[0] === "o" && name[1] === "n") {
+        if (!element.events) {
+          element.events = {}
+        }
+        element.events[(name = name.slice(2))] = value
+        if (value) {
+          if (!oldValue) {
+            element.addEventListener(name, eventListener)
+          }
+        } else {
+          element.removeEventListener(name, eventListener)
+        }
+      } else if (name in element && name !== "list" && !isSVG) {
         element[name] = value == null ? "" : value
       } else if (value != null && value !== false) {
         element.setAttribute(name, value)
@@ -178,7 +188,7 @@ export function app(state, actions, view, container) {
     var attributes = node.attributes
     if (attributes) {
       if (attributes.oncreate) {
-        events.push(function() {
+        lifecycle.push(function() {
           attributes.oncreate(element)
         })
       }
@@ -218,9 +228,9 @@ export function app(state, actions, view, container) {
       }
     }
 
-    var cb = isFirstRender ? attributes.oncreate : attributes.onupdate
+    var cb = isRecycling ? attributes.oncreate : attributes.onupdate
     if (cb) {
-      events.push(function() {
+      lifecycle.push(function() {
         cb(element, oldAttributes)
       })
     }
@@ -301,7 +311,7 @@ export function app(state, actions, view, container) {
           continue
         }
 
-        if (newKey == null || isFirstRender) {
+        if (newKey == null || isRecycling) {
           if (oldKey == null) {
             patch(element, oldElements[i], oldChildren[i], children[k], isSVG)
             k++
--- a/src/index.js
+++ b/src/index.js
@@ -56,9 +56,11 @@ export function app(state, actions, view, container) {
   }
 
   function resolveNode(node) {
-    return typeof node === "function"
-      ? resolveNode(node(globalState, wiredActions)) || h(node, null)
-      : node
+    return node === null
+      ? h(node, null)
+      : typeof node === "function"
+        ? resolveNode(node(globalState, wiredActions)) || h(node, null)
+        : node
   }
 
   function render() {
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,4 @@
-export function h(name, attributes) {
+export function h(name, attributes /*...rest*/) {
   var rest = []
   var children = []
   var length = arguments.length
@@ -7,7 +7,10 @@ export function h(name, attributes) {
 
   while (rest.length) {
     var node = rest.pop()
-    if (node && node.pop) {
+    if (
+      node &&
+      node.pop // isArray
+    ) {
       for (length = node.length; length--; ) {
         rest.push(node[length])
       }
@@ -16,9 +19,7 @@ export function h(name, attributes) {
     }
   }
 
-  return name === null
-    ? null
-    : typeof name === "function"
+  return typeof name === "function"
     ? name(attributes || {}, children) // h(Component)
     : {
         nodeName: name,
@@ -30,11 +31,11 @@ export function h(name, attributes) {
 
 export function app(state, actions, view, container) {
   var map = [].map
+  var events = []
   var element = (container && container.children[0]) || null
-  var oldNode = element && recycleElement(element)
-  var lifecycle = []
+  var oldNode = element && recycleNode(element)
   var skipRender
-  var isRecycling = true
+  var isFirstRender = true
   var globalState = clone(state)
   var wiredActions = clone(actions)
 
@@ -42,24 +43,22 @@ export function app(state, actions, view, container) {
 
   return wiredActions
 
-  function recycleElement(element) {
+  function recycleNode(element) {
     return {
       nodeName: element.nodeName.toLowerCase(),
       attributes: {},
       children: map.call(element.childNodes, function(element) {
         return element.nodeType === 3 // Node.TEXT_NODE
           ? element.nodeValue
-          : recycleElement(element)
+          : recycleNode(element)
       })
     }
   }
 
   function resolveNode(node) {
-    return node === null
-        ? h(node, null)
-        : typeof node === "function"
-          ? resolveNode(node(globalState, wiredActions))
-          : node
+    return typeof node === "function"
+      ? resolveNode(node(globalState, wiredActions))
+      : node || ''
   }
 
   function render() {
@@ -69,9 +68,9 @@ export function app(state, actions, view, container) {
       element = patch(container, element, oldNode, (oldNode = node))
     }
 
-    skipRender = isRecycling = false
+    skipRender = isFirstRender = false
 
-    while (lifecycle.length) lifecycle.pop()()
+    while (events.length) events.pop()()
   }
 
   function scheduleRender() {
@@ -143,10 +142,6 @@ export function app(state, actions, view, container) {
     return node ? node.key : null
   }
 
-  function eventListener(event) {
-    return event.currentTarget.events[event.type](event)
-  }
-
   function updateAttribute(element, name, value, oldValue, isSVG) {
     if (name === "key") {
     } else if (name === "style") {
@@ -154,19 +149,10 @@ export function app(state, actions, view, container) {
         element[name][i] = value == null || value[i] == null ? "" : value[i]
       }
     } else {
-      if (name[0] === "o" && name[1] === "n") {
-        if (!element.events) {
-          element.events = {}
-        }
-        element.events[(name = name.slice(2))] = value
-        if (value) {
-          if (!oldValue) {
-            element.addEventListener(name, eventListener)
-          }
-        } else {
-          element.removeEventListener(name, eventListener)
-        }
-      } else if (name in element && name !== "list" && !isSVG) {
+      if (
+        typeof value === "function" ||
+        (name in element && name !== "list" && !isSVG)
+      ) {
         element[name] = value == null ? "" : value
       } else if (value != null && value !== false) {
         element.setAttribute(name, value)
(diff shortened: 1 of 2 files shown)


jorgebucaran:
@jonlov What happens when you return `null` from your ~lazy component~ → **subview*? Can you check the HTML in the dev console?

src/index.js
```diff
@@ -57,7 +57,7 @@ export function app(state, actions, view, container) {
 
   function resolveNode(node) {
     return typeof node === "function"
-      ? resolveNode(node(globalState, wiredActions))
+      ? resolveNode(node(globalState, wiredActions)) || node
```
frenzzy: When `node` is a function:
1. `node(gstate, gactions)` => `null`
2. `resolveNode(null)` => `null`
3. `null || function` => `function`

`function` is not a valid vnode type

jonlov:
@jorgebucaran It shows
```javascript
Cannot read property 'nodeName' of null
```
and the line that throws the error is [line 268]
I created a [Codepen](https://codepen.io/0n/pen/jZRMwP)

jorgebucaran:
@jonlov No, I mean that question with your changes, not the current behavior. I already know the current behavior.

jonlov:
@jorgebucaran there are no <null></null> or <undefined></undefined> garbage tags. Is a weird behaviour. You can check on this [Codepen](https://codepen.io/0n/pen/BYeOKO) and play with it while using console log on the resolveNode function.

@frenzzy is right

> When node is a function:
> 
> 1. node(gstate, gactions) => null
> 2. resolveNode(null) => null
> 3. null || function => function
> 
> function is not a valid vnode type, expected object: { nodeName, attributes, children }

src/index.js
```diff
@@ -57,7 +57,7 @@ export function app(state, actions, view, container) {
 
   function resolveNode(node) {
     return typeof node === "function"
-      ? resolveNode(node(globalState, wiredActions)) || h(node(globalState, wiredActions), null))
+      ? resolveNode(node(globalState, wiredActions)) || h(node, null)
```
frenzzy: `h(node)` same as `h(Component)`. For component which returns null the `h` function will also return `null`.
This line become `null || null`

src/index.js
```diff
@@ -59,7 +59,7 @@ export function app(state, actions, view, container) {
     return node === null
       ? h(node, null)
```
jonlov: @frenzzy What do you think if we check in all the nodes if there's any null node ?

src/index.js
```diff
@@ -56,9 +56,11 @@ export function app(state, actions, view, container) {
   }
 
   function resolveNode(node) {
-    return typeof node === "function"
-      ? resolveNode(node(globalState, wiredActions))
-      : node
+    return node === null
+      ? h(node, null)
```
frenzzy: `h(null)` returns `{nodeName: null, attributes: {}, children: [], key: undefined}` which is invalid vnode for `patch` function

src/index.js
jonlov: Yes @frenzzy, I just realize that, thank you!

src/index.js
jonlov: @frenzzy I see, any suggestions? Do you think we should modify it on the patch function?

frenzzy: We may add support for vnode type=array inside patch function or implement proper fragments support #530 and return fragment node here

jonlov: When you mean is an invalid vnode for `patch` function, what it'll happen when `nodeName: null`? I'm using it right now with the route and it will just render `<null></null>` on the dom but then is removed. It works perfectly and does not show any error or that it renders `<null></null>` is bad? What sort of errors could happen?

jonlov:
@frenzzy @jorgebucaran Could you guys check this [CodePen](https://codepen.io/0n/pen/BYeOKO) ? It has the fix that I made. The weird part is that a component that has state and actions renders `<null></null>` on the dom. However, a normal component doesn't create any element on the dom (how it should be when returning null). 
Could anyone explain this behaviour?

src/index.js
frenzzy: `null` should render nothing

Swizz:
`NormalComponent` is used as `h(NormalComponent)` that return `null`.
`ComponentWithStateAndActions` is used as `h(ComponentWithStateAndActions(globalState, wiredActions))`, that will be `h(null)` and this return a vnode with `null` as `nodeName`.

I suggest `h` to skip when `name` is `null`. This will avoid further troubles and fixs in core.

https://github.com/hyperapp/hyperapp/blob/62feb73302da9c02d04c16670804b472609c566f/src/index.js#L22-L25

frenzzy:
> I suggest `h` to skip when `name` is `null`. This will avoid further troubles and fixs in core.

@Swizz it is already happens on this line:

https://github.com/hyperapp/hyperapp/blob/62feb73302da9c02d04c16670804b472609c566f/src/index.js#L17

Swizz:
`h(null)` do not return `null`, but a vnode with `null` as `nodeName` instead : 

```jsx
Object {
  nodeName: null,
  attributes: {},
  children: [],
  key:  undefined
}
```
And there is no guard either in `h` nor `patch` nor `createElement`, that lead to 
https://github.com/hyperapp/hyperapp/blob/62feb73302da9c02d04c16670804b472609c566f/src/index.js#L176

And `document.createElement(null)` return `<null></null>`

frenzzy:
@Swizz there are no any valid reason why you should call `h(null)` instead of just `null`, `h` accepts node name or a component as the first argument and may return `null` if component returns null. So, `h` is perfect, no need to change it.

Swizz:
So `null` is a correct `nodeName` ? And should produce a `<null></null>` vnode (then DOM node) ?

frenzzy:
Component which returns `null` should not generate nullable vnode, it should render nothing. If you trying to use `null` as a vnode name `h(null)`, then probably `<null></null>` behavior is expected.

src/index.js
Swizz: Following our discussion in this PR, this line should definitely not lead to `h(null)`

Swizz: Following our discussion in this PR, this line should definitely not lead to h(null)

jonlov:
I understand what's happening, NormalComponent returns null at first like this

```javascript
var NormalComponent = function NormalComponent(attributes, children) {
    return null; //this return null at first which makes 
};
```
which makes this line be true
https://github.com/hyperapp/hyperapp/blob/62feb73302da9c02d04c16670804b472609c566f/src/index.js#L17

On the other hand ComponentWithStateAndActions returns a function and it doesn't make that line to be true
```javascript
var ComponentWithStateAndActions = function ComponentWithStateAndActions(attributes, children) {
    return function(actions, state) {//it returns a function and then returns null
        return null;
    };
};
```

src/index.js
jonlov: If I remove that line and I `return null` into `ComponentWithStateAndActions` it will throw an error on the console because it can't get `nodeName` from undefined `node.nodeName`

jonlov:
@frenzzy I agree with @Swizz we should avoid `h(null, null)` or `h(undefined, undefined)`. If someone wants to use `null` as a tag name they could do `h("null", null)` or in jsx `<null></null>` which will be a string inside the `h` function.

Another issue is that this line just removes the child elements, so then doesn't get rendered.
https://github.com/hyperapp/hyperapp/blob/62feb73302da9c02d04c16670804b472609c566f/src/index.js#L17

Example:
```es6
const NormalComponent = (attributes, children) => {
  return null;
}

const Component = (attributes, children) => {
  return <NormalComponent/>;
}
```
it will append a `<null></null>` element to the dom since `NormalComponent` is not a child.

Now the question would be, do we need to change the patch function so it allows null and skip the element? Doing this not only childs but parents woudn't render. I don't wanna add a bunch of code just for this small feature, is not really super important but probably there's a good way of implementing it. Do have anything in mind?

jorgebucaran:
@jonlov The feature is to be able to return `null` from a subview?

jonlov:
@jorgebucaran Yes, basically you should be able to `return null` into any component or a component with states and view. Then hyperapp should avoid rendering `<null></null>`.

jimfranke:
I'm experiencing this issue when I try to return a Redirect component from the Hyperapp-Router inside a subview/lazy-component. It happens when returning nothing (undefined) or null.

jorgebucaran:
@jonlov We are not required to support returning null from subviews. The top-level view can't be null either. So, we could simple close as wontfix.

jonlov:
@jorgebucaran I created this PR because I was getting the same error as @jimfranke on the router. What would return the Redirect component? A empty div?

jorgebucaran:
@jonlov But components can return `null`. The issue is with the not-yet-published subviews returning <samp>null</samp>. Right?

jorgebucaran:
> @jonlov A empty div?

I wouldn't want to return an empty div either. 👍

frenzzy:
As a workaround here, you may just return an empty string when you want to render nothing.
Or maybe it is a better idea to create a component for this case:
```js
const Empty = () => <span />
```
and return `<Empty />` instead of `null`. But later when hyperapp will support returning nulls from (sub)views you will be able to just change one place in your app:
```js
const Empty = () => null
```

src/index.js
```diff
@@ -16,8 +16,10 @@ export function h(name, attributes) {
     }
   }
 
-  return typeof name === "function"
-    ? name(attributes || {}, children)
+  return name === null
```
jorgebucaran: @jonlov Why would `name` be null here? I don't want to support that.

jonlov:
@jorgebucaran Yes, I wouldn't like to return an empty element either. I found the latest fix that I could make. You can check it out here [Codepen](https://codepen.io/0n/pen/BYeOKO). I'm not sure if it's the best approach but because the patch function is a bit complex I didn't want to mess it too much.

src/index.js
jorgebucaran: Now, I see that this is just how you implemented the feature overall. Hmm, thanks for trying. I need to collect my thoughts and see how to go about allowing nulls from subviews. Maybe just return `""`?

frenzzy:
@jonlov FYI, the simplest fix for this issue is `|| ''` on this line:
https://github.com/hyperapp/hyperapp/blob/ec5759566e31afcd5fe1fec76675f9e2f348c724/src/index.js#L58
But I am pretty sure @jorgebucaran will not accept it because it is not a proper fix. The only proper fix is modifying `patch` function.

jonlov:
@frenzzy how come I never tried that! Haha, probably I was too tired. That works perfectly. I think that's a really simple solution and doesn't mess too much with the patch function. I think we could merge that as a temporary solution. I will push your solution just in case you would like to merge it as a temporary fix.

jorgebucaran:
@jonlov @frenzzy Okay, let's consider the downsides to this approach. Can we get away with it and what would be the [technical debt](https://en.wikipedia.org/wiki/Technical_debt) if we concede to this easy way out?

frenzzy:
The `h` function does not remove text nodes with an empty string from the tree:
https://github.com/hyperapp/hyperapp/blob/ec5759566e31afcd5fe1fec76675f9e2f348c724/src/index.js#L14
so, this change also make sense IMO.
But the right way probably filter out empty strings too, to improve batch performance.

jorgebucaran:
> @frenzzy But the right way probably filter out empty strings too. It may improve patch performance.

I am okay with this. 👍

jorgebucaran:
Are ready to merge this or do we want to wait for something else? @frenzzy @jonlov

jonlov:
@jorgebucaran It's ready to merge, I added the filtering for empty nodes.

src/index.js
```diff
@@ -11,7 +11,7 @@ export function h(name, attributes) {
       for (length = node.length; length--; ) {
         rest.push(node[length])
       }
-    } else if (node != null && node !== true && node !== false) {
+    } else if (node != null && node !== true && node !== false && node !== '') {
```
jorgebucaran: @jonlov 🤔 Why?

jorgebucaran:
@jonlov Include your tests in the same PR, please. 👍

src/index.js
jonlov: @jorgebucaran as freenzy mentioned before the h function does not remove text nodes with an empty string from the tree so by filtering out empty strings may improve patch performance.

jonlov:
@jorge sure! When I arrive home I’ll do it

src/index.js
frenzzy: But for now you are returning empty string `""` from components anyway.
And second, who explicitly returns empty strings from components? :)
Maybe change in this PR only `node || ''` plus a test case for it?

jonlov: @freenzy but why do we want to push an empty node? It makes a slightly difference on the benchmark

jorgebucaran: @jonlov What benchmark?

jonlov: Sorry, benchmarks I meant performance wise. We’re pushing into an array an empty string that is not going to be used which is unnecessary I think.

jorgebucaran: Right, but how often do people return `null` from lazy/wired components?

jonlov: That’s true, so we rather safe some kbs and get some penalty on performance just if anyone `return null`. I’ll do the test and revert the last commit then!

jonlov:
@jorgebucaran This is ready to merge!

src/index.js
```diff
@@ -55,7 +55,7 @@ export function app(state, actions, view, container) {
   function resolveNode(node) {
     return typeof node === "function"
       ? resolveNode(node(globalState, wiredActions))
-      : node
+      : node || ''
```
jorgebucaran: @jonlov This will fail when the node resolves to the number 0.

Let's try:

```js
node == null ? "" : node
```

jonlov: @jorgebucaran I change it. May I ask, why would ever a node be 0?

jorgebucaran: @jonlov A node can be also a string or a number that will be casted as a TextNode. Maybe you have a view that returns the multiplication table and a specific subview that returns only a number. It's a shame we need to complicate the expression, but nodes can be strings or numbers, so if this code was written, e.g., in TypeScript we would get a type mismatch error.

src/index.js
```diff
@@ -55,7 +55,9 @@ export function app(state, actions, view, container) {
   function resolveNode(node) {
     return typeof node === "function"
       ? resolveNode(node(globalState, wiredActions))
-      : node || ''
+      : node != null
```
jorgebucaran: @jonlov Did you run `npm run format`? Not sure if prettier would do this.

okwolf: @jorgebucaran if you're interested in checking the format using the build/CI: https://github.com/hyperapp/fx/blob/7be65b2983a6b0c9e732840343be4b458d8e34e4/package.json#L27

jonlov: @jorgebucaran no I didn't. I wrote it like that. It looks nicer to have the empty string at the end. Is it wrong to do it like this?

src/index.js
jonlov: @jorgebucaran oh got it! Thanks for the explanation.

test/app.test.js
```diff
@@ -63,3 +63,45 @@ test("lazy components", done => {
 
   app(state, actions, view, document.body)
 })
+
+
+test("returning null on a component", done => {
```
okwolf: Isn't this test the lazy component?

test/app.test.js
```diff
@@ -63,3 +63,45 @@ test("lazy components", done => {
...
+      [nullComponent]
+    )
+
+  const view = () => h(Component)
+
+  app(null, null, view, document.body)
+})
+
+
+test("returning null on a lazy component", done => {
```
okwolf: Isn't this test the non-lazy component?

jonlov: @okwolf I swap the names lol. Thank you!

jonlov: @okwolf Is fixed now and ready to merge!

src/index.js
jorgebucaran: @jonlov  https://github.com/hyperapp/hyperapp/blob/master/CONTRIBUTING.md#style

> We use Prettier to format the code. If you don't have Prettier installed in your code editor or IDE, you can use npm run format before adding a commit.

jonlov: @jorgebucaran sorry, didn't know.

jorgebucaran:
This looks good, now the only issue still bothering me a bit is, if we are going to turn `null` into `""` for you, why not let you (the user) do it explicitly? What is the gain here?

jonlov:
Just for me? If it’s going to be just for me then we don’t do it. This is how other frameworks work and it was working with the normal components. Now with the lazy components it will break. For example the Redirect on the router broke and that’s why I created this. But now that we know how to solve it. The Redirect component could return an empty string. But I think there’s a lot of people that uses `return null`. Because that’s how it’s done in other frameworks. If hyperapp won’t support it then how the user is going to know that hyperapp don’t allow `return null` and it uses `return ””` instead? Add it to documentation? 

I’m fine with saving kbs and using either of the solutions. I’m just a bit concern about future users that come from other frameworks and try to do some common jsx like `return null` and the app will crash. Because hyperapp doesn’t have specific errors they wouldn’t know what’s going on. So that means they need to do research and to see what’s going on. So it’s time that could be save for them. 

What would be the best way to document this or should we make a development hyperapp that throws errors ?

jorgebucaran:
No, not you, I mean users in general.

Thanks, for your input, you raise some valid concerns. 

I'll think about it :)

jonlov:
@jorgebucaran haha that's fine. It's the same with the height on the img tag. It works without it but if I'm a new user that doesn't get specific errors and probably will move on because of the time. If we're not going to use this then we should create a development version with specifics errors if we want new developers that are not that advanced to be learning hyperapp. Just let me know what you decide and if I can help with something else. No rush 😊

jorgebucaran:
Sure! 👋

jorgebucaran:
@jonlov Let's merge it! Could you please resolve the conflicts?

Also, now that we are using JSX in tests, let's use it here as well. 🙏

jonlov:
@jorgebucaran awesome!! Is ready to merge now :)

PR #638, opened 2018-03-06
Delegate DOM events to proxy.

jorgebucaran:
Instead of setting event handlers every time we patch the DOM like <samp>button.onclick = handler</samp> for every element, add a single event listener (only once) to every element with an event handler and only the first time we see it. 

Since all elements share the same listener, which merely acts as a proxy, we can expect moderate perf improvements, nothing like what #499 is aiming to do, but something.

This change also goes back to using addEventListener/removeEventListener, which may have some other favorable side effects like the ability to set certain events that only work via addEventListener (I can't find the issue now, perhaps it was a convo on Slack).

**+** 72 bytes 😅

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,4 @@
-export function h(name, attributes /*...rest*/) {
+export function h(name, attributes) {
   var rest = []
   var children = []
   var length = arguments.length
@@ -7,10 +7,7 @@ export function h(name, attributes /*...rest*/) {
 
   while (rest.length) {
     var node = rest.pop()
-    if (
-      node &&
-      node.pop // isArray
-    ) {
+    if (node && node.pop) {
       for (length = node.length; length--; ) {
         rest.push(node[length])
       }
@@ -20,7 +17,7 @@ export function h(name, attributes /*...rest*/) {
   }
 
   return typeof name === "function"
-    ? name(attributes || {}, children) // h(Component)
+    ? name(attributes || {}, children)
     : {
         nodeName: name,
         attributes: attributes || {},
@@ -31,11 +28,11 @@ export function h(name, attributes /*...rest*/) {
 
 export function app(state, actions, view, container) {
   var map = [].map
-  var events = []
   var element = (container && container.children[0]) || null
-  var oldNode = element && recycleNode(element)
+  var oldNode = element && recycleElement(element)
+  var lifecycle = []
   var skipRender
-  var isFirstRender = true
+  var isRecycling = true
   var globalState = clone(state)
   var wiredActions = clone(actions)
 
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -142,17 +142,30 @@ export function app(state, actions, view, container) {
...
     } else {
-      if (
-        typeof value === "function" ||
-        (name in element && name !== "list" && !isSVG)
-      ) {
+      if (name[0] === "o" && name[1] === "n") {
+        if (!element.events) {
+          element.events = {}
+        }
+        element.events[(name = name.slice(2))] = value
```
okwolf: What do you think of adding `.toLowerCase()` for supporting the common variations like `onClick`?

jorgebucaran:
@okwolf The gain is _not_ touching the DOM element's event handler.

jorgebucaran:
@okwolf I edited my cryptic answer, see again please.

jorgebucaran:
> @okwolf I'm not sure what black magic the browser does when you touch elements in each case. 

This.

frenzzy:
Looks like this PR unlocks an ability to use events such as [`focusin`](https://developer.mozilla.org/en-US/docs/Web/Events/focusin), [`focusout`](https://developer.mozilla.org/en-US/docs/Web/Events/focusout) and so on.
```js
<input
  onfocus={e => console.log('focus')}
  onfocusin={e => console.log('focus in')} // does not work in current version
/>
```

src/index.js
```diff
@@ -142,17 +142,30 @@ export function app(state, actions, view, container) {
...
     } else if (name === "style") {
       for (var i in clone(oldValue, value)) {
         element[name][i] = value == null || value[i] == null ? "" : value[i]
       }
     } else {
-      if (
-        typeof value === "function" ||
-        (name in element && name !== "list" && !isSVG)
-      ) {
+      if (name[0] === "o" && name[1] === "n") {
```
SkaterDad: This will add lifecycle events and any other properties starting with "on" to the `element.events` object, right?   Any downside to this?  I suspect not, but thought it was worth asking.

jorgebucaran: Yes, you are right. I can't think of any downsides, can you?

* oncreate/onupdate/onremove/ondestroy: Will get added too, although never used. However, this is still better than our _current_ behavior, where we add each of them them directly on the element.

SkaterDad: I agree, it's probably a non-issue.

jorgebucaran:
Merge? :shipit:

SkaterDad:
> Merge? :shipit:

Might as well!

Just curious, though, if anyone has done some before/after CPU profiling or benchmarking to put real numbers to this?

jorgebucaran:
@SkaterDad I run it through js-framework-benchmarks and there was a wonderful (moderate) improvement. 20% better perf overall.

SkaterDad:
Impressive!

src/index.js
frenzzy: Doesn't this line introduce memory leak? I do not see `element.events` cleanup...

jorgebucaran: @frenzzy Good question. We are doing the clean up here:

```js
element.events[(name = name.slice(2))] = value
```

...when `value` is null or undefined, as well as calling `removeEventListener`.

frenzzy:
probably because regex is slower than `===`

SkaterDad:
@infinnie & @frenzzy 

The regex version is 70% slower in JSPerf benchmarking in both IE11 and Chrome on my machine:
https://jsperf.com/regex-vs-triple-eq121321321

PR #639, opened 2018-03-06
Add docs for subviews.

jorgebucaran:
Add some basic documentation for subviews. Needs a lot more work, the current example is not very realistic either. It's just a start.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,17 +181,16 @@ setInterval(main.down, 500, 1)
 
 ### View
 
-Every time your application state changes, the view function is called so that you can specify how you want the DOM to look based on the new state.
+Every time your application state changes, the view function is called with the state and actions so that you can specify how you want the DOM to look based on the new state.
 
 ```js
-import { h } from "hyperapp"
-
-const view = (state, actions) =>
-  h("div", {}, [
-    h("h1", {}, state.count),
-    h("button", { onclick: () => actions.down(1) }, "-"),
-    h("button", { onclick: () => actions.up(1) }, "+")
-  ])
+const view = (state, actions) => (
+  <div>
+    <h1>{state.count}</h1>
+    <button onclick={() => actions.down(1)}>-</button>
+    <button onclick={() => actions.up(1)}>+</button>
+  </div>
+)
 ```
 
 The view returns your specification in the form of a plain JavaScript object known as a [virtual DOM](#virtual-dom) and Hyperapp takes care of updating the actual DOM to match it.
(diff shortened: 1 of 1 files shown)


SkaterDad:
I finally did some playing around with these sub-views, and I'm really liking it.  I expect my application bundle size will go down significantly without all the property-passing :+1: 

Here's a quick demo I just built in Codepen (pointing at the rawgit src file).  It demonstrates having 2 sub-views nested a few elements deep into the main view.  Too big for the docs, but perhaps it will serve as inspiration.
https://codepen.io/SkaterDad/pen/ddBEdg?editors=0110

A small, but practical example of how these subviews are useful is a router <Link> component which needs to be wired up to an action (like the official one used to do).  You can now sprinkle those around your app without passing the `route`/`navigate` actions everywhere.

okwolf:
This makes me wonder if I should update my blog article or perhaps write a new one.

What I’m looking forward to are solutions built on this to act like memorized selectors for you automatically.

My main concern is the complexity this introduces for testing. Might want some more helpers there also.

jorgebucaran:
@okwolf What problem do you have testing subviews?

okwolf:
@jorgebucaran the struggle is that it’s fairly simple to unit test stateless components, but sub views beg for more of an integration test since you potentially need state and actions across multiple slices.

Once I update `@hyperapp/fx` to support this I’ll see how the testing story looks. 🔬

jorgebucaran:
@okwolf Testing anything is difficult because we are _"pragmatic"_. That's the cost we pay. 😭

jorgebucaran:
Thanks, @SkaterDad. That's exactly the feedback I was looking for. 

Anyone else has feedback? 🙏

lukejacksonn:
Yup, all looks good. As @SkaterDad says, it is a simple enough concept to grasp, especially considering actions behave the same. I haven't had time to try it out yet but I look forward to it.

jorgebucaran:
@lukejacksonn Good point about actions behaving the same! 💯

By the way, I went with term "lazy components" (7617d8b7713f53a7cd8d9c9ffd840f8babdc83a1) as I found it easier to document. These are all only docs so if anyone has a better suggestion I am open to change it to whatever people want.

jorgebucaran:
Another idea for this term could be Wired Components, in line with our Wired Actions concept and taken from @lukejacksonn's [comment above](https://github.com/hyperapp/hyperapp/pull/639#issuecomment-371450437).

frenzzy:
I would prefer `subview` term and document it something like this:

`subview` is a function which is identical to `[`view`](#view)` but could be nested in any place of your virtual dom tree. Unlike to components, subview is called only after all other conditions are resolved which allows you to skip bunches of code during the construction of a virtual tree:
```js
const condition = false
<div>
  {condition && <Component message="always executed" />)
  {condition && (state, actions) => <Component message="not executed" />}
</div>
```
The most common use case for subview is providing access to global state and actions to your components:
```js
// example from this PR here
```

jorgebucaran:
Thanks, @frenzzy. I _personally_ found it easier to document as Lazy Components. But I wouldn't oppose to go back to subviews if you or someone else makes a PR based on your comment with spotless and less vernacular wording.

SkaterDad:
@frenzzy In your example, I don't think either `<Component />` would get executed since JS stops evaluating after the false `condition`.  Codepen wouldn't let put the arrow function after `&&`, so it's defined as a separate lazy component.   Increment the counter and watch logs. https://codepen.io/SkaterDad/pen/jzObVq?editors=0011

I like `Lazy Components` as a term, but `Wired Components` is also nice and illustrates their real benefit a bit better.  Either way, I'm excited for this feature to get released. 🥂

zaceno:
Saying "Hyperapp has lazy-components" only tells you that Hyperapp supports late evaluation of components. It doesn't really tell you why you'd want that (i e to access state/actions). So imo calling it "lazy components" distracts from the point of the feature.

PR #640, opened 2018-03-06
fix types according with ultradom types

dancespiele:


hyperapp.d.ts
```diff
@@ -6,48 +6,34 @@ export as namespace hyperapp
  *
  * @memberOf [VDOM]
  */
-export interface VNode<Props> {
-  name: string
-  props: Props
-  children: VNodeChild<object | null>[]
+export interface VNode<Attributes = {}> {
+  nodeName: string
+  attributes?: Attributes
```
frenzzy: but `attributes` are not optional, they are required, isn't it?
maybe `key` is optional...

okwolf: I know they can be `null` but I’m not sure if they can be absent entirely.

jorgebucaran: This is okay.
```js
h(nodeName, null, null)
```

But this is not.

```
{
  nodeName, attributes: null, children: null 
}
```

jorgebucaran: @dancespiele ☝️

okwolf: Ah, so the ~~`h`~~ `createNode` function handles that for you?

Issue #641, opened 2018-03-06
Use Ultradom as a dependency?  🤔


jorgebucaran:
Like the title says. 
- https://github.com/jorgebucaran/ultradom

okwolf:
Should we use it as a `devDependency` so it’s still all bundled into a single dependency for users to install?

jorgebucaran:
Hmm? I don't think that's how it works. 🤔

okwolf:
Oh yeah, that would only work for the UMD build. 🙃

zaceno:
I think it would be a good thing! But since youve been opposed to this in the past, I’m curious: what changed your mind?

mrozbarry:
I think it's better to have to separate things. It makes it easier for people to contribute to one or the other, plus it avoids duplicate code (especially if you're back-porting all updates to hyperapp to ultradom).

SkaterDad:
How would that look?  I assumed the reason you kept them separate until now was that hyperapp needed to be able to hook deeper into v-dom to make it smaller & easier to customize?

zaceno:
@mrozbarry I found your comment a bit ambiguous: Do you mean “separate things” as in “Hyperapp should keep its own vdom implementation separate from Ultradom so they are free to diverge” or did you mean: “Hyperapp should reuse a separate vdom (Ultradom) rather than implement its own”

zaceno:
@mrozbarry sorry, reread your post. The end made it more clear you meant the latter (... right?)

okwolf:
Are there any downsides to this other than no longer having a single self-contained source file you can import the Hyperapp module from?

Swizz:
If Ultradom become a dependency, I would want to add on table the fact of making Hyperapp more like a state manager and Ultradom the renderer.

jorgebucaran:
@SkaterDad True. For this to work, I would need to complicate Ultradom's API a bit more.

Now

```js
patch(VNode, Element)
```
After
```js
patch(VNode, Element, { resolveNode, proxyEvent, ...other, ...options })
```

Mytrill:
I see this change as a good thing as it reduce repetition.
Regarding ultradom's API change, isn't it possible to rewrite `render()` to re-use ultradom's `patch()` as-is?

okwolf:
This wouldn't hurt performance by pulling in a more general-purpose Ultradom, correct?

jorgebucaran:
[...]

PR #642, opened 2018-03-06
Add Size Limit

ai:
It is awesome, that you try to take care of library size. Maybe [Size Limit](https://github.com/ai/size-limit) could help you with it.

Size Limit is some sort of benchmark but for size.

It is a special tool to prevent JS libraries bloating. It calculates real cost of using the library as a dependency. It runs for every PR and will warn you if the size will be changed very quickly.

@jorgebucaran MobX, Autoprefixer, Material-UI uses it too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ install:
 
 script:
   - npm test
+  - npm run size
   - codecov
 
 notifications:
--- a/package.json
+++ b/package.json
@@ -24,17 +24,25 @@
     "bundle": "rollup -i src/index.js -o dist/hyperapp.js -m -f umd -n hyperapp",
     "minify": "uglifyjs dist/hyperapp.js -o dist/hyperapp.js -mc pure_funcs=['Object.defineProperty'] --source-map includeSources,url=hyperapp.js.map",
     "prepare": "npm run build",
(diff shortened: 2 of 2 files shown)


Swizz:
Thank you for this awesome contribution !
I wanted to see `size-limit` used here for a while. 

We used `bundlesize` a while ago, I do not remember why we took the choice to remove it.

jorgebucaran:
Thanks, @ai! Like @Swizz commented, we've had this before, but I removed it because I want to keep the README clean and simple. Now, I may re-consider this now that we're slowing moving away from a fixed 1 kB limit into a more flexible ~ 1.6 kB or so and I really want to stop golfing JavaScript for no good reason. 😅

jorgebucaran:
@ai See [this comment](https://github.com/hyperapp/hyperapp/issues/641#issuecomment-371383655) for some background as to why I am not going to merge this.

> I am no longer that fervently obsessed about the file size and as many of you may have noticed, I've been improving the code readability at the expense of more bytes here and there. I am more open to changes than I was ever before.

We're still going to be tiny & ultra lightweight though, but obsessing about the size is no longer my first priority. It's important, but not the most important aspect of the project! 👋

PR #643, opened 2018-03-07
fix h types

dancespiele:
everything is explain here https://github.com/jorgebucaran/ultradom/pull/105

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -6,48 +6,34 @@ export as namespace hyperapp
  *
  * @memberOf [VDOM]
  */
-export interface VNode<Props> {
-  name: string
-  props: Props
-  children: VNodeChild<object | null>[]
+export interface VNode<Attributes = {}> {
+  nodeName: string
+  attributes?: Attributes
+  children: Array<VNode | string>
+  key: string
 }
-
-/** In the VDOM a Child can be either a VNode or a string.
- *
- * @memberOf [VDOM]
- */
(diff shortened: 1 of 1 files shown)


dancespiele:
I update this according with this https://github.com/jorgebucaran/ultradom/pull/105

hyperapp.d.ts
```diff
@@ -21,6 +21,11 @@ export interface Component<Attributes = {}> {
   (attributes: Attributes, children: Array<VNode | string>): VNode<Attributes>
 }
 
+/**
+ * Possibles children types
+ */
+export type Children = VNode | string | number | null;
```
frenzzy: ..and function (subview) #639

jorgebucaran: I think that belongs to the definition of VNode.

frenzzy: Why then `string`, `number` and `null` are not VNode definitions?

Virtual Tree:
```js
{ // VNode
  key,
  nodeName,
  attributes,
  children: [
    { key, nodeName, attributes, children }, // VNode
    "string", // text node
    123, // text node
    null, // currently not supported
    undefined, // currently not supported
    () => {}, // subview (just function)
    () => ({ key, nodeName, attributes, children }), // subview may return VNode
    () => "string", // subview may return text node
    () => 123, // subview may return text node
    () => null, // currently not supported
    () => undefined, // currently not supported
  ]
}
```

jorgebucaran:
Better than nothing! I'll merge and work on them later myself. Let's see if I can figure it out.

PR #644, opened 2018-03-07
Enter the dark side: rewrite some tests using JSX. 

jorgebucaran:
Use JSX in tests (like I used to back in the day). I tried to fall in love with `h`, but old habits kept getting in the way. I confess I personally find JSX a lot easier to understand than nested `h` calls. Perhaps my brain is wired to look at HTML-like trees and instantly visualize what's going on. Sorry, JSX detractors, I just can't get used to `h`. 

> P.S: `@hyperapp/html` is also a valid alternative, but it's easier to use JSX with Jest and since we're using Jest...

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,9 +28,18 @@
     "release": "npm run build && npm test && git commit -am $npm_package_version && git tag $npm_package_version && git push && git push --tags && npm publish"
   },
   "babel": {
-    "presets": "env"
+    "presets": "env",
+    "plugins": [
+      [
+        "transform-react-jsx",
+        {
+          "pragma": "h"
+        }
+      ]
+    ]
   },
   "devDependencies": {
+    "babel-plugin-transform-react-jsx": "^6.24.1",
     "babel-preset-env": "^1.6.1",
     "jest": "^22.2.0",
     "prettier": "^1.10.2",
--- a/test/actions.test.js
+++ b/test/actions.test.js
@@ -15,21 +15,18 @@ test("sync updates", done => {
     up: () => state => ({ value: state.value + 1 })
   }
 
-  const view = state =>
-    h(
-      "div",
-      {
-        oncreate() {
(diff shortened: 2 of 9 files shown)


lukejacksonn:
You know my stance on this topic.. but in case I have not said enough, it is one thing supporting a proprietary syntax and for app developers to construct vnodes (because old habits die hard), it is another to integrate this dependency at such a low level as testing.

Now we have to transpile tests? What is JSX changes, all your tests break? Don't take this as overly negative.. but this proposal is absurd.

jorgebucaran:
> @lukejacksonn Now we have to transpile tests? What if JSX changes, all your tests break?

I don't think JSX is going to change in a way that could break any of these tests.

frenzzy:
Personally, I do not understand why this was not done from the beginning, JSX in the tests are really much more readable!

Here for example tests for [`@hyperapp/render`](https://github.com/hyperapp/render) using JSX: https://github.com/hyperapp/render/blob/master/test/index.test.js

jorgebucaran:
We used to, but then I changed it back when I thought I could also get rid of babel. The problem is we need babel with jest so that tests can be written using ES modules.

jorgebucaran:
The docs use JSX and we've been "promoting" JSX (while offering alternatives too) for a while. For better or for worse, the best way to introduce HA to a frontend developer coming from React is as an opinionated, tiny React & Redux. It's too late to pretend JSX is not a VIP first-class citizen in HA. 😄

Swizz:
I would love to keep tests bare metal. 
Using JSX in tests seem a little bit too overpowered.

jorgebucaran:
Sorry, folks, but I am nearly done porting the tests and made up my mind about this one. 😈  

Look, I am always afraid to touch the tests because they're too difficult to maintain and JSX made almost all the pain go away.

PR #646, opened 2018-03-09
Another alternative to JSX.

slacktracer:
The DX using hyperapp with ijk has been really great. Perhaps more people could benefit of it with the extra exposure.


slacktracer:
Let me know if that's an OK PR to make (to ask for I mean) or if I missed some existing guideline instruction or best practice. Thank you very much.

jorgebucaran:
@slacktracer Absolutely! 👍 

/cc @lukejacksonn

lukejacksonn:
Great work @slacktracer and thanks for the support @jorgebucaran 🙇

PR #647, opened 2018-03-10
Improve Actions documentation

warwickgrigg:
As requested and promised via slack: some small improvements to the documentation for actions: clarifying that each state argument is a separate and distinct clone of the current state, adding getState tip to "Interoperability" , and a few tweaks for better ease of understanding.

I don't actually know whether each state argument really is a separate and distinct clone of the current state, but that's how I now understand it to be, or how it is in effect. I might be talking rubbish, or I might need to add the phrase "in effect". Please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,8 +123,6 @@ const actions = {
 
 Each state argument is a separate and distinct clone of the current state. Therefore state updates are always immutable. Immutability enables time-travel debugging, helps prevent introducing hard-to-track-down bugs by making state changes more predictable, and allows cheap memoization of components using shallow equality === checks.
 
-Immutability enables time-travel debugging, helps prevent introducing hard-to-track-down bugs by making state changes more predictable, and allows cheap memoization of components using shallow equality <samp>===</samp> checks.
-
 Do not mutate the state object argument within an action and return it: the results are not what you expect. 
 
 #### Asynchronous Actions
--- a/README.md
+++ b/README.md
@@ -109,20 +109,23 @@ const state = {
 
 ### Actions
 
-The way to change the state is via actions. An action is a unary function (accepts a single argument) expecting a payload. The payload can be anything you want to pass into the action.
+The only way to change the state is via actions. An action is a function that takes a payload as its only argument. The payload can be anything you want to pass into the action. To update the state, an action must return a partial state object. The new state will be the result of a shallow merge between this object and the current state.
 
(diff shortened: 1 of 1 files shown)


jorgebucaran:
Thanks, @warwickgrigg. I want to merge some of this stuff, but not everything, so to save you some trouble I'll merge this now and tweak it myself.


> @warwickgrigg ...clarifying that each state argument is a separate and distinct clone of the current state...

Why do we need to clarify this? We already say, in the documentation for state, that Hyperapp's state is immutable. Therefore, I don't want to make this clarification.

warwickgrigg:
I think it would be better for the docs to clarify this because it wasn't obvious to me (and still isn't) that those two statements are equivalent. I think other future readers of the docs would benefit: avoids misconceptions; aids understanding.

warwickgrigg:
Further to my previous reply, the docs as they currently stand say that "state updates" are immutable; they don't say that Hyperapp's state is immutable. 

I think it's true to say that Hyperapp's state is mutable, at least in this sense:

const state = { o1: {}, o2: {} };
const actions = { o1: dt => (s, a) => { s.o2.a = dt; // mutates hyperapp's state ....

unless Hyperapp's state is defined as just the shallow first level of the state object.

The concept I hold in my head is that the state is immutable if I write my actions a particular recommended way (to be defined).

jorgebucaran:
@warwickgrigg Good point. I'll make sure to stay _"the state is immutable"_ when I tweak this.

warwickgrigg:
But the state is not immutable. This jsfiddle demonstrates so ( https://jsfiddle.net/WarwickGrigg/yktgo2LL/ ) . I understand why, and I don't see this as a bug in the code.

Irrespective of absolute immutability, it's not equivalent to "each state argument is a separate and distinct clone of the current state". Consider two possible alternative ways of implementing the functionality as currently described in the docs:

1. As now, Hyperapp makes a shallow copy of Hyperapp's state and passes it to each action. Here each state argument is a separate and distinct clone (actually a shallow copy) of the current state. I think that's what Hyperapp does, but I might be wrong.

2. Alternatively, Hyperapp maintains a "shadow" state object, passes this object to actions, then merges the objects that the actions return into Hyperapp's state object and into Hyperapp's shadow state object.

Immutable state, in some sense, applies to both.

The alternative method in point 2, is not a ridiculous notion at all. I'm not saying whether it's better or not, but it has some merits: it is a plausible notion at the very least. My main point is that the readers of the docs would benefit from knowing (explicitly and unambiguously) that Hyperapp works the first way, not the second way, otherwise they won't know (unless they read the code, ask, experiment etc).

PR #651, opened 2018-03-16
Fix typos in README

SkaterDad:
Some minor doc tweaks 🤓 

- The raw VNode example was incorrect since `children` must be an array.
- I felt it was more clear to say that components can return an a Array, rather than just "multiple", which someone could mistakenly think is fragments.
- Children composition was missing `to`, and I think the comma makes it read better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,7 +307,7 @@ export const view = (state, actions) => (
 )
 ```
 
-If you don't know all the attributes that you want to place in a component ahead of time, you can use the [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator). Note that Hyperapp components can return an Array of elements as in the following example. This technique lets you group a list of children without adding extra nodes to the DOM.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -307,7 +307,7 @@ export const view = (state, actions) => (
 )
 ```
 
-If you don't know all the attributes that you want to place in a component ahead of time, you can use the [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator). Note that Hyperapp components can return multiple elements as in the following example. This technique lets you group a list of children without adding extra nodes to the DOM.
+If you don't know all the attributes that you want to place in a component ahead of time, you can use the [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator). Note that Hyperapp components can return an Array of elements as in the following example. This technique lets you group a list of children without adding extra nodes to the DOM.
```
jorgebucaran: > array

Please? 🙏

jorgebucaran:
Thanks, @SkaterDad! 🎉

PR #659, opened 2018-03-23
Fix d.ts file as component can return VNode or View

yuku-t:
Since v1.2, components can return VNode or View.

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -13,12 +13,12 @@ export interface VNode<Attributes = {}> {
   key: string
 }
 
-/** A Component is a function that returns a custom VNode.
+/** A Component is a function that returns a custom VNode or View.
  *
  * @memberOf [VDOM]
  */
-export interface Component<Attributes = {}> {
-  (attributes: Attributes, children: Array<VNode | string>): VNode<Attributes>
+export interface Component<Attributes = {}, State = {}, Actions = {}> {
+  (attributes: Attributes, children: Array<VNode | string>): VNode<Attributes> | View<State, Actions>
 }
 
 /**
@@ -35,7 +35,7 @@ export type Children = VNode | string | number | null
  * @memberOf [VDOM]
  */
 export function h<Attributes>(
-  nodeName: Component<Attributes> | string,
+  nodeName: Component<Attributes, any, any> | string,
   attributes?: Attributes,
   ...children: Array<Children | Children[]>
 ): VNode<Attributes>
@@ -96,7 +96,7 @@ export function app<State, Actions>(
 
 declare global {
   namespace JSX {
-    interface Element<Data> extends VNode<object> {}
+    interface Element extends VNode<any> {}
     interface IntrinsicElements {
       [elemName: string]: any
     }
--- a/test/ts/Component.tsx
+++ b/test/ts/Component.tsx
@@ -1,0 +1,24 @@
+import { h, Component } from "hyperapp"
+
+interface State {
+  count: number
+}
+
+interface Actions {
+  down(): State
+  up(value: number): State
+}
+
+const Counter: Component<{ state: State, actions: Actions }> = ({ state, actions }) =>
+  <div>
+    <div>{state.count}</div>
+    <button onclick={actions.down}>-</button>
+    <button onclick={actions.up}>+</button>
+  </div>
+
+const LazyCounter: Component<{}, State, Actions> = () => (state, actions) =>
+  <div>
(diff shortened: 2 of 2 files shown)


hyperapp.d.ts
```diff
@@ -13,12 +13,12 @@ export interface VNode<Attributes = {}> {
...
 }
 
-/** A Component is a function that returns a custom VNode.
+/** A Component is a function that returns a custom VNode or View.
  *
  * @memberOf [VDOM]
  */
-export interface Component<Attributes = {}> {
-  (attributes: Attributes, children: Array<VNode | string>): VNode<Attributes>
+export interface Component<Attributes = {}, State = any, Actions = any> {
```
Swizz: Maybe we should prefer the use of `{}` instead of `any`, here. 
What do you think ?

yuku-t: First I used `{}` then changed to `any` in order to avoid breaking change... but, wait, there isn't breaking change (because `View<State, Actions>` is added by myself). Get back to `{}`

hyperapp.d.ts
```diff
@@ -96,7 +96,7 @@ export function app<State, Actions>(
 
 declare global {
   namespace JSX {
-    interface Element<Data> extends VNode<object> {}
+    interface Element extends VNode<any> {}
```
yuku-t: This change came from [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f2982c175477516d52bdc40191d7829d0531ba35/types/react/index.d.ts#L3742).

dancespiele: could you explain me why you need to add the state and actions object into the generics for a `components` and `h`, is there a reason for that? what doesn't work without that currently?

yuku-t: It is because `View` type is generics:

https://github.com/hyperapp/hyperapp/blob/9031077fcb9bf760fd4e86593fda284ce000bb3b/hyperapp.d.ts#L75-L77

This PR aims to allow `Component` type to return `View` so I had to add `State` and `Actions` to `Component`.

Swizz: @dancespiele See https://github.com/hyperapp/hyperapp/releases/tag/1.2.0

Component are not able to return a `View`. If you are not using the Lazy Components feature or you prefer to type the `View` arguments by your self it is also possible.

hyperapp.d.ts
```diff
@@ -34,8 +34,8 @@ export type Children = VNode | string | number | null
  *
  * @memberOf [VDOM]
  */
-export function h<Attributes>(
-  nodeName: Component<Attributes> | string,
+export function h<Attributes, State = {}, Actions = {}>(
```
jorgebucaran: @yuku-t Is it required to add State={}, Actions={} to `h< ...>`? The `h` function has never nothing to do with the current lazy component implementation.

Swizz: `nodeName` can contain a `Component` that require `State` and `Actions`

And both are optional so, I do not see any point against this.

jorgebucaran: @Swizz The lazy component feature has really nothing to do with actual components — it's just the Vdom's ability to handle nodes of type function.

Swizz: You can give to `h` as first parameters a `Component` that can be lazy, right ?

jorgebucaran: @Swizz You can give any function to `h` and `h` will call it. So, it's not like `h` will do anything special for lazy components.

yuku-t: To remove `State` and `Actions`, we have to change the next line as:

```ts
nodeName: Component<Attributes, any, any> | string,
```

In this case, developers who directly use `h` (not via JSX) cannot specify state and actions. I think almost developers are using JSX, so it is ok to remove them. How do you think?

Swizz: So do we need to remove the `Component` type of `h` parameter to `Function` ?

jorgebucaran: @yuku-t People using `h` will be calling components like `MyComponent(...)` and not like `h(MyComponent)`, so I think it's okay.

yuku-t: @Swizz If we replace `Component` with `Function`, TypeScript compiler no longer checks direct `h` call:

image

so we shouldn't do that.

@jorgebucaran 🙆‍♂️

yuku-t: done

jorgebucaran:
Thanks, @yuku-t! Published as 1.2.2 🎉

Issue #657, opened 2018-03-22
Parcel HMR doubles up event handlers.


Squeegy:
I tried using the Parcel packager, which does hot-reloading out of the box. And it seems to work great with hyperapp, aside from one thing. On each hot-reload, it seems to re-add event handlers, causing them to eventually be double/triple/quad called from a single click.

My view is basically just the example code:

```ts
import { h, View } from 'hyperapp'
import { state, State } from './state'
import { actions, Actions } from './actions'

export const view: View<State, Actions> = (state, actions) => (
  <div>
    <h1>{state.count}</h1>
    <button onclick={() => actions.down(1)}>-</button>
    <button onclick={() => actions.up(1)}>+</button>
  </div>
)
```

If I change any source code file, I see the changes in my browser immediately, but then one click triggers the action as many times as hot reloads have happened.

Given that `addEventListener` appears exactly once in the whole framework, I'm guessing there is a bug here wher ethe previous event handers are not being cleared before new ones are installed: https://github.com/hyperapp/hyperapp/blob/master/src/index.js#L170

jorgebucaran:
@Squeegy Is your mounting element `document.body`?

okwolf:
I have a feeling that this is causing multiple calls to `app`, which has been discussed already in another thread: https://github.com/hyperapp/hyperapp/issues/533#issuecomment-355809201.

Swizz:
If I understand correctly. 

The app is mounted, the view is renderered and events handlers are added.  
After the HMR stuff, the app is mounted again, the hydration do its job by keeping the dom, then the view is rerendered and events handlers added again. 

So, hydration should remove events handlers on reused node ?

Squeegy:
@jorgebucaran I'm mounting it to `#app` with this `index.html`, so I don't think thanks the problem.

```html
<html>
  <head>
    <link rel="stylesheet" href="./styles/index.scss">
  </head>
  <body>
    <div id="app"></div>
    <script src="./src/index.tsx"></script>
  </body>
</html>
```

okwolf:
This has the potential to cause other issues. The problem isn't limited to event handlers, since you actually have an entire other **_app_** still running. If external events were wired to call actions in the previous `app` and they are both using the same container, then they will both be fighting for control of that part of the DOM.

[...]

Issue #637, opened 2018-03-05
Is there React Portal concept?


lifeiscontent:
Hi there, I just started reading about hyperapp I love the idea, I was wondering if you guys have a concept of being able to render into different areas of an app. e.g. something like a React Portal.

Is this possible? Or if its not, is it something of interest within the community?

SkaterDad:
Hyperapp itself doesn't have any concept of portals.

My first question would be:  What do portals give you that hyperapp can't already do?

Since your entire app state is shared/global, you can already render state in multiple places.  I think React needed portals to do this because of local component state?

The party trick of using portals to render content outside of the application's DOM tree is more interesting.  I recall an article showing that you can render a React portal into a popup window and maintain the interactivity.   I haven't looked into how that's implemented.

lifeiscontent:
@SkaterDad yeah, one of the things I have done is used portals on the server to render contents into the head of a document which is nice for setting things like titles/metadata.

zaceno:
@lifeiscontent Interesting! I was just actually thinking about how one would implement something like Vue.js's "slots" and the problem seems very similar to the problem of implementing something like React's portals.

I don't have a solution yet, but I think a solution would require something like React's context -- which also doesn't exist in hyperapp. ( I have a package you could add add it with though: https://github.com/zaceno/hyperapp-context)

lifeiscontent:
@zaceno is hyperapp capable of doing things like `React.fragment`? just to give you an example in pseudo react code.

``` jsx
import Head from './components/Head';

export default () => {
  return (
    <React.Fragment>
      <Head>
        <title>My Page</title>
       </Head>
       <h1>Welcome to my site...</h1>
    </React.Fragment>
  );
}
```

the use case here is to obviously make it so you don't have to render a div or something with the page you're rendering.

jorgebucaran:
@lifeiscontent Your components can return arrays of nodes, but there must be a top-level container for the view. So, it can do what you want, but we don't have proper fragments like React which renders the children to a DocumentFragment which can be keyed.

lifeiscontent:
@jorgebucaran is this something that would potentially be a feature in the future?

jorgebucaran:
@lifeiscontent Fragments? Possibly. (See related: https://github.com/hyperapp/hyperapp/issues/530). Depends on how much code it involves.

Portals? I imagine you are new to Hyperapp, but if and when you know more about it, you'll understand that API doesn't make sense in Hyperapp.

[...]

PR #662, opened 2018-03-24
Multiple app calls

frenzzy:
refs #533, #657, 7585a402626894cb1146044c77be42f177b885e2

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -29,10 +29,10 @@ export function h(name, attributes) {
 export function app(state, actions, view, container) {
   var map = [].map
   var rootElement = (container && container.children[0]) || null
-  var oldNode = rootElement && recycleElement(rootElement)
+  var oldNode = rootElement && (rootElement.node || recycleElement(rootElement))
   var lifecycle = []
   var skipRender
-  var isRecycling = true
+  var isRecycling = !(rootElement && rootElement.node)
   var globalState = clone(state)
   var wiredActions = wireStateToActions([], globalState, clone(actions))
 
@@ -65,6 +65,7 @@ export function app(state, actions, view, container) {
 
     if (container && !skipRender) {
       rootElement = patch(container, rootElement, oldNode, (oldNode = node))
+      rootElement.node = node
     }
 
     isRecycling = false
@@ -108,9 +109,10 @@ export function app(state, actions, view, container) {
 
   function wireStateToActions(path, state, actions) {
     for (var key in actions) {
-      typeof actions[key] === "function"
+      var action = actions[key].h || actions[key]
+      typeof action === "function"
         ? (function(key, action) {
-            actions[key] = function(data) {
+            var wiredAction = function(data) {
               var result = action(data)
 
               if (typeof result === "function") {
@@ -129,11 +131,13 @@ export function app(state, actions, view, container) {
 
               return result
             }
-          })(key, actions[key])
+            wiredAction.h = action
+            actions[key] = wiredAction
+          })(key, action)
         : wireStateToActions(
             path.concat(key),
             (state[key] = clone(state[key])),
-            (actions[key] = clone(actions[key]))
+            (actions[key] = clone(action))
           )
     }
 


jorgebucaran:
@frenzzy Why?

frenzzy:
For Hot Module Replacement, just call `app` with new state/actions/view on the same node as parcel do.

jorgebucaran:
Thanks, @frenzzy, but I don't understand why I'd want to merge this, but if you can properly explain why then I might be interested.

frenzzy:
Oh.. looks like dynamic actions already works!
Demo: https://codepen.io/frenzzy/pen/NYvvKm?editors=0010
Nevermind then :blush:

jorgebucaran:
🙃

jorgebucaran:
@frenzzy A bit better.

```js
loadAction: () => state => {
  app(state, {...actions,
    double: () => state => ({ count: state.count * 2 })
  }, view, document.body)
```

Does this PR help improve the above and how? That would make sense.

frenzzy:
Dynamic actions demo:
- master branch: https://codepen.io/frenzzy/pen/RMZxGY?editors=0010
- current branch: https://codepen.io/frenzzy/pen/MVvrQb?editors=0010

Steps:
1. Initial state: `10` => `<h1>10</h1>`
2. Click `load action` => `<h1>10</h1>`
3. Click `x2` => `<h1>20</h1>`
4. Click `+` =>
    - master branch: `<h1>11</h1>` // <= Wrong!
    - current branch: `<h1>21</h1>`

okwolf:
Thanks for looking into this @frenzzy. Unfortunately, there are still more problems with multiple `app` calls as I've [previously written](https://github.com/hyperapp/hyperapp/issues/657#issuecomment-375777875).

To see one of the issues this doesn't solve take a look a this pen that simulates a second `app` call where each one wires up external events to call different actions (imagine a HMR scenario where that part of the code changed). This change does not help in this scenario as you can see if you switch the comments on the pen: https://codepen.io/okwolf/pen/vRJQpZ?editors=0010

Off the top of my head the only true solution I can think of would be to have some way of unwiring/stopping our previous `app` instance. Maybe attach this to the container element and call this if it exists when mounting a new `app`? 🤔

jorgebucaran:
> Maybe attach this to the container element and call this if it exists when mounting a new app? 

Yeah, similarly to what I do in Ultradom.

frenzzy:
> Off the top of my head the only true solution I can think of would be to have some way of unwiring/stopping our previous `app` instance.

But they are the different features:
- **An ability to call app multiple times** is the way how you can fire update your app with new state/actions/view without the actual stopping the app. You don't want to completely stop and run again the app because it is too much overhead. For example your components have to destroy and initialize instances of third-party libs when you have the opportunity to avoid it.
- **Stop/destroy/unmount** feature at the same time is needed to stop the application and allocate resources. Just removing a container DOM node is not enough because we need to destroy all instances and global event handlers by calling remove/destroy lifecycle events. The good news is that we can implement this feature in the user land, for example using a higher-order app approach. This feature also must remove all mounted DOM nodes from the container.

Suggested way to store VDom tree inside `element.node` sounds good to me and allows to easily implement unmount feature and helps multiple app calls to work properly. I am worried only about possible memory leaks with this approach because VDom tree also usually contains links to functions in VNode attributes.

jorgebucaran:
We will not need to call app multiple times in 2.0, but we'll definitely want an unmount function or similar, eventually.

Since this code just deals with multiple app calls, I am going to close here as wontfix.

frenzzy:
Should we also revert changes from this commit? 7585a402626894cb1146044c77be42f177b885e2

If multiple app calls are not supported, then changes from the commit just increase bundle size and useless for users.

jorgebucaran:
That's also fixing an HMR bug, so I think I'll leave it.

frenzzy:
@jorgebucaran the problem is HMR is not configured properly and by default it calls `app` multiple times on the same container. You are fixed events problem but do not fix actions and recycling problem from this PR.

In my mind if you want to support multiple app calls for default HMR, then do it properly or do not support at all. Am I wrong?

For example:
```js
let container = document.getElementById('app')

if (module.hot) {
  module.hot.dispose(() => {
    container.parentNode.removeChild(container)
  });

  module.hot.accept(function () {
    container = document.body.appendChild(document.createElement('div'))
  });
}
```

jorgebucaran:
@frenzzy I know what you mean. We'll see.

jorgebucaran:
@frenzzy On a second thought, you're calling app twice in your example.

frenzzy:
@jorgebucaran yes, but each time with a different container which is fine as I understand.

frenzzy:
You always can preserve previous state if you want...
```js
const main = app(state, actions, view, container)

if (module.hot) {
  module.hot.dispose(function (hotData) {
    hotData.prevState = main.getState()
  });

  module.hot.accept(function (hotData) {
    const nextState = Object.assign({}, state, hotData.prevState) // or deep merge
    const nextContainer = document.body.appendChild(document.createElement('div'))
    app(nextState, actions, view, nextContainer)
  });
}
```

frenzzy:
With [unmount feature](https://github.com/hyperapp/hyperapp/issues/637#issuecomment-378230188) you even may use the same container and avoid memory leaks:
```js
const main = withUnmount(app)(state, actions, view, container)

if (module.hot) {
  module.hot.dispose(function (hotData) {
    hotData.prevState = main.getState()
    main.unmount()
  });

  module.hot.accept(function (hotData) {
    const nextState = Object.assign({}, state, hotData.prevState) // or deep merge
    app(nextState, actions, view, container)
  });
}
```

PR #663, opened 2018-03-25
Keyed children diff optimization

### LINKED ISSUE #660: Keyed <input> loses focus when earlier sibling is removed
I've noticed some unexpected behavior with keyed children, particularly `<input>`.

Removing a sibling **before** the currently focused input causes it to lose focus.  Removing a sibling **after** the input is okay.

For example, removing the `<span>` or one of the `<input>` in this mock dynamic form:
```jsx
<div>
  { state.visible ? <span key={"theSpan"}>I might disappear</span> : null }
  { state.inputs.map(inputName => (
      <input
          key={name}
          oninput={ (event) => action.mightRemoveAnInputOrTheSpan(event.target.value)}
      />
    )
  }
</div>
```

An `element.insertBefore` call is being made for every keyed child beyond the removed one, which is likely when focus is lost.   In this case, it is also an unecessary perf hit, since a single `removeElement` operation could have been performed instead.
```js
          if (oldKey === newKey) {
            patch(element, keyedNode[0], keyedNode[1], children[k], isSvg)
            i++
          } else if (keyedNode[0]) {
            patch(
              element,
              element.insertBefore(keyedNode[0], oldElements[i]), // <--- THIS
              keyedNode[1],
              children[k],
              isSvg
            )
```

@jorgebucaran Is this an issue that will be resolved with the diffing re-write you've talked about?   It seems like this could be solved by determining which keyed items are being added & removed before doing any DOM operations?

**Reproduction (with instructions):**  https://codepen.io/SkaterDad/pen/jzLbBX

Similar demo in React which seems fine: https://codesandbox.io/s/z68rm906jx

SkaterDad:
### Summary of the change
 1.  Determine if the new keyed child matches the next old child (thanks to `preact` for the inspiration here).
 2. If the current old child is keyed, leave it alone.
 3. If the current old child is non-keyed, remove it.  This keeps the "mixed keyed/nonkeyed" test happy.
 4. Increment the old children index and go back to the start of the while loop.

### Benefits
 * Closes #660: Keyed `<input>` stays focused if siblings before it are removed ([Proof on Codepen](https://codepen.io/SkaterDad/pen/KoXzwN)).  I tried writing test cases for this, but the `<input>` never fired a "blur" event like it did in-browser.
 * Better keyed list modification performance.  Results in fewer "insertBefore" calls when a keyed item is removed but other items are in the same order.  Open the "Details" below to see a screenshot from `JS Framework Benchmark`.  Row removal is 2x faster.  Row swap is 5x faster.

     <details>

      image

     </details>

### Bytes
Minified file size changes from 3.45KB to 3.51KB.
Gzipped file size changes from 1.54KB to 1.56KB (using default options in linux gzip command).

> Edit:  Added benchmark screenshot.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -320,9 +320,18 @@ export function app(state, actions, view, container) {
 
       while (k < children.length) {
         var oldKey = getKey(oldChildren[i])
+        var nextOldKey = getKey(oldChildren[i + 1])
         var newKey = getKey((children[k] = resolveNode(children[k])))
 
         if (newKeyed[oldKey]) {
+          i++
+          continue
+        }
+
+        if (newKey != null && newKey === nextOldKey) {
+          if (oldKey == null) {
+            removeElement(element, oldElements[i], oldChildren[i])
+          }
           i++
           continue
         }
--- a/src/index.js
+++ b/src/index.js
@@ -320,7 +320,6 @@ export function app(state, actions, view, container) {
 
       while (k < children.length) {
         var oldKey = getKey(oldChildren[i])
-        var nextOldKey = getKey(oldChildren[i + 1])
         var newKey = getKey((children[k] = resolveNode(children[k])))
 
         if (newKeyed[oldKey]) {
@@ -328,7 +327,7 @@ export function app(state, actions, view, container) {
           continue
         }
 
-        if (newKey != null && newKey === nextOldKey) {
+        if (newKey != null && newKey === getKey(oldChildren[i + 1])) {
           if (oldKey == null) {
             removeElement(element, oldElements[i], oldChildren[i])
           }


SkaterDad:
I've done a little more experimenting to see how flexible this change is.

As the code implies, this is an optimization for single item removals.   It also handles cases like this, where single items are removed in multiple parts of the list.  Here's another demonstration which is automated:  https://codepen.io/SkaterDad/pen/WzZPxr

```
[ 1, 2, 3 ] ==> [ 2, 3 ]
[ 1, 2, 3, 4, 5] ==> [ 2, 4, 5] 
```

I believe these are common operations in real apps, and are worth optimizing for.  The benchmark results are a nice bonus.

This change does not handle multiple consecutive items being removed, or if a non-keyed item is mixed in.  The diffing will go back to its current behavior and do a bunch of "insertBefore" DOM operations.

Those cases are where a more complicated diffing algorithm like the one in `ivi` can make better decisions, but for now I'm happy to write my views in a way that avoids those scenarios when possible.

SkaterDad:
@jorgebucaran Do you see this going anywhere before 2.0?  I've got code I'm waiting to PR for thunks/lazy VNodes which touches code in the same region.  I would imagine 2.0 is touching some of the patch  & resolveNode code also?

I'm fine waiting, just curious what your plans are.

jorgebucaran:
@SkaterDad I didn't know this was blocking you, I'll get this merged and published tomorrow.

src/index.js
```diff
@@ -320,13 +320,22 @@ export function app(state, actions, view, container) {
 
       while (k < children.length) {
         var oldKey = getKey(oldChildren[i])
+        var nextOldKey = getKey(oldChildren[i + 1])
```
jorgebucaran: @SkaterDad Can we inline this call with the `if (newKey != null && newKey === nextOldKey) {` later below?

jorgebucaran:
> @SkaterDad If the current old child is non-keyed, remove it. This keeps the "mixed keyed/nonkeyed" test happy.

Do we want to keep those tests happy though? 🤔

SkaterDad:
> @jorgebucaran Do we want to keep those tests happy though? thinking

If you don't remove the non-keyed child at that point in the code, it stays in the DOM since cleanup doesn't happen until the end and assumes the only children you need to remove are at the end of the oldChildren array.

I think mixed keyed/non-keyed is a nice feature to support.  I do find myself only keying specific children sometimes, and it behaves intuitively.

src/index.js
SkaterDad: You bet.  Looks like that shaves ~3 bytes off after gzip.

SkaterDad: Done :+1:

jorgebucaran:
> @cjh9 ...having non-keyed elements together with keyed elements in the same parent...

This hasn't changed.

jorgebucaran:
@SkaterDad I published 1.2.6 by the way. Sorry for the incredible delay. 🙇

SkaterDad:
Haha, no worries!

PR #664, opened 2018-03-25
Format improvements

okwolf:
Since I'm using [`prettier`](https://prettier.io) in my editor with format on save enabled - I noticed that our configuration for `prettier` is sent as a command line argument instead of stored in the `"prettier"` property of our `package.json` (or we could have a [`.prettierrc`](https://prettier.io/docs/en/configuration.html) if you'd rather add a new file). Having this encourages folks to keep auto format enabled in their editors, which will eliminate the problem of forgetting to format the code. I'm already using this config in [`@hyperapp/logger`](https://github.com/hyperapp/logger/blob/634e52ba8ae9fc6c2b38271ba15bdd9d2aabdbfc/package.json#L29) and [`@hyperapp/fx`](https://github.com/hyperapp/fx/blob/80d20cb6356644af37991cd2561340f63f9e0d0f/package.json#L38).

Also, some of the existing code wasn't formatted, so I'm including that fix.

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -18,7 +18,9 @@ export interface VNode<Attributes = {}> {
  * @memberOf [VDOM]
  */
 export interface Component<Attributes = {}, State = {}, Actions = {}> {
-  (attributes: Attributes, children: Array<VNode | string>): VNode<Attributes> | View<State, Actions>
+  (attributes: Attributes, children: Array<VNode | string>):
+    | VNode<Attributes>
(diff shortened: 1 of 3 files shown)


jorgebucaran:
Thanks, @okwolf! 💯

Swizz:
Which Prettier plugin are you using for your editor ? Prettier Now ?

jorgebucaran:
I am using this one: https://github.com/prettier/prettier-vscode

PR #666, opened 2018-03-26
Adding build:es task

### LINKED ISSUE #665: Building hyperapp as minified es6 module
## Problem Statement
Currently Hyperapp is bundled using rollup (generating ES5 library) and uglified using uglify-js. So For those who want to consume Hyperapp as pure ES6 module which works natively on Chrome at present there is no way out

## Solution Proposal
We should use babel-minify that takes away whitespace and do other optimization to generate pure ES6 module that can be consumed directly in chrome. Refer this sample code snippet to better understand the requirement:

__index.html__
```html
<!DOCTYPE html>
<html>
    <head>
        <script type="module" src="./app.js"></script>
    </head>
    <body>
        <div id="app"></div>
    </body>
</html>
```

__app.js__
```js

import { h, app } from './hyperapp.es.js';

const state = {
  count: 0
}

const actions = {
  down: value => state => ({ count: state.count - value }),
  up: value => state => ({ count: state.count + value })
}

const view = (state, actions) =>
  h("div", {}, [
    h("h1", {}, state.count),
    h("button", { onclick: () => actions.down(1) }, "-"),
    h("button", { onclick: () => actions.up(1) }, "+")
  ])

app(state, actions, view, document.getElementById('app'))
```

As seen above I have kept `hyperapp.es.js` in current directly, but once we agree and this PR is merged then in future we can write statement like this:
```
import { h, app } from 'https://cdnjs.cloudflare.com/ajax/libs/hyperapp/1.2.6/hyperapp.es.js';
```

kumarabhishek:
## Fixes #665 
Adding build:es npm script task for generating hyperapp.es.js (optimised ES6 module) for direct consumption via ES6 module import in modern browser like Chrome.

**Note**: babel-minify currently does not generate sourcemap file. In future if it does such mangling that need sourcemap, then it can be added accordingly at that point of time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
   ],
   "scripts": {
     "test": "jest --coverage --no-cache && tsc -p test/ts",
-    "build": "npm run bundle && npm run minify",
+    "build": "npm run bundle && npm run minify && npm run build:es",
+    "build:es": "minify src/index.js --mangle.keepClassName --outFile dist/hyperapp.es.js",
     "bundle": "rollup -i src/index.js -o dist/hyperapp.js -m -f umd -n hyperapp",
     "minify": "uglifyjs dist/hyperapp.js -o dist/hyperapp.js -mc pure_funcs=['Object.defineProperty'] --source-map includeSources,url=hyperapp.js.map",
     "prepare": "npm run build",
@@ -42,6 +43,7 @@
     ]
   },
   "devDependencies": {
+    "babel-minify": "^0.3.0",
     "babel-plugin-transform-react-jsx": "^6.24.1",
     "babel-preset-env": "^1.6.1",
     "jest": "^22.2.0",


jorgebucaran:
Thanks, @kumarabhishek. To clarify again, the use case here is using Hyperapp as an ES module directly from a CDN?

frenzzy:
### 666 :smiling_imp:

kumarabhishek:
@jorgebucaran Yes. You got it right.

jorgebucaran:
@kumarabhishek And what is the disadvantage of using https://unpkg.com/hyperapp@1.2.5/src/index.js?module?

kumarabhishek:
## CDN:

image

## Localhost Example:

image

If you see above pictures, notice the size. CDN(2.6 KB) and localhostexample(1.8 KB). Hope this clarify the motivation. I know it may sound crazzy to run for another 0.8 KB, but in future it may increase.

jorgebucaran:
@kumarabhishek Would you consider this less relevant if unpkg had a way to minify modules on the fly? Just curious about your use case.

> ...but in future it may increase

Hyperapp? Maybe, but not for much.

kumarabhishek:
@jorgebucaran If unpkg does what babel-minify is doing on the fly then its awesome and we can drop this change. In near future many browser will start supporting ES6/ES7/ES8 natively and whole concept of bundling and chunking that we do today using webpack and similar tools may go away. So its very usefull if we have some way to consume Hyperapp in a modern way (ES6 way). I know the current implementation of Hyperapp is also in ES5 but to save more space and keep up with the 1 KBish starting motto there may be need to write some part in ES6/ES7. My idea is just to bring to our notice that Hyperapp consumption as ES6 module is future and we should consider it. Which ever way we do is pure our choice (babel-minify or unpkg or something else).

vladshcherbin:
Related issue directly in `unpkg ` - https://github.com/unpkg/unpkg/issues/57

kumarabhishek:
@jorgebucaran So what is that we conclude on this matter ? Is it worth discussing further or shall we close this for now and revisit in future if required ?

jorgebucaran:
@kumarabhishek Let's leave it open for now! 😄

oskarrough:
Hey, just wanted to ping that jsdelivr offers minification on the fly. But it doesn't work for this file. Probably because of uglify+es6 trouble? 

- https://cdn.jsdelivr.net/npm/hyperapp@1.2.5/src/index.min.js
- https://github.com/jsdelivr/jsdelivr/issues/18043

jorgebucaran:
> @oskarrough But it doesn't work for this file. Probably because of uglify+es6 trouble?

I think so too. 😓

rbiggs:
@kumarabhishek, you do realize that the benefits of http2 will make the need to minifying a library the size of Hyperapp superfluous? Import of ES6 modules in browsers is still not entirely ironed out, but when it is done, you'll be able to easily import whatever part of Hyperapp you need without having to use a bundled version. That's still down the road. Therefore I don't think it warrants making changes to Hyperapp's build process to support this yet. My two cents.

kumarabhishek:
@rbiggs You are correct. Http2 is really a nice thing but it is usually done at page load when server pushes more files as part of document response. In scenario where we need on demand loading upon page navigation, dynamic ES6 module import is what will be more relevant.

jorgebucaran:
#665  

Thank you, @kumarabhishek, but I am not going to take this.

People who want a non-UMD minified bundle can simply minify Hyperapp in their own dev environment.

PR #667, opened 2018-03-26
Make data have a generic type

pspeter3:
This might make it so the actions are a little clearer?

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -54,8 +54,8 @@ export type ActionResult<State> = Partial<State> | Promise<any> | null | void
  *
  * @memberOf [App]
  */
-export type ActionType<State, Actions> = (
-  data?: any
+export type ActionType<Data, State, Actions> = (
+  data?: Data
 ) =>
   | ((state: State, actions: Actions) => ActionResult<State>)
   | ActionResult<State>
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -66,7 +66,7 @@ export type ActionType<Data, State, Actions> = (
  */
 export type ActionsType<State, Actions> = {
   [P in keyof Actions]:
-    | ActionType<State, Actions>
+    | ActionType<any, State, Actions>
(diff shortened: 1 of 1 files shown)


jorgebucaran:
Hey @pspeter3, thanks. Can you explain why and how?

Also, see https://github.com/hyperapp/hyperapp/pull/659#discussion_r176727888

pspeter3:
In TypeScript, `any` is an extremely permissive type.

For example:

```ts
const increment: ActionType<{ count: number }, {}> = (amount: number) => (
    state,
    actions
) => state.count + amount;

// Should be a type error
increment("3");
```

With the idea I wanted to play with:

```ts
const increment: ActionType<number, { count: number }, {}> = amount => (
    state,
    actions
) => state.count + amount;

// Is a type error
increment("3");
// Is not a type error
increment(3);
```

The downside is you only get type safety on the first parameter.

pspeter3:
I'll fix the tests if people think this is a good idea. This is a WIP.

Swizz:
Im in favor of it. 
Can you provide an example of an actions without the need of data ?

pspeter3:
@Swizz an empty action would have type `void`

```ts
const increment: ActionType<void, { count: number }, {}> = () => () => { return { count: 0 } };
```

SkaterDad:
> @pspeter3 The downside is you only get type safety on the first parameter.

Actions can only take one argument anyway, so that shouldn't be an issue.

pspeter3:
That's good to know

jorgebucaran:
@pspeter3 Is this still a work in progress? What else do we need to resolve this issue?

pspeter3:
I just resolved the test. I wish the `ActionsType` could carry through the generic prop but that's ok. Maybe something that could be handled with conditional types or more thought.

jorgebucaran:
Long time no see! The current focus is on shipping V2, but I think we can merge here. I'll publish a new V1 patch at some point when I'm less busy. Thank you! 👋

PR #670, opened 2018-03-30
Fix TS fails because of the use of ES6

Swizz:
https://stackoverflow.com/a/39280186


jorgebucaran:
@Swizz Awesome! Thanks a lot.

PR #671, opened 2018-03-30
Add ci test for prettier use

Swizz:
No more 
"Run prettier on the files please"
Now, the CI will tell us if it the PR used prettier on files

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ install:
 script:
   - npm test
   - codecov
+  - npm run format -- --list-different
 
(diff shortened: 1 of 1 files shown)


jorgebucaran:
Haha, thanks. So, will this cause CI to fail if we don't run npm format on files?

Swizz:
Absolutely. If the formatted files are different to the committed files, the CI will raise an error.

jorgebucaran:
Ok, let's give it a shot!

PR #674, opened 2018-04-02
Avoid nested condition

frenzzy:
```js
if () {
} else {
  if () {
  } else if () {
  } else {}
}
```
=>
```js
if () {
} else if () {
} else if () {
} else {}
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -165,34 +165,32 @@ export function app(state, actions, view, container) {
           element[name][i] = style
         }
       }
-    } else {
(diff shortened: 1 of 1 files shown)


jorgebucaran:
I think it's safe, but it bothers me a little we're going to run:

```js
if (value == null || value === false) {
  element.removeAttribute(name)
}
```

...even if the attribute is a `key` or `style`.

frenzzy:
Oh.. you are right! Missed this stuff :)

PR #675, opened 2018-04-02
Inline styles

frenzzy:
It allows us to use bunch of valid html/svg code inside vdom tree without the need to convert a lot of inline styles into objects.

Fixes https://github.com/hyperapp/hyperapp/issues/672#issuecomment-377783572

Demo: https://codepen.io/frenzzy/pen/OvEyeL/left/?editors=0010

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -156,7 +156,7 @@ export function app(state, actions, view, container) {
 
   function updateAttribute(element, name, value, oldValue, isSvg) {
     if (name === "key") {
-    } else if (name === "style") {
+    } else if (name === "style" && typeof value === "object") {
       for (var i in clone(oldValue, value)) {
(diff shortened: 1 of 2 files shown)


jorgebucaran:
I'd like to compare this with using cssText when style is a string.

jorgebucaran:
@frenzzy Close here?

frenzzy:
@jorgebucaran why do you think `cssText` version is better? This version looks shorter.
Possible issue could be a browser support here, but I haven't checked it yet.

jorgebucaran:
Thanks, @frenzzy. I am going to abandon this feature for the same reason given [here](https://github.com/facebook/react/issues/5397#issuecomment-153954732) as I hold the same opinion.

PR #676, opened 2018-04-02
Inline styles using cssText

frenzzy:
Same as #675 but using [`cssText`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/cssText)

Demo: https://codepen.io/frenzzy/pen/PRaPrM/left/?editors=0010

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -156,17 +156,21 @@ export function app(state, actions, view, container) {
 
   function updateAttribute(element, name, value, oldValue, isSvg) {
     if (name === "key") {
-    } else if (name === "style") {
-      for (var i in clone(oldValue, value)) {
-        var style = value == null || value[i] == null ? "" : value[i]
-        if (i[0] === "-") {
-          element[name].setProperty(i, style)
-        } else {
-          element[name][i] = style
-        }
-      }
     } else {
-      if (name[0] === "o" && name[1] === "n") {
+      if (name === "style") {
+        if (typeof value === "string") {
+          element[name].cssText = value
+        } else {
+          for (var i in clone(oldValue, value)) {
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -157,13 +157,17 @@ export function app(state, actions, view, container) {
   function updateAttribute(element, name, value, oldValue, isSvg) {
     if (name === "key") {
     } else if (name === "style") {
```
jorgebucaran: @frenzzy Let's use `typeof value === "string"` instead (below) 🙏

src/index.js
```diff
@@ -156,17 +156,21 @@ export function app(state, actions, view, container) {
...
     if (name === "key") {
-    } else if (name === "style") {
-      for (var i in clone(oldValue, value)) {
-        var style = value == null || value[i] == null ? "" : value[i]
-        if (i[0] === "-") {
-          element[name].setProperty(i, style)
+    } else {
+      if (name === "style") {
+        if (typeof value === "string") {
+          element[name].cssText = value
```
jorgebucaran: What if we also check if `value == null`? 🤔

frenzzy: maybe just `typeof value !== "object"` instead of `typeof value === "string" || value == null`?

frenzzy: or you are talking about `cssText = value == null ? '' : value`?
This could work too, but it seems it changes nothing

jorgebucaran:
Thanks, @frenzzy. I am going to abandon this feature for the same reason given [here](https://github.com/facebook/react/issues/5397#issuecomment-153954732) as I hold the same opinion.

PR #679, opened 2018-04-13
add Contribution and support to README

kidandcat:
Contribution and support


jorgebucaran:
Interesting! 🤔😄

jorgebucaran:
Thank you @kidandcat! I'll include something like this in the V2 docs.

PR #683, opened 2018-04-26
Update CONTRIBUTING.md: ES3 → ES5

infinnie:
Things like `Array.prototype.map()` are features of ECMAScript 5 (though they could be shimmed to ES 3 too).

Also exports are only available as of ES 6.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ Our open source community strives to be nice, welcoming and professional. Instan
 
 ## Style
 
-* Hyperapp is written in ES3. Please don't create an issue asking us to rewrite it in ES6 or TypeScript.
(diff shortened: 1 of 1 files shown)


jorgebucaran:
@infinnie You are totally right about <samp>Array.prototype.map</samp>! 🤔

jorgebucaran:
> Also exports are only available as of ES 6.

This was the only exception.

jorgebucaran:
@infinnie What if we rewrite `recycleElement` to not need `.map`? 😏

infinnie:
like rewriting to `jQuery.map()`?

jorgebucaran:
@infinnie Only `recycleElement` would need to be rewritten because it's the only place using `.map`.  This is far from a priority, so I'll go ahead and merge this anyway.

jorgebucaran:
Thanks! @infinnie

Issue #681, opened 2018-04-23
Remove onremove event.


jorgebucaran:
There is no `onremove` equivalent in React/Preact/Vue that [I am aware of](https://github.com/jorgebucaran/ultradom/issues/68#issuecomment-382780383).

> **Note**: Hyperapp has the same problem as https://github.com/jorgebucaran/ultradom/issues/68. This can be and **will be fixed** by storing the element in the virtual node instead of using children position which is totally unreliable if you are adding / removing elements asynchronously.

mindplay-dk:
Are you proposing to remove this for good, or just from the next release? (pending rewrite of patch?)

The feature isn't *useless* as it is - just unreliable.

It might make more sense to just log it as a bug, since it sounds like the patch rewrite will fix it? And the patch rewrite presumably is an internal change that doesn't need to be a breaking change - whereas removing the feature now is a breaking change.

zaceno:
At least the way it works in petit-dom (the only vdom I’ve played much with besides Ultradom), as far as I can tell, the way you do any lifecycle events is to provide a component in the form of an object which defines the mount, unmount and patch methods.

When you do that, how the tree is handled from that point on is entirely up to you, including removing or deferring removal of the element.

As such, petit-dom’s “unmount” is closer to ultradom’s onremove than ondestroy.

I suspect it works similarly in React - or how else would exit transitions even be possible?

jorgebucaran:
While investigating how to fix jorgebucaran/ultradom#68, I discovered two things: (1) how to fix the problem and (2) that there is no `onremove` equivalent in any of the mentioned frameworks.

The fix consists of storing the element in the virtual node as previously described, then retrieving it inside the async `done` handler. Even if there was no `onremove` I am still going to to store the element in the node at some point in the future when I have time to rewrite patch.

The reason I created this issue is because of (2) above. If others can get away without an `onremove` event, I'd like to explore how we could get rid of it too.

infinnie:
But most other frameworks/libraries put lifecycle methods on *components* instead of DOM elements. And I think that is the difference.

jorgebucaran:
@infinnie So, what's the difference?

infinnie:
So if you want to put lifecycle methods on representations of DOM elements (which I actually think is better) and you want it to support custom element removal, you have to do something.

infinnie:
And `onremove()` or the like is that “something”.

localvoid:
[...]

PR #684, opened 2018-04-26
Fix onremove() ordering

infinnie:
Fixing #681.


jorgebucaran:
Partly? :D

infinnie:
Partly for it being untested.

jorgebucaran:
Gotcha. This is nice try attempt. Ultimately, storing the element in the virtual node will solve the problem, help simplify diffing and improve performance, so while this is fun and you are free to keep doing your best, this is not the route I'll choose.

infinnie:
As of commit https://github.com/infinnie/hyperapp/commit/b12f9bd5ccdb00540b8d2630a8cba256f3571377 tests could be passed now. But I still don’t know why it even works.

jorgebucaran:
@infinnie Because there's no test documenting the said onremove bug.

infinnie:
@jorgebucaran I mean the original tests.
Also maybe @mindplay-dk should be held responsible for documenting the `onremove()` test.

infinnie:
@jorgebucaran ok I’ve added a preliminary test for that purpose.

infinnie:
Seems that much of the problem of reordering should have been fixed as of #663, but that brought forth a problem as I stated in https://github.com/hyperapp/hyperapp/issues/681#issuecomment-384141348:
>> Just by the way: what would we expect to happen if two asynchronous element removals happen consecutively, so that when one element is being removed we start to remove the second one?
And this PR then solves that new problem.

infinnie:
@jorgebucaran #685

mindplay-dk:
@infinnie did you try this on the [nightmare fiddle](https://jsfiddle.net/mindplay/omovhf25/) I came up with to ruin Jorge's day? 😀

infinnie:
@mindplay-dk An error occurred that
> Assignment to read-only properties is not allowed in strict mode.

infinnie:
@mindplay-dk An error occurred that
> 'ultradom' is not defined

infinnie:
@mindplay-dk See also https://jsfiddle.net/y7a2jc55/ that shows where the original Ultradom is wrong.

mindplay-dk:
@infinnie I wrote this test a while back to document the reordering issue with deferred removals:

https://github.com/mindplay-dk/picodom/commit/2b0784b05c4d12f9cdf2471b2eccae327c92aa51

I'm not sure if this covers the more complex issue demonstrated by the nightmare fiddle - it was a long time ago and my attentions are elsewhere at the moment.

I wrote a [more aggressive test](https://bitbucket.org/mindplaydk/domsync/src/7150a9028fd1ba0cbb6279126f0ea0aa21e7ec89/src/test.tsx?at=master&fileviewer=file-view-default#test.tsx-187) (for my own dead project) for reordering, but that doesn't cover the issue with deferred removals.

Probably what's needed is a more minimal/controlled test of some sort, along the lines of [this](https://jsfiddle.net/mindplay/gcd663fx/) fiddle, which you could probably turn into a test by hoisting the removal callbacks, same as I did in my first test.

PR #688, opened 2018-05-04
Recycle innerHTML attribute correctly

frenzzy:
Fixes https://github.com/hyperapp/render/issues/8

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -157,7 +157,7 @@ export function app(state, actions, view, container) {
   }
 
   function updateAttribute(element, name, value, oldValue, isSvg) {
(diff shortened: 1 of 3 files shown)


jorgebucaran:
@frenzzy What if we abort patch if innerHTML is used?
```js
if (node.attributes.innerHTML) {
  return element
}
```

jorgebucaran:
Oh, but then we need to change `updateAttribute` so that innerHTML is not inadvertently reset when the new node's innerHTML is null.

jorgebucaran:
I will not support innerHTML in Hyperapp.

PR #689, opened 2018-05-14
Add examples on Hyperstart

hspotorno:
Changing the readme.md file to link to Hyperstart examples, and added a couple that might be useful as well.


infinnie:
Firebase is blocked in China due to its server using Google’s IP address.

PR #693, opened 2018-05-23
Add size badges to README

styfle:
This adds two badges to the `README.md` file:

- gzip size: the size of the compressed bundle
- install size: the size of running npm install

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 [![Travis CI](https://img.shields.io/travis/hyperapp/hyperapp/master.svg)](https://travis-ci.org/hyperapp/hyperapp)
 [![Codecov](https://img.shields.io/codecov/c/github/hyperapp/hyperapp/master.svg)](https://codecov.io/gh/hyperapp/hyperapp)
 [![npm](https://img.shields.io/npm/v/hyperapp.svg)](https://www.npmjs.org/package/hyperapp)
+[![gzip size](http://img.badgesize.io/https://unpkg.com/hyperapp?compression=gzip)](https://unpkg.com/hyperapp)
+[![install size](https://packagephobia.now.sh/badge?p=hyperapp)](https://packagephobia.now.sh/result?p=hyperapp)
 [![Slack](https://hyperappjs.herokuapp.com/badge.svg)](https://hyperappjs.herokuapp.com "Join us")
 
 Hyperapp is a JavaScript micro-framework for building web applications.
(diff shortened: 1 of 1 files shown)


Swizz:
I would a way prefer the bundlephobia badge promoted by [shields.io](https://shields.io). (like the other badges we use)

```
https://img.shields.io/bundlephobia/minzip/hyperapp.svg
```

styfle:
@Swizz Thanks, I updated the badge

okwolf:
Since we’re less focused on size going forward, I’m not sure it makes sense to add this.

jorgebucaran:
Thanks, @styfle, but I'll pass this time. 

Unrelated, but close, I merged https://github.com/jorgebucaran/ultradom/pull/116 because size is a theme going forward in _Ultradom_. 

So closing for now. 👋 

@okwolf Exactly my thoughts!

atomiks:
Even though size is less of a focus (i.e., as a marketing slogan currently), it's still useful to know what the size of the lib is at a glance. I think it's a nice addition.

jorgebucaran:
What could be another way to market the size going forward _without_ a badge?

okwolf:
@jorgebucaran silly way to measure this, but I believe we have ~9 🌟/B and React has ~3 ⭐️/B.

jorgebucaran:
lol

atomiks:
Well you mentioned the size is less of a focus, so it don't be marketed anymore. But the file size should still be displayed for people as the badge, it's just not a main focus.

jorgebucaran:
Let me try again: I don't like overusing badges, so that's why I closed here. Now, 2.0 will be more than 1 kB, but we should still find a way to mention the byte size, just not using a badge. What could that way be was my question.

For example, instead of our current "1 kB JavaScript library..." tagline, change it to "2 kB JavaScript library...".

atomiks:
>1 kB JavaScript library..." tagline, change it to "2 kB JavaScript library...

Dealbreaker for everyone who uses Hyperapp for its size 😅

2 kB is still tiny so it's fine to mention there I think

jorgebucaran:
Should point out that 2.0's core will not be a lot larger than 1.0, the extra ultrathin layer of fat is from picking the effects and subscriptions you'll use.

Fun FYI fact: An app in **1.0** could potentially be larger than its equivalent in 2.0 as you'd need to define your own way to handle effects or subscriptions anyway.

frenzzy:
For me as for a regular user the size information of a client-side library makes much more sense than for example travis ci/build status. Just an opinion, don't beat me tomatoes :)

jorgebucaran:
@frenzzy Hmm, yes, I agree now. That info should go on the website, though (which I am planning to update after 2.0 too.)

Swizz:
I am with @frenzzy, size were the main argument for my and my company Hyperapp's adoption. 

And our marketing at start were to promote the small size of Hyperapp. 

Even if it out of concern, I think it is mainly because of the awesome goal, it has been reached in the early stage of the project. 
Because now we can work safely on Hyperapp without worrying of adding some bytes. 

Because our code base is so tiny that will never add too much to go further than the 2 KB.

PR #697, opened 2018-05-24
Lazy components can return arrays now

### LINKED ISSUE #695: Lazy components cannot return arrays
```jsx
var h = hyperapp.h, app = hyperapp.app;
/*@jsx h*/
var SomeComponent = function () {
    return function () {
        return [<p>Hello</p>];
    };
};

app({}, {}, function () {
    return (<div oncreate={function (e) {
        setTimeout(function () { alert(e.innerHTML); }, 0); // <undefined></undefined> would be alerted.
    }}>
        <SomeComponent/>
    </div>);
}, document.getElementById("app"));
```

See JSFiddle: https://jsfiddle.net/vmvaywrr/2/

infinnie:
Fix #695 only (with corresponding tests but without other code in #684)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -53,6 +53,9 @@ export function app(state, actions, view, container) {
   }
 
   function resolveNode(node) {
+    if (Array.isArray(node)) {
+      return [].concat.apply([], node.map(resolveNode))
+    }
     return typeof node === "function"
       ? resolveNode(node(globalState, wiredActions))
       : node != null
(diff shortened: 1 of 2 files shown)


frenzzy:
Does it add support for returning an array from a root view?

infinnie:
No, @frenzzy, not now.

jorgebucaran:
Thanks, @infinnie, but I'm not going to add this.

PR #700, opened 2018-05-27
readme: clarify local component state with example

atomiks:
I feel like this needs clarification. Currently it exists only as an issue (https://github.com/hyperapp/hyperapp/issues/147) but I think it needs to be more prominently noted.


jorgebucaran:
Thanks, @atomiks!

PR #718, opened 2018-07-02
Fix typing issue with Component<>

### LINKED ISSUE #714: Type checking for props
I love this framework and wish to use it in more projects.

But I am wondering is there a recommanded way to do **type checking** for props (It is very important in some cases) like what [prop-types](https://github.com/facebook/prop-types) do.

I konw it is not a good idea to embed this functionality into the main library, but writing a plugin may be fine.

Alber70g:
Fixes #714

`Component` return type is `VNode<Attributes> | View<State, Actions>`. These
two aren't compatible with each other. Therefore the `Component` can only be a
function that returns a `VNode`. This has to do with the `h` only returning a
`VNode<Attributes>`.

Another fix could be to allow `h<Attributes()` to return `VNode<..> | View<..>`
but this will trickle down other issues with the typings.

This way there is a distinct way of creating a `VNode` from `Component` and
creating the `View(state, actions)` function typed as `View<TState, TActions>`

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -20,7 +20,6 @@ export interface VNode<Attributes = {}> {
 export interface Component<Attributes = {}, State = {}, Actions = {}> {
   (attributes: Attributes, children: Array<VNode | string>):
     | VNode<Attributes>
-    | View<State, Actions>
 }
 
 /**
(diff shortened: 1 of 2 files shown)


jorgebucaran:
@Alber70g :) 👍

PR #719, opened 2018-07-03
readme: add community packages/projects

atomiks:
I think there are some Hyperapp-related things currently missing from the docs. The README is seen as the one-stop source for everything Hyperapp (to me), so it makes sense that it should include these for visibility reasons.

### Starter Kit section

Why? Although beginners might use the "browser-only" method for toying around, they will want to also get started with a production app as well. Setting up Webpack can be difficult and they might not realise these exist. Linking to these starter kits lets them know there are ways get started with a build setup quickly.

### Packages section

Why? Solutions to common problems (routing, transitions, SSR) should be visible in the docs so readers know they exist without needing to externally search for them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,13 @@ If you don't want to set up a build environment, you can download Hyperapp from 
 <script src="https://unpkg.com/hyperapp"></script>
 ```
 
+### Starter Kits
+
+If you are looking to get started with a production app using a build setup quickly and easily, there are community starter kits to help.
+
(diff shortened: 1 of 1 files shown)


selfup:
I can dig it 🙏

jorgebucaran:
@atomiks Would it be feasible/convenient/a good idea to merge these two sections? And can we move them to the bottom?

okwolf:
Would it be better to just add a single link to [awesome-hyperapp](https://github.com/hyperapp/awesome-hyperapp) and only keep that one up-to-date?

jorgebucaran:
Looks good. V2 docs will have some of this stuff and then some other. Thanks!

Issue #721, opened 2018-07-03
V2 Lazy lists/components


jorgebucaran:
The idea is to avoid building a virtual DOM patching altogether. Think of it as built-in component memoization. Lazy views, not be confused with V1's deprecated "lazy components", is a way to tell Hyperapp to reuse a subtree from the old virtual DOM when patching the DOM and skip the diff.

Applications are not lazy out of the box. Instead, we designate what components will be lazy based on one or more properties in the state and Hyperapp will evaluate them only when the specified properties change. Large lists are usually great candidates for laziness.

To create a lazy view, import `Lazy` from the `hyperapp` package and wrap a view function like this.

```jsx
import { Lazy } from "hyperapp"
import { Foo } from "./components/foo"

export const LazyFoo = props => <Lazy view={Foo} foo={props.foo} bar={props.bar} />
```
Hyperapp will render `Foo` only when `props.foo` and `props.bar` change. 

To understand why laziness is useful, you need to remember that Hyperapp calculates the virtual DOM from scratch whenever the state changes. Then, compares it against the actual DOM, applying changes. While this is already highly-efficient, laziness allows us to optimize this process further.

Here's a working example. Try it in this [code playground](https://codepen.io/anon/pen/NVKMwz?editors=0010) first to see what it does.

```jsx
import { h, app, Lazy } from "hyperapp"

const randomColor = () =>
  Array.from({ length: 6 }).reduce(
    (color, i) => color + "0123456789ABCDEF"[Math.floor(Math.random() * 16)],
    "#"
  )

const List = props => (
  <div style={{ color: randomColor() }}>
    <h1>List</h1>
    {props.list.map(i => (
      <span>{i}</span>
    ))}
  </div>
)

const LazyList = props => <Lazy view={List} list={props.list} />

const Increment = state => ({ ...state, count: state.count + 1 })
const MoreItems = state => ({
  ...state,
  list: state.list.concat(state.list.length + 1)
})

app({
  init: () => ({
    list: [1, 2, 3],
    count: 0
  }),
  view: state => (
    <main>
      <button onclick={Increment}>{state.count}</button>
      <button onclick={MoreItems}>More Items</button>
      <List list={state.list} />
      <LazyList list={state.list} />
    </main>
  ),
  node: document.getElementById("app")
})
```

[...]

Issue #717, opened 2018-07-02
What are lifecycle events for?


jorgebucaran:
What are lifecycle events for and **can we do without them**? Below is each lifecycle event and its use cases. 

## `oncreate(element)`
1. Create fade/slide-in transitions when an element appears on the screen.
2. Add global event listeners, start timers, create websockets.
3. Fetch some data when the element appears on the screen.
4. Simulate local component state by storing data in the element.
5. Modify the element attributes, children or inner html (integration with third party libraries).
6. Set innerHTML.

### Alternatives
1. Store the initial style in the state, `style.opacity: 0` and use a tick subscription to create an interval. The interval will last **until** you go past/pass a threshold `style.opacity >= 1`. Update the state interactively at every interval `=> opacity+.2`. The same strategy can be used for fade/slide-out transitions.

2. Use a subscription for global event listeners, timers, websockets, etc.

3. Use the `init` function to set the initial state and immediately start an effect that will fetch your data and deliver it to another action. Inside that action you can set the next state. You could even set the next state and start another effect creating effectively a chain of effects.
    - Elements appearing or disappearing on the screen is a by-product of the state. The state changing is the result of user interacting with the DOM like clicks, etc., or external events, like global events, timers, url changes, geolocation changes, etc. Therefore, you can fetch data which will be "consumed" by some component via the same actions that update the state so that those components are displayed in the first place! 
    - Remember that actions can return not only the `=> nextState` but also start an effect `=> [nextState, someEffect]`.

4. [Custom Elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements)
      - Your custom elements can trigger [CustomEvents](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent) and Hyperapp can respond to them via subscriptions (receiving). 
  
    - Conversely,  you can trigger effects that communicate with Custom Elements via Custom Events (sending/giving).

5. You can watch for changes being made to the DOM tree using [Mutation Observers](https://developers.google.com/web/updates/2012/02/Detect-DOM-changes-with-Mutation-Observers). It's not as subtle as using lifecycle events, but third party integrations are not subtle anyway. In fact, the result will be more maintainable code. Clean, declarative, pure Hyperapp code in one side, and imperative, hard-knocks code in the other side.

6. Provide better innerHTML support. Current support is questionable. For example: https://github.com/hyperapp/hyperapp/pull/688 /cc @frenzzy.

[...]

PR #668, opened 2018-03-28
Support xlink:href attribute

ngryman:
🎉💃🎶🍻🎈😎

fixes #211

PR #726, opened 2018-07-09
Hyperapp V2

jorgebucaran:
## Summary

👋 V2 is the pet name for the next version ([2.0.0](https://github.com/jorgebucaran/hyperapp/milestone/7)) of Hyperapp. Be sure to subscribe to this PR if you want to be informed of the progress towards the release. I'm still open to feedback and ideas, so share your feedback freely! 

To learn more about the upcoming architecture changes check the issue list below.

- [Actions API](https://github.com/jorgebucaran/hyperapp/issues/749)
  - [Payload creators](https://github.com/jorgebucaran/hyperapp/issues/830)
- [Effects API](https://github.com/jorgebucaran/hyperapp/issues/750)
- [Subscriptions API](https://github.com/jorgebucaran/hyperapp/issues/752)
- [Lazy Lists/sCU](https://github.com/jorgebucaran/hyperapp/issues/721)
- [Middleware/DevTools](https://github.com/jorgebucaran/hyperapp/issues/753)

## Status Progress

- [ ] Docs (65%) 🌶 
- [ ] Tests
- [x] Core
  - [x] Implement Action API #750 #751 
  - [x] Implement the Effect API #750 
      - [x] Add built-in batching for effects
  - [x] Implement Subscriptions API #752 
  - [x] Optimize `defer` function (`setTimeout` to use faster `Promise.resolve`)
- [x] VDOM
  - [x] Improve performance #499
  - [x] Lazy lists #721 
  - [ ] ~~Dynamic components #760~~
  - [x] Add built-in [classcat](https://github.com/JorgeBucaran/classcat)-like support for `className` attribute
- [ ] ~~Typings~~

## [Try it!](https://www.npmjs.com/package/hyperapp/v/alpha)

<pre>
npm i <a href=https://www.npmjs.com/package/hyperapp/v/beta>hyperapp@beta</a>
</pre>


mindplay-dk:
I also noticed there's [a block of 3 commented-out functions](https://github.com/hyperapp/hyperapp/blob/V2/src/index.js#L516) left behind in the source.

src/index.js
```diff
@@ -1,390 +1,586 @@
...
+  }
+
+  var render = function() {
+    updateInProgress = false
+
+    if (subscribe) {
+      subscriptions = listen(subscribe(state), subscriptions, dispatch)
+    }
+
+    if (view) {
```
frenzzy: Why view is optional now?

SteveALee:
REAME seems unchanged. As I've not been part of the V2.0 discussion I need a way get context and understand the goals. At least I prefer that before reading code.

okwolf:
Wow, still single file source. Is that the plan to keep that going forward?

jorgebucaran:
@okwolf For starters, yeah. I had everything in its own file until the last moment.

lukejacksonn:
I was really looking forward to seeing it all split up.. I feel the one file things conceals different functions reliance on certain globals. You can't get away with that when everything is split up. Still nice code though. Much more comprehensible that the last one, great job.

jorgebucaran:
> I feel the one file things obfuscates different functions reliance on the various globals.

We'd just have a `constants.js` file and export the vnode types and every function required by any given function would need to import it at the top of the file.

lukejacksonn:
Yeh.. I know how it would work. I was looking forward to that structure, that is all. Thought it would make it more clear to people what is going on and what relies on what. It is ok though, I will do it myself sometime as a learning exercise.

jorgebucaran:
@lukejacksonn Do you think it would be better to split up the code then?

jorgebucaran:
> @lukejacksonn Yeh.. I know how it would work.

For sure, but then I misunderstood this:

> You can't get away with that when everything is split up because you have to pass a function everything it needs

---

My point is, while everything is crammed in one file, every function is independent and already expects as input everything it needs to work on and it doesn't rely on stuff which is in the global scope, other than the stuff you'd import to the top-level scope of the module in which the function is declared if it was in a file.

In other words, the amount of refactoring needed to put everything into files is minimal, that's what I meant. Sorry, if I said something confusing! 😄

lukejacksonn:
You didn't say anything confusing 😅 and I can see that it would easily refactor. I love the single file thing too, but I am starting to wonder if it would be easier to grok in individual files.

Swizz:
I love the one function per file too.
But it could be confusing too, you will need to dive into an `src` folder with multitude of files, to find the entrpypoint, mostly `index` then to jump from file to file to catch all the function calls.

Here, it is like reading a story, you learn how the function works, then, when it is later called, you already know the function and how it works.

I was for the individual function files, but now I am thinking about it, I am not sure anymore.

atomiks:
Maybe add the "utility"-type functions in their own file as named exports to be imported into the main file that exports `h` and `app`?

```
utils.js
---- export function functions() {}
---- export function needed()
index.js
---- import { functions, needed } from './utils'
---- export function h() {}
---- export function app() {}
```

src/index.js
```diff
@@ -1,390 +1,586 @@
...
+var XLINK_NS = "http://www.w3.org/1999/xlink"
+var SVG_NS = "http://www.w3.org/2000/svg"
 
-  while (rest.length) {
-    var node = rest.pop()
-    if (node && node.pop) {
-      for (length = node.length; length--; ) {
-        rest.push(node[length])
+var EMPTY_OBJECT = {}
+var EMPTY_ARRAY = []
```
Swizz: :thinking:

src/index.js
```diff
@@ -1,390 +1,586 @@
...
+var merge = function(a, b) {
+  var target = {}
+
+  for (var i in a) target[i] = a[i]
+  for (var i in b) target[i] = b[i]
+
+  return target
+}
+
+var resolved = typeof Promise === "function" && Promise.resolve()
```
Swizz: So now Promises are used in core ? (if supported)

src/index.js
```diff
@@ -1,390 +1,586 @@
...
+    }
+  }
 
-      var oldKeyed = {}
-      var newKeyed = {}
-      var oldElements = []
-      var oldChildren = oldNode.children
-      var children = node.children
+  return typeof name === "function"
+    ? name(props, (props.children = children))
```
Swizz: A Component will received children property and children argument ?

src/index.js
```diff
@@ -1,390 +1,586 @@
...
 
-      while (i < oldChildren.length) {
-        if (getKey(oldChildren[i]) == null) {
-          removeElement(element, oldElements[i], oldChildren[i])
-        }
-        i++
-      }
+  var dispatch = function(obj, data) {
+    typeof obj === "function"
+      ? dispatch(obj(state, data))
```
Swizz: Clever.

But the whole codeblock is pretty unreadable :thinking:

src/index.js
```diff
@@ -1,390 +1,586 @@
...
+    }
+
+    if (view) {
+      node = patch(container, element, node, view(state), eventProxy)
     }
-    return element
   }
+
+  dispatch(props.init)
 }
```
Swizz:
```jsx
return dispatch
```
?

src/index.js
```diff
@@ -1,390 +1,586 @@
...
+export var listen = function(sub, oldSub, dispatch) {
+  return isArray(sub) || isArray(oldSub)
+    ? batch(toArray(sub), toArray(oldSub), dispatch)
+    : sub
+      ? oldSub ? update(sub, oldSub, dispatch) : start(sub, dispatch)
+      : oldSub ? cancel(oldSub) : oldSub
+}
+*/
+
+export function app(props) {
```
Swizz: How about instead of inline exports, using the following block at the top of the bottom of the file
```jsx
export { h, app }
```
To make clear that `h` and `app` will be exported.

I am using this pattern and I found it verry clever. (I learned it from Erlang).

src/index.js
jorgebucaran: Nope, this is just a trick to call your callback as quickly as possible.

src/index.js
jorgebucaran: A small optimization, mainly used for text vnodes that don't have children, but still use the same VNode shape.

src/index.js
jorgebucaran: Good catch. It's an implementation detail. If you need `children` for some reason, use `props.children`. I am still open to change this.

jorgebucaran: @frenzzy

src/index.js
jorgebucaran: I agree. I'm working on a new implementation. 👍

src/index.js
jorgebucaran: Yes, I'm still not sure if I should return it or not.

src/index.js
jorgebucaran: I _could_ like that. Let me do some research.

src/index.js
dangerousdan: `dispatch` doesn't return anything, so it will be `undefined` either way, right?

src/index.js
```diff
@@ -1,390 +1,586 @@
...
-      while (i < oldChildren.length) {
-        if (getKey(oldChildren[i]) == null) {
-          removeElement(element, oldElements[i], oldChildren[i])
-        }
-        i++
-      }
+  var dispatch = function(obj, data) {
+    typeof obj === "function"
+      ? dispatch(obj(state, data))
+      : Array.isArray(obj)
```
dangerousdan: Should this be `isArray(obj)` for the sake of consistency?

Swizz: I would say Yes!

https://github.com/hyperapp/hyperapp/blob/2b3d85625321038ed2369ddb4ecc320a9e20984e/src/index.js#L12

jorgebucaran: Yup! Good catch and thanks. I'll push updated code soon.

src/index.js
```diff
@@ -1,390 +1,586 @@
-export function h(name, attributes) {
-  var rest = []
-  var children = []
-  var length = arguments.length
+var DEFAULT = 0
+var RECYCLED_NODE = 1
+var TEXT_NODE = 2
```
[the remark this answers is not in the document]
jorgebucaran: @dangerousdan Haha you do have a point. I'll fix it.

jorgebucaran:
> @atomiks https://github.com/hyperapp/hyperapp/pull/726#issuecomment-403784063 

I considered something like it, but I feel now that a mixed approach would break the balance. It's all or nothing. Split the code or not, but not in between is what I mean :)

src/index.js
TheNando: Chaining ternaries is not so bad if you change the formatting a little. I know this is unorthodox, but it definitely helps with clarity.
```
typeof obj === "function" ?
    dispatch(obj(state, data))
: isArray(obj) ?
    obj[1].effect(obj[1].props, dispatch, setState(obj[0]))
: obj != null && typeof obj.effect === "function" ?
    dispatch([state, obj])
: setState(obj)
```

src/index.js
```diff
@@ -1,390 +1,586 @@
...
+
+var updateProperty = function(
+  element,
+  name,
+  lastValue,
+  nextValue,
+  eventProxy,
+  isSvg
+) {
+  if (name === "key") {
```
Mytrill: Could also be 
```js
if(name === "style") {
  ...
} else if (name !== "key") {
  ...
}
```
To avoid an empty if

TheNando: I'm a fan of the "exit early" pattern
```
if (name === "key") return

// Do other ifs
```

src/index.js
frenzzy: How to get the state of your application at specific moment of time? For example for HMR

src/index.js
frenzzy: It is definitely more clear when all exports goes near to each other, but I found out that for example webpack generate more noise when you export a name instead of an expression.

src/index.js
frenzzy: https://eslint.org/docs/rules/no-nested-ternary

src/index.js
frenzzy: `props.children` definitely cooler then the second argument, see pros and cons here: https://github.com/hyperapp/hyperapp/issues/569
I am worried only about performance and memoization:
https://github.com/hyperapp/hyperapp/blob/2b3d85625321038ed2369ddb4ecc320a9e20984e/src/index.js#L458

If everything is ok with `props.children` would be nice to remove `children` from the second argument IMO.

src/index.js
```diff
@@ -1,390 +1,586 @@
...
+var createElement = function(node, lifecycle, eventProxy, isSvg) {
+  var element =
+    node.type === TEXT_NODE
+      ? document.createTextNode(node.name)
+      : (isSvg = isSvg || node.name === "svg")
+        ? document.createElementNS(SVG_NS, node.name)
+        : document.createElement(node.name)
+
+  var props = node.props
+  if (props.onCreate) {
```
okwolf: So lifecycle events are still supported and use camelCase now? 🤔

jorgebucaran: I gave in to camelCase :)

> ...lifecycle events

They're in this branch, for now. I still haven't decided whether to remove them or not. I know we don't need them, but having them can be useful. You can help me decide by adding your feedback to #717.

src/index.js
jorgebucaran: I wonder if this is just personal taste or there's something to it. My brain finds it a lot easier to think if something is true, instead of if not something do something.

src/index.js
okwolf: So should interop be done with `subscribe` then?

jorgebucaran: @frenzzy There is currently no way other than using subscriptions. But I am open to exploring alternatives if need be.

jorgebucaran: @okwolf Bingo.

jorgebucaran: Like Elm. But still, if we can think of an alternative, e.g., app.getState(), I am happy to discuss it.

okwolf: @jorgebucaran we should start by trying to build things that need interop and find where this approach gets ugly.

okwolf:
Will `hyperapp@2.0.0-beta.0` be published from this branch or not until merged to master?

jorgebucaran:
@okwolf Published from master after it's merged (with docs). Future V1 releases will be published from the V1 branch, though.

src/index.js
```diff
@@ -1,390 +1,586 @@
...
+      subscriptions = listen(subscribe(state), subscriptions, dispatch)
+    }
+
+    if (view) {
+      node = patch(container, element, node, view(state), eventProxy)
     }
-    return element
   }
+
+  dispatch(props.init)
```
znk: ```dispatch(props.init || {})```

Providing `init` with an empty state permits to omit it in case of concise example or stateless app. We can write like:
```
app({
  view: () => (
    <div>
      Lorem ipsum dolor si amet
    </div>
  ),
  container: document.body
})
```
instead of:
```
app({
  init: () => ({ }),
  view: () => (
    <div>
      Lorem ipsum dolor si amet
    </div>
  ),
  container: document.body
})
```

Swizz: `props.init` is a function, so it should be 
```jsx
dispatch(props.init || function() {})
```
or in a more concise way
```jsx
dispatch(props.init || Function())
```

Swizz: Nevermind, `dispatch` works with objects too.

_(part of the unreadable ternary :roll_eyes: )_

src/index.js
```diff
@@ -1,390 +1,586 @@
...
+  var subscriptions = []
+  var container = props.container
+  var element = container && container.children[0]
+  var node = element && recycleElement(element)
+
+  var setState = function(newState) {
+    if (state !== newState) {
+      state = newState
+
+      if (!updateInProgress) {
```
kumarabhishek: What will happen if updateInProgress is true ? Are we ignoring all setState() call till updateInProgress is true ? Just a wild thought of saving it to tempState(keep overwriting till updateInProgress is true) and apply the latest value of tempState once updateInProgress becomes false.

src/index.js
```diff
@@ -1,390 +1,586 @@
...
-    }
+  return nextNode
+}
 
-    var cb = node.attributes && node.attributes.onremove
-    if (cb) {
-      cb(element, done)
-    } else {
-      done()
+export var h = function(name, props) {
```
kumarabhishek: As we are exporting h object here so this source file becomes ES6 module. So why then we are not using all power of ES6. To name a few, using const, Object.assign, spread operator, etc which will make code more readable, smaller (not larger at least) and simple without adding any burden to performance. Making methods will easily become one liners.

src/index.js
```diff
@@ -1,390 +1,586 @@
...
+  })
+}
 
-      while (k < children.length) {
-        var oldKey = getKey(oldChildren[i])
-        var newKey = getKey((children[k] = resolveNode(children[k])))
+var listen = function(sub, oldSub, dispatch) {
+  if (isArray(sub) || isArray(oldSub)) {
+    var out = []
+    var subs = isArray(sub) ? sub : [subs]
```
kumarabhishek: Here you are doing duplicate isArray check for sub which is already done in if condition. So the tradeoff is having to variables to store result of isArray for sub and oldSub versus calling isArray two times. I will consider more variables as cheaper than more function call in runtime. My idea is something like this:
```js
const isArraySub = isArray(sub)
const isArrayOldSub = isArray(oldSub)
if (isArraySub || isArrayOldSub) {
  let out = []
  let subs = isArraySub ? sub : [subs];
  ...
}
```

src/index.js
```diff
@@ -1,390 +1,586 @@
...
+*/
+
+export function app(props) {
+  var state
+  var updateInProgress
+  var view = props.view
+  var subscribe = props.subscribe
+  var subscriptions = []
+  var container = props.container
+  var element = container && container.children[0]
```
kumarabhishek: Is container.children always defined and thus container.children[0] is always valid ?

src/index.js
```diff
@@ -1,390 +1,586 @@
...
+    key: key,
+    type: type
+  }
+}
 
-      for (var name in attributes) {
-        updateAttribute(element, name, attributes[name], null, isSvg)
-      }
-    }
+var createTextVNode = function(text, element) {
```
kumarabhishek: If you use ES6 default parameter and organise params like below then createTextVNode is not needed at all:
```js
var createVNode = function(name, element, props={}, children=[],  key=null, type=TEXT_NODE) {
 ...
};
```

src/index.js
```diff
@@ -1,390 +1,586 @@
...
-    return out
+  var cb = node.props && node.props.onRemove
+  if (cb != null) {
+    cb(node.element, remove)
+  } else {
+    remove()
   }
+}
+
+var getKey = function(node) {
```
kumarabhishek: This method will not be needed if we ensure that node.key is by default null. This can be done using default parameter in functions related to node.

src/index.js
```diff
@@ -1,390 +1,586 @@
...
+  lastProps,
+  nextProps,
+  lifecycle,
+  eventProxy,
+  isSvg,
+  isRecycled
+) {
+  for (var name in merge(lastProps, nextProps)) {
+    if (
+      (name === "value" || name === "checked"
```
kumarabhishek: Just a matter of taste, i like single quotes for strings in JS and double quotes inside JSX/HTML.

joseluisq: It's not bad suggestion, in fact I prefer single quotes for strings too.

joseluisq: But in that case, It's  Hyperapp @jorgebucaran options :sweat_smile:

src/index.js
TheNando: Well, that won't work with `let`, only with var. Var hoists the definition to the top of the scope, so when it's referenced at `[subs]` it returns `undefined`. But with `let` it will throw a ReferenceError, since it is self referencing and won't exist until the line has executed. I actually think this is a typo and you were intending to ensure that `sub` is an array. If that's the case, you can just do:
```
var subs = [].concat(sub)
```
Which will ensure that `subs` ends up an array and works if `sub` is already and array or just a value.

src/index.js
TheNando: I don't think this file gets transpiled. Rollup supports specifying UMD, which is how this project is configured, which is probably where the export support comes from.
https://github.com/hyperapp/hyperapp/blob/2b3d85625321038ed2369ddb4ecc320a9e20984e/package.json#L27

TheNando:
@kumarabhishek Regarding using ES6 over ES5...

> My major comment is to use as much ES6 as possible to make code simpler, smaller and performant.

Simpler and smaller, maybe, but definitely not more performant.

Ultimately, it's per-browser implementation, but in most cases, ES6 is less performant. Declaring arrow functions are a little slower in Chrome and much slower FF. Destructuring is much slower across the board. Classes are slower except in FF. The new loops, generators, map/set, Object.assign, spread...pretty much all slower than es5 implementations.

http://incaseofstairs.com/six-speed/

jorgebucaran:
> @kumarabhishek My major comment is to use as much ES6 as possible to make code simpler, smaller and performant.

Remember https://github.com/hyperapp/hyperapp/issues/665? 😉 

For a longer response, see my comment in https://github.com/hyperapp/hyperapp/issues/486#issuecomment-350542197.

And can't help to remind you about our CONTRIBUTION style guidelines.

- https://github.com/hyperapp/hyperapp/blob/master/CONTRIBUTING.md#style

---

Having said that, I'm definitely open to revise and improve! So, thank you, for your feedback! 🙏

package.json
```diff
@@ -43,12 +43,12 @@
   },
   "devDependencies": {
     "babel-plugin-transform-react-jsx": "^6.24.1",
-    "babel-preset-env": "^1.6.1",
-    "jest": "^22.4.3",
-    "prettier": "^1.11.1",
-    "rollup": "^0.57.1",
-    "typescript": "2.8.1",
-    "uglify-js": "3.3.16"
+    "babel-preset-env": "^1.7.0",
```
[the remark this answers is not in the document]
jorgebucaran: @k-murakami0609 Do you have any idea when these packages will become official (not beta)? Just curious.

src/index.js
jorgebucaran: Yup.

src/index.js
Swizz: @TheNando I have already made this suggestions one month ago on Slack 
see https://hyperapp.slack.com/archives/C5D86NQ3X/p1529772214000096?thread_ts=1529772214.000096&cid=C5D86NQ3X

If @jorgebucaran made the decision to do not use this implementation, he should have some good reasons.

Swizz: @TheNando I have already made this suggestions one month ago on Slack 
see https://hyperapp.slack.com/archives/C5D86NQ3X/p1529772214000096?thread_ts=1529772214.000096&cid=C5D86NQ3X

If @jorgebucaran made the decision to do not use this implementation, he should have some good reasons.

jorgebucaran: Thanks, @Swizz. Basically `[].concat` is a lot slower.

src/index.js
jorgebucaran: > @k-murakami0609 Are we ignoring all setState() call till updateInProgress is true ?

Nope! We're avoiding patching, that's it. In other words, state updates are pretty much immediate.

src/index.js
```diff
@@ -1,390 +1,586 @@
...
-      done()
+export var h = function(name, props) {
+  var node
+  var rest = []
+  var children = []
+  var length = arguments.length
+
+  while (length-- > 2) rest.push(arguments[length])
+
+  if ((props = props == null ? {} : props).children != null) {
```
joseluisq: More simplified version:

jorgebucaran: @joseluisq Yes, you are very much right! But there's new directive in town. Now, I'm trying to optimize for speed (while still optimizing for small bundles —this is not gonna end good hehe—) so I'm trying to avoid implicit coercion if possible.

src/index.js
```diff
@@ -1,390 +1,586 @@
...
+    if (rest.length <= 0) {
+      rest.push(props.children)
     }
+    delete props.children
```
joseluisq: A tip from TS core about `delete`:

```js
// Using 'delete' on an object causes V8 to put the object in dictionary mode.
// This disables creation of hidden classes, which are expensive when an object is
// constantly changing shape.
map.__ = undefined;
delete map.__;
```

Source: [TS core](https://github.com/Microsoft/TypeScript/blob/master/src/compiler/core.ts#L73-L78)
Perf: [delete vs undefined vs null](https://jsperf.com/delete-vs-undefined-vs-null/16)

Thoughts?

src/index.js
joseluisq: Awesome!

src/index.js
TheNando: @jorgebucaran But `: [subs]` supposed to be `: [sub]`, right?

jorgebucaran: @TheNando Yes, totally, that's a bug! Fixing it now :)

src/index.js
frenzzy:
```js
const Component = (props) => <div>{props.children}</div>

app({
  init: { children: 'Hello' },
  view: (state) => (
    <main>
      <Component {...state}>World</Component>
      <div>{state.children}</div>
    </main>
  ),
  container: document.body,
})
```
Can you guess what this will render without opening the demo?
1. `<main><div>Hello</div><div>World</div></main>`
2. `<main><div>Hello</div><div>Hello</div></main>`
3. `<main><div>World</div><div>World</div></main>`
4. throws an error

Demo: https://codepen.io/frenzzy/pen/Zjbqpq/left/?editors=0010

jorgebucaran: @frenzzy So, don't use `children` in your state?

frenzzy: lol, v2 allows everything in state but you should not use `children` key in it to prevent internal state mutation by hyperapp components implementation...

It is pretty common when an api returns data with `children` keys, for example for tree structure of pages, nav items, breadcrumbs, etc. It is not cool when you need to change api as a workaround for hyperapp.

If you still want to disallow `children` key, maybe it must throw an error instead of silently mutating state? Because it will inevitably lead to difficult to perceptible or hidden/uncatched errors in UI.

jorgebucaran: @frenzzy So you'd prefer to drop ({children,...props}) and go back to (props, children)?

frenzzy: yes, because it is less hacky implementation, we already discussed this in #569

src/index.js
jorgebucaran: Definitely `delete` is what we want _if_ we want ({ children, ...props }). If we decide to go with (props, children), then sure.

src/index.js
Swizz: 👌

jorgebucaran: Any alternatives? Why can preact get away with the same hack? 🤔

frenzzy: Purity forever! Say NO to mutations :D

jorgebucaran: Okay, so to clarify, we're back to (props, children)?

frenzzy: Yes!
- no mutations
- no pitfalls for `children` key
- simple memoization with `===`

src/index.js
joseluisq: Sure,but in that case, does it make sense to you to set undefined only?

package.json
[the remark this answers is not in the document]
ggassmann: @jorgebucaran I would not anticipate babel 7 leaving beta soon. They have many open issues.

src/index.js
SkaterDad: > but having them can be useful.

I'm glad the hooks are still there so I don't yet have to use custom elements and allow `this` in my codebase :nauseated_face:

src/index.js
```diff
@@ -1,390 +1,586 @@
...
-      }
+  var eventProxy = function(event) {
+    dispatch(event.currentTarget.events[event.type], event)
+  }
+
+  var render = function() {
+    updateInProgress = false
+
+    if (subscribe) {
+      subscriptions = listen(subscribe(state), subscriptions, dispatch)
```
SkaterDad: Coincidentally, I just read a post on the elm forum yesterday asking about when subs are updated. 
 They seem to be updated immediately after every state update.  https://discourse.elm-lang.org/t/when-is-subscriptions-called/1524

If hyperapp waits until a re-render, there could be brief moments where subscriptions that rely on the state are incorrect since they don't have the current state.

Moving this block of code to `setState`, directly after `state = newState` could prevent that.

package.json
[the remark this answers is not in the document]
jorgebucaran: Thanks for clarifying that! @ggassmann

src/index.js
jorgebucaran: Let's experiment with that! 👍

SkaterDad:
@jorgebucaran I've started messing around with the V2 Codepen & effects.

What are the `props` intended to be used for on effects?  Are they just a way to pass an argument to the "effect" function without resorting to higher order functions?

The way the souce reads, you define an effect like this:
```js
// definition
const MyEffect = {
  props: { testProp: true },
  effect: (props, dispatch) => {  // ignoring the 3rd arg which receives `undefined` from `setState`
    // do something else
    getStuff(props.url).then(data => dispatch(SaveData, data)).catch(e => dispatch(SaveError, e.message))
  }
}

// use
const MyAction = (state) => {
  return [newState, MyEffect]
}

// --------------------------------------------
// or construct the effect using a function
const MyEffect = props => {
   props,
   effect: () => {  /*stuff*/  }
}
const MyAction = (state) => {
  return [newState, MyEffect({ testProp: true})]
}
```

Alternative approach?  This seems way simpler (for both the app code & library code), but I could be missing something important about how effects are intended to work.

```js
// definition
const MyEffect = (props) => (dispatch) => {
    // do something else
    getStuff(props.url).then(data => dispatch(SaveData, data)).catch(e => dispatch(SaveError, e.message))
  }
}

// use
const MyAction = (state) => {
  const newState = { ...state, change: true }
  return [newState, MyEffect({ testProp: newState.change })]
}
```

Do you have any example effects prepared yet?

okwolf:
Does this have performance similar to https://github.com/hyperapp/hyperapp/issues/499#issuecomment-403257843?

jorgebucaran:
@okwolf Should be the same or _better_ because of less intermediate arrow functions. The #499 results are using the lazy feature, which is not yet implemented in V2.

okwolf:
IMHO the docs should be stripped bare in this branch so it's blatantly obvious they are missing. How can we start contributing to them? Start another branch off of this one?

package.json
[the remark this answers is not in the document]
frenzzy: Maybe today? https://twitter.com/left_pad/status/1031700991186489344

src/index.js
```diff
@@ -1,390 +1,604 @@
...
+        name !== "spellcheck" &&
+        name !== "translate" &&
+        !isSvg
+      ) {
+        element[name] = nextValue == null ? "" : nextValue
+        if (nullOrFalse) {
+          element.removeAttribute(name)
+        }
+      } else {
+        var ns = isSvg && name !== (name = name.replace(/^xlink:?/, ""))
```
frenzzy: Isn't this is too heavy to apply it for all attributes of all svg elements? Why don't you like #668 approach?

src/index.js
```diff
@@ -1,390 +1,604 @@
...
+    }
+  }
+  return oldSub
+}
 
-      var i = 0
-      var k = 0
+var start = function(sub, dispatch) {
+  return merge(sub, {
+    cancel: sub.effect(sub.props, dispatch)
```
okwolf: Why are subs still using `.props` but FX aren't? This comment appears wrong: https://github.com/hyperapp/hyperapp/issues/748#issuecomment-415935138.

jorgebucaran: Yup, still working on it! 🙇

src/index.js
okwolf: Please read [this](https://github.com/hyperapp/hyperapp/blob/master/CONTRIBUTING.md#style) regarding which ES spec we follow.

okwolf:
@jorgebucaran anything us contributors can do to help move this along? Test coverage? Rewriting docs?

jorgebucaran:
@okwolf Docs would be wonderful, but in the meantime, we could create a #docs channel on Slack. What do you think?

src/index.js
```diff
@@ -1,390 +1,583 @@
...
-        var oldKey = getKey(oldChildren[i])
-        if (oldKey != null) {
-          oldKeyed[oldKey] = [oldElements[i], oldChildren[i]]
-        }
-      }
+var refresh = function(sub, oldSub, dispatch) {
+  if (isArray(sub) || isArray(oldSub)) {
+    var out = []
+    var subs = isArray(sub) ? sub : [sub]
+    var oldSubs = isArray(oldSub) ? oldSub : [oldSubs]
```
Seldszar: I think there's an error with this line. In fact, the line should be `var oldSubs = isArray(oldSub) ? oldSub : [oldSub]` instead, no?

jorgebucaran: @Seldszar Yes! Thanks for catching this. 🎉🎉

okwolf:
@jorgebucaran to get the unit tests going, we should be able to port over all the VDOM/`patch` ones already, correct?

jorgebucaran:
@okwolf From Superfine? Yes, kind of. I also want to take the opportunity to revise the tests to see if there's room for improvement and if there is, make them better! And I want to rewrite everything for testmatrix too.

attack-monkey:
Heya - not sure if it's useful, but in my hyperapps (any apps really), I use an immutable update utility I made for updating state - https://github.com/attack-monkey/immutable-update. Helps to update deeply nested objects quickly. Happy for you to take it and use it / modify (if it's useful at all).

src/index.js
```diff
@@ -1,390 +1,603 @@
...
-          ? element.nodeValue
-          : recycleElement(element)
-      })
+      if (
+        name in element &&
+        name !== "list" &&
+        name !== "draggable" &&
+        name !== "spellcheck" &&
+        name !== "translate" &&
+        !isSvg
```
shiljopaulson: can we not use it like below instead of multiple AND conditions
```suggestion
        var array = ["list","draggable","spellcheck", "translate"]
       if(
       name in element &&  
       !array.includes(name) && 
       name && !isSvg)
```

src/index.js
```diff
@@ -1,390 +1,603 @@
...
+    })
+  }
 
-  function resolveNode(node) {
-    return typeof node === "function"
-      ? resolveNode(node(globalState, wiredActions))
-      : node != null
-        ? node
-        : ""
+  for (var i = 0, length = node.children.length; i < length; i++) {
```
shiljopaulson: Please store the length in a variable 
```suggestion
var len = node.children.length
  for (var i = 0, length = len; i < length; i++) {
```

src/index.js
```diff
@@ -1,390 +1,603 @@
...
+  if (cb != null) {
+    lifecycle.push(function() {
+      cb(element, lastProps)
+    })
+  }
+}
 
-    while (lifecycle.length) lifecycle.pop()()
+var removeChildren = function(node) {
+  for (var i = 0, length = node.children.length; i < length; i++) {
```
shiljopaulson: node.children.length in length in variable as menionted in the previous comment

src/index.js
```diff
@@ -1,390 +1,603 @@
...
+      }
     }
   }
+}
+
+var createElement = function(node, lifecycle, eventProxy, isSvg) {
+  var element =
+    node.type === TEXT_NODE
+      ? document.createTextNode(node.name)
+      : (isSvg = isSvg || node.name === "svg")
```
shiljopaulson: why do we want to do isSvg = isSvg

src/index.js
jorgebucaran: Embedding all the conditions in the same `if` expression should be less expensive than a method call.

src/index.js
jorgebucaran: `node.children.length` is stored in the `length` variable, declared alongside `i` in the setup part of the `for` loop. Or do you mean something else?

src/index.js
jorgebucaran: Do you mean the length of `lifecycle`? In this case, we can't store its length in a variable because we're mutating the array every time we call `.pop()` on it.

src/index.js
```diff
@@ -1,390 +1,603 @@
...
-          var keyedNode = oldKeyed[newKey] || []
+  var dispatch = function(obj, data) {
+    if (obj == null) {
+    } else if (typeof obj === "function") {
+      dispatch(obj(state, data))
+    } else if (isArray(obj)) {
+      if (typeof obj[0] === "function") {
+        dispatch(obj[0](state, obj[1], data))
+      } else {
+        obj[1].effect(obj[1], dispatch, setState(obj[0]))
```
okwolf: Wasn't there talk of using tuples for the effect and props instead of an object container?
```suggestion
         obj[1].[0](obj[1][1], dispatch, setState(obj[0]))
```

Or was that already debated and rejected? :thinking:

jorgebucaran: @okwolf Not rejected at all. Going to use tuples as originally suggested by @sergey-shpak.

src/index.js
kristianmandrup: Why still use `var` in 2019? Please use `const`. No need to be ES3/5 compliant

src/index.js
jorgebucaran: @shiljopaulson We're setting the value and checking it at the same time.

kristianmandrup:
Have a look at https://meiosis.js.org as well. Looks like a really awesome stream state management library that is highly composable.

src/index.js
jorgebucaran: @kristianmandrup Check out CONTRIBUTION.md—it's written there.

kristianmandrup:
When can we expect an alpha or beta release?

jorgebucaran:
@kristianmandrup https://github.com/jorgebucaran/hyperapp/issues/791

jorgebucaran:
@kristianmandrup Hopefully soon. I've been very busy, but I haven't forgotten about this. See https://github.com/jorgebucaran/hyperapp/issues/791

By the way, this PR is to talk about the PR. For anything else better to create an issue, join the Slack or reach me on Twitter if you like.

src/index.js
```diff
@@ -1,390 +1,604 @@
...
+    : createVNode(name, props, children, null, props.key, DEFAULT)
+}
 
-        var oldKey = getKey(oldChildren[i])
-        if (oldKey != null) {
-          oldKeyed[oldKey] = [oldElements[i], oldChildren[i]]
-        }
-      }
+var cancel = function(sub) {
+  sub[1][2]()
```
okwolf: I think this is actually a typo and should instead be:
```js
sub[2]()
```

jorgebucaran:
You can install the V2 branch from npm now. This should download the latest alpha release. An alpha tag doesn't necessarily imply there will be a beta tag and it's possible we'll go straight to 2.0 from here.  

<pre>
npm i <a href=https://www.npmjs.com/package/hyperapp/v/alpha>hyperapp@alpha</a>
</pre>

will-stone:
Hi Jorge. Thanks for all the hard work on v2. Could you (or another volunteer) get a CodeSandbox up-and-running with the v2 README example? I'm struggling.

zaceno:
@will-stone If by "v2 README example" you mean a simple counter, here's one version: https://codepen.io/zaceno/pen/ebgvGW

will-stone:
Thanks @zaceno, I meant this: https://github.com/jorgebucaran/hyperapp/tree/V2#getting-started

zaceno:
@will-stone Ah, hehe. Hadn't seen that one. Well yeah, it's wrong. Here's an up to date version: https://codepen.io/zaceno/pen/daLbQz

jorgebucaran:
@will-stone Sorry! Error is fixed now.

zaceno:
@will-stone In case you're interested, here's the same example boiled down to the least amount of code that has the intended behavior (but doesn't work well as a teaching-example):

https://codepen.io/zaceno/pen/PVgqqJ

src/index.js
mrozbarry: Good catch, I'll PR something in the next hour

src/index.js
```diff
@@ -505,116 +526,67 @@ export var h = function(name, props) {
...
-  }
-}
-
-var cancel = function(sub) {
-  sub[2]()
+var isSameAction = function(a, b) {
+  return isArray(a) && isArray(b) && typeof a[0] === "function" && a[0] === b[0]
 }
 
-var isSameValue = function(a, b) {
```
SkaterDad: @jorgebucaran This function was used inside of `resolveNode`, so the current alpha release is broken.

jorgebucaran: @SkaterDad Thanks, I'll fix that.

src/index.js
```diff
@@ -1,390 +1,648 @@
...
 
-    var node = resolveNode(view)
+var resolved = typeof Promise === "function" && Promise.resolve()
 
-    if (container && !skipRender) {
-      rootElement = patch(container, rootElement, oldNode, (oldNode = node))
+var defer = !resolved
+  ? setTimeout
+  : function(cb) {
+      return resolved.then(cb)
```
frenzzy: If I understand correctly this function produce an infinite promise chain, is it a memory leak?

mrozbarry: Not quite, `resolved` is never updated, so it's essentially forcing js to wait a single tick. The alternative is something like `setTimeout(cb, 0)` which I believe isn't as fast as a quick-resolve like this.

jorgebucaran: Yeah. `setTimeout` schedules a task, while our `defer` schedules a "micro" task, which has more priority. All this is untested science. I don't if it'd make a difference to use `setTimeout` instead. I can remove this until we've seen positive benchmark results if you want.

SkaterDad: For what it's worth, `ivi` also uses `Promise.resolve` to schedule a microtask. 

https://github.com/localvoid/ivi/blob/cdaf62c521b1d32c601ab5ed0a0cd226ed5921c5/packages/ivi-scheduler/src/index.ts#L1-L7

frenzzy: Yes, but they create a new `Promise.resolve()` every tick and not use `.then()` for the same resolved promise up to infinity.

I thought it may cause a memory leak and created a small demo with a try to reproduce it: https://frenzzy.github.io/test/then.html (or [codepen](https://codepen.io/frenzzy/pen/gEPdRa?editors=1010))
But looks like there are no problem at least in the latest Chrome.

SkaterDad: @frenzzy  I just tried your example in IE11 with a Promise polyfill, and there was no memory leak.

localvoid: @SkaterDad `ivi` is using [its own microtask queue](https://github.com/localvoid/ivi/blob/cdaf62c521b1d32c601ab5ed0a0cd226ed5921c5/packages/ivi/src/scheduler/index.ts#L91) and `Promise.resolve` is invoked only when microtask is added [outside of a scheduler context](https://github.com/localvoid/ivi/blob/cdaf62c521b1d32c601ab5ed0a0cd226ed5921c5/packages/ivi/src/scheduler/index.ts#L131). There are edge cases when some DOM operations are blocked when they executed outside of an event context, that is one of the reasons why React and ivi executes event handlers inside of a scheduler context.

jorgebucaran: Okay, folks, I changed it to:

```js
var defer =
  typeof Promise === "function"
    ? function(cb) {
        Promise.resolve().then(cb)
      }
    : setTimeout
```

jorgebucaran: Okay, folks, I changed it to:

```js
function defer(cb) {
  Promise.resolve().then(cb)
}
```

src/index.js
selfup: Another reason: `ES2016 Specifications included the includes() method` 😭 

We would have to use `indexOf` and as @jorgebucaran said, `if` expression will be less expensive 🙏

mrozbarry:
In terms of removing the lifecycle hooks (which I think is perfectly fine), I do think that we should have some sort of escape hatch to grab the **vdom** element, like a react ref. I think it should run a callback that doesn’t give you access to dispatch or anything. It should be up to the dev to have a subscription to an event that you could run off the `vdom.node` or main app container.

**Pros:**

 - Could help facilitate some types of testing
 - It's a single callback, not a series of lifecycle hooks
 - It could work like some lifecycle hooks as a last resort
 - May be a pragmatic choice, since mutation observer may be slow or difficult to work with in some instances.

**Cons:**

 - It's still callback that behaves similar to lifecycle hooks
 - It's further away from the DOM (and Elm Architecture)
 - It's a side-effect

**Alternatives:**

 - Just like the current solution with mutation observer subscriptions, elements could be tagged and selected in a subscription and managed internally there.
   - This turns the above implementation into a controlled side effect
   - This does not give us the vdom element, which may contain important information in props

---

I just wanted to dump this here to think through a bit. I'm not convinced this is better than lifecycle hooks or the current mutation observer subscription solution.

SkaterDad:
@mrozbarry Can you give an example of how this might look?

The code I've read from React libraries that grab refs always looks way over-complicated and gross.

mrozbarry:
@SkaterDad maybe something like this:

```
const handleRef = (vdom, container) => {
  const event = new Event('node.capture', { node: vdom.node, props: vdom.props });
  container.dispatchEvent(event); // Some subscription can listen for this.
};

// ...
h('div', { ref: handleRef })
```

jorgebucaran:
I've given alternatives to typical use cases for lifecycle events in https://github.com/jorgebucaran/hyperapp/issues/717#issue-337438224, so I don't see what problem the things @mrozbarry suggested would solve.

mrozbarry:
@jorgebucaran the only thing is solves is getting access to the vdom data rather than just the DOM node. Not sure that it's any benefit over the current solutions, though.

src/index.js
```diff
@@ -619,7 +619,9 @@ export function app(props) {
       if (typeof obj[0] === "function") {
         dispatch(obj[0](state, obj[1], props))
       } else {
-        obj[1][0](obj[1][1], dispatch, setState(obj[0]))
+        obj.slice(1).map(function(fx) {
+          fx[0](fx[1], dispatch)
+        }, setState(obj[0]))
```
diontools: It may be premature optimization, but I think slice is expensive.
https://jsperf.com/slice-1-vs-for-loop-vs-foreach

jorgebucaran: This is a size tradeoff I think we can afford as I don't think you'd need to dispatch more than a few effects in parallel at any given time. We can definitely revisit this in the future; it isn't saving us a lot of bytes anyway.

Thank you, @diontools. 💯

src/index.js
```diff
@@ -436,120 +412,70 @@ export var Lazy = function(props) {
...
-    if (!(state === newState || renderLock)) {
-      defer(render, (renderLock = true))
+    if (!(state === newState || lock)) {
+      defer(render, (lock = true))
     }
     state = newState
   }
 
   var dispatch = function(obj, props) {
-    if (obj == null) {
```
SkaterDad: Was this removed purely for size?

I was use/abusing that null check to avoid needing to wrap `dispatch` calls in conditionals sometimes.

jorgebucaran: @SkaterDad 

> Was this removed purely for size?

Partly, yes. The other reason is that I don't think there is a use case for returning `null` or `undefined` from an action. Is there?

> I was use/abusing that null check to avoid needing to wrap dispatch calls in conditionals sometimes.

Show me an example, please. 🙏

SkaterDad: Before the recent commits, I could do this:
```js
function myEffect(props, dispatch) {
  // props.init could be undefined.
  // but dispatch was checking for `obj == null`, so it was okay
  dispatch(props.init);

 // do other stuff
  setTimeout(() => {
    dispatch(props.action);
  }, 100)
}
```

Now my effect code will have to to wrap optional props in an `if...then` statement:
```js
function myEffect(props, dispatch) {
  // props.init could be undefined.
  // If we don't check for truthiness here, dispatch will end up doing `setState(undefined)`
  if (props.init) {
    dispatch(props.init);
  }

 // do other stuff
  setTimeout(() => {
    dispatch(props.action);
  }, 100)
}
```

jorgebucaran: @SkaterDad Ah, I see. Thanks. Convenient, but maybe not so if you needed to do something when `props.init` is `undefined` instead. In that case, we'll end up with two checks: Hyperapp's and yours. I'll note this on the documentation at some point. 👍

SkaterDad:
> SandBox/CodePen?

Here's a sandbox demonstrating how SVG sprites break when you don't check for `xlink:href`.
https://codesandbox.io/s/186m022m57

Lately I've come to prefer SVG sprites for icon sets because of how small they are, so it would be great to have this supported.

SkaterDad:
Found another broken thing after the latest updates.  If a subscription is removed from the array completely, not just made false-y, the cancel function doesn't get called anymore.  The example is a little contrived, but I was working on something which requires the previous behavior.

https://codesandbox.io/s/oo98pkp3wq

mrozbarry:
@SkaterDad Yup, looks like subscriptions are currently implemented to be toggled, not fully removed; https://github.com/jorgebucaran/hyperapp/blob/V2/src/index.js#L50 . That seems wrong, especially if you are building subscriptions from a state array map.

SkaterDad:
@mrozbarry Glad I'm not the only one to see it.

Back to `alpha.8` for now.

jorgebucaran:
> @SkaterDad Here's a sandbox demonstrating how SVG sprites break when you don't check for xlink:href.

The only main offenders are Safari and iOS right now. 

```html
<use href="https://codesandbox.io/s/k93mkj3j7v"/>
``` 

...works on Chrome and Firefox. I might add it back, though. Let 's see.

jorgebucaran:
> @SkaterDad  If a subscription is removed from the array completely, not just made false-y, the cancel function doesn't get called anymore.

Not difficult to fix, but to be sure, did you know this works?

```js
return [some, subs, before, state.enable && [[mySub]]]
```

> @mrozbarry That seems wrong, especially if you are building subscriptions from a state array map.

Ah, you may be onto something. Any idea of how that might look?

SkaterDad:
@jorgebucaran Yep, I know that works, but I've been building something which generates the array of subs based on state.

jorgebucaran:
@SkaterDad How does that work or look like? If you can paste the relevant code here, it would be helpful.

jorgebucaran:
@SkaterDad How does that work or look? If you can paste the relevant code here, it would be helpful.

You should be careful about how you generate subscriptions. Array indices are to subscriptions what keys are to nodes in the VDOM. If you reduce or filter the subscriptions array, it grows or shrinks, and we end up resetting some subscriptions with didn't have to.

SkaterDad:
> @SkaterDad How does that work or look? If you can paste the relevant code here, it would be helpful.
> 
> You should be careful about how you generate subscriptions. Array indices are to subscriptions what keys are to nodes in the VDOM. If you reduce or filter the subscriptions array, it grows or shrinks, and we end up resetting some subscriptions with didn't have to.

Here's some psuedocode, but it illustrates what I'm doing.  Using the indices as keys works perfectly fine, as I'm only adding/removing from the end of the array dynamically.  I'll be careful not to add/remove before the dynamic ones.

```js
subscriptions: function(state) {
  return [
    StaticSub1,
    StaticSub2,
    state.cond && OptionalSub1,
    ...computedSubs(state)
  ]
}
```

Will give the new alpha a try today 👍

mrozbarry:
The latest commits might fix it, @jorgebucaran I'll take a look over my lunch break.

SkaterDad:
@jorgebucaran The new commits (alpha-10) are working great.  Thanks for the quick fix :1st_place_medal:

src/index.js
```diff
@@ -1,390 +1,478 @@
...
+        element[name].setProperty(i, style)
+      } else {
+        element[name][i] = style
       }
-
-      if (value == null || value === false) {
-        element.removeAttribute(name)
+    }
+  } else {
+    if (name[0] === "o" && name[1] === "n") {
```
diontools: It looks like we can #674 in alpha.10.

jorgebucaran: @diontools Good catch!

src/index.js
```diff
@@ -1,390 +1,465 @@
...
+}
 
-    return out
+var createElement = function(node, eventCb, isSvg) {
+  var element =
+    node.type === TEXT_NODE
+      ? document.createTextNode(node.name)
+      : (isSvg = isSvg || node.name === "svg")
+      ? document.createElementNS("http://www.w3.org/2000/svg", node.name)
+      : document.createElement(node.name)
```
dennis84: Can we support customized built-in elements here? For example:
```
<p is="word-count">
```
and
```
document.createElement("p", { is: "word-count" })
```

jorgebucaran: @dennis84 Hyperapp sets attributes/properties after creating the element. This code:

```jsx
<p is="word-count">
```

...creates a `p` element with an `is` attribute exactly as declared. Is that what you needed?

Screen Shot 2019-04-15 at 21 44 55

dennis84: Ah, that doesn't work for custom elements. As Far as I know, you have to pass the "is" attribute as options to the createElement function.

If you run this in the conosle nothing is logged:
```javascript
class WordCount extends HTMLParagraphElement {
  connectedCallback() {
    console.log('AAAA')
  }
}

window.customElements.define('word-count', WordCount, {extends: 'p'})

var p = document.createElement('p')//, {is: 'word-count'})
p.setAttribute('is', 'word-count')
p.appendChild(document.createTextNode('test'))
document.body.appendChild(p)
```

jorgebucaran: @dennis84 Does React/Inferno/Preact/etc. handle this?

dennis84: I could only find it in React and Mithril:
- https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOMComponent.js#L429
- https://github.com/MithrilJS/mithril.js/blob/master/mithril.js#L554

jorgebucaran: @dennis84 Awesome, thanks. I'll figure out what's the best way to add this after V2 is out. Thank you.

okwolf:
@jorgebucaran is there going to be a new alpha release with `decorateDispatch`/middleware?

jorgebucaran:
@okwolf Yes, working on it! Hang on. :)

README.md
```diff
@@ -7,63 +7,229 @@
...
+You want to build high-quality, rich, and fast single-page applications. You heard about declarative UIs, unidirectional data-flow, immutable state, effects as data, functional reactive programming and thought you might try them out, but don't know how or where to get started. Hyperapp combines all these concepts into a single apparatus, clean architecture and tiny foundation. Once you learn how to use it, you'll never go back to DOM traversal and manipulation.
 
-## Getting Started
+- **Minimal**—We have aggressively minimized the concepts you need to learn to be productive while remaining on par with what other frameworks can do. Dynamic, purely functional frontend applications without the headache.
+- **Declarative**–Design your UI as a function of the state, and Hyperapp will efficiently update the DOM as your data changes. Declarative views, managed effects, and subscriptions make your code predictable and easy to debug.
+- **Standalone**—Do more with less. Hyperapp includes state management and a [state-of-the-art] Virtual DOM engine that supports keyed updates, functional components & tree memoization—all without dependencies.
```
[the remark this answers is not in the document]
jorgebucaran: Hyperapp itself is a dependency. Is that what you mean? :)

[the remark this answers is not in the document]
okwolf: Alternatively you could say all without additional dependencies. But yes.

src/index.js
```diff
@@ -54,411 +57,418 @@ var isSameAction = function(a, b) {
...
-var patchSub = function(sub, newSub, dispatch) {
-  for (var i = 0, a, b, out = []; i < sub.length || i < newSub.length; i++) {
-    a = sub[i]
-    out.push(
-      (b = newSub[i])
-        ? !a || b[0] !== a[0] || shouldRestart(b[1], a[1])
-          ? [b[0], b[1], b[0](b[1], dispatch), a && a[2]()]
-          : a
-        : a && a[2]()
+var patchSubs = function(oldSubs, newSubs, dispatch) {
```
okwolf: Is this code golfed enough? I can almost read it 😆

jorgebucaran: @okwolf I just pushed to this branch like a few seconds ago. Are you sure you're looking at the right code? :)

README.md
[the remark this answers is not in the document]
jorgebucaran: I went with "all without extra dependencies". 😅

README.md
```diff
@@ -1,4 +1,4 @@
-# <img height=24 src=https://cdn.rawgit.com/jorgebucaran/f53d2c00bafcf36e84ffd862f0dc2950/raw/882f20c970ff7d61aa04d44b92fc3530fa758bc0/Hyperapp.svg> Hyperapp
+# Hyperapp
 
 [![Travis CI](https://img.shields.io/travis/hyperapp/hyperapp/master.svg)](https://travis-ci.org/hyperapp/hyperapp)
```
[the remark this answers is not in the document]
jorgebucaran: Thanks! Will fix.

jorgebucaran:
We're β now! Still tons of documentation todos, but I think the API is very likely done at this point.

```
npm i hyperapp@beta
```

## Changes

- ~`container`~ → `node`
  - Mount by replacing the supplied DOM node, not appending #589 

## New features

- Payload creators #830 
- Declarative event options (`preventDefault`, `stopPropagation`, `passive`)
  - `<form onsubmit={{ action: SubmitForm, preventDefault: true }}> ...</form>`

## Size
~1.9 kB

src/index.js
okwolf: I was referring to the new version of `patchSubs` being more readable. Hopefully not at the cost of too many bytes 🙃

jorgebucaran: @okwolf Which one is more readable?

okwolf: Most folks seem to think the nested ternaries are less readable. I don't think it's by much though 🤷‍♂

PR #739, opened 2018-08-01
Improve TS  component example/test

SteveALee:
I felt the test file `Component.tsx` demonstrating using components in TS left something to be desired. It was incomplete and I think inaccurate and confused me initially. Also it was not runnable which I did do to properly test it (but this PR does not included changes required to make it run as currently the test only compile)

See my inline comments for more details

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -17,10 +17,19 @@ export interface VNode<Attributes = {}> {
  *
  * @memberOf [VDOM]
  */
-export interface Component<Attributes = {}, State = {}, Actions = {}> {
-  (attributes: Attributes, children: Array<VNode | string>):
-    | VNode<Attributes>
-    | View<State, Actions>
+export interface Component<Attributes = {}> {
+  (attributes: Attributes, children: Array<VNode | string>): VNode<Attributes>
+}
+
+/** A LazyComponent is a Component that returns a view function.
+ *
+ * @memberOf [VDOM]
+ */
+export interface LazyComponent<Attributes = {}, State = {}, Actions = {}> {
+  (attributes: Attributes, children: Array<VNode | string>): View<
+    State,
+    Actions
+  >
 }
 
 /**
@@ -37,7 +46,7 @@ export type Children = VNode | string | number | null
  * @memberOf [VDOM]
  */
 export function h<Attributes>(
-  nodeName: Component<Attributes, any, any> | string,
+  nodeName: Component<Attributes> | string,
   attributes?: Attributes,
   ...children: Array<Children | Children[]>
 ): VNode<Attributes>
@@ -74,9 +83,7 @@ export type ActionsType<State, Actions> = {
  * @returns A VNode tree.
  * @memberOf [App]
  */
-export interface View<State, Actions> {
-  (state: State, actions: Actions): VNode<object>
-}
+type View<State, Actions> = (state: State, actions: Actions) => VNode<object>
 
 /** The app() call creates and renders a new application.
  *
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -18,19 +18,21 @@ export interface VNode<Attributes = {}> {
  * @memberOf [VDOM]
  */
 export interface Component<Attributes = {}> {
-  (attributes: Attributes, children: Array<VNode | string>): VNode<Attributes>
+  (attributes: Attributes, children: Array<VNode | string>): VNode<Attributes>  
 }
 
 /** A LazyComponent is a Component that returns a view function.
  *
  * @memberOf [VDOM]
  */
+
 export interface LazyComponent<Attributes = {}, State = {}, Actions = {}> {
   (attributes: Attributes, children: Array<VNode | string>): View<
     State,
     Actions
   >
 }
+
 
 /**
  * Possibles children types
@@ -46,7 +48,7 @@ export type Children = VNode | string | number | null
  * @memberOf [VDOM]
  */
 export function h<Attributes>(
-  nodeName: Component<Attributes> | string,
+  nodeName: Component<Attributes> | LazyComponent<Attributes, any, any> | string,
   attributes?: Attributes,
   ...children: Array<Children | Children[]>
 ): VNode<Attributes>
@@ -83,7 +85,9 @@ export type ActionsType<State, Actions> = {
  * @returns A VNode tree.
  * @memberOf [App]
  */
-type View<State, Actions> = (state: State, actions: Actions) => VNode<object>
+export interface View<State, Actions> {
+  (state: State, actions: Actions): VNode<object>
+}
 
 /** The app() call creates and renders a new application.
  *
--- a/test/ts/Component.tsx
+++ b/test/ts/Component.tsx
@@ -1,29 +1,70 @@
-import { h, Component } from "hyperapp"
+import {
+  h,
+  app,
+  LazyComponent,
+  Component,
+  View,
+  ActionType,
+  ActionsType
+} from "hyperapp"
 
 interface State {
   count: number
 }
 
+const state: State = { count: 10 }
+
 interface Actions {
-  down(): State
-  up(value: number): State
+  down: ActionType<State, Actions>
+  up: ActionType<State, Actions>
 }
 
-const Counter: Component<{ state: State; actions: Actions }> = ({
-  state,
-  actions
-}) => (
+const actions: ActionsType<State, Actions> = {
+  down: () => state => ({ count: state.count - 1 }),
+  up: (value: number) => state => ({
+    count: state.count + value
+  })
+}
+
+interface ComponentParams {
+  count: number
+  up: () => void
+  down: () => void
+}
+const Counter: Component<ComponentParams> = ({ count, up, down }) => (
   <div>
-    <div>{state.count}</div>
-    <button onclick={actions.down}>-</button>
-    <button onclick={actions.up}>+</button>
+    <div>{count}</div>
+    <button onclick={down}>-</button>
+    <button onclick={up}>+</button>
   </div>
 )
 
-const LazyCounter: Component<{}, State, Actions> = () => (state, actions) => (
+interface LazyComponentParams {
+  label: string
+}
+const LazyCounter: LazyComponent<LazyComponentParams, State, Actions> = ({
+  label
+}) => (state, actions) => (
   <div>
-    <div>{state.count}</div>
+    <div>
+      {label}: {state.count}
+    </div>
     <button onclick={actions.down}>-</button>
-    <button onclick={actions.up}>+</button>
+    <button
+      onclick={() => {
+        actions.up(2)
+      }}
+    >
+      +
+    </button>
   </div>
 )
+
+const view: View<State, Actions> = (state, actions) => (
+  <main>
+    <Counter count={state.count} up={() => actions.up(2)} down={actions.down} />
+    <LazyCounter label={"Lazy"} />
+  </main>
+)
+
+app<State, Actions>(state, actions, view, document.body)


hyperapp.d.ts
```diff
@@ -17,10 +17,19 @@ export interface VNode<Attributes = {}> {
  *
  * @memberOf [VDOM]
  */
-export interface Component<Attributes = {}, State = {}, Actions = {}> {
-  (attributes: Attributes, children: Array<VNode | string>):
-    | VNode<Attributes>
-    | View<State, Actions>
+export interface Component<Attributes = {}> {
```
SteveALee: So here I split out lazy component from component. I'm not sure of maintainers view but normal components should only be passed attributes (aka properties) as describe in the readme.

To my mind Lazy Component should really be called Context Components as they let you pass the global context down directly (like React context). This appears to be the main attribute. Lazy to me means leaving till last possible minute and has always confused me

hyperapp.d.ts
```diff
@@ -37,7 +46,7 @@ export type Children = VNode | string | number | null
  * @memberOf [VDOM]
  */
 export function h<Attributes>(
-  nodeName: Component<Attributes, any, any> | string,
+  nodeName: Component<Attributes> | string,
```
SteveALee: I'm pretty sure this is always a component never a lazy one.

hyperapp.d.ts
```diff
@@ -74,9 +83,7 @@ export type ActionsType<State, Actions> = {
  * @returns A VNode tree.
  * @memberOf [App]
  */
-export interface View<State, Actions> {
-  (state: State, actions: Actions): VNode<object>
-}
+type View<State, Actions> = (state: State, actions: Actions) => VNode<object>
```
SteveALee: I find `type` much better as it's a function not an object. No need to export but I should have done for consistency :(

hyperapp.d.ts
```diff
@@ -17,10 +17,19 @@ export interface VNode<Attributes = {}> {
  *
  * @memberOf [VDOM]
  */
-export interface Component<Attributes = {}, State = {}, Actions = {}> {
-  (attributes: Attributes, children: Array<VNode | string>):
-    | VNode<Attributes>
```
SteveALee: Initially I found the initial  '|' confusing. I did not know you could do this and looked for a preceding token. It does give better layout though so I'll leave it as a stylistic issue and not say anything else :)

jorgebucaran:
cc @Alber70g @tetradice (pon) & @Swizz do you want to helo out @SteveALee with this? 😄

hyperapp.d.ts
```diff
@@ -17,10 +17,19 @@ export interface VNode<Attributes = {}> {
...
-    | View<State, Actions>
+export interface Component<Attributes = {}> {
+  (attributes: Attributes, children: Array<VNode | string>): VNode<Attributes>
+}
+
+/** A LazyComponent is a Component that returns a view function.
+ *
+ * @memberOf [VDOM]
+ */
+export interface LazyComponent<Attributes = {}, State = {}, Actions = {}> {
```
Alber70g: This shouldn't be a framework typing. It's not describing the API and is only for convenience. 

Also the naming is confusing. I think `ComponentCreator` would be a better name, since it's a function returning a component

hyperapp.d.ts
Alber70g: Don't do this. It disallows declaration merging of interfaces/types. 

You can do this (below) with interfaces, but not with types because the names need to be unique. Don't use _types_ for typings that are exposed to the world. Generally, always use interfaces, they're a bit more flexible, unless you want it to be unique and 'protected' in some way

```typescript
interface Box {
  size: number;
}
interface Box {
  height: number;
  width: number;
}
```

hyperapp.d.ts
Alber70g: This has to do with the max-width of 80 columns i guess...

test/ts/Component.tsx
```diff
@@ -1,29 +1,70 @@
...
+    <button onclick={down}>-</button>
+    <button onclick={up}>+</button>
   </div>
 )
 
-const LazyCounter: Component<{}, State, Actions> = () => (state, actions) => (
+interface LazyComponentParams {
+  label: string
+}
+const LazyCounter: LazyComponent<LazyComponentParams, State, Actions> = ({
```
Alber70g: As I mentioned, `LazyComponent` (or `ComponentCreator`) should be a userland typing. Or better, define it in your own project.

test/ts/Component.tsx
```diff
@@ -1,29 +1,70 @@
...
+  })
 }
 
-const Counter: Component<{ state: State; actions: Actions }> = ({
-  state,
-  actions
-}) => (
+interface ComponentParams {
+  count: number
+  up: () => void
```
Alber70g: Actually, an `action` returns the new `State`. 

Why is this not typed as `Action`?

test/ts/Component.tsx
```diff
@@ -1,29 +1,70 @@
...
+  up: (value: number) => state => ({
+    count: state.count + value
+  })
 }
 
-const Counter: Component<{ state: State; actions: Actions }> = ({
-  state,
-  actions
-}) => (
+interface ComponentParams {
```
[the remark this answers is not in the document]
Alber70g: It's mostly preference. To test this, what this file made for, there's no need to define a separate Interface.

hyperapp.d.ts
SteveALee: it's more the leading | than the wrapping ie
```
| VNode<Attributes> | View<State, Actions>
```
rather than 
```
VNode<Attributes> | View<State, Actions>

```

hyperapp.d.ts
SteveALee: Hmm, the read me explicitly calls out Lazy Components as a distinct concept. Ie are passed state and actions not just attributes like non lazy components). Though it doesn't explicityly say that,

I just split it out the type from component typing as State and Actions are note part of a non lazy component so I thought is clearer.

It returns a View not a Component. And yes that is a bit confusing as a component returning a view seems weird to me given a view is something the App takes and manages. Still that's the way it is. :)

hyperapp.d.ts
SteveALee: Interesting. That's new to me. I read the typescript manual to say that interfaces are for objects not functions (they have separate sections) . Given this, using an interface to declare a function signature seems plain weird. It's not a interface that contains a function - it IS a function. Or am I missing somehting ?

https://www.typescriptlang.org/docs/handbook/interfaces.html
https://www.typescriptlang.org/docs/handbook/functions.html

To my mind functions do things to the values that Interfaces define. These are not the same interfaces as C++ or C#. But Interfaces can have function members of course. But that's not the case with View - it's a function nothing more.

Your example is clearly an object not a function

test/ts/Component.tsx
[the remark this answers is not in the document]
SteveALee: Agree but test files are VERY  likely to be used as examples  so should prove clarity to using the library.

test/ts/Component.tsx
SteveALee: Simply as its a property function passed to a component (ie callback if you like). it takes no parameters and returns nothing. The parent conmpenent has the logic in . This is a common pattern in React etc. You have a parent "component" with logic and state and dump child components. 

Perhaps we should not be introducing patterns here.

Also it might not be a hyperapp pattern you could pass a global action directly i guess , I just baulked at passing all Actions as you can;t tel which are used by the component form it's signature

test/ts/Component.tsx
SteveALee: Explained above - it seems to be a framework concept to me

SteveALee:
@Alber70g Thanks for the feedback. I might just be POV differences :) I tried to explain.

hyperapp.d.ts
SteveALee: Hmm, the read me explicitly calls out Lazt Components as a distinct concept (ie are passed state and actions not just attributes like non lazy components). I just split out the type from component.

It returns a View not a Component.

hyperapp.d.ts
Alber70g: It's the language. So get used to Typescript

hyperapp.d.ts
Alber70g: It's still limiting the end user to modify and extend the typing by using the concept of declaration merging. Generally, when typing a library, always use interfaces, since they're more flexible for the end user.

> using an interface to declare a function signature **seems plain weird**

I wonder what the creators of Typescript would feel about this. There's clearly a reason to be able to define functions as interfaces. To you it might feel *plain weird*, because you don't fully understand the reasoning behind it.

hyperapp.d.ts
Alber70g: I see it's a concept. I wasn't aware it's in the README. Still, it's not a description of a part of the API. Therefore I feel it shouldn't be in here.

/cc @jorgebucaran

test/ts/Component.tsx
Alber70g: > it takes no parameters and returns nothing
Where you use the component it clearly passes a WiredAction. Especially the 'down' action, but also the 'up' action returns the new state. 

```
<Counter count={state.count} up={() => actions.up(2)} down={actions.down} />
```

> Perhaps we should not be introducing patterns here.
What pattern do you mean? I'm simply referring to the typings, they're incorrect this way.

hyperapp.d.ts
SteveALee: > To you it might feel plain weird, because you don't fully understand the reasoning behind it.

That's often the case, for sure. However I can't find that reasoning in the typescript docs or ever seen other examples.

hyperapp.d.ts
SteveALee: like I said - it's stylistic so not part of the PR

test/ts/Component.tsx
SteveALee: That is an error. Good spot. I intended 

```
<Counter count={state.count} up={() => {actions.up(2)}} down={actions.down} />
```

I use {....} to indicate no return - ie invoked for side effects only

This [pattern](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0)

SteveALee:
I'm going to close this given the discussion ended up with nothing of much value to add. Just difference of opinion.

test/ts/Component.tsx
SteveALee: > Where you use the component it clearly passes a WiredAction. Especially the 'down' action, but also the 'up' action returns the new state.

The callee has no idea it is an action - It just gets a function. That's intentional decoupling. passing directly is a short cut for

```
down={()=>{actions.down()}}
```

The call does not pass any args and ignores the return so type script is OK with it. You might want to be explicit about it an use long for

jorgebucaran:
@SteveALee I don't have a strong opinion on any of this since (1) I don't use TS (much) and (2) I am trying to focus on V2 at the moment.

tetradice:
I tried to comment and it was already closed...:sob:
I agree with some content, so it's a pity that the pull was closed.

SteveALee:
@tetradice thank you. I 're opened as would like your thoughts.

hyperapp.d.ts
tetradice: I agree that TypeScript recommends using interface, but I still think that type should be used in this case. Because hyperapp's view is expected to behave as a simple function.

The reason interface is preferable to type is that users can add properties and methods by using it. But is hyperapp's view like something the user wants to add properties and methods? I do not think so.

hyperapp.d.ts
tetradice: Why does Component remain interface? In hyperapp, Component is just a function that returns VDOM Node as well as View, so I think that Component should be of type as well as View.

SteveALee: No argument from me - I just split it out and duplicated :)

tetradice:
@SteveALee Sorry, it was late to notice that it was open. Thank you.

hyperapp.d.ts
SteveALee: I'm obviously going to have to get to understand this better. to my mind the previous declaration with interface was a object with a function member. I'm obviously got that wrong. But agree in this case would never catenate extra items - it's a stand alone function not an interface

hyperapp.d.ts
tetradice: I think this code is incorrect. Because, according to the code transpiled from Component.tsx to js by TypeScript, the h function receives LazyComponent.

```javascript
const view = (state, actions) => (h("main", null,
    h(Counter, { count: state.count, up: () => actions.up(2), down: actions.down }),
    h(LazyCounter, { label: "Lazy" })));
```

Considering this fact, I think that the definition here is correct.

```ts
export function h<Attributes>(
  nodeName: Component<Attributes> | LazyComponent<Attributes, any, any> | string,
  attributes?: Attributes,
  ...children: Array<Children | Children[]>
): VNode<Attributes>
```

By the way, the reason why the definition before modification does not become a build error is unknown. I think that probably because the type check of tsx syntax is too loose... :confused:

hyperapp.d.ts
tetradice: OK :)

hyperapp.d.ts
SteveALee: Interesting as i built and tested and it seemed to run correctly. Imay have used different options e.g. target and module but that should not impact.

tetradice: Yes, the test code (`test/ts/Component.tsx`) that exists in the repository certainly will be built without errors in my environment. But this is built without error because it is not because the definition is correct, but unfortunately TypeScript performs only looser type checking against jsx syntax than normal code.

By referring to the result of transpiling by TypeScript, if you create a ts file like this and save it as `test/ts/Component.ts`, you will see that it is an error to call the `h` function .
https://gist.github.com/tetradice/95cb50eb787eb63d37872ab2028883a6

(Actually, even the original hyperapp.d.ts will generate an error, so I'm trying to create a new pull request if SteveALee's pull request is accepted)

SteveALee: That's definitely not right, but it did seem to run correctly for me. Hidden bug! Thanks for the gist.

I'll tidy up the PR now and see.

SteveALee: > I think that probably because the type check of tsx syntax is too loose

That's interesting and if true pushes one to use hyperscript / functioncalls for good type safety

hyperapp.d.ts
SteveALee: Now I understand the different between a function interface and an object with a function member! It's a pure declaration without a name!

However, while it is not going to get extended,  there is one advantage of interface over type alias - the "View" name will appear in error messages not just the type details.

hyperapp.d.ts
SteveALee: @tetradice I confirm you suggested fix works in your gist.
BUT I got extra null assignment errors as well on other lines. I left details on your gist.

SteveALee:
I updated the typings.

Am wondering if we should use h() as @tetradice gist rather than JSX in typescript tests as JSX seems to stop much of the type checking we use typescript for? Indeed I'm nor concerned JSZ + Typescript are not a good match.

@tetradice - can you comment on those other errors I mention in the gist? I think may point at needed changes in either the example or typings.

tetradice:
@SteveALee Thanks. I commented my idea and amendment on gist.
https://gist.github.com/tetradice/95cb50eb787eb63d37872ab2028883a6

> Am wondering if we should use h() as @tetradice gist rather than JSX in typescript tests as JSX seems to stop much of the type checking we use typescript for? Indeed I'm nor concerned JSZ + Typescript are not a good match.

As a matter of fact, I do not think there is so much harm to users using JSX. It's just a bit loose TypeScript. :grin: So I think that there is no problem with using TypeScript + JSX as before.

However, there is harm to users who do not use JSX, so it may be better to add a test code without JSX.

aminnairi:
Funny, I was having an issue when using the `Component` interface with my lazy and non-lazy components. Decided to make an issue and then a PR. Found that you did exactly what I had to do in my own project to make it work under TypeScript with Rollup.

Eager to see this PR getting approved and merged because this is, IMO, the good way of handling both cases.

jorgebucaran:
The current focus is on shipping V2, so I don't think this is going to get merged soon. Thank you, @SteveALee and everyone else who participated in this discussion for sharing your thoughts. 👋

PR #755, opened 2018-09-03
Fix descriptions about transform-react-jsx

wtetsu:
If one follows the document, "transform-react-jsx" causes a error .

```
Error: Cannot find module 'babel-plugin-transform-react-jsx' from '...'
```

* https://babeljs.io/docs/plugins/transform-react-jsx
* https://babeljs.io/docs/en/babel-plugin-transform-react-jsx

At present, the former redirects to the latter. And the latter is about "@babel/plugin-transform-react-jsx" and shows .babelrc like this.

```js
{
  "plugins": ["@babel/plugin-transform-react-jsx"]
}
```


jorgebucaran:
Thanks, @wtetsu! LGTM

PR #758, opened 2018-09-13
Custom Elements

### LINKED ISSUE #757: Provide a way to keep existing markup as is (Custom Elements, Hydration)
I'm looking for a way to make hyperapp respect existing markup thats inside the app and leave it untouched. My problem is quite similar to the issues @frenzzy and @gyopiazza already have pointed out (#741, #688).

Parts of my application are made up of sub-applications provided through Custom Elements. Due to browser support be can not rely on ShadowDOM, so these Custom Elements use the Light DOM to render their UI. Hyperapp is overwriting their content in every render cycle. I've tried the [hydration workaround](https://github.com/hyperapp/render/issues/8#issuecomment-385974406). But this is not really a solution. You are able to restore the markup, but the innerHTML gets completely rewritten so you loose all events.

I've compiled a little JSBin the shows one concrete scenario I'm having problems with:
https://jsbin.com/dogujuq/2/edit?html,js,output

Using react I was able to make this work using `shouldComponentUpdate`. But I currently don't see any solution how to make this happen in hyperapp.

frenzzy:
Custom elements usually control their children nodes themselves and hyperapp should not touch them during hydration and lifecycle.

Fixes #757

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -321,6 +321,8 @@ export function app(state, actions, view, container) {
         (isSvg = isSvg || node.nodeName === "svg")
       )
 
+      if (node.nodeName.indexOf("-") !== -1) return element
+
       var oldKeyed = {}
       var newKeyed = {}
       var oldElements = []
--- a/test/recycling.test.js
+++ b/test/recycling.test.js
@@ -50,3 +50,25 @@ test("recycle markup against keyed vdom", done => {
     document.getElementById("app")
   )
 })
+
+test("recycle custom elements", done => {
+  const SSR_HTML = `<div id="app"><main><custom-element>foo</custom-element></main></div>`
+
+  document.body.innerHTML = SSR_HTML
+
+  app(
+    null,
+    null,
+    state => (
+      <main>
+        <custom-element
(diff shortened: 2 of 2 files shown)


jorgebucaran:
Thank you, @frenzzy. It looks hacky, won't you agree? How infallible do you think this is? 😄

frenzzy:
Oh, [exceptions...](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name)

> _name_ must not be any of the following:
> - annotation-xml
> - color-profile
> - font-face
> - font-face-src
> - font-face-uri
> - font-face-format
> - font-face-name
> - missing-glyph

zaceno:
I don't think we want this. Sure custom-elements *control* their children, but this takes away our ability to *compose* with custom elements. Like what if I had a custom element that displays whatever in a modal:

```jsx
...
<custom-modal titleBarText="Blah">
  <Foo state={state.foo} />
  <p>{state.bar}</p>
</custom-modal>
...
```

frenzzy:
Yes, you are right! Closing here.

Issue #179, opened 2017-04-07
input readonly list-attribute 


marcusasplund:
Fails in rendering an input with attribute 'list' user for targeting datalist-element
Removing the attribute renders both input and datalist elements, then unconnected and useless of course.

Console message:
Cannot assign to read only property 'list' of object '#<HTMLInputElement>'
    at setElementData...

Reproduced in glitch
https://glitch.com/edit/#!/wry-autocomplete

````js
import { h, app } from 'hyperapp' // 0.7.1
// import {h, app} from '../hyperapp/index' // 0.8.0, unreleased


const dataList = () =>
  <datalist id='browsers'>
    <option value='Chrome'></option>
    <option value='Firefox'></option>
    <option value='Internet Explorer'></option>
    <option value='Opera'></option>
    <option value='Safari'></option>
    <option value='Microsoft Edge'></option>
  </datalist>

const view = () =>
  <div class='container'>
    <label>Choose a browser from this list:
    <input list='browsers' /></label>
    {dataList()}
  </div>            

app({view})

jbucaran:
@marcusasplund Hi 👋!

I just woke up so I haven't looked at this in good detail, but smells like another use case we fixed with _keys_ <https://github.com/hyperapp/hyperapp/pull/172>?

@zaceno Do you have any comments?

zaceno:
@jbucaran - Looked at the glitch-example @marcusasplund provided. As far as I can tell, it's got not nothing to do with keys.

Not sure exactly what the problem is. The relevant lines of code are (in `setElementData`)

~~~~
     element.setAttribute(name, value)
     if (element.namespaceURI !== "http://www.w3.org/2000/svg") {
         // SVG element's props are read only in strict mode.

        element[name] = value
     }
~~~~

the line that throws the exception is `element[name] = value`. Not sure why. Possibly because the `list` attribute is newish HTML5. Maybe there's some special way browsers treat it differently to other attributes.

The thing I don't get, though, is: We already do `element.setAttribute(name, value)` higher up. What does `element[name] = value` give us?

Even though using keys won't fix the issue, the `setElementData` function has been changed on master since 0.7.1, so it'd be interesting to see how this behaves on master.

marcusasplund:
@zaceno Yup. I tried to switch from 0.7.1 to src from master in the  glitch provided. Same behaviour. But commenting out line 184 in src/hyperapp/app.js (in the glitch) all elements renders and works as expected.
`//  element[name] = value
`
[...]

PR #762, opened 2018-09-24
add form attribute to readonly attributes list.

kiyonori-matsumoto:
form attribute for input tag is readonly, so following errors occurred and nothing is rendered.

```
Uncaught TypeError: Cannot assign to read only property 'form'
 of object '#<HTMLInputElement>'
```

I also created codepen samples that fails for rendering.
https://codepen.io/middleearth/pen/EeJoyK

I think this issue is same as #179 , but i wondered this fix is correct.

please tell me if something is wrong.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -194,6 +194,7 @@ export function app(state, actions, view, container) {
       } else if (
         name in element &&
         name !== "list" &&
+        name !== "form" &&
         name !== "type" &&
         name !== "draggable" &&
         name !== "spellcheck" &&
--- a/test/dom.test.js
+++ b/test/dom.test.js
@@ -648,6 +648,13 @@ testVdomToHtml("input list attribute", [
   }
 ])
 
+testVdomToHtml("input form attribute", [
+  {
+    vdom: <input form="foobar" />,
+    html: `<input form="foobar">`
+  }
+])
+
 testVdomToHtml("events", [
   {
     vdom: (


jorgebucaran:
👋 Hi! @kiyonori-matsumoto Thank you, for your PR. Would it be possible to have a more complete example? An example using at least one action and submitting the form would help.

kiyonori-matsumoto:
Hi. @jorgebucaran thank you for your reviewing.

I created another sample that uses action to send data (to example.com).

https://codepen.io/middleearth/pen/zmYNRz

jorgebucaran:
@kiyonori-matsumoto @frenzzy Could we discuss the merits of adding this feature? I'd like to know why anyone would ever need to write the \<input\> elements **outside** the form element.

selfup:
Not everything needs to be wrapped in `<form>`

To be honest I haven't wrapped any inputs in a <form> for like 3 years now and haven't run into any issues 😂 

However I don't see why an input needs to be treated as a form? 🤔

lukejacksonn:
@selfup forms also handle all the submission logic for you. So `onsubmit` will handle interactions like hitting enter in an input and `form.elements` will give you all the inputs in a given form which is handy when it comes to serialising inputted data. I'd highly recommend using `form` elements!

selfup:
Oh interesting, I'll check that behavior out. Thanks for explaining 🎉

kiyonori-matsumoto:
@lukejacksonn 
thank you for your comments.

@selfup @jorgebucaran 
It is now not rendered if we use `form` tag inside each table row from new chrome version(may be 69), and also firefox.

so if I use hyperapp, form and table rows, I must do like 2nd line of example shown below.

```html
<table><form><tr><td><input value="not shown" /></td></tr></form></table>
<table><form id="baz" /><tr><td><input value="shown" form="baz" /></td></tr></table>
```

I also updated my sample codepen to show the difference of the example.

https://codepen.io/middleearth/pen/zmYNRz

jorgebucaran:
Does anyone know what React/Vue/Preact do in this case? 🤔

frenzzy:
> ```html
> <table><form><tr><td><input value="not shown" /></td></tr></form></table>
> <table><form id="baz" /><tr><td><input value="shown" form="baz" /></td></tr></table>
> ```

@kiyonori-matsumoto this is [not valid html](https://validator.w3.org/), so of course it may not work properly:

image

If you will use valid html, maybe you don't need to use `form` attribute...

```html
<form><table><tr><td><input value="not shown" /></td></tr></table></form>
<table><tr><td><form><input value="not shown" /></form></td></tr></table>
```

frenzzy:
> Does anyone know what React/Vue/Preact do in this case? 🤔

No runtime errors in React

React demo: https://codepen.io/frenzzy/pen/rqOQqB/left/?editors=0010
Hyperapp demo: https://codepen.io/frenzzy/pen/OByrba/left/?editors=0010

kiyonori-matsumoto:
@frenzzy 
thank you for your comment.

I know that my html is invalid, but I want to apply form to each table row, like following.

```html
<table>
  <form><tr><td><input value="not shown" /></td><td><input /></tr></form>
  <form><tr><td><input value="not shown" /></td><td><input /></tr></form>
</table>
```

this html is also invalid, so I really wanted to use form attributes.

frenzzy:
@kiyonori-matsumoto what about using non-table html markup?
```html
<div class="table">
  <form class="tr"><div class="td"><input /></div></form>
  <form class="tr"><div class="td"><input /></div></form>
</div>
```
```css
.table { display: table }
.tr { display: table-row }
.td { display: table-cell }
```
Demo https://codepen.io/frenzzy/pen/JmXmQm/left/?editors=1000

frenzzy:
Personally I think this PR is good and I would accept it. Looks good to me.

jorgebucaran:
Does this problem still happen?

jorgebucaran:
@kiyonori-matsumoto Ping. 

> Does this problem still happen?

🤔

PR #772, opened 2018-10-25
Allow lazy component to return null in TypeScript

hasparus:
Types #636.


jorgebucaran:
Thanks, @hasparus! 👋

infinnie:
They should also be able to return arrays.

jorgebucaran:
@infinnie Why?

infinnie:
@jorgebucaran Normal components can return arrays. Lazy components should behave similarly.

jorgebucaran:
@infinnie But what does that have to with this PR?

PR #776, opened 2018-11-02
Set text node type to Node.TEXT_NODE

frenzzy:
Fixes https://github.com/jorgebucaran/hyperapp/pull/726#discussion_r201364435

Would be nice to standardize VNode types as early as possible to be able to build third-party extensions which can rely on this type without the need to change them in the future.


jorgebucaran:
@frenzzy Looks good. 👍

PR #777, opened 2018-11-09
Add Lazy Views (V2)

SkaterDad:
This PR adds lazy views to V2.  See Issue #721 for a detailed explanation of how this works.

#### Brief Summary
If you are familiar with [elm's Html.lazy](https://guide.elm-lang.org/optimization/lazy.html), or the new React `memo` function, this is a similar concept.

Any function you wrap inside of `Lazy` will only re-render if the props passed in have changed.  It works by creating a special object which the VDOM patching algorithm uses to compare the "old" node props with the "new" node props.  If the props are the same (`===`), the old VNode is used without having to compute it again.  When the VDOM patching notices that `oldNode === newNode`, it skips traversing that part of the DOM tree, potentially saving a lot of work!

#### Use cases

This is very useful for performance optimization, in cases where the parts of your view rely on separate parts of the state tree.

#### Example
```js
import { h, Lazy, app } from 'hyperapp'

const ExpensiveComponent = ({ bigArray, other }) => h('div', {}, /* some complicated stuff */)

app({
  init: {
    count: 1,
    bigArray: [ /* lots of stuff */],
    other: {  test: true }
  },
  view: state =>
    h('main', {}, [
      Counter({ count: state.count }),

      // Uh-oh! This will re-render on every state change, even if it was just the counter
      ExpensiveComponent({ bigArray: state.bigArray, other: state.other  })

      // But wait!  We can make it lazy!
      // This will only re-renders if the "bigArray" or "other" properties change.
      Lazy({
        render: ExpensiveComponent,
        key: "my-lazy-comp",
        bigArray: state.bigArray,
        other: state.other
      })
    ])
})
```
JSX should also work:
```jsx
<Lazy
  render={ExpensiveComponent}
  key="my-lazy-comp"
  bigArray={state.bigArray}
  other={state.other}
/>
```

#### Required Props

**render** - Must be a function that returns some Virtual DOM nodes.  _This must not be an anonymous function._
**key** - For maximum benefit, assign a key.  This ensures the VDOM patching will pair up the "old" and "new" versions of the component.

#### Quirks?
For the sake of code simplicity, whatever props you pass to `Lazy` will get passed into the underlying view function.  That includes the `render` prop.


SkaterDad:
Just pushed another commit fixing bugs which occurred during keyed list diff/patch.

I discovered the bugs while setting up JS Frameworks Benchmark, which I ran on my Windows 10 desktop.  I pushed up the code to Github if anyone is interested.
https://github.com/SkaterDad/js-framework-benchmark

You'll see two benchmarks `v2-lazy` and `v2-lazy-opt`.  One of them used a [different `isSameValue` function](https://github.com/moroshko/shallow-equal/blob/master/src/objects.js) which [benchmarked as 2x faster on its own](https://esbench.com/bench/5c0b26744cd7e6009ef61da2).  Those two implementations must be optimized in Chrome, since they trade places in the results table every time I run it.  I would bet that older browsers like IE11 will show vastly different results.

This is probably why people tell you not to do microbenchmarks 😆 

This screenshot is the results from `npm run selenium -- --count 10 --framework ivi hyperapp`.

The results illustrate pretty well that wrapping something in `Lazy` does incur some overhead for creation, since it's an extra function call or two.  The trade-off _can be_ faster partial updates, depending on your view.

The results also illustrate how insanely fast Hyperapp v2 is going to be out of the box!

image

jorgebucaran:
@SkaterDad I know this is probably not the right place to discuss it, but since adding this feature means going back to `resolveNode`, what do you think about delaying the creation of component nodes until calling the view? (I can create a new issue from this later.)

I mean for: `h(MyComponent, props)` a.k.a `<MyComponent ...props />`.

SkaterDad:
> @SkaterDad I know this is probably not the right place to discuss it, but since adding this feature means going back to `resolveNode`, what do you think about delaying the creation of component nodes until calling the view? (I can create a new issue from this later.)
> 
> I mean for: `h(MyComponent, props)` a.k.a `<MyComponent ...props />`.

I can see that going two ways:

1. Defer execution of the component function until just before DOM needs to be touched.
2. All components essentially become Lazy, and only re-execute when props are different.

I spent some time trying to make `#1` work, and it seems difficult, since all of the DOM stuff assumes you have a fully resolved VNode.

If you have to resolve the component either way on every render, I don't think there will be any perf difference between doing it in the `h` function, or doing it later before touching the DOM. :man_shrugging: 

`#2` could be interesting, but I'm sure there is a trade-off in performance somewhere since you have to do more `isSameValue` comparisons on each render.

jorgebucaran:
@SkaterDad Thanks! So, let me clarify. This is not about performance or trying to make all components lazy, hence the "this discussion may be out of place" notice.

I agree that if we resolve components either way on every render, there will be no performance gains, but it would defer calling component functions, changing the order how the component's VDOM unwraps (resolves into an object).

@mindplay-dk has some interesting ideas about this (not evaluating components eagerly like we do right now). (Rasmus if you want to add your thoughts here, be my guest.)

See: https://github.com/jorgebucaran/superfine/issues/135#issuecomment-402996786

SkaterDad:
> it would defer calling component functions, changing the order how the component's VDOM unwraps (resolves into an object).

Hmmm...  The TabList example he posted makes a very interesting point on execution order.  I hadn't thought of that before, but it makes a lot of sense.

mrozbarry:
Huzzah :tada:

jorgebucaran:
@SkaterDad Is https://github.com/SkaterDad/js-framework-benchmark ready to clone, `npm i` and use to run the benchmarks or is there any additional setup required? :)

SkaterDad:
@jorgebucaran Anything is possible!  Haha. 

Running those benchmarks is such a pain.

You'll need to change the hyperapp v2 implementations to actually use the library from npm.  I had modified versions in each folder.

I also don't remember what the difference was between the two lazy versions.  Glancing over the code, it looks like it could still work with the latest alphas. 

I can probably find some time to work on it, if you're low on time.

PR #784, opened 2018-12-02
Fix anoying error from interface Element not being able to implement …

nmpribeiro:
…VNode and ReactElement simultaneously due to Key.
Basically, I was having the following error:
```
 Interface 'Element' cannot simultaneously extend types 'VNode<any>' and 'ReactElement<any>'.
  Named property 'key' of types 'VNode<any>' and 'ReactElement<any>' are not identical.
```

This pull request fixes it.

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -10,7 +10,7 @@ export interface VNode<Attributes = {}> {
   nodeName: string
   attributes?: Attributes
   children: Array<VNode | string>
-  key: string | number | null
+  key: string | number
 }
 
 /** A Component is a function that returns a custom VNode or View.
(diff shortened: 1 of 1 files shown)


jorgebucaran:
@nmpribeiro Were you using a number as a key?

Swizz:
One day, React will no longer be the facto standard for JSX...

jorgebucaran:
@Swizz Please use Slack, a forum, your twitter or your internal thought process, not this pull request, to discuss your opinions about JSX.

@nmpribeiro Is `null` necessary? Could we get away with just `string | number`? Just wondering.

nmpribeiro:
@jorgebucaran indeed `null` is not necessary. I have updated this pull request accordingly.

jorgebucaran:
Great, thanks! @nmpribeiro 🎉

nmpribeiro:
Sorry @jorgebucaran we do need `null`...

PR #795, opened 2019-01-13
clean code

uxitten:
clean code remove { } in single line


jorgebucaran:
@uxitten Thanks, but no thanks! I like it the way it is. 👋

PR #801, opened 2019-02-12
could make our virtual DOM a little bit "purer" by using Object.creat…

cbbfcd:
…e(null)

Object literals (e.g. { a: 3 }) automatically inherit from Object. This means that the object created by object literals will have methods defined in the Object.prototype like hasOwnProperty, toString, etc.

We could make our virtual DOM a little bit "purer" by using Object.create(null). This will create a truly plain object that doesn't inherit from Object but null instead.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -402,14 +402,15 @@ var patchElement = function(
 }
 
 var createVNode = function(name, props, children, element, key, type) {
-  return {
+  var pureVNode = Object.create(null)
+  return Object.assign(pureVNode, {
     name: name,
     props: props,
     children: children,
     element: element,
     key: key,
     type: type
-  }
+  })
 }
 
 var createTextVNode = function(text, element) {
--- a/src/index.js
+++ b/src/index.js
@@ -20,6 +20,12 @@ var merge = function(a, b) {
   return target
 }
 
+var assign = function(a, b) {
+  for(var i in b) if(b.hasOwnProperty(i)) a[i] = b[i]
+  
+  return a
+}
+
 var resolved = typeof Promise === "function" && Promise.resolve()
 
 var defer = !resolved
@@ -403,7 +409,7 @@ var patchElement = function(
 
 var createVNode = function(name, props, children, element, key, type) {
   var pureVNode = Object.create(null)
-  return Object.assign(pureVNode, {
+  return assign(pureVNode, {
     name: name,
     props: props,
     children: children,


src/index.js
```diff
@@ -402,14 +402,15 @@ var patchElement = function(
 }
 
 var createVNode = function(name, props, children, element, key, type) {
-  return {
+  var pureVNode = Object.create(null)
+  return Object.assign(pureVNode, {
```
jorgebucaran: @cbbfcd Why are you using `Object.assign`?

cbbfcd:
i can add a polyfill，so it's ok?

jorgebucaran:
@cbbfcd Why do you need this?

src/index.js
cbbfcd: I realized this, so I can add a simple polyfill or use another method.

jorgebucaran:
@cbbfcd Let me do some research. I'll keep this in mind as I work towards the V2 release. I'll let you know if this will be accepted or not. 

Thanks! 👋😄

cbbfcd:
> We can't use `Object.assign`.



> @cbbfcd Let me do some research. I'll keep this in mind as I work towards the V2 release. I'll let you know if this will be accepted or not.
> 
> Thanks! 👋😄

@jorgebucaran Thank you very much, I am very honored.

b-fuze:
@cbbfcd did you benchmark this? How is the perform compared to a literal?

whaaaley:
Reading over the MDN spec for `Object.create`, it looks like it returns a new object, so would assign be needed at all? Couldn't you just `return Object.create(null, { ... }`?

I'm also concerned about performance though.

b-fuze:
Yes, but it'd have to look like this
```js
Object.create({}, {
    name: { value: name, },
    props: { value: props, },
    children: { value: children, },
    element: { value: element, },
    key: { value: key, },
    type: { value: type }
})```

**Comment by b-fuze** [comment]
So I threw this together to test four different scenarios... Mixed results across Firefox and Chrome but they pretty much point to the same thing: Perf goes down the drain

https://jsperf.com/object-freeze-vs-object-create-null-vs-object-literal

**Comment by lukejacksonn** [comment]
It is funny how these things often turn out exactly the opposite of what you would expect 😅 

image

**Comment by cbbfcd** [comment]
> @cbbfcd did you benchmark this? How is the performance compared to a literal?

In fact, my concern is not the difference in performance, but the side effects. @b-fuze

**Comment by cbbfcd** [comment]
> Reading over the MDN for `Object.create`, it looks like it returns a new object, so would assign be needed at all? Couldn't you just `return Object.create(null, { ... }`?
> 
> I'm also concerned about performance though and this would affect view code that relies on regular object prototypes.

Yes, you can use Obeject.create(null, {}) directly; just use assign I personally feel a little more elegant, 😂。
There is a slight performance overhead compared to the direct use of object literals, but side effects can be avoided. This is a balanced question, see how to choose.

**Comment by cbbfcd** [comment]
> It is funny how these things often turn out exactly the opposite of what you would expect 😅
> 
> image

Good! friend

**Comment by cbbfcd** [comment]
> Yes, but it'd have to look like this
> 
> ```js
> Object.create({}, {
>     name: { value: name },
>     props: { value: props },
>     children: { value: children },
>     element: { value: element },
>     key: { value: key },
>     type: { value: type }
> })
> ```

Yes, it doesn't look very beautiful.

**Comment by cbbfcd** [comment]
> So I threw this together to test four different scenarios... Mixed results across Firefox and Chrome but they pretty much point to the same thing: Perf goes down the drain
> 
> I included `Object.freeze` because `Object.create(null)` doesn't actually make your objects pure... You can still mutate them later (impure) `Object.freeze` would actually make your objects immutable thus pure.
> 
> https://jsperf.com/object-freeze-vs-object-create-null-vs-object-literal

Making the virtual dom more pure is just an idea. First we need to discuss whether it is necessary to do so and then decide which one is more performance advantage. Thank you for your test case!

**Comment by b-fuze** [comment]
> > Reading over the MDN for `Object.create`, it looks like it returns a new object, so would assign be needed at all? Couldn't you just `return Object.create(null, { ... }`?
> > I'm also concerned about performance though and this would affect view code that relies on regular object prototypes.
> 
> Yes, you can use Obeject.create(null, {}) directly; just use assign I personally feel a little more elegant, 。
> There is a slight performance overhead compared to the direct use of object literals, but side effects can be avoided. This is a balanced question, see how to choose.

The overhead isn't that small, but yes, avoiding side effects is a good idea. However, removing `Object` from the prototype chain doesn't prevent side effects. For one, if you look at the (inherited methods on MDN)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object#Methods], the only methods that cause side effects are getter/setter methods, and those are being deprecated. 

Secondly side effects are merely achieved by assigning a property on the object a new value. The following code is an example of why it's not pure (by just inheriting from `null`):
```js
pureVNode.children = [];
```
That's the simplest way to achieve side effects, and generally the only ways to prevent that is by using `Object.prototype.freeze` or--God forbid--a `Proxy`.

The sad truth is that JS was never really designed with immutability in mind. That's why we have all these functional-to-JS transpilers everywhere.

cbbfcd:
> > > Reading over the MDN for `Object.create`, it looks like it returns a new object, so would assign be needed at all? Couldn't you just `return Object.create(null, { ... }`?
> > > I'm also concerned about performance though and this would affect view code that relies on regular object prototypes.
> > 
> > 
> > Yes, you can use Obeject.create(null, {}) directly; just use assign I personally feel a little more elegant, 。
> > There is a slight performance overhead compared to the direct use of object literals, but side effects can be avoided. This is a balanced question, see how to choose.
> 
> The overhead isn't that small, but yes, avoiding side effects is a good idea. However, removing `Object` from the prototype chain doesn't prevent side effects. For one, if you look at the [inherited methods on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object#Methods), the only methods that cause side effects are getter/setter methods, and those are being deprecated.
> 
> Edit: So that's not entirely true, you can get rid of the `__proto__` property which prevents changing the property chain, thus preventing _some_ weird side effects, but that's a very obscure issue in my opinion. I wouldn't think that anyone uses it.
> 
> Secondly side effects are merely achieved by assigning a property on the object a new value. The following code is an example of why it's not pure (by just inheriting from `null`):
> 
> ```js
> pureVNode.children = [];
> ```
> 
> That's the simplest way to achieve side effects, and generally the only ways to prevent that is by using `Object.prototype.freeze` or with--God forbid--a `Proxy`.
> 
> The sad truth is that JS was never really designed with immutability in mind. That's why we have all these functional-to-JS transpilers everywhere.

@b-fuze Thanks a lot!

the key point of side effects is getter/setter, so freeze (proxy) is a good choice. 

Maybe you can provide a configuration like [immer](https://github.com/mweststrate/immer#auto-freezing).

I also think that I am not thinking about this issue very well. If there is a regret drug, I would rather just guarantee that the vnode object is a plain object， that's enough for now.

jorgebucaran:
The V2 codebase has changed a bit since this PR was created. I'm happy to revisit after V2 is released. I'm going to close for now.

PR #802, opened 2019-02-15
Update effect and subscription signatures

mrozbarry:
## Tasks

 - [x] Implement tuple-based effects
 - [x] Implement tuple-based subscriptions

## Discussion

 - [Part of issue 765](https://github.com/jorgebucaran/hyperapp/issues/765#issuecomment-426542026)

#### Current V2 Effects:

```js
{ effect: someEffectFunction, ...otherParams }
```

#### Changes for V2 Effects:

```js
[someEffectFunction, { ...otherParams }]
```

#### Current V2 Subscriptions:

```js
subscriptions: state => [
  { effect: someSubscriptionEffect, ...otherParams, cancel: injectedCancelFn },
]
```

#### Changes for V2 Subscriptions:

```js
subscriptions: state => [
  [someSubscriptionEffect, { ...otherParams }, injectedCancelFn],
]
```

#### Other notable changes for subscriptions:

 - Previous implementation allowed infinite array nesting, this implementation requires a flat list of subscription tuples.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -482,7 +482,7 @@ export var h = function(name, props) {
 }
 
 var cancel = function(sub) {
-  sub.cancel()
+  sub[1][2]()
 }
 
 var isSameValue = function(a, b) {
@@ -503,8 +503,7 @@ var isSameAction = function(a, b) {
 
 var restart = function(sub, oldSub, dispatch) {
   for (var k in merge(sub, oldSub)) {
-    if (k === "cancel") {
-    } else if (sub[k] === oldSub[k] || isSameAction(sub[k], oldSub[k])) {
+    if (sub[k] === oldSub[k] || isSameAction(sub[k], oldSub[k])) {
     } else {
       cancel(oldSub)
       return start(sub, dispatch)
@@ -514,31 +513,33 @@ var restart = function(sub, oldSub, dispatch) {
 }
 
 var start = function(sub, dispatch) {
-  return merge(sub, {
-    cancel: sub.effect(sub, dispatch)
-  })
+  return [
+    sub[0],
+    sub[1],
+    sub[0](sub[1], dispatch)
+  ]
 }
 
 var refresh = function(sub, oldSub, dispatch) {
-  if (isArray(sub) || isArray(oldSub)) {
-    var out = []
-    var subs = isArray(sub) ? sub : [sub]
-    var oldSubs = isArray(oldSub) ? oldSub : [oldSub]
+  var current = [].concat(sub);
+  var previous = [].concat(oldSub);
+  var out = [];
 
-    for (var i = 0; i < subs.length || i < oldSubs.length; i++) {
-      out.push(refresh(subs[i], oldSubs[i], dispatch))
-    }
-
-    return out
+  for(var i = 0; i < current.length || i < previous.length; i++) {
+    var cSub = current[i];
+    var pSub = previous[i];
+    out.push(
+      cSub
+        ? pSub
+          ? restart(cSub, pSub, dispatch)
+          : start(cSub, dispatch)
+        : pSub
+          ? cancel(pSub)
+          : pSub
+    );
   }
 
-  return sub
-    ? oldSub
-      ? restart(sub, oldSub, dispatch)
-      : start(sub, dispatch)
-    : oldSub
-      ? cancel(oldSub)
-      : oldSub
+  return out;
 }
 
 export function app(props) {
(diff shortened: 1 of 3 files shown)


mrozbarry:
This should warrant discussion from @jorgebucaran and @okwolf since it dramatically changes expectations on how effects and subscriptions should be implemented.

test/subscriptions.test.js
```diff
@@ -0,0 +1,51 @@
+import { app, h } from '../src';
+
+test('runs a subscription', () => {
```
mrozbarry: I don't know how much ceremony this test requires. Further more, this needs to be more thoroughly tested, including a test that explicitly tests cancel behaviour, and another for restart behaviour.

src/index.js
```diff
@@ -570,7 +573,7 @@ export function app(props) {
       if (typeof obj[0] === "function") {
         dispatch(obj[0](state, obj[1], data))
       } else {
-        obj[1].effect(obj[1], dispatch, setState(obj[0]))
+        obj[1][0](obj[1], dispatch, setState(obj[0]))
```
okwolf: Shouldn't this be the following?
```js
obj[1][0](obj[1][1], dispatch, setState(obj[0]))
```

mrozbarry: I can get behind that. I was hesitate to make the change since previously the first param was the whole effect object.

mrozbarry: With that in mind, I'll update subscriptions to behave the same way.

okwolf: I think when the whole effect is an object it makes more sense than having an array that users will always need to be doing `props[1]` with.

okwolf: Unless there's a use case I'm not thinking of for the effect to need a reference to its own function while running for some reason 🤔

mrozbarry: I can't think of a good reason, since I should know which function is being called at this point.

src/index.js
```diff
@@ -514,31 +514,34 @@ var restart = function(sub, oldSub, dispatch) {
 }
 
 var start = function(sub, dispatch) {
-  return merge(sub, {
-    cancel: sub.effect(sub, dispatch)
-  })
+  return [
+    sub[0],
+    merge(sub[1], {
```
selfup: So in V2 we provide all new state to the engine. Is there a reason we still have `merge`? No need to `for` through two objects and replace if we can just swap new state provided by the action right?

Just curious.

jorgebucaran: > @selfup I would be in favor of something like `replace` instead of merge.

We totally are. V2's state change strategy is by replace, not by merge like in V1. This instance of `merge` is for a different thing. It's a way to add the `cancel` property to the subscription, so we know how to cancel it later.

selfup: Yea I see that now. We use merge for checks mostly. I updated my comment 😄

selfup: Yea I guess updating the object directly is impure. Nvm, merge makes sense, I was just hoping we wouldn't have to do all that work just to add cancel 😂

okwolf: Since we're already putting the sub function and props into an array, we could just add the cancel as another element if that's not too confusing to keep track of 🤷‍♂️

mrozbarry: I've got mixed feelings on that. We wanted a tuple data structure, which a third item violates, **but** it's not really a argument we should pass in and a subscription shouldn't be able to cancel itself.

SkaterDad: Tuples can have three members in a few languages, including Elm.

I'd prefer that to having to merge a cancel prop to my subscription props.

mrozbarry: I can certainly make the change of we're all behind it

jorgebucaran: @mrozbarry Sounds good. If I understand this correctly, this is only an implementation detail, not a change to the API.

mrozbarry: Done

lukejacksonn:
Is the description of this PR accurate still; specifically the `cancel` prop part? Are we proposing that an effect no longer returns it's own cancel function?

```
subscriptions: state => [
  [someSubscriptionEffect, { ...otherParams, cancel: injectedCancelFn }],
]
```

jorgebucaran:
> @lukejacksonn Are we proposing that an effect no longer returns its own cancel function?

No. Subscriptions must still return their cancel function. The `injectedCancelFn` is appended to the array behind the scenes after the subscription starts (so we know how to cancel it later).

mrozbarry:
@jorgebucaran I think @lukejacksonn was just saying that since I updated subscriptions internally to be `[effectFn, params, cancel]` that I should update my description. I think we're all on the same page 😄

lukejacksonn:
Yeh, my bad.. I thought the description was describing an external API but it is not 👍

test/subscriptions.test.js
jorgebucaran: @mrozbarry Yeah, that's why I'd prefer adding tests later.

Issue #753, opened 2018-08-31
V2 Middleware API


jorgebucaran:
## Summary

I've made up my mind about this issue. I'm going with the dev-only middleware option. That would be the 🎉 in @okwolf's two-choice poll. This should tie in with https://github.com/jorgebucaran/hyperapp/issues/417. I'll close there and summarize everything here.

Thank you everyone, for your feedback! ❤️ 

> Technically, this isn't middleware, I've been calling it so incorrectly. Hyperapp will have no middleware feature as we know it from other projects such as Redux or Express. In the past we've used _plugins_ (pre-V1) to integrate to ReduxDevTools and later V1 wiredActions for the same purpose. V2 will provide its own DevTools instead. 

We want a top-class development experience when working with Hyperapp. Tools like [okwolf/logger](https://github.com/okwolf/logger), [mrozbarry/debug](https://github.com/mrozbarry/hyperapp-debug) and [hyperstart/devtools](https://github.com/hyperstart/hyperapp-devtools) have helped us along the way and will be the inspiration for what's coming next. 

Starting with V2 we'll be able to compile our apps for production _or_ development. Development mode will have useful warnings, logging, session import/export and time-traveling built right into your app!

This will work by [setting an environment variable](https://en.wikipedia.org/wiki/Env), e.g., <samp>NODE_ENV=production</samp> or <samp>NODE_ENV=development</samp>. What will be default? 

For CDN users, <samp>hyperapp.js</samp> will be the development build and a new <samp>hyperapp.min.js </samp> will be available.

okwolf:
I've held off on making the first to comment here since I felt I sucked up most of the oxygen in the [last issue about middleware](https://github.com/jorgebucaran/hyperapp/issues/703).

I think the right way to kick this off is to decide the scope of when middleware should be used in the first place. 

The @jorgebucaran perspective on this has already been voiced:

> [The kind of middleware that I want to see is the least invasive kind, the one that allows you to build a logger or a time travel debugger, but not modify how dispatch works in order to, among other things, add Promise support to actions, etc.](https://github.com/jorgebucaran/hyperapp/issues/703#issuecomment-416205392)

Perhaps we should kick this off with a small poll (inspired by the [action parameter comment](https://github.com/jorgebucaran/hyperapp/issues/751#issuecomment-416974789)). Please react to this comment accordingly:

- 🎉 I only want to enable development/debug middleware with Hyperapp
[...]

PR #803, opened 2019-02-17
Timetravel debugger

mrozbarry:
## Notes

This is **not** ready for production yet. I'm putting this PR up to get the ball rolling, but it's going to require a lot more thought and feedback.

## Tasks

 - [x] Implement a timetravel debugger
 - [x] Make it always enabled
 - [ ] Make it look nice
 - [ ] Come up with a good way to test it other than manual/visual testing
 - [x] Come up with a way to strip out the debugger on minified builds

## Screenshots

screen shot 2019-02-16 at 6 46 52 pm
screen shot 2019-02-16 at 6 47 53 pm


demo/app.js
```diff
@@ -0,0 +1,19 @@
+import { app, h } from '../src/index.js';
```
mrozbarry: We can safely ignore this, I don't think it should be bundled in hyperapp, this is just for me to demo locally and not have to constantly stash push/pop.

package.json
```diff
@@ -28,10 +28,13 @@
     "minify": "uglifyjs dist/hyperapp.js -o dist/hyperapp.js -mc pure_funcs=['Object.defineProperty'] --source-map includeSources,url=hyperapp.js.map",
     "prepare": "npm run build",
     "format": "prettier --write {src,test}/**/*.js {,test/ts/}*.{ts,tsx}",
-    "release": "npm run build && npm test && git commit -am $npm_package_version && git tag $npm_package_version && git push && git push --tags && npm publish"
+    "release": "npm run build && npm test && git commit -am $npm_package_version && git tag $npm_package_version && git push && git push --tags && npm publish",
+    "demo": "parcel ./demo/index.html"
```
mrozbarry: Ignore this

package.json
```diff
@@ -28,10 +28,13 @@
...
   },
   "babel": {
-    "presets": "env",
+    "presets": [
```
mrozbarry: Although this change is for parcel, I think for it to be valid, it needs to be in an array.

package.json
```diff
@@ -42,9 +45,11 @@
     ]
   },
   "devDependencies": {
+    "babel-core": "^6.26.3",
```
mrozbarry: Parcel change, we can ignore this

package.json
```diff
@@ -42,9 +45,11 @@
...
     "babel-plugin-transform-react-jsx": "^6.24.1",
     "babel-preset-env": "^1.7.0",
     "jest": "^23.5.0",
+    "parcel-bundler": "^1.11.0",
```
mrozbarry: Ignore this.

rollup.config.js
```diff
@@ -0,0 +1,16 @@
...
+  input: 'src/index.js',
+  output: {
+    file: 'dist/hyperapp.js',
+    format: 'umd',
+    name: 'hyperapp',
+    sourcemap: true
+  },
+  plugins: [
+    replace({
+      ENVIRONMENT: JSON.stringify(process.env.NODE_ENV || 'production')
```
okwolf: Why not use the same `NODE_ENV` for the identifier to replace?

src/index.js
```diff
@@ -581,6 +581,13 @@ export function app(props) {
     dispatch(event.currentTarget.events[event.type], event)
   }
 
+  if (ENVIRONMENT === 'development') {
```
okwolf: Does Uglify have dead code elimination on to actually remove this from the bundle? How about the `makeDebugger` function?

mrozbarry: Looks like it, but I'll attach the build files so we can all double check

rollup.config.js
mrozbarry: Done

src/index.js
mrozbarry: @okwolf @jorgebucaran here's the build outputs: https://gist.github.com/mrozbarry/b42ccf9704e7eee5ffc281b72f384440

We'll want to double check that makeDebug is not present in hyperapp.js or hyperapp.min.js, but is in hyperapp.dev.js

rollup.config.js
okwolf: Thanks, this is what I've done on other projects that needed to branch based on build environment and I think it will probably be more obvious where values come from.

okwolf:
Anyone who wants to check this out is probably going to want [this](https://github.com/mrozbarry/hyperapp/tree/timetravel-debugger) link.

mrozbarry:
@jorgebucaran asked me to put up some alternative middleware discussion here, also linking to #753 as this is relevant.

What are some alternatives to built in middleware, with pros and cons?

Well, first, for built in:

Pros:

 - Tight integration
 - Guaranteed version compatibility
 - No third-party exposure

Cons:

 - Updates mean new hyperapp releases
 - There might be good use-cases for third-party middleware

Second, expose some information through a single callback, like:

```js
import { app } from 'hyperapp';

app({
  // ...
  middleware: (actionFn, actionArgs, previousState, nextState) => {
  },
});
```

**Note:** Not 100% on what the params would actually be.

Pros:

 - Easy to deploy new middleware without a new release of hyperapp
 - Can bring in third-party dependencies to make the middleware work nicer (ie for debugger, styling comes to mind)
 - Can be run after the hyperapp render/events in a setTimeout to prevent middleware from blocking the main activity.
 - Can limit middleware so it can't change the behaviour of hyperapp
 - Can prevent middleware from running in production more easily
 - Pattern could allow for HoF middleware, to make it harder to trace and expose middleware the general developer.

Cons:

 - Forces extra dependencies on devs who want to use middleware
 - Relies on the end-developer to find or implement middleware
 - May not be clear that it's dev-only (what is dev and what is not)
 - Not sure if treeshaking will catch that HA doesn't run middleware, and allow the end-developer to have a smaller bundle (ie removes whatever middleware calls).

If there are other solutions/ideas, we'd be very interested in discussing those, too.

zaceno:
@mrozbarry My only comment on the non-built-in option, is: let's not call the prop `middleware`. Since it doesn't actually allow "ware" to get in the "middle" of anything, it's more of an inspection opportunity for logging/timetravel et c. Let's rather call it `probe`, `log`, `inspect` or `tap` (or anything else implying that things go in but don't come out).

That's not to say I would be opposed to proper middleware, but that's not what we're talking about here I think.

mrozbarry:
@zaceno I'm with you. I think in a previous issue, I think you (maybe someone else) mentioned `tap` - I like that, but I'm biased because ruby. I might even go as far as calling it `devTap` so it's more obvious on when this method is even relevant/called.

sergey-shpak:
Is there any possibility of [integration](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/Integrations.md) with [Redux DevTools](https://github.com/zalmoxisus/redux-devtools-extension)?

mrozbarry:
@sergey-shpak if we expose some sort of middleware integration, then I don't see why not.

mrozbarry:
Closing in favor of discussion in #753

PR #805, opened 2019-02-18
V2 css enhance

cbbfcd:
just like this:

```js
// the props
{
  style: {
     color: 'orange !important'
  }
}
```

according to this 👉[setProperty](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty)

⚠️：value must not contain "!important" -- that should be set using the priority paramete

and i have a  test: 
```
// it's no use
someNode.style.color = 'orange !important'
```

So I think there should be a test about importance here.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -402,14 +402,15 @@ var patchElement = function(
 }
 
 var createVNode = function(name, props, children, element, key, type) {
-  return {
+  var pureVNode = Object.create(null)
+  return Object.assign(pureVNode, {
     name: name,
     props: props,
     children: children,
     element: element,
     key: key,
     type: type
-  }
+  })
 }
 
 var createTextVNode = function(text, element) {
--- a/src/index.js
+++ b/src/index.js
@@ -20,6 +20,12 @@ var merge = function(a, b) {
   return target
 }
 
+var assign = function(a, b) {
+  for(var i in b) if(b.hasOwnProperty(i)) a[i] = b[i]
+  
+  return a
+}
+
 var resolved = typeof Promise === "function" && Promise.resolve()
 
 var defer = !resolved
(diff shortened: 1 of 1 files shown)


lukejacksonn:
Using important is usually a last resort.. is using important on inline styles ever advised? Or are you just going for complete/closeness to the spec?

mrozbarry:
Is this necessary? Inline styles always take precedence unless a css style was already marked as `!important` elsewhere. Furthermore, like @lukejacksonn said, marking `important` styles is always a last resort, so maybe we shouldn't encourage it.

@cbbfcd  Are you running into an issue where you can't use css and must flag the inline style as important?

frenzzy:
Have you tried to use `cssText`? Why don't you like it?
```jsx
<div style={{ cssText: 'color:orange !important' }} />
```
Your approach will not work with styles in lowerCamelCase like this one:
```jsx
<div style={{ fontSize: '32px !important' }} />
```

cbbfcd:
@lukejacksonn

Sorry, I did some research. My idea is indeed wrong. Thank you for your correction.

cbbfcd:
> Have you tried to use `cssText`? Why don't you like it?
> 
> ```js-jsx
> <div style={{ cssText: 'color:orange !important' }} />
> ```
> 
> Your approach will not work with styles in lowerCamelCase like this one:
> 
> ```js-jsx
> <div style={{ fontSize: '32px !important' }} />
> ```

yep,

My idea is indeed wrong. thank u very much.

cbbfcd:
> Is this necessary? Inline styles always take precedence unless a css style was already marked as `!important` elsewhere. Furthermore, like @lukejacksonn said, marking `important` styles is always a last resort, so maybe we shouldn't encourage it.
> 
> @cbbfcd Are you running into an issue where you can't use css and must flag the inline style as important?
> 
> Another thing to consider is that [react has not put any stance on using !important](https://github.com/facebook/react/issues/1881) either.

My idea is indeed wrong. thank u very much

jorgebucaran:
@cbbfcd Could you merge or remove your duplicate comments? Thanks.

PR #806, opened 2019-02-20
Fix sub tuple bug

okwolf:
See https://github.com/jorgebucaran/hyperapp/pull/726#discussion_r258327167


jorgebucaran:
Thanks, @okwolf! 🎉

This fix is available in `hyperapp@2.0.0-alpha.2`.

PR #807, opened 2019-02-20
Fix subscription cancel

mrozbarry:
## Tasks

 - Fix subscription cancel reference from `sub[1][2]()` to `sub[2]()`, thanks @okwolf


mrozbarry:
Oops, saw that @okwolf beat me to it 🥇

PR #809, opened 2019-02-21
Memoized subscriptions

mrozbarry:
## Tasks

 - [x] Memoize `subscription()` method - if state doesn't change, don't try and change subscriptions
 - [x] Move subscription handling to setState
 - [x] Add utilities to compare subscriptions
 - [x] Refactor refresh to manage changes in subscriptions
 - [x] Decouple subscriptions from ordering (may occasionally do odd things?)
 - [x] Re-introduce array nesting in subscriptions
 - [x] Fix bug where subscription arguments weren't always being deep-compared properly

## Notes

 - Because this implementation cares _less_ about ordering, this could mean that a state change that both removes a subscription while also changing another subscription to match the same props could result in the changed subscription being re-used rather than reset. This can be avoided by passing unique key/value pairs in as props (ie an id).

## Test App

 I built a small app to manage a series of timers to run intervals at different timings. Sorry for the es6 jibberish, I'm using parcel, and just wanted to hammer this out quickly. [Same code, but on codepen](https://codepen.io/mrozbarry/pen/JxVqrY?editors=0010).

```js
// Tuple sub fix
// With the original sub code, a second subscription can't even start
import { app, h } from 'https://rawcdn.githack.com/okwolf/hyperapp/ff6c78805bb9911088ee503170e0096241de2412/src/index.js'

// This is my subscription refactor
// Now subs behave the way we'd expect
// import { app, h } from 'https://rawcdn.githack.com/mrozbarry/hyperapp/e5138ac26c4bcbd08b037cec1409b675c7f60416/src/index.js';

const updateTimer = (timers, index, update) => [
  ...timers.slice(0, index),
  { ...timers[index], ...update },
  ...timers.slice(index + 1),
];

const timerSecondsAction = (index) => {
  const setTimerSeconds = (state, event) => {
    const number = parseInt(event.target.value, 10);
    const seconds = Math.min(10, Math.max(1, number));
    return {
      ...state,
      timers: updateTimer(state.timers, index, { seconds }),
    };
  };

  return setTimerSeconds;
};

const timerEnabledAction = (index) => {
  const setTimerEnabled = (state, event) => {
    return {
      ...state,
      timers: updateTimer(state.timers, index, { enabled: event.target.checked }),
    };
  };

  return setTimerEnabled;
};

const timerTickAction = (index) => {
  const setTimerLast = (state, arg) => ({
    ...state,
    timers: updateTimer(state.timers, index, { ticks: state.timers[index].ticks + 1 }),
  });

  return setTimerLast;
};

const timerFx = (props, dispatch) => {
  const action = props.actionBuilder(props.index);
  const tick = () => dispatch(action, null);

  const handle = setInterval(tick, props.interval);

  return () => {
    clearInterval(handle);
  };
};

const addTimerAction = state => ({
  ...state,
  timers: state.timers.concat({ enabled: false, seconds: 1, ticks: 0 }),
});

const removeTimerAction = index => {
  const removeTimer = (state) => ({
    ...state,
    timers: [
      ...state.timers.slice(0, index),
      ...state.timers.slice(index + 1),
    ]
  });

  return removeTimer;
};

const initialState = {
  timers: [{ enabled: false, seconds: 1, ticks: 0 }]
};

const timerView = (timer, index) => h(
  'div',
  null,
  [
    h('button', { type: 'button', onclick: removeTimerAction(index) }, 'Remove'),
    'Tick every ',
    h('input', { type: 'number', min: 1, max: 10, step: 1, value: timer.seconds, oninput: timerSecondsAction(index) }),
    ' seconds. ',
    h('label', null, [
      h('input', { type: 'checkbox', checked: timer.enabled, onchange: timerEnabledAction(index) }),
      ' Enable?'
    ]),
    ' ',
    h('span', null, ['Ticks: ', timer.ticks]),
  ]
);


const view = state => h(
  'div',
  null,
  [
    h('div', null, state.timers.map(timerView)),
    h('div', null, [h('button', { type: 'button', onclick: addTimerAction }, 'Add Timer')]),
  ]
);

app({
  init: initialState,
  view: view,
  container: document.getElementById('app'),
  subscriptions: state => [
    ...state.timers.map((timer, index) => timer.enabled ? [timerFx, { interval: timer.seconds * 1000, index, actionBuilder: timerTickAction }] : false).filter(Boolean),
  ],
});
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -502,14 +502,21 @@ var flattenSubs = function(subs) {
 }
 
 var isSameFn = function(a, b) {
-  return a === b || (a.name && (a.name === b.name))
+  const result = a === b || (a.name && (a.name === b.name))
+
+  return result
 }
 
-var isSameSubArg = function(a, b, deep) {
-  if (a && b && a !== b && typeof a === 'object' && deep) {
+var canDeepCheck = function(a) {
+  return isArray(a)
+    || (Object(a) === a && Object.getPrototypeOf(a) === Object.prototype)
+}
+
+var isSameSubArg = function(a, b) {
+  if (a && b && a !== b && (canDeepCheck(a) && canDeepCheck(b))) {
     var combined = merge(a, b)
     for(var key in combined) {
-      if (!isSameSubArg(a[key], b[key], false)) {
+      if (!isSameSubArg(a[key], b[key])) {
         return false
       }
     }
@@ -522,7 +529,7 @@ var isSameSubArg = function(a, b, deep) {
 }
 
 var isSameSub = function(a, b) {
-  return isSameFn(a[0], b[0]) && isSameSubArg(a[1], b[1], true)
+  return isSameFn(a[0], b[0]) && isSameSubArg(a[1], b[1])
 }
 
 var refresh = function(subs, oldSubs, dispatch) {
--- a/src/index.js
+++ b/src/index.js
@@ -25,8 +25,7 @@ var resolved = typeof Promise === "function" && Promise.resolve()
 var defer = !resolved
   ? setTimeout
   : function(cb) {
-      return resolved.then(cb)
-    }
+      return resolved.then(cb) }
 
 var updateProperty = function(
   element,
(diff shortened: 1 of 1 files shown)


SkaterDad:
@mrozbarry Can you explain the benefits of this PR?

This part is a little concerning: " Decouple subscriptions from ordering (may occasionally do odd things?)"

mrozbarry:
@SkaterDad yeah, but the ordering still matters. The issue is if I have something like:

Initial subs:

```
[
  [subA, { foo, bar }],
  [subB, { baz }],
  [subA, { foo, bar }],
]
```

Then I subsequently remove the first subA,

```
[
  [subB, { baz }],
  [subA, { foo, bar }],
]
```

Since both `subA` subscriptions had the same params, there isn't a good way to know which one to remove, but we do know we can remove one of them. My algorithm tries to find the closes array index to the original to remove, but also finds the first one to keep. This is easily avoided by adding some unique thing (like an id) if you have multiple subs with the same arguments. I think that's the only edge case, though.

mrozbarry:
@SkaterDad in terms of benefits, the current V2 subscriptions are broken, and you cannot have multiple subscriptions running. Check out the codepen I have linked. Using the tuple fix branch (which is essentially V2), you can't run more than 1 timer. Using this fix, you can run as many as you want, and even remove or disable them live, and things keep working.

zaceno:
@mrozbarry @SkaterDad 

I'll confirm this fixes the bug I was noticing too(which I only notice on @okwolf sub tuple fix PR, because that fixed a problem preceding this one).

https://codepen.io/zaceno/pen/yZrxdq

mrozbarry:
I guess the main thing about the positioning note I made is that you can safely concat/slice off subscriptions without having pre-established subs resetting.

SkaterDad:
@mrozbarry Thanks for clarifying.  My only subscription testing has been with a history popstate listener.

I think it would make sense to require people to pass a unique key if they want multiple subs of the same type.  You already have to do that with the VDOM to get predictable results.

If keys were required, could the implementation be simplified at all?

mrozbarry:
Internally, we could certainly assign an id, but that doesn't change the initial issue. I think for the most part, having multiple subscriptions with the same effect and params is going to be very uncommon in real-world applications.

I've been trying to digest any potential reasons why you'd have repeated subs from apis or data structures, and I honestly can't think of any that couldn't have pre-assigned ids passed into the sub.

zaceno:
The thing about keys in the vdom is that identical vendors are not always equivalent. That’s (one reason) why we need keys to tell them apart.

In the “subdom” however, I believe it’s ok to assume that identical subscription nodes are 100% equivalent, so no need to distinguish between them. 

I could be wrong of course, but if I am, it’s easy enough for the user to pass along a unique dummy prop to make them non-identical. So essentially like keys from the users standpoint, but from our standpoint we don’t need to look for a prop called key and treat it differently.

zaceno:
Hm... seems unfortunately there is something incomplete about this PR. It fixed the issue in my reduced example (above).

In the original example (from my upcoming HAv2 article), there is still some strange behavior:

https://codepen.io/zaceno/pen/vzMbLQ

Notice that if you start the first timer and then the second, the second does not move. However, it fixed *something*, because if you pause the first (running) timer, and then unpause it, suddenly both timers work as expected, for all time until the page reloads. So there's something stuck at first, that "jiggles loose" when you fiddle with the subscriptions. Odd...

jorgebucaran:
Was there a bug in the original implementation before the API changes introduced in #802?

zaceno:
@jorgebucaran 

*Before* #802, no. Not afaik. With 802 came a bug (Bug A) that was fixed in #806, but that revealed another bug (B) (either also in 802, or that came with 806, can't say). This #809 appears to *mostly* fix bug B, but not completely.

zaceno:
But I wonder how it is that this PR about memoizing subs (semi-)fixes the issue I'm seeing. Is it a coincidence that memoization fixes the issue, or does this PR deliberately contain 2 changes? Either way I feel there should be 2 PRs: one that fixes the bug (because that should definitely be included) and one for the memoization (because that is a more debatable change)

mrozbarry:
Hmm. Yeah, it seems that there wasn't an issue before #802 - forked my [codepen from an earlier commit](https://codepen.io/mrozbarry/pen/yZdLvL?editors=0010), and it works as expected. Now I'm not sure what the issue is there. I did get rid of a thing that ignored a cancel key, but I couldn't figure out what that was. Also, my tests only covered single subscriptions, which was a bit of a failing on my part.

To address why this PR fixes things, this PRs name is maybe a bit incorrect - the subscriptions were already memoized indirectly by storing subs. I did, however, re-write the code that figures out which subs are new, continued, or cancelled. I had a lot of trouble reasoning through the original V2 code, and I wanted to try and figure out how to fix the new bugs I introduced.

zaceno:
Thanks @mrozbarry, then I understand. This PR is intended to fix the bug in #806 and no need for another PR. 👍 

even after your latest commit, my codepen has the same weird "almost working" behavior. Do you have any leads, or can I help somehow reduce the example? (If you have a clue as to what might be going wrong I could try... I'm just not sure how atm)

mrozbarry:
@zaceno Fixed the bug you were experiencing. The short story is I was preventing checking arguments that were too deep in an object or array, which is probably mostly wrong. Now the code checks n-deep (instead of fixed 1-deep) unless the variable is not a POJO or a real Array. Everything else is by value or reference, which should cover primitives and ref comparison (ie if you have a DOM node, function, or non-standard-js object in your state).

jorgebucaran:
@mrozbarry Sorry for the late reply, but I'm still confused by this PR. I think I want to take a stab myself. Could you please describe the following two items from your list? 

- Memoize subscription() method - if state doesn't change, don't try and change subscriptions
- Fix bug where subscription arguments weren't always being deep-compared properly

mrozbarry:
@jorgebucaran no worries. Once I started, I realized that your were already memoizing subscriptions - the bug I introduced made me think we weren't. I just checked it to confirm that is the behaviour.

The most recent commit fixes an issue with my own implementation where subscription arguments weren't being compared properly. Check out my comment above or the last commit for more details.

dzgank:
To address why this PR fixes things, this PRs name is maybe a bit incorrect - the:Check out my comment above or the last commit for??,

zaceno:
@mrozbarry Just dropping in to confirm that the latest fix did the trick. With it, tuple-subscriptions appear to be fully working as intended.

zaceno:
@jorgebucaran 

As I've come to understand it, the confusion about what this PR is about stems from the fact that really #802, #806 and #809 are *all* about implementing the tuple-signature for subscriptions. The first one introduced it, and the rest were about fixing bugs. The latest force-push in this PR seems to complete the picture.

So I don't think it would make sense for you to "take a stab at" your implementation of this particular PR. Rather, you could take a stab at implementing tuple signature for subs in your own preferred way.

mrozbarry:
Closed since it looks like V2 subscriptions are behaving as expected now.

PR #811, opened 2019-02-26
Lifecycle hooks as Events (no delayed remove)

lukejacksonn:
I shared a snippet of an alternative approach to lifecycle hooks I have been playing with in a fork of V2 being used on personal project of mine. This raised a few questions and a more detailed explanation was requested, so here is the implementation. A couple of things are assumed:

- Firstly @frenzzy suggested that:
> I think delayed remove is not needed, but oncreate and ondestroy are important https://hyperapp.slack.com/archives/C41ECC0V6/p1551190367214200

I'm not sure how you can do transition out style animations without them? But removing this certainly [simplifies](https://github.com/jorgebucaran/hyperapp/compare/V2...lukejacksonn:V2#diff-1fdf421c05c1140f6d71444ea2b27638L179) hyperapp core so `onRemove` is omitted here.

- Secondly by nature of the implementation the remaining lifecycle events get treated like any user interaction (like `onclick` for example) and so now have the same signature and behave the same:

```js
oncreate | onupdate | onremove:  (state, e) => ...,
```

~I will create a codepen for this build now.~ Here is a demo on [CodePen](https://codepen.io/lukejacksonn/pen/EMawmX?editors=0010).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -9,6 +9,10 @@ var SVG_NS = "http://www.w3.org/2000/svg"
 var EMPTY_OBJECT = {}
 var EMPTY_ARRAY = []
 
+var CREATE_EVENT = new Event('create')
+var UPDATE_EVENT = new Event('update')
+var REMOVE_EVENT = new Event('remove')
+
 var map = EMPTY_ARRAY.map
 var isArray = Array.isArray
 
@@ -104,9 +108,9 @@ var createElement = function(node, lifecycle, eventProxy, isSvg) {
       : document.createElement(node.name)
 
   var props = node.props
-  if (props.onCreate) {
+  if (props.oncreate) {
     lifecycle.push(function() {
-      props.onCreate(element)
+      element.dispatchEvent(CREATE_EVENT)
     })
   }
 
@@ -154,10 +158,10 @@ var updateElement = function(
     }
   }
 
-  var cb = isRecycled ? nextProps.onCreate : nextProps.onUpdate
+  var cb = isRecycled ? nextProps.oncreate : nextProps.onupdate
   if (cb != null) {
     lifecycle.push(function() {
-      cb(element, lastProps)
+      element.dispatchEvent(UPDATE_EVENT)
     })
   }
 }
@@ -167,25 +171,15 @@ var removeChildren = function(node) {
     removeChildren(node.children[i])
   }
 
-  var cb = node.props.onDestroy
-  if (cb != null) {
-    cb(node.element)
+  if (node.props.onremove != null) {
+    node.element.dispatchEvent(REMOVE_EVENT)
   }
 
   return node.element
 }
 
 var removeElement = function(parent, node) {
-  var remove = function() {
-    parent.removeChild(removeChildren(node))
-  }
-
-  var cb = node.props && node.props.onRemove
-  if (cb != null) {
-    cb(node.element, remove)
-  } else {
-    remove()
-  }
+  parent.removeChild(removeChildren(node))
 }
 
 var getKey = function(node) {


src/index.js
```diff
@@ -104,9 +108,9 @@ var createElement = function(node, lifecycle, eventProxy, isSvg) {
       : document.createElement(node.name)
 
   var props = node.props
-  if (props.onCreate) {
+  if (props.oncreate) {
```
SkaterDad: Thank you for this :clap:

lukejacksonn: Hah 😅 unintentional side effect of trying to keep things as uniform/simple as possible.

SkaterDad:
Played around a bit with your Codepen.

I like the new function signature.  Having the app state injected should allow us to create these handlers outside of the view function, which could have small perf wins.

Removing the delayed onRemove is going to make animations more challenging, but it's workable.

Here is a forked codepen showing one way to achieve a fade-in/fade-out:  https://codepen.io/SkaterDad/pen/rRaGoz?editors=0010

I cloned the exiting node and inserted it into the parent in the same position.  I then remove the node when the animation is done.

jorgebucaran:
@SkaterDad Could you add onRemove back? 🙏

jorgebucaran:
@lukejacksonn  Could you add onRemove back? 🙏

src/index.js
frenzzy: if remove this line at all, it will be possible to use both `onCreate` and `oncreate` similar to all other events like `onClick` and `onclick`.

lukejacksonn:
Humm.. adding back `onRemove` is slightly problematic as we can't pass the `remove` function through an Event 🤔

jorgebucaran:
@lukejacksonn IMO the most important thing about this PR is the user-facing API changes making lifecycle events act like other events, i.e., implicit dispatch callers. If using `Event` creates an implementation problem, we can use `dispatch` directly in the original implementation.

jorgebucaran:
@lukejacksonn IMO the most important thing about this PR is the user-facing API changes making lifecycle events act like other events, i.e., implicit dispatch callers. If using `Event` creates an implementation problem, we can use `dispatch` directly in the current code. Do you want to make those changes?

We need to have a discussion about how this changes the way lifecycle events can be used. @loteoo You were telling me how you've been using this in your own projects. Can you share it here as well?

I'd love to hear what @zaceno thinks about this too.

src/index.js
```diff
@@ -9,6 +9,10 @@ var SVG_NS = "http://www.w3.org/2000/svg"
 var EMPTY_OBJECT = {}
 var EMPTY_ARRAY = []
 
+var CREATE_EVENT = new Event('create')
```
mrozbarry: Depending on browser support, we may want to use `event = new CustomEvent('build', { detail: elem.dataset.time });`.

Another thing I think we need to consider is running in headless - will this still work as intended?

zaceno:
@jorgebucaran I think this proposal is good. Especially the fact that this makes it possible (...or, well, easier) to manage state and dispatch effects in relation to particular dom elements. I think it should be perfectly adequate to have just "oncreate" and "ondestroy". (Perhaps all we need, even is "oncreate")

(Also I am 100% on board with all-lowercase at least as an option, for all events incl lifecycle)

src/index.js
lukejacksonn: From what I have read.. `Event` is supported better than `CustomEvent` but we might need to *not* use the constructor. Something like this (which is not ideal I guess but will work):

```
var event;
if(typeof(Event) === 'function') {
    event = new Event('submit');
}else{
    event = document.createEvent('Event');
    event.initEvent('submit', true, true);
}
```

lukejacksonn:
Just to clarify one thing.. I didn't change the capitalisation of the lifecycle events out of preference, rather out of necessity – as they are treated the same as any other event which gets lowercased. That said, I do prefer it like this 😅

jorgebucaran:
> @zaceno I think it should be perfectly adequate to have just "oncreate" and "ondestroy".

I'm not following. We need onUpdate and onRemove too.

jorgebucaran:
@lukejacksonn I can still use camelCase correct?

SkaterDad:
> @zaceno  I think it should be perfectly adequate to have just "oncreate" and "ondestroy". (Perhaps all we need, even is "oncreate")

I agree that the delayed onremove isn't strictly necessary, but I think `onupdate` and `ondestroy` are pretty important if you're using an non-hyperapp component lib or trying to do animation.

jorgebucaran:
I have no intention of removing some lifecycle events. It's all or nothing.

mrozbarry:
This probably isn't explicitly for this PR, but I don't think this will work for headless mode.  Then again, most events wouldn't work unless you used jsdom or similar (which is essentially not headless mode), so maybe that's okay, but probably warrants a discussion.

zaceno:
> I have no intention of removing only some lifecycle events. It's all (Create, Update, Remove, and Destroy) or nothing (no lifecycle events at all #717).

That's entirely up to you of course, although I think a halfway approach would be a win. We kan keep convenient access to particular DOM-nodes of interest, while still reducing the complexity our current lifecycle methods add.

Like @SkaterDad says, maybe Update is necessary (but perhaps we can simplify it to only run if attributes change). I no longer think both onDestroy and onRemove are necessary. Just one that works like onDestroy does should be enough.

B-Teague:
I would like to use this solution for simplifying css transitions with onCreate and onDestroy by changing the global state. I was looking for this exact solution earlier, but had to instead manually add an attribute to the element and a custom css selector (yuck). Maybe I’m going about transitions the wrong way?

jorgebucaran:
There's disagreement over what lifecycle methods we should keep and which ones are okay to lose as if we needed to simplify Hyperapp when there's no pressing need to do that.

Implementation details aside, do we want to change the lifecycle API to dispatch actions like we can from other DOM events?

SkaterDad:
I vote yes on the new function signature. Seems like a big usability win, so long as it isn't adding too many bytes.

lukejacksonn:
I have been migrating continent these past two days so only just catching up with this thread. Will be back soon with some responses and an updated PR!

loteoo:
I'm probably late to the discussion, but as requested, what I'm doing on some of my projects is dispatching a DOM event on all dom nodes created via javascript.

[Codepen example](https://codepen.io/lot3oo/pen/LgRbjj?editors=0010)

I actually used this on the hyperapp.rocks website:
https://github.com/loteoo/hyperapp-rocks/blob/master/src/app/components/ProjectViewer/index.jsx#L26

I might move to using a Custom Element, and dispatch the `mount` event on the node from there in the `connectedCallback`, which is a little more work but seems cleaner.

And as for why I'm pro life-cycle events:
They allow you to re-use logic from the view to dispatch actions/side-effects.

(from slack, for sharing purposes)

> If you have a component that needs to start a side effect when it becomes visible, and you have already some sort of logic in the view that handles displaying / hiding that component, it is very easy to "time" that side-effect right there, instead of duplicating that logic in some subscription.

> Common example, if you have some sort of routing inside the view, and you want to fetch data when that piece of view appears, you would need a similar (duplicate) routing in another place, probably in a subscription, that handles dispatching the side-effects.

> But it's not always a routing problem, sometimes it's for animations / transitions and all sorts of view logic that might need side effects.

The Custom Elements solution definitively does the job, but it doesn't feel as good as it would be if the hook was natively integrated in hyperapp.

jorgebucaran:
@lukejacksonn I'm going to hold on this for now. I may circle back to this eventually, as I don't consider the current alternatives final answers to the problem either.

@loteoo Thank you for your input, I appreciate it. I've decided to remove lifecycle events. I also plan on getting rid of other features that can be used as "escape hatches" to V2's declarative model. I know we can (and will) find solutions to problems where lifecycle events were convenient (https://github.com/jorgebucaran/hyperapp/issues/717). These solutions may not be perfect at first, but we'll iterate on them and learn how to improve them. 

One situation where lifecycle events were really convenient was when animating removing elements (delayed removals), which this PR removes. But even for that, I don't want to use lifecycle events.

PR #812, opened 2019-03-05
Fix bug with Lazy when used in JSX

SkaterDad:
When Lazy is passed through `h` (like when used in JSX), it was being assigned a `children` property, set to an empty array.

This unnecessary "children" property was preventing the `shouldUpdate` function from returning `false` when it should have.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -462,7 +462,7 @@ export var h = function(name, props) {
   }
 
   return typeof name === "function"
-    ? name(props, (props.children = children))
+    ? name(props, name !== Lazy && (props.children = children))
     : createVNode(name, props, children, null, props.key, DEFAULT_NODE)
 }
 


jorgebucaran:
@SkaterDad 👍

src/index.js
```diff
@@ -462,7 +462,7 @@ export var h = function(name, props) {
   }
 
   return typeof name === "function"
-    ? name(props, (props.children = children))
+    ? name(props, name !== Lazy && (props.children = children))
```
frenzzy: This is not the only problem with a current particular line of code.
For example your immutable state may become mutable if you will use a spread operator like this:
```js
const stateOrSlice = { a: 1 } // immutable state will be mutated by view
<div {...stateOrSlice} />     // dangerous usage
console.log(stateOrSlice)     // oh no => { a: 1, children: [] }
```
A discussion about the problem is here: https://github.com/jorgebucaran/hyperapp/pull/726#discussion_r201326488

jorgebucaran: @frenzzy Do you know how other frameworks solve this problem?

frenzzy: by copying instead of mutation? https://github.com/facebook/react/blob/v16.8.3/packages/react/src/ReactElement.js#L175

jorgebucaran: @frenzzy We should do this right. Want to send me a PR? :)

frenzzy: and after it they have to use shallow compare instead of `===` for React.memo =/

frenzzy: here we go: https://github.com/jorgebucaran/hyperapp/pull/813

PR #813, opened 2019-03-05
Never mutate component props

frenzzy:
Alternative solution to https://github.com/jorgebucaran/hyperapp/pull/812

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -435,19 +435,19 @@ export var Lazy = function(props) {
   }
 }
 
-export var h = function(name, props) {
+export var h = function(name, attrs) {
   var node
+  var props = merge(attrs)
   var rest = []
   var children = []
   var length = arguments.length
 
   while (length-- > 2) rest.push(arguments[length])
 
-  if ((props = props == null ? {} : props).children != null) {
+  if (attrs && attrs.children != null) {
     if (rest.length <= 0) {
-      rest.push(props.children)
+      rest.push(attrs.children)
     }
-    delete props.children
   }
 
   while (rest.length > 0) {


SkaterDad:
This PR breaks `Lazy` again, since it is allowing the `children` prop to be attached, and it's a new array every time.

frenzzy:
There are no tests in V2 branch 😢

SkaterDad:
It should be fine again if you re-add the code I just PR'd.

SkaterDad:
@frenzzy It's working again, thanks :+1:  

Tests really would be nice, but I can understand why they were deleted until the APIs are settled.

I'm a bit concerned about the overhead of copying props on every invocation of `h`.  Seems like an expensive thing to do.

If the component signature had stayed `(props, children)`, this wouldn't be an issue at all, right?

frenzzy:
> If the component signature had stayed `(props, children)`, this wouldn't be an issue at all, right?

Exactly!

jorgebucaran:
@SkaterDad @frenzzy What if we just go back to the (props, children) syntax? That would fix this no?

jorgebucaran:
@SkaterDad @frenzzy What if we just go back to the (props, children) syntax? That would fix this no?

Sorry, I'm busy with other Hyperapp stuff and haven't gotten to this problem yet.

SkaterDad:
> @SkaterDad @frenzzy What if we just go back to the (props, children) syntax? That would fix this no?
> 
> Sorry, I'm busy with other Hyperapp stuff and haven't gotten to this problem yet.

I think that'll fix it, since you won't have any need to mutate the `props` object being passed in.

I can't remember why the signature was changed to begin with.  Was it to make it easier to copy a JSX component from React to Hyperapp?

jorgebucaran:
@SkaterDad Yes, but am I worrying too much? Performance-wise, how much this branch affects the benchmarks?

frenzzy:
@jorgebucaran yes, `(props, children)` signature will solve the issue (HAv1 uses it and does not have this problem at all)

@SkaterDad reasons to change signature where here: https://github.com/jorgebucaran/hyperapp/issues/569

Back to `(props, children)` component signature: PR https://github.com/jorgebucaran/hyperapp/pull/814

SkaterDad:
@frenzzy  Of the reasons listed, the one I think is most appealing is the potential to pass in context as another argument.  However, that could also be done V1-style in the future if needed, without a breaking change. 

```jsx
const comp = (props, children) => state => (<div>{children}</div>)
```

jorgebucaran:
@SkaterDad What is this `context`? 🤔

SkaterDad:
> @SkaterDad What is this `context`?

@frenzzy mentioned it in #569 as a speculative thing.

Could be anything you want, I suppose?  Theoretically it could be the global state, but we'd probably have to thread that all over the place to make it work.

For my own projects, I think the `(props, children)` signature for components is perfectly fine, and it avoids this mutated props issue.

frenzzy:
Closing in favor of #814

PR #818, opened 2019-03-17
Build hyperapp to root instead of dist

mrozbarry:
## Summary

 - Add npmignore to prevent some files from being published to npm
 - Update gitignore now that there is deviation between git and npm ignore lists
 - Update scripts to target root rather than dist

## Benefits

This will allow browsers and nodejs to include things like `import { interval } from 'hyperapp/time';` or `import { interval } from 'https://unpkg.com/hyperapp@2.0.0-alpha.X/time`.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -16,6 +16,7 @@ coverage
 
 # Dist
 *.gz
-hyperapp.js
-hyperapp.js.map
+dist/
 
+# Not necessary for distribution
+.travis.yml
--- a/package.json
+++ b/package.json
@@ -2,14 +2,13 @@
   "name": "hyperapp",
   "description": "JavaScript micro-framework for building web applications.",
   "version": "2.0.0-alpha.10",
-  "main": "dist/hyperapp.js",
+  "main": "hyperapp.js",
   "module": "src/index.js",
   "license": "MIT",
   "repository": "jorgebucaran/hyperapp",
   "homepage": "https://github.com/jorgebucaran/hyperapp",
   "files": [
-    "src",
-    "dist"
+    "src"
   ],
   "author": "Jorge Bucaran",
   "keywords": [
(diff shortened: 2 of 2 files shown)


okwolf:
I like what this is trying to achieve, but I'm curious - would this mean we would also publish UMD bundles of every `hyperapp/*`?

We might want to consider using a build script instead of trying to inline everything in `package.json` 🤔

jorgebucaran:
Thank you, @mrozbarry. Did you check if it is already possible?

okwolf:
I took a look at how [`material-ui`](https://github.com/mui-org/material-ui) creates their releases (they offer components in their own folders) and instead of creating build files at the root of the repo, they create _**only**_ the files to publish in a `build` folder (including a copy of `package.json`) and then publish the contents of that folder with [`npm publish build`](https://github.com/mui-org/material-ui/blob/02b6dab6032d0bfe9473847c366e5de6731f4d74/packages/material-ui/package.json#L30). 

This gets cleaner the more complicated your package is.

mrozbarry:
@jorgebucaran I'll double check today and let you know.

mrozbarry:
@jorgebucaran doesn't look like importing directly from dist is possible, at least from node. Steps to reproduce:

 1. Checkout `V2`
 2. `npm i && npm run build`
 3. `echo "console.log('test.js');" > dist/test.js`
 4. `npm link`
 5. go to a new demo directory
 6. `npm init && npm link hyperapp`
 7. `node` - once in the repl, first type `require('hyperapp')`, which works
 8. Now try `require('hyperapp/test')` - this does not work

Running node 10.15.3, npm 6.4.1.

mrozbarry:
That said, I like @okwolf 's publish directory way better than the solution in this PR.

okwolf:
My vote is don't change the build until we have other "folders" to import like FX and subs. Then we will have real-world test cases.

jorgebucaran:
I agree with @okwolf. Also, I need to figure out exactly how I'm going to lay out `@hyperapp/pkg` packages in the repository. I'm not going to merge this PR, but I may come back here in the future. 

Thank you! @mrozbarry 😄

PR #820, opened 2019-03-25
EXPERIMENT: Docco

### LINKED ISSUE #819: js.org Source Annotations Website
See [slack conversation here](https://hyperapp.slack.com/archives/C41ECC0V6/p1553165862002900)

I think we have full buy-in from @jorgebucaran on having the hyperapp source code annotated for the https://hyperapp.js.org site, so I think we should explore creating this annotated source code.

It's important that we do this for both V1 and V2, since there are many V1 sites alive in the wild that may not be upgraded.

mrozbarry:
## Summary

This may be closed if we're not happy with this. I can't really see a simpler way to have well synced annotated source code, and comments could still be stripped for distribution.

That said, if we do go with this, comments should **not** be:
```
// Makes a thing
var makeThing = function() {}
```

Annotations should be high-level comments for difficult to reason through code. Some comments may be more technical in nature (ie discussing why we don't use `Object.assign` or `Array.prototype.flat`). Not all functions need to be annotated, either.

## Tickets

 - Closes #819

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
   ],
   "scripts": {
     "test": "exit",
+    "docco": "docco src/index.js",
     "build": "npm run bundle && npm run minify",
     "bundle": "rollup -i src/index.js -o dist/hyperapp.js --no-esModule -m -f umd -n hyperapp",
     "minify": "terser dist/hyperapp.js -o dist/hyperapp.js -mc --source-map includeSources,url=hyperapp.js.map",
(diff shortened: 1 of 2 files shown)


jorgebucaran:
Thank you, @mrozbarry. I'll participate in this discussion after I am done writing docs for the V2 branch.

okwolf:
@mrozbarry you should probably turn off semicolons in your formatter to cut down on the diff noise 🙃

jorgebucaran:
@mrozbarry Thank you, but I don't want to add docco to Hyperapp. 🙏

PR #822, opened 2019-04-02
Create HAv2 Debugger Interface

mrozbarry:
## Tasks

 - [x] Track action and props on dispatch
 - [x] Call middleware with action and props when doing a setState

## Example Debugger Middleware

**Note:** This isn't optimized or as well written as it could be, so keep that in mind, this is just an example, a real time travel debugger would obviously need a lot more polish.

```js
import {
  app,
  h
} from 'hyperapp';

const makeEmitter = () => {
  const listeners = [];
  const queue = [];
  const listen = fn => {
    listeners.push(fn);
    while (queue.length > 0) {
      trigger(queue.pop());
    }

    const index = listeners.length - 1;
    return () => listeners.splice(index, 1);
  }
  const trigger = state => {
    if (listeners.length > 0) {
      listeners.forEach(fn => fn(state));
    } else {
      queue.push(state);
    }
  };

  return {
    listen,
    trigger,
  };
};

const AddHistoryAction = (state, appData) => {
  if (appData.action === InjectHistoryAction) return state;
  const history = state.history.slice(0, state.timeTravel + 1).concat(appData);

  return ({
    ...state,
    history,
    timeTravel: history.length - 1,
  })
};

const InjectHistoryAction = (state, newState) => newState;

const InjectDebuggerIntoAppSub = (props, dispatch) => {
  const cancel = props.emitter.listen((appData) => {
    dispatch(InjectHistoryAction, appData);
  });

  return () => {
    cancel();
  };
};

const SendHistoryFx = (props, dispatch) => {
  props.emitter.trigger(props.state);
};

const TimeTravel = (state, arg) => {
  const timeTravel = Math.min(Math.max(arg.time, 0), state.history.length - 1);
  const current = state.history[timeTravel];

  return [{
      ...state,
      timeTravel,
    },
    [SendHistoryFx, {
      state: current.state,
      emitter: arg.emitter
    }]
  ];
}

const StepBack = (state, emitter) => TimeTravel(state, {
  time: state.timeTravel - 1,
  emitter
});
const StepForward = (state, emitter) => TimeTravel(state, {
  time: state.timeTravel + 1,
  emitter
});

const ListenToAppSub = (props, dispatch) => {
  const cancel = props.emitter.listen((appData) => {
    dispatch(AddHistoryAction, appData);
  });

  return () => {
    cancel();
  };
};

const initialState = {
  history: [],
  timeTravel: -1,
};

const dispatchDecorator = appEmitter => dispatch => {
  const isAction = obj => typeof obj === "function" || (Array.isArray(obj) && typeof obj[0] === "function");

  let middlewareArgs = {};

  return (obj, props) => {
    if (isAction(obj)) {
      middlewareArgs = {
        action: obj,
        props
      };
    } else {
      appEmitter.trigger({
        ...middlewareArgs,
        state: obj
      });
      middlewareArgs = {};
    }
    return dispatch(obj, props);
  };
};

export default app => (props, otherMiddleware) => {
  const appEmitter = makeEmitter();
  const debugEmitter = makeEmitter();

  const handle = window.open('about:blank', 'HAv2Debugger', 'dependent,dialog,width=400,height=400');
  window.onbeforeunload = () => {
    handle.close();
  };

  const KeySub = (props, dispatch) => {
    const listener = (e) => {
      switch (e.key) {
        case 'ArrowDown':
          return dispatch(StepBack, debugEmitter)
        case 'ArrowUp':
          return dispatch(StepForward, debugEmitter)
      }
    };

    handle.document.addEventListener('keydown', listener);

    return () => {
      handle.document.removeEventListener('keydown', listener);
    };
  };

  app({
    init: initialState,
    view: state => h('div', null, [
      h('h1', null, 'DEBUGGER'),
      h('button', {
        type: 'button',
        onclick: [StepBack, debugEmitter],
        disabled: state.timeTravel === 0,
      }, 'Step Back'),
      h('button', {
        type: 'button',
        onclick: [StepForward, debugEmitter],
        disabled: state.timeTravel >= state.history.length - 1,
      }, 'Step Forward'),
      h('ul', null, state.history.map((s, index) => h('li', {
        style: {
          backgroundColor: (index === state.timeTravel) ? 'yellow' : 'inherit'
        }
      }, [
        h('button', {
          type: 'button',
          onclick: [TimeTravel, {
            time: index,
            emitter: debugEmitter
          }]
        }, s.action ? s.action.name : 'INIT'),
        JSON.stringify(s.state)
      ])).reverse()),
    ]),
    container: handle.document.body,
    subscriptions: () => [
      [ListenToAppSub, {
        emitter: appEmitter
      }],
      [KeySub],
    ],
  });

  const config = {
    init: props.init,
    view: props.view,
    subscriptions: (state) => {
      const appSubs = props.subscriptions ? props.subscriptions(state) : [];
      return [
        [InjectDebuggerIntoAppSub, {
          emitter: debugEmitter
        }],
      ].concat(appSubs);
    },
    container: props.container,
  };

  return app(config, dispatchDecorator(appEmitter));
};
```

## Demo

 - [Youtube Video](https://youtu.be/Pioi_319oI8) (sorry for the bad quality on that!)
 - [Surge link](http://spotless-mice.surge.sh/)
   - **Notes:**
   - Uses a popup
   - ~Used parcel build, and it trimmed down "AddOneToCount" to "e" which sort of makes it harder to read the debugger. You'll have a better time running this locally with the bundler/dev server of your choosing.~ Fixed, not minified any more.

### Code:

```js
import {
  app,
  h
} from 'hyperapp';

import withDebugger from './debugger.js'; // The thing up above ^^^

const AddOneToCount = state => ({
  ...state,
  count: state.count + 1
});

const Randomize = state => ({
  ...state,
  count: Math.round(Math.random() * 1000)
});

const SetText = (state, e) => ({
  ...state,
  text: e.target.value,
});


withDebugger(app)({
  init: {
    count: 0,
    text: 'foo',
  },
  view: state => {
    return h('div', null, [
      h('h1', null, state.count),
      h('button', {
        onclick: AddOneToCount,
      }, 'Add one'),
      h('button', {
        onclick: Randomize,
      }, 'Randomize'),
      h('div', null, [
        h('input', {
          oninput: SetText,
          value: state.text
        }),
      ]),
    ]);
  },
  container: document.getElementById('app'),
});
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -427,7 +427,7 @@ export var h = function(name, props) {
     : createVNode(name, props, children, null, props.key, DEFAULT_NODE)
 }
 
-export var app = function(props) {
+export var app = function(props, decorateDispatch) {
   var container = props.container
   var element = container && container.children[0]
   var node = element && recycleElement(element)
@@ -436,6 +436,7 @@ export var app = function(props) {
   var lock = false
   var state = {}
   var sub = []
+  var decorator = decorateDispatch || (function (d) { return d })
 
   var eventCb = function(event) {
     dispatch(event.currentTarget.events[event.type], event)
@@ -448,7 +449,7 @@ export var app = function(props) {
     state = newState
   }
 
-  var dispatch = function(obj, props) {
+  var dispatch = decorator(function(obj, props) {
     if (typeof obj === "function") {
       dispatch(obj(state, props))
     } else if (isArray(obj)) {
@@ -462,7 +463,7 @@ export var app = function(props) {
     } else {
       setState(obj)
     }
-  }
+  })
 
   var render = function() {
     lock = false


jorgebucaran:
Thank you, @mrozbarry. Good work! It looks awesome. Could we make it less invasive, though? Do you think something like this would work?

```jsx
function app(props, mw) {
  // ...

  var dispatch = (mw ||
    function(obj) {
      return obj
    })(function(obj, props) {
    if (typeof obj === "function") {
      dispatch(obj(state, props))
    } else if (isArray(obj)) {
        dispatch(obj[0](state, obj[1], props))
      } else {
        flatten(obj.slice(1)).map(function(fx) {
          fx && fx[0](fx[1], dispatch)
        }, setState(obj[0]))
      }
    } else {
      setState(obj)
    }
  })

  // ...
}
```

src/index.js
```diff
@@ -436,6 +436,7 @@ export var app = function(props) {
   var lock = false
   var state = {}
   var sub = []
+  var middlewareArgs = { action: null, props: null }
```
mrozbarry: Might make sense for this to be an array, then we could use `middleware.apply(null, middlewareArgs)(state)` a few lines down.

mrozbarry:
@jorgebucaran I think that would work, other than I'd probably hoist that up and do `var mw = middleware || function(obj) { return obj; };` in the app var initializations, just to avoid building an adhoc function.

mrozbarry:
Update implementation and initial PR comment with updated stuff.

mrozbarry:
@jorgebucaran do we need anything else in this PR to get it validated and merged? I also wouldn't mind a larger discussion here for other potential usage candidates other than a debugger.

jorgebucaran:
@mrozbarry I think this is good, but can we discuss it a bit more? I'm still not ready to release V2 since I'm still working on documentation. The question I'd like to ask you: how would you simplify the DX for middleware authors?

mrozbarry:
I think the form of data you get from dispatch is not human friendly. It would be great to have helpers to know if dispatch would set state, run an effect, etc.

okwolf:
@mrozbarry I vote any such helpers should be a 3rd party library and not part of Hyperapp core.

mrozbarry:
I think a funny(?) middleware could be just what I described:

```js
withDispatchWatcher(app({
  dispatchWatchers: {
    onAction((dispatch, action, args) => {}),
    onState((dispatch, state) => {}),
    onEffects((dispatch, fx) => {}),
  }
})
```

The big thing is making sure authors don't have to have the same boilerplate checking for state/action/effects in every dispatchDecorator they write. That said, other than the debugger, what things do we expect to be written for HAv2 dispatch decoration? I think two that I've read that I could see early are @okwolf 's error handling (wrap dispatch in try catch?) and automatic state persistence (using a sub seems wrong since passing the new state would trigger a sub restart).

Both seem easy enough with the HoF approach, but what other decorators would be good fits for our approach?

jorgebucaran:
I was trying to imagine how a debugger would look without middleware and I think it's possible; for example, a higher order function of `app` that traverses the virtual DOM tree returned by the `view` function, wrapping every on-event handler function/action. Also, the array returned by `subscriptions`,  wrapping each subscription. My point is that it should be possible to inject a debugger without changing Hyperapp or exposing a way to "hook into" `dispatch`. 

Just brainstorming.

selfup:
Only concerns about `try/catch` would be performance. Same for traversing the dom to capture events. Opening a debugger is already quite the performance hit, there must be a more elegant way.

Either way this won't super simple 😂

mrozbarry:
@jorgebucaran that's probaby a non-starter. Properly wrapping event handlers would be tough (but not impossible) to track (ie second call to view, how do I know which event handlers I've already wrapped?). I think that is also a pretty bad developer experience, and increases the points of failure.

I consider an idea of functional programming to have the least public interface surface area. If we say that to handle any middleware, we need intimate knowledge of the vdom and user subscriptions, that's almost 100% surface area. Wrapping dispatch is considerably less, as it only exposes one piece.

That said, I don't like wrapping dispatch, I preferred having a read-only interface. It would also be more performant if we didn't have a n-deep dispatch HoF.

jorgebucaran:
I went ahead and changed the title of this PR to reflect what I believe is a misconception that we continue to make every time we talk about middleware. It's my fault for not being clear every time. Let me correct it. Hyperapp will not have middleware. I'm not interested in it. It's not happening. I definitely want a better debugging experience, though! :)

> @mrozbarry How do I know which event handlers I've already wrapped?

You wrap on-event handlers on every call to the view function, of course. Traversing he VDOM is orders of magnitude faster than traversing the DOM. The debugger won't have bad performance.

I agree, the API (or lack of one) isn't great, but this is not for developers to suffer from. Hyperapp needs a better debugging experience, that's all.

jorgebucaran:
I went ahead and changed the title of this PR to reflect what I believe is a misconception that we continue to make every time we talk about middleware. It's my fault for not being clear every time. Let me correct it. Hyperapp will not have middleware. I'm not interested in it. It's not happening. I want a better debugging experience, though! :)

> @mrozbarry How do I know which event handlers I've already wrapped?

You wrap on-event handlers on every call to the view function, of course. Traversing the VDOM is orders of magnitude faster than traversing the DOM. The debugger won't have bad performance.

I agree, the API (or lack of one) isn't great, but this is not for developers to suffer. Hyperapp needs a better debugging experience, that's all.

okwolf:
I started looking at writing a helper library for making v2 middleware without any assistance from Hyperapp (because what else would I do with my time?) and I'm beginning to agree with @mrozbarry.

For starters, you would have to reimplement and bypass the entire `dispatch` function in your wrapper in order to get all the info that would be useful for a debugger including the recursive `dispatch` calls that give you updated state values, FX results, etc.

When it comes to subscriptions, you would either need some tricky way of wrapping them that somehow maintained the `isSameAction` equality check for the wrapper or would need to reimplement and bypass all the subscription logic. You would already need to do the latter if you wanted to know when subs started, restarted, and stopped so that would be the more likely path.

What you'd be left with would be essentially just the VDOM/Superfine parts and a forked implementation of all the state management code since `dispatch` is entangled deeply with all of that. There would be ongoing maintenance cost as any behavior change to the internals would likely be a breaking change.

I still plan to finish the effort so there's at least _something_ available in the absence of first party support, but it's turning out to be much uglier than I had thought.

jorgebucaran:
Great. I'd love to see that. 

> @okwolf What you'd be left with would be essentially just the VDOM/Superfine parts and a forked implementation of all the state management code since dispatch is entangled deeply with all of that. 

There's no entanglement. We can put every function in a file and it would make no difference. I'm talking about V2, though. In V1 everything is entangled, indeed.

okwolf:
Entanglement is probably the wrong term.

I'm referring to `dispatch` not being a separable concept from actions, FX, tuples, subscriptions, init, and VDOM events. Writing middleware this way requires very intimate knowledge of the implementations of all these things, and will break much any time they are refactored.

jorgebucaran:
@okwolf I think it is pretty neat how V2 wires actions to DOM events makes it possible to implement a debugger without changing Hyperapp. Unfortunately, built-in effects/subscriptions batching makes it more challenging than it should be, but I still think I have a "decent" proposal.

> Writing middleware this way requires very intimate knowledge of the implementations of all these things, and will break much any time they are refactored.

Middleware is not a Hyperapp feature, though. I _just_ want a debugger! :)

okwolf:
I've completed [a first pass](https://github.com/okwolf/hyperapp-middleware) at a minimal middleware that could support the debugger use case. I made [a couple](https://codepen.io/okwolf/pen/GLMdGM?editors=0010) of code [pens](https://codepen.io/okwolf/pen/wZpLrV?editors=0010) to show using it to build a logger for state, actions, and FX.

I'm looking forward to hearing how I could improve this approach, and if we think it might be a viable approach.

okwolf:
One unique insight that having an FX runtime gives us is knowing the full "thread" of what `dispatch`s caused everything. Imagine, if you will, that every state change comes with a stack of events, actions, and FX that caused it to happen:
Screen Shot 2019-04-24 at 19 48 00

I have two [new versions](https://codepen.io/okwolf/pen/mgGroj?editors=0010) of the same [code pens](https://codepen.io/okwolf/pen/ROqLwL?editors=0010) showing this approach.

Issue #830, opened 2019-04-23
Consolidate default and custom payload into one payload


jorgebucaran:
Let's start with a quick recap to make sure we're all on the same page.

1. An action can take a default payload.

```jsx
const Change = (state, event) => ({ ...state, value: event.target.value })

// ...

<input type="text" value={state.value} oninput={Change} />
```

2. An action can take a custom payload.

```jsx
const Remove = (state, { id }) => ({ /* ... */ })

// ...

<button onclick={[Remove, { id: item.id }]} />
```

3. Sometimes you need both a custom and the default payload. What happens to the default payload if an action takes a custom payload? The default payload shifts one position to the right!

```jsx
const UpdateItem = (state, { id }, event) => ({
  ...state,
  items: state.items.map(item =>
    item.id === id
      ? {
          ...item,
          value: event.target.value
        }
      : item
  )
})

// ...

<input
  type="text"
  value={state.value}
  oninput={[UpdateItem, { id: todo.id }]}
/>
```

Wouldn't be better if default and custom payloads were only one payload? The other day, I wanted to wrap an action and noticed I needed a conditional inside the wrapper function to determine if the action was called with or without a custom payload since that affects how the action is dispatched. The fact that the action signature changes depending on how you use it can be difficult to understand at first. Destructuring the event object inside an action is awkward too. I'd rather define actions that take a value and not think about events at all.

I found a way to solve all the above by consolidating both payloads into a one payload.

Here's the generalized action signature before and after the proposed changes.

```jsx
// Before
const anAction = (state, [customPayload,] defaultPayload) => newState
```

```jsx
// After
const anAction = (state, payload) => newState
```

Let's see how it works.

- Without a custom payload, the action takes the default payload precisely as described in case (1).
- With a custom payload, the action takes the custom payload precisely as described in case (2).

What about case (3)? How to consolidate the default and custom payload into one as promised? 

```jsx
<input
  type="text"
  value={state.value}
  oninput={[UpdateItem, event => ({ id: todo.id, value: event.target.value })]}
/>
```

Notice this allows you to create a custom payload with the `event.target.value` instead of `event`, so we can write our action this way:

```jsx
const UpdateItem = (state, { id, value }) => ({
  ...state,
  items: state.items.map(item => (item.id === id ? { ...item, value } : item))
})
```

The ability to create a custom payload can also be useful when you need to use the payload multiple times. Destructing can help, in fact:

```jsx
[...]

PR #831, opened 2019-04-24
Initial TypeScript declarations for V2 API.

petejohanson:
- Add new subscriptions API.
- Add new Effects API.
- Refactor Actions API to account for changes.
- Refactor `app` to account for changes.

Some outstanding potential items for discussion:

[ ] - I used conditional types to account for bare `SomeAction` versus tuple `[SomeAction, actionProps]` style of actions, which seems to work nicely. The only caveat that I seem to have found is the no-props version requires explicitly specifying the generic parameter to the effect constructor, e.g. `observe<void>(MyAction)` versus the inferred `observe([MyAction, someProps])` which infers the proper `observe<MyPropType>([MyAction, someProps])`. The advantage of the conditional types is you can't mistakenly omit props for an action that requires them and vice versa. I *believe* TypeScript is incorrectly inferring the `data` parameter in the non-props `ActionFunc` to be the attributes and that we've omitted the final `data` parameter. If you attemept to omit the `<void>` it infers the type to be the type from the `data` parameter.
[ ] - I renamed a few types, e.g. `ActionType` -> `Action`, to be consistent with other naming of `Effect` and `Subscription` types.
[ ] - I used `Props` for the generic argument in the `Action`, `Effect`, and `Subscription` types, since that's the language used in the various RFCs on here about the V2 APIs. Happy to change back to `Attributes` if requested.

Thanks for the work on this!

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -1,0 +1,149 @@
+export as namespace hyperapp;
+
+
+/** @namespace [VDOM] */
+
+/** The VDOM representation of an Element.
+ *
+ * @memberOf [VDOM]
+ */
+export interface VNode<Attributes = {}> {
+    nodeName: string;
+    attributes?: Attributes;
+    children: (VNode | string)[];
+    key: string | number | null;
+}
+
+/**
+ * Possibles children types
+ */
+export type Children = VNode | string | number | null
+
+/** A Component is a function that returns a custom VNode or View.
+ *
+ * @memberOf [VDOM]
+ */
+export interface Component<Attributes = {}> {
+    (attributes: Attributes, children: Children[]): VNode<Attributes>;
+}
+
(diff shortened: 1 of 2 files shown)


jorgebucaran:
@petejohanson This looks like a promising start. :tada:

By the way, the action signature is going to change a bit, see #830. The tl;dr is that actions default and custom payload arguments will be consolidated into a single argument.

petejohanson:
@jorgebucaran With the refactored Actions API, I would probably type it something like: https://github.com/petejohanson/hyperapp/blob/v2/typescript-declarations-action-ref-tweaks/hyperapp.d.ts#L81

jorgebucaran:
@petejohanson This got closed automatically when I switched back to `master` or when I pushed to delete the V2 branch. You can send your PR to `master` now! 😄

PR #837, opened 2019-05-08
Recycle only ELEMENT_NODE and TEXT_NODE

### LINKED ISSUE #825: Merge subsequent text nodes into one
Hello! When doing some **server-side rendering** experiments (using the **v2 alpha** and **hyperapp-render**) I've noticed that – on hydration – text nodes trigger an unnecessary render when containing interpolated strings. 

In my example the children become `['Title: ', 'Some Title', '!']` instead of `['Title: Some Title!']` and a new `createTextVNode()` is created for each array item. This is probably caused by bundling the `h()` pragma which results in a diff mismatch that triggers the unnecessary render:

`return Object(hyperapp__WEBPACK_IMPORTED_MODULE_0__[\"h\"])(\"h3\", null, \"Title: \", props.title,  \"!\")`

I applied a patch to the while() loop of the h() function: instead of creating a text node for each word it combines the previous and the current, if they are both text nodes. This is also like browsers render HTML by default, by combining text nodes when they are siblings. Also React seems to do something similar because the text nodes are kept together and no extra render is done during hydration.

Patched version:
```js
while (rest.length > 0) {
    var previousNode = children[children.length-1]
    if (isArray((node = rest.pop()))) {
      for (i = node.length; i-- > 0; ) rest.push(node[i])
    } else if (node === false || node === true || node == null) {
    } else if (typeof node === 'string' && previousNode && previousNode.type === 3) {
      previousNode.name += node
    } else {
      children.push(typeof node === "object" ? node : createTextVNode(node))
    }
  }
```

Current version:
```js
while (rest.length > 0) {
    if (isArray((node = rest.pop()))) {
      for (i = node.length; i-- > 0; ) rest.push(node[i])
    } else if (node === false || node === true || node == null) {
    } else {
      children.push(typeof node === "object" ? node : createTextVNode(node))
    }
  }
```

Here is an online demo with a MutationObserver listener to show that the DOM is being recreated even though the server-side and the client-side HTML are the same:
https://hyperapp-ssr-demo.glitch.me/

Please let me know if this issue is caused by the bundler (webpack) and/or the proposed patch makes sense (for which I could create a PR).

Thank you!

frenzzy:
Fixes #825

This PR allows to hydrate text nodes without DOM changes during startup.
See https://github.com/jorgebucaran/hyperapp/issues/825#issuecomment-489261749 for details.

Demo: https://codepen.io/frenzzy/pen/qGbZyB?editors=0010

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -4,7 +4,7 @@ var TEXT_NODE = 3
 var EMPTY_OBJ = {}
 var EMPTY_ARR = []
 
-var map = EMPTY_ARR.map
+var reduce = EMPTY_ARR.reduce
 var isArray = Array.isArray
 var defer = requestAnimationFrame || setTimeout
 
@@ -377,7 +377,13 @@ var recycleNode = function(node) {
     : createVNode(
         node.nodeName.toLowerCase(),
         EMPTY_OBJ,
-        map.call(node.childNodes, recycleNode),
+        reduce.call(node.childNodes, function(out, node) {
+          return out.concat(
+            node.nodeType === TEXT_NODE || node.nodeType === RECYCLED_NODE
+            ? recycleNode(node)
+            : EMPTY_ARR
+          )
+        }, EMPTY_ARR),
         node,
         null,
         RECYCLED_NODE


jorgebucaran:
@frenzzy Thanks! How can we go about benchmarking if this will be worth it or not? Positive results would help me decide whether I'm going to merge this or not.

src/index.js
```diff
@@ -377,7 +377,13 @@ var recycleNode = function(node) {
     : createVNode(
         node.nodeName.toLowerCase(),
         EMPTY_OBJ,
-        map.call(node.childNodes, recycleNode),
+        reduce.call(node.childNodes, function(out, node) {
+          return out.concat(
+            node.nodeType === TEXT_NODE || node.nodeType === RECYCLED_NODE
```
jorgebucaran: It took me a while to realize you were using `RECYCLED_NODE` to mean `ELEMENT_NODE`. Haha, no problem.

jimfranke: @jorgebucaran Is this going to be merged any time soon?

jorgebucaran: @jimfranke Not now, but I will definitely revisit again later, probably after V2 is released.

jorgebucaran:
@frenzzy Care to explain what this is for again? 😄

frenzzy:
About multiple subsequent text nodes recycling (hydration) without DOM manipulations.
See https://github.com/jorgebucaran/hyperapp/issues/825#issuecomment-489261749 for the detailed explanation.

jorgebucaran:
@frenzzy Thanks! There must be a way to measure if this can hurt performance more than improve it.

jorgebucaran:
Why should I merge here? What's the gain? @frenzzy

PR #839, opened 2019-05-14
Initial TypeScript definition for master

petejohanson:
Follow up on my original TS PR. Some initial notes on areas where eventually existential types would help with things like the new actions API.

### DIFF
--- a/hyperapp.d.ts
+++ b/hyperapp.d.ts
@@ -1,0 +1,149 @@
+export as namespace hyperapp;
+
+
+/** @namespace [VDOM] */
+
+/** The VDOM representation of an Element.
+ *
+ * @memberOf [VDOM]
+ */
+export interface VNode<Attributes = {}> {
+    nodeName: string;
+    attributes?: Attributes;
+    children: (VNode | string)[];
+    key: string | number | null;
+}
+
+/**
+ * Possibles children types
+ */
+export type Children = VNode | string | number | null
+
+/** A Component is a function that returns a custom VNode or View.
+ *
+ * @memberOf [VDOM]
+ */
+export interface Component<Attributes = {}> {
+    (attributes: Attributes, children: Children[]): VNode<Attributes>;
+}
+
+/** The soft way to create a VNode.
+ * @param name      An element name or a Component function
+ * @param attributes     Any valid HTML atributes, events, styles, and meta data
+ * @param children  The children of the VNode
+ * @returns A VNode tree.
+ *
+ * @memberOf [VDOM]
+ */
+export function h<Attributes>(
+    nodeName: Component<Attributes> | string,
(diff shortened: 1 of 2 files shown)


petejohanson:
Sample simple project using the types for things: https://gitlab.com/petejohanson/hyperworm/tree/master/src

petejohanson:
Note: I saw the effect parameters changed in the recent release. I'll work on a rebase that accounts for that.

hyperapp.d.ts
```diff
@@ -0,0 +1,149 @@
...
+ * @returns The current subscription(s) given the current state
+ * @memberOf [App]
+ */
+export type Subscriptions<State> = (state: State) => SubscriptionsResult<State>;
+
+/** The set of properties that define a Hyperapp application.
+ * @memberOf [App]
+ */
+export interface App<State> {
+    init?: (() => State) | State;
```
okwolf: The type for `init` should be the same as anything you can pass to `dispatch`, since that's what it does on app start. This definition doesn't allow for using FX in `init`.

petejohanson: @okwolf Indeed. Good catch. Will push a tweak momentarily.

petejohanson: Thinking about this more though, the initial `init` is the only one where the `state` passed into it by by `dispatch` will be null/undefined, compared to all other `Action`s, which will get the current `state` passed directly. I don't really want to loosen the type assurances on `Action<State>` just for the version in the initial `init`, so perhaps the type for init would be something like:

```
init?: InitAction<State> | State
```

where `InitAction` doesn't expect `State` as an input to the function?

hyperapp.d.ts
```diff
@@ -0,0 +1,149 @@
...
+    view: View<State>;
+    node?: Element;
+    subscriptions: Subscriptions<State>;
```
tetradice: `subscriptions` is optional. Also, as long as you read the hyperapp V2 code, `node` looks like it is required. So, isn't it correct to do this section as follows?

```ts
    node: Element;
    subscriptions?: Subscriptions<State>;
```

hyperapp.d.ts
```diff
@@ -0,0 +1,149 @@
...
+ */
+export function h<Attributes>(
+    nodeName: Component<Attributes> | string,
+    attributes?: Attributes,
+    ...children: (Children | Children[])[]
+): VNode<Attributes>
+
+/** @namespace [App] */
+
+/** An effect as the result of an ation
```
ritschwumm: typo "ation"

hyperapp.d.ts
jorgebucaran: Probably, yeah.

jorgebucaran:
Closing in favor of #969.

PR #848, opened 2019-06-05
Detached state fix (subscriptions)

sergey-shpak:
State wasn't properly set for subscriptions after [moving from render func](https://github.com/jorgebucaran/hyperapp/commit/5e2b64a4977b7b574e4ab3bc1a6eb393b934c7e6). 
When state updated and `patchSubs(subs, batch([subscriptions(newState)]), dispatch)` was running, 
`dispatch` function still used old state value (kinda detached state)
As result subscriptions with immediately dispatched actions run into `maximum call stack`

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -431,12 +431,12 @@ export var app = function(props, enhance) {
 
   var setState = function(newState) {
     if (state !== newState) {
+      state = newState
       if (subscriptions) {
         subs = patchSubs(subs, batch([subscriptions(newState)]), dispatch)
       }
       if (!lock) defer(render, (lock = true))
-    }
-    return (state = newState)
+    } else return state
   }
 
   var dispatch = (enhance ||


src/index.js
```diff
@@ -431,12 +431,12 @@ export var app = function(props, enhance) {
...
   var setState = function(newState) {
     if (state !== newState) {
+      state = newState
       if (subscriptions) {
         subs = patchSubs(subs, batch([subscriptions(newState)]), dispatch)
       }
       if (!lock) defer(render, (lock = true))
-    }
-    return (state = newState)
+    } else return state
```
frenzzy: `setState` function should always return a state?
```suggestion
    }
    return state
```

jorgebucaran:
Sorry, @sergey-shpak, I already had a fix in the works and now is out! Please upgrade to the latest beta and let me know if https://github.com/jorgebucaran/hyperapp/issues/844 is still solved.

sergey-shpak:
@jorgebucaran upgrading, thanks!

PR #849, opened 2019-06-06
Simplified deferred exec

sergey-shpak:
Simplified deferred execution

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,11 +5,7 @@ var EMPTY_OBJ = {}
 var EMPTY_ARR = []
 var map = EMPTY_ARR.map
 var isArray = Array.isArray
-var nextFrame = requestAnimationFrame || setTimeout
-var nextTask =
-  typeof Promise == "function"
-    ? Promise.resolve().then.bind(Promise.resolve())
-    : nextFrame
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -438,8 +434,8 @@ export var app = function(props, enhance) {
 
   var setState = function(newState) {
     if (state !== newState) {
-      if (subscriptions) nextTask(subscribe)
-      if (view && !lock) nextFrame(render, (lock = true))
+      if (subscriptions) subscribe((state = newState))
```
sergey-shpak: since subscriptions can immediately dispatch actions (dispatch internally has to use latest state), 
so state has to be assigned before subscriptions initialized

jorgebucaran:
@sergey-shpak Cool, thanks!

src/index.js
frenzzy: Why not
```suggestion
      state = newState
      if (subscriptions) subscribe(state)
```
Or we should not update the state if there are no subscriptions and it is fine?

jorgebucaran: Yup, that's how I refactored it after merging here. 👍

PR #850, opened 2019-06-10
Decoders

sergey-shpak:
#### Usage examples

**preventDefault:**
```javascript
import { preventDefault } from '@hyperapp/decoders'

const action = (state, event) => {
  // event is prevented
  // event.defaultPrevented => true
  return state
}

app({
  view: state => <input
    onclick={[action, preventDefault()]}
    type='button'
    value='Click!'
  />,
  node: document.body
})

```

**formData + preventDefault (composition):**

```javascript
import { preventDefault, formData } from '@hyperapp/decoders'

const submit = (state, data) => {
  // data => { text: 'value' }
  return state
}

app({
  view: state => <form 
    onsubmit=[submit, formData(preventDefault())] >
      <input type='text' name='text' value='value' />
      <input type='submit' value='submit' />
    </form>,
  node: document.body
})

```

### DIFF
--- a/lib/decoders/README.md
+++ b/lib/decoders/README.md
@@ -1,0 +1,47 @@
+# @hyperapp/decoders
+
+## Installation
+
+`npm install @hyperapp/decoders`
+
+## Usage
+
+Import decoders and use it with actions, as you usually do
+
+```javascript
+import { app, h } from 'hyperapp'
+import { preventDefault } from '@hyperapp/decoders'
+
+const action = (state, event) => {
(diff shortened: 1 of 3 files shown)


jorgebucaran:
@sergey-shpak This is very cool, but it doesn't belong to Hyperapp's core. Please consider creating a library out of this.

I do want to provide a mechanism to prevent default and stop propagation via `@hyperapp/events`, but it isn't based on the payload.

Cheers!

PR #852, opened 2019-06-14
timeout unsubscription

ravshansbox:
make timeout effect return function for unsubscribing

### DIFF
--- a/lib/time/src/index.js
+++ b/lib/time/src/index.js
@@ -5,7 +5,7 @@ var timeFx = function(fx) {
 }
 
 export var timeout = timeFx(function(dispatch, props) {
-  const id = setTimeout(function() {
+  var id = setTimeout(function() {
     dispatch(props.action)
   }, props.delay)
   return function() {
--- a/lib/time/src/index.js
+++ b/lib/time/src/index.js
@@ -5,9 +5,12 @@ var timeFx = function(fx) {
 }
 
 export var timeout = timeFx(function(dispatch, props) {
-  setTimeout(function() {
+  const id = setTimeout(function() {
     dispatch(props.action)
   }, props.delay)
+  return function() {
+    clearTimeout(id)
+  }
 })
 
 export var interval = timeFx(function(dispatch, props) {


lib/time/src/index.js
```diff
@@ -5,9 +5,12 @@ var timeFx = function(fx) {
 }
 
 export var timeout = timeFx(function(dispatch, props) {
-  setTimeout(function() {
+  const id = setTimeout(function() {
```
frenzzy:
```suggestion
  var id = setTimeout(function() {
```

ravshansbox: Ough, thanks. Forgot that we are strict es5 :)

jorgebucaran:
Hey @ravshansbox! If you need a recurring task, you should use the `interval` subscription. `timeout` is supposed to be used an effect, not a subscription, so we likely don't need this. 

I'm still curious, though, what was your line of reasoning here?

ravshansbox:
I think some cases may require to cancel the job that was timeouted, no?

mshgh:
my 2c. Let say you want to show alert which should be auto-closed in few seconds. If user decides to close it manually execution of the scheduled function can become pointless.
Sure, it can be solved by some check in the function to quit the execution immediately. So probably hard to find case when you **must** cancel the timeout. Just matter of cleaner code I'd say.

jorgebucaran:
@ravshansbox This deserves some explanation. First of all, back in vanilla JavaScript, you clear a timeout before the browser calls the callback function _so_ it won't run. 

```js
clearTimeout(setTimeout(() => console.log("You won't see this."), 1000))
```

Otherwise, you don't need to clear the timeout. Read more [here](https://stackoverflow.com/questions/32580940/is-cleartimeout-necessary-after-settimeout-with-0ms).

```js
const id = setTimeout(() => {
  console.log("You won't see this.")
  clearTimeout(id) // This doesn't make sense!
}, 1000)
```

So, what's the `timeout` effect for then? The `timeout` effect is not very useful, but you can use it as a simple way to defer calling an action.

```jsx
app({
  init: () => [
    null,
    timeout(() => console.log("Hello."), {
      delay: 1000
    })
  ]
})
```

Most of the time, you want to use an `interval` subscription instead:

```js
app({
  subscriptions: state => state.goesOn && interval(DoThatThing, { 
    delay: state.timerDelay 
  })
})
```
The interval will start when `goesOn` is true, stop when `goesOn` it becomes false, and so on. It will restart if `timerDelay` changes too.

pekeler:
I have `action_1` whose state change causes a CSS transition. At the completion of that transition plus a bit of a pause, `action_2` needs to be called. Using a timeout seems the easiest way to get this working. Now there's also a `quit` action that could be called at any time. Being able to clear the pending timeout seems much preferable to complicating `action_2` with a check that it's still in the expected state. In fact, if I'd get a quick succession of `action_1`->`quit`->`action_1`, state management would become extra complicated to prevent `action_2` from being called too often.

I guess I could implement my own timeout side effect, but this somehow seems like a not uncommon use case. What am I missing?

jorgebucaran:
@pekeler I'd use an `interval` subscription, (probably an `onAnimationRequest` instead) and manage it through the state. Now, if you can go ahead and create your own effect and show us—maybe there's some interesting pattern you can come up with.

pekeler:
@jorgebucaran an interval seems to make sense for recurring animations. Mine are one-offs, closely related to specific actions, and of different durations. My state-transition logic is currently neatly encapsulated within actions. Splitting that up between subscriptions and actions seems like it would make my program less clear - my app currently has 34 actions and more than half of them may dispatch to another action after some delay. However, I'm new to HyperApp so I'm probably not understanding some important basics yet.

@ravshansbox's proposal seemed like it should help my needs. But after spending some time trying to understand the implementation of `dispatch()` (those nested tertiaries and parameter overloading don't make it easy :), it seems to me that anything returned by side effects gets lost. Specific example with intervals:

1) an event causes `listener()` to dispatch my action
2) my action returns `[newState, intervalFx]`
3) `dispatch()` executes the intervalFX and returns interval's result (the `clearInterval()` function)
4) `listener()` ignores the returned value from `dispatch()`, i.e. no one can actually call the `clearInterval()`

So intervals are only useful for subscriptions and not for actions, unless you don't ever need to clear the interval. Is that intended?

jorgebucaran:
@pekeler Maybe it's better if you show us some code.

pekeler:
Problem:
```js
const action1 = (state) =>
  state.foo
    ? {...state, bar: 1}
    : [{...state, bar: 2}, timeout(action2, {delay: 250})]
const reset = (state) => INITIAL_STATE
```
The issue is that `action2` can still get dispatched after the user dispatched `reset`.

My current solution is to wrap `setTimeout` and `clearTimeout` in a pool which keeps track of IDs for later clearing:
```js
const newTimerPool = () => {
  let pool = []
  const timeout = (fn, duration) => pool.push(setTimeout(fn, duration))
  const clearAll = () => {
    pool.forEach(clearTimeout)
    pool = []
  }
  return {
    timeout,
    clearAll
  }
}

const timerPool = newTimerPool()
const timeFx = (fx) => (action, props) => [
  fx,
  {action: action, delay: props.delay}
]
const timeout = timeFx((dispatch, props) =>
  timerPool.timeout(() => dispatch(props.action), props.delay)
)
const clearFx = (fx) => () => [fx]
const clearTimeouts = clearFx(timerPool.clearAll)

const action1 = (state) =>
  state.foo
    ? {...state, bar: 1}
    : [{...state, bar: 2}, timeout(action2, {delay: 250})]
const reset = (state) => [INITIAL_STATE, clearTimeouts()]
```

jorgebucaran:
> @pekeler The issue is that action2 can still get dispatched after the user dispatched reset.

That's why I suggest using `interval`.

I can't emphasize this enough: `timeout` is essentially useless. It's probably there for "consistency" with JavaScript `setTimeout`, but we could (and perhaps should) get rid of it.

pekeler:
I'm not getting it. Why would I want to use an interval for a non-periodic effect?

I looked at the example and tutorial and the way they use an interval subscription for fetching new stories and moving the clock forward makes sense. But those seem like different use cases from needing to delay the dispatching of an action.

jorgebucaran:
> @pekeler But those seem like different use cases from needing to delay the dispatching of an action. 

Could you explain what you are trying to do? Why do you need to delay dispatching an action? What are you building?

zaceno:
@pekeler 

> I guess I could implement my own timeout side effect, but this somehow seems like a not uncommon use case. What am I missing?

You're correct, it is common in the general sense. But I think it will be hard to make an official solution that works well for everyone. The devil is in the detail. It depends how many outstanding timers you will have, and how you want to cancel them. Since making your own custom effects for cancellable timeouts is easy enough, might as well do that IMO.  That way you can make a solution optimized for your own use case.

pekeler:
> @jorgebucaran Could you explain what you are trying to do? Why do you need to delay dispatching an action? What are you building? 

I'm implementing a cards game. Every card is an absolutely positioned SVG. My actions are basically a finite state machine - every stage of the game is represented by one or a couple of actions. For example `determine1stPlayer`, `shuffleDeck`, `drawCard`, etc. For every stage of the game, cards have specific positions on the screen (taken care of by the `view()`).

Sometimes the state machine moves forward as a result of user actions (click events), other times as a result of the random side effect.

So that cards don't just flash up in new locations across the screen, they've been styled with a css transition. The state machine needs to wait for the completion of these transitions plus some extra time to allow the user to orient themselves before proceeding to the next stage.

For example, the `drawCard` action calls itself after a delay, each time a card is moving (animation from the CSS transition) from the deck to the hand, until 6 cards have been drawn at which point `drawCard` calls the `revealHand` action (which in turn moves the full hand to a different position, and also moves the deck out of the way).

I considered using the `transitionend` event instead of the timeout side effect to move the state machine forward. But since I want to wait for a little bit longer past the ending of each animation, just using a timeout seems to be the simplest solution.

jorgebucaran:
Thanks for clarifying what you are trying to do! 💯 

> @pekeler But since I want to wait for a little bit longer past the ending of each animation, just using a timeout seems to be the simplest solution.

And why do you need to cancel this "bit longer" delay? Is it even noticeable? You mention that there is also a quit action that can be called at any time, so I'm guessing my answer is probably there, but what say you?

pekeler:
> @jorgebucaran why do you need to cancel this "bit longer" delay?

Consider these two actions:
```js
const drawCard = (state) => {
  const nextState = {
    ...state,
    stage: 'drawCard',
    hand: [...state.hand, state.deck[0]],
    deck: state.deck.slice(1)
  }
  return nextState.hand.length < 6
    ? [nextState, timeout(drawCard, {delay: SHORT_DURATION})]
    : [nextState, timeout(revealHand, {delay: LONG_DURATION})]
}
const reset = (state) => INITIAL_STATE
```

Let's say we've drawn 3 cards so far, then the user is suddenly clicking the Reset button. If `reset` doesn't clear any outstanding timers, `drawCard` would be called 3 more times, causing weird animations.

I could make `drawCard` more robust by first checking whether the game is still going on and, if not, just return the current state (i.e. the timeout's invocation is essentially ignored). However, this check would be needed in all actions which might be invoked by a timeout. Also, if you consider that the user might quickly click Reset followed by clicking StartGame, this check becomes complicated. On the other hand, just clearing the timeouts allows me to keep my actions small and simple.

jorgebucaran:
@pekeler You may need to create your own, fine-grained delay effect. BTW, how would you propose `@hyperapp/time` handled this if it were to handle it?

pekeler:
Thanks @jorgebucaran for talking this through with me. My main intention was to find out if I'm doing something wrong or should do this differently. And it seems my solution https://github.com/jorgebucaran/hyperapp/pull/852#issuecomment-544182207 is not totally stupid.

I think that cancellation would be useful not just for timeouts but for any sideeffect that does something asynchronous, such as making HTTP requests or launching a web worker. How to do this in a consistent matter I'm not sure. Looking at how various sideeffect libs for Redux provide cancellation doesn't translate easily to hyperapp's architecture.

I'd imagine that all functions that create a sideeffect should return either a function or some identifier that can be used to later do things to this sideeffect. Spitballing:

```js
const action = (state) => {
  const nextState = {/**/}
  const effect = newEffect(completionAction, {/*params*/})
  return [{...nextState, myCurrentEffect: effect.serialized()}, effect]
}
```
and `serialized` includes all the info that can be used to look up the current side effect (for example for cancellation) and can also be used to recreate the side effect (like when restarting the app with a previously persisted state).

PR #855, opened 2019-07-03
Effect dispatch

sergey-shpak:
PR allows to compose actions which return effects
```javascript

// effectA, effectB 

const actionA = state =>
  [state, [effectA]]

const actionB = state =>
  [actionA(state), [effectB]]

```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -458,7 +458,7 @@ export var app = function(props, enhance) {
           )
         : (batch(action.slice(1)).map(function(fx) {
             fx && fx[0](dispatch, fx[1], props)
-          }, setState(action[0])),
+          }, dispatch(action[0])),
           state)
       : setState(action)
   })


jorgebucaran:
@sergey-shpak I see what you mean. This is interesting, but I think it's too premature to add now since I don't have a use case for it. Maybe in the future.

sergey-shpak:
@jorgebucaran it could be cherry on the hyperapp#2 cake :cake: :cake: :cake:

sergey-shpak:
This PR is also quite good fit for simple as possible immutable enhancer (without trying to get state from effect tuple `setState(action[0])`)
```
const freeze = obj => {
  obj && Object.getOwnPropertyNames(obj).forEach(name =>
    typeof obj[name] == 'object' && freeze(obj[name]))
  return Object.freeze(obj)
}

const immutable = dispatch => 
  (action, props, obj) => dispatch(
    typeof action == 'object' && !Array.isArray(action)
    ? freeze(action)
    : action,
    props,
    obj
  )
// and then
app({...}, immutable)
```

jorgebucaran:
I still think this is too hacky.

sergey-shpak:
@jorgebucaran it is the same as `dispatch(action(state, props), obj || props)` or `dispatch(action[0], typeof (action = action[1]) === "function" ? action(props) : action, props)`, isn't it?

jorgebucaran:
Is it? That's kind of hard to read hehe.

jorgebucaran:
@sergey-shpak Not this time! But maybe in the future if there is a need for something like it.

PR #856, opened 2019-07-03
Fix the first example

krzychukula:
I'm not sure why but the default import `import { h, app } from "https://unpkg.com/hyperapp"` doesn't work. 
I get: `SyntaxError: import not found: app`

I can fix the import by changing it to `import { h, app } from "https://unpkg.com/hyperapp?module"`
but I get a **blank page** and no error to debug it. 🤷‍♂ 

What works for me is: 
`import { h, app } from "https://unpkg.com/hyperapp@beta"`

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Don't want to set up a build step? Import Hyperapp in a `<script>` tag as a modu
 
 ```html
 <script type="module">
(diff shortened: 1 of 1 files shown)


jorgebucaran:
@krzychukula Thanks! :)

Issue #857, opened 2019-07-03
Infinite loop when subscription immediately dispatches


mrozbarry:
I have a Countdown subscription, which for simplicity, looks like this:

```js
const CountdownFx = (dispatch, { id, seconds, progress, complete }) => {
  let remaining = seconds + 1; // this is contrived to make this example smaller
  let handle = null;
  const tick = () => {
    remaining -= 1; // again, contrived
    if (remaining > 0) {
      dispatch(progress, { id, remaining });
      handle = setTimeout(tick, 1000);
    } else {
      dispatch(complete, { id });
    }
  }

  tick();

  return () => clearTimeout(handle);
}

export default props => [CountdownFx, props]
```

When I enable this countdown, it triggers setState immediately, which runs subscriptions immediately, and does an infinite loop.

If I make a timeout around the initial tick, everything works fine.

My real example makes more sense that tick happens immediately, but this example also triggers the problem.

---

The problem is [here](https://github.com/jorgebucaran/hyperapp/blob/master/src/index.js#L438).

A solution (which is maybe incorrect) is to mimick the render deferring for subscriptions, where running the next subscription patch is deferred/locked.

jorgebucaran:
We deferred subscriptions before, but it created other problems. 

/cc @zaceno @sergey-shpak 

See also:

- https://github.com/jorgebucaran/hyperapp/issues/844
- https://github.com/jorgebucaran/hyperapp/issues/844

zaceno:
> A solution (which is maybe incorrect) is to mimick the render deferring for subscriptions, where running the next subscription patch is deferred/locked.

A lock mechanism on patchSubs which preventing infinte looping, but _without_  deferring to next tick, should be possible I think, and would solve the problem for @mrozbarry as well.

jorgebucaran:
> By lock mechanism I mean a recursion guard.

I'd love to see that in practice in a PR.

mrozbarry:
> We deferred subscriptions before, but it created other problems.

Just for context, what problems did it create?

mrozbarry:
@zaceno not sure that guard will work if there is a legitiment state update that should either (re)start another subscription.

zaceno:
@mrozbarry Hmm, yeah you're right. But deferring won't help in that case either. It will just be a slightly slower-spinning infinite loop, won't it?

So if a subscription can dispatch immediately, it needs to be up to the subscription itself to prevent infinite looping, I guess. Regardless of deferral.

[...]

Issue #844, opened 2019-05-30
Hyperapp#2 deferred render issue 


sergey-shpak:
Hi, 
I'm running hyperapp#2 in context of chrome extension background script,
chrome background scripts are headless, they are running without any view,
there is no screen redrawing event at all (requestAnimationFrame exists, but never called),
as result app won't call deferred render (update state, run subscriptions, etc)

Any workaround?

jaeh:
if requestAnimationFrame is falsy,
hyperapp will fall back to setTimeout

it's years since i built an extension,
do you have access to some kind of global like window or global?
if so,
maybe setting 
```
window.requestAnimationFrame = false
```
 before importing hyperapp helps :)

alternatively maybe try:
```
window.requestAnimationFrame = window.setTimeout
```
please note that this both are to be considered pretty fragile hacks,
which might break at any time in the future.

jorgebucaran:
@jaeh That would work. Now, I'm considering changing when we evaluate subscriptions. Right we do it inside the render loop, but I think it makes more sense to do it inside `setState`. That should address your issue. 

@zaceno I think you'd prefer this too, no?

sergey-shpak:
@jaeh I've already tried the similar hack, but there is some possibility when app is bundled, that hyperapp will be defined earlier then module with `requestAnimationFrame = setTimeout` and it won't work )

jaeh:
hm, do you not have some kind of index.js file that imports everything else?

```
window.requestAnimationFrame = window.setTimeout
import { h, app } from 'hyperapp'
```
in the same file should get around this?

i personally would rather rebuild my build process to correctly keep files ordered as they were than trying to work around this, because, with my limited information, it sounds like a bug in your build. 

imho [idempotency](https://www.restapitutorial.com/lessons/idempotency.html) should be an integral part of any build chain.

sergey-shpak:
@jaeh `imports` are hoisted
```javascript
requestAnimationFrame = setTimeout
import { h, app } from 'hyperapp'
```
is bundled into something similar to (and won't work as you expect) 
```javascript
var _hyperapp = require("hyperapp");
requestAnimationFrame = setTimeout
```
If I could fix the issue with simple build process change I wouldn't ask for a workaround

sergey-shpak:
@jorgebucaran moving to setState makes sense because deferred/locked view rendering and subscriptions evaluation are different processes

sergey-shpak:
[...]

PR #858, opened 2019-07-04
Give subscriptions a deferred version of dispatch

mrozbarry:
## Summary

Subscriptions with an immediate action/state dispatch cause patchSub to go into an infinite loop.
To prevent this, I gave subscriptions a copy of dispatch that defers for 0ms (next tick).

## Possible issues

 - Prevents dispatch from returning the current state in subscriptions
 - May introduce unexpected behaviour in subscriptions
 - May conflict with #844 ?

## Tickets

 - Issue #857

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -64,6 +64,7 @@ var shouldRestart = function(a, b) {
 }
 
 var patchSubs = function(oldSubs, newSubs, dispatch) {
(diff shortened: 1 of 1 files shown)


jorgebucaran:
https://github.com/jorgebucaran/hyperapp/issues/857#issuecomment-508904420 👋

PR #864, opened 2019-07-29
Propagate prevented event

sergey-shpak:
Propagate prevented event to make decoders properly work
```javascript
onsubmit: preventDefault([onsubmit, formData]) // or any other decoder
```


lib/events/src/index.js
```diff
@@ -65,7 +65,7 @@ export var targetValue = function(event) {
 export var eventOptions = fx(function(dispatch, props) {
   if (props.preventDefault) props.event.preventDefault()
   if (props.stopPropagation) props.event.stopPropagation()
-  if (props.action != undefined) dispatch(props.action)
+  if (props.action != undefined) dispatch([props.action, props.event])
```
jorgebucaran: @sergey-shpak Why not `[props.action, props]`?

sergey-shpak:
this won't work, because `preventDefault([onsubmit, formData])` leads to `dispatch([onsubmit, formData], props)` which is not correct and fails with error. 
My bad, have to find another way to combine event effects and decoders

jorgebucaran:
@sergey-shpak It would help if you post an example of what you tried to do. Even better: a pen/repro of the problem you had.

sergey-shpak:
now that should work and event is propagated to decoders )
```javascript
const formData = (event) => {
  const form = new FormData(event.currentTarget)
  return Array.from(form).reduce((obj, [key, value]) => ({ ...obj, [key]: value }), {})
}

const submit = (state, data) => {
  /* submitting form data */
}

h('form', {
  onsubmit: preventDefault([submit, formData])
}, [ /* inputs */ ])
```

PR #866, opened 2019-08-02
fallback 'requestAnimationFrame'  to 'Timeout'

sergey-shpak:
When running hyperapp#2 at env that doesn't provide 'requestAnimationFrame' function, 
you could expect that `var defer = requestAnimationFrame || setTimeout` will fallback 'defer' to timeout. But it's not, instead you get 'Reference Error' (by trying to get **undeclared** 'requestAnimationFrame').
This PR makes proper 'requestAnimationFrame' check and 'setTimeout' fallback

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,7 +5,9 @@ var EMPTY_OBJ = {}
 var EMPTY_ARR = []
 var map = EMPTY_ARR.map
 var isArray = Array.isArray
-var defer = requestAnimationFrame || setTimeout
+var defer = typeof requestAnimationFrame !== 'undefined' 
+  ? requestAnimationFrame 
+  : setTimeout
 
 var createClass = function(obj) {
   var out = ""


jorgebucaran:
> @sergey-shpak When running hyperapp#2 at env that doesn't provide 'requestAnimationFrame' function...

What's an environment that doesn't provide `requestAnimationFrame`? If you're using JSDOM, there's an option [pretendToBeVisual](https://github.com/jsdom/jsdom#pretending-to-be-a-visual-browser) we can use to mock `request|cancelAnimationFrame`.

sergey-shpak:
@jorgebucaran nodejs doesn't provide `requestAnimationFrame`, also 
I was working on chrome extension, and it's background script doesn't have any kind of attached DOM, so I was using hyperapp without view layer (headless, only subscriptions and actions, so I didn't need any DOM), and since effects/subs are deferred proper `requestAnimationFrame` is required

jorgebucaran:
@sergey-shpak Would it be possible to mock/patch `requestAnimationFrame` in this context? 

```js
global.requestAnimationFrame = setTimeout

// or

global.requestAnimationFrame = _ => _
```

sergey-shpak:
@jorgebucaran no, it's not possible
```
window.requestAnimationFrame = setTimeout
import { app } from 'hyperapp'
// app initialization
```
but `import` is always hoisted as result real execution order is:
```
import { app } from 'hyperapp'
window.requestAnimationFrame = setTimeout
// app initialization
``` 
and it fails because hyperapp defer tries to get `requestAnimationFrame` which is undefined at the moment of importing hyperapp

jorgebucaran:
@sergey-shpak How does one mock a global then?

jorgebucaran:
@sergey-shpak Maybe we can do this:

```jsx
var defer = window.requestAnimationFrame || setTimeout
```

or is `window` not available inside a chrome extension env either?

jorgebucaran:
Here's a alternative (proposed by @zaceno):

```js
// Other polyfills, globalThis, etc.

globalThis.requestAnimationFrame =
  globalThis.requestAnimationFrame || setTimeout
```

Then let's assume we put that in a module called "hyperapp-compat":

```js
import "hyperapp-compat"

// Succes
```

sergey-shpak:
@jorgebucaran, latest example with additional module import to define `requestAnimationFrame` can possibly solve the issue, even though it looks little bit 'hacky' I'm ok with this). I believe 'requestAnimationFrame' can be sacrificed for something more greater

sergey-shpak:
@jorgebucaran just curious, what are the benefits of
```js
var defer = requestAnimationFrame || setTimeout
```
comparing to 
```js
var defer = typeof requestAnimationFrame !== 'undefined' 
  ? requestAnimationFrame 
  : setTimeout
```

frenzzy:
@sergey-shpak in case when `requestAnimationFrame` is not available:
```js
var defer = requestAnimationFrame || setTimeout
// => Uncaught ReferenceError: requestAnimationFrame is not defined
```
```js
var defer = typeof requestAnimationFrame !== 'undefined' 
  ? requestAnimationFrame 
  : setTimeout
// => ok
```
This one makes no sense, since [globalThis](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#Browser_compatibility) is supported in less envs then [raf](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame#Browser_compatibility) does.
```js
globalThis.requestAnimationFrame =
  globalThis.requestAnimationFrame || setTimeout
```
Userland polyfill is the better choice IMO:
```js
import "request-animation-frame-polyfill"
```

sergey-shpak:
@frenzzy yes I know, I created this PR to fix `requestAnimationFrame is not defined` and this is my piece of code:
```js
var defer = typeof requestAnimationFrame !== 'undefined' 
  ? requestAnimationFrame 
  : setTimeout
```
so I know how things work, my point is - if we want to roll back to `requestAnimationFrame || setTimeout` or use `globalThis` - what are benefits?

PR #867, opened 2019-08-04
Proposal for improvement of type definition (based on "Initial TypeScript definition for master")

tetradice:
A few days ago, I saw the PR #839 "Initial TypeScript definition for master" and I thought that there are some potential improvements in the TypeScript definition (`hyperapp.d.ts`) included in it. So I put together a proposal for improvement and send a new pull request.

(This PR is a fork from #839)

### Summary

1. ~~Add `a` and `an` functions (Details are explained below)~~
2. Add typing support of JSX event handler attributes
    - e.g.) `onclick`, `onchange`, ...
3. Add typing support with default payload
4. Modify type definition of `init` in arguments of `app` by reviews in the PR #839
5. Fix typo by reviews in the PR #839
6. Solve TODO in the PR #839  
7. Rename and separate some types
    - `ActionFunc `-> **`Action`, `ActionWithEffects`** 
        - I thought the short name is better because `Action` is likely to be used from application code.
        - `Action` must return only new state. `ActionWithEffects` can return new state with effects. `ActionWithEffects` has default payload type information, but `Action` doesn't.
    - `Action` -> `Dispatchable`
    - `App` -> `AppProps`
8. ~~Add `PureAction` type (it is a function that does not include effects in the return value)~~
9. Merge latest master branch
10. Add libraries type definition -- `@hyperapp/time`, `@hyperapp/events`, `@hyperapp/dom` and `@hyperapp/random`

~~You can try it online [here](https://codesandbox.io/s/hyperapp-v2-type-definition-improvement-minimal-bnzln?fontsize=14&module=%2Findex.tsx). (on CodeSandbox)~~
You can also see the difference with PR #839 from the commit 90c99e98.

### ~~What is  `a` and `an` functions?~~

(this section are deleted)


jorgebucaran:
@tetradice Thank you so much for working on this. Do you think we can move `a`, `an` to userland? I don't feel comfortable exporting these extra functions, but I'm open to exploring all the possibilities.

tetradice:
@jorgebucaran 
OK. 👌 No problem.

I'm thinking of splitting it so that it can be imported as shown below. How about that?

```sh
npm install --save-dev @hyperapp/type
```

```ts
import { a, an } from "@hyperapp/type";
```

Another plan:

```ts
import { a, an } from "hyperapp-type";
import { a, an } from "hyperapp-ts";
```

jorgebucaran:
@tetradice "Another plan" looks good. 😁 
Sidenote: I recommend exporting a single function rather than two. Some suggestions: `$`, `_`, `run`, `call`, `action`.

Here's a fun thing we can do with JSX to take advantage of transpilation.

```js
const Run = props => props.with ? [props.action, props.with] : props.action
```


```jsx
<input onchange={<Run action={Increment} with={{ value: 100 }} />}></input>
```

tetradice:
@jorgebucaran Thank you reply. Then, I will proceed according to another plan. 😁

By the way, if another plan is better, is it better to separate from the official repository of hyperapp? Should `hyperapp-type` be in `tetradice/hyperapp-type` instead of `jorgebucaran/hyperapp`?

> Sidenote: I recommend exporting a single function rather than two. Some suggestions: $, _, run, call, action.

Excuse me, could you tell me why it is better to have one function? (I'm not a JavaScript professional, so I'm sorry if this is common sense for JavaScript professionals )

I'm trying to name the package `hyperapp-type`, but the `a` and `an` functions doesn't seem to be the only feature in the package. These functions are part of the package and someone may add other functions to `hyperapp-type`. 

So I think it's better not to have only one default export. (This is similar to `@hyperapp/time`) Please let me know if there are any other advantages or aspects that you have missed.

> Here's a fun thing we can do with JSX to take advantage of transpilation.

Unfortunately, I found that using JSX like that in a TypeScript environment would be difficult. :cry:

As far as I can see, in the TypeScript world, JSX user components can only return one type. Therefore, it is not possible to define that `<TodoList />` returns a VNode and `<Run />` returns an Action.

jorgebucaran:
@tetradice What's the difference between `a` and `an` functions?

tetradice:
@jorgebucaran these functions are same. (`an` is alias of `a`) The reason for these two is that they resemble English indefinite articles.

However, I am not particular about these names. For example, I think that there is no problem in changing this into a single function called `run`.

```ts
import { run } from "hyperapp-type";
```

jorgebucaran:
Including type declarations in this repo is okay, but exporting a new function like `a/an` isn't.

> However, I am not particular about these names. For example, I think that there is no problem in changing this into a single function called run.

Yup, that's what I meant. Rather than `a` or `an`, a single export like `run`, or `t`, or `x` is better IMO.

I don't want to expose this function in core and be responsible for maintaining it. That's why I suggest a `hyperapp-*` naming scheme. Examples:

- hyperapp-x
- hyperapp-t
- hyperapp-type

frenzzy:
I do not understand why are you trying to introduce new `a`/`and`/`run` function which makes no sense for me? Isn't better to just provide proper typings? Something like:
```js
interface DOMAttributes<...> {
  onClick?: Dispatchable<...>
}
```

See [@types/react](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts) for inspiration.

tetradice:
@jorgebucaran Thanks. I am going to proceed the work as follows based on your opinion. Please let me know if there is anything different from your intention.

- Remove the `a`, `an` functions from the pull request. (remove from `index.js` and `hyperapp.d.ts`)
- The removed function will be published as `tetradice/hyperapp-xxx` on my GitHub repository and npm.

@frenzzy Thank you reply. The reason I proposed functions like `a` and `an` is that strict typing would not be possible without function calls. In other words, using a function call instead of a tuple has two advantages:

1. A build error occurs when an unmatched payload is passed.
2. Input completion is enabled for payload.

For example:

```tsx
type State = { count: number };

const Increment = (state: State, payload: {value: number}) => ({
    ...state,
    value: state.count + payload.value
});

<input onchange={[Increment, { value: 'invalid' }]}></input>;
```

The `<input>` above should cause a build error. Because Increment expects a number as payload, but it passes a string. However, as far as I know, the current TypeScript JSX support cannot make this code an error. The reason is that TypeScript can infer types by Generics only when calling a function, and JSX attributes cannot infer what the `Increment` requires.

By using function calls, type inference is enabled and the above example can also be a build error. Also, when using Visual Studio Code, input completion of payload is also enabled.

If there were a way to be as strict without a function call, this function would be completely unnecessary.

tetradice:
Supplement on type inference:

```ts
interface DOMAttributes<State, Payload> {
  onClick?: Dispatchable<State, Payload>
}

declare global {
    namespace JSX {
        interface Element extends VNode<any> {}
        interface IntrinsicElements {
            [elemName: string]: DOMAttributes<State, Payload>;
        }
    }
}
```

If written as above, `[elemName: string]: DOMAttributes <State, Payload>;` will raise an error. Because `State` and` Payload` are undefined at this point.

tetradice:
@jorgebucaran  I completed the work and made some improvements.

1. Remove `a` and` an` functions
2. Add libraries type definition -- `@hyperapp/time`, `@hyperapp/events`, `@hyperapp/dom` and `@hyperapp/random`
3. Improve and fix definitions
4. Merge latest master branch -- no conflicts, all ready for merging!

If it is merged, I would also like to send a new PR, including documentation and sample code for TypeScript.

tetradice:
I forgot to add support for Lazy and middleware, so I added it with some fixes.

hyperapp.d.ts
```diff
@@ -0,0 +1,212 @@
...
+ */
+export interface View<State> {
+    (state: State): VNode<object> | null;
+}
+
+/** The possible response types for the subscription callback for an application
+ * 
+ * @memberOf [App]
+ */
+export type SubscriptionsResult<State> = | (Subscription<State> | boolean)[] | Subscription<State>;
```
diontools: I think `subscriptions` can returns a boolean without array.
Currently, the following code is compilable, but it is best to specify it in the type definition.

```tsx
subscriptions: state => state.enabled && subs()
```

hyperapp.d.ts
```diff
@@ -0,0 +1,212 @@
...
+ * @memberOf [App]
+ */
+export type Subscriptions<State> = (state: State) => SubscriptionsResult<State>;
+
+
+/** The lazy view. {@link https://github.com/jorgebucaran/hyperapp/issues/721#issuecomment-402150041}
+ *
+ * @memberOf [App]
+ */
+export function Lazy<P extends object>(props: { view: (props: P) => VNode<object>, key?: string | number | null } & { [key in keyof P]: P[key] }): VNode<object>;
```
diontools: `{ [key in keyof P]: P[key] }` can be replaced with `P`.

hyperapp.d.ts
```diff
@@ -0,0 +1,212 @@
...
+ * @memberOf [App]
+ */
+export type Dispatch<State, NextPayload> = (obj: Dispatchable<State, NextPayload>, data: NextPayload) => void;
+
+
+/** An effect runner. It is actually invoked when effect is reflected.
+ * 
+ * @memberOf [App]
+ */
+export type EffectFunc<State, Props, AcceptDPayload = any, NextPayload = void> = (dispatch: Dispatch<State, NextPayload>, props: Props, payload: AcceptDPayload) => void;
```
diontools: I think effect runner has only 2 arguments. (subscription runner too)
https://github.com/jorgebucaran/hyperapp/blob/master/src/index.js#L463
https://github.com/jorgebucaran/hyperapp/blob/master/src/index.js#L85

tetradice:
@diontools Thank you for the review! I corrected them. If there is no problem, please resolve conversations.

diontools:
@tetradice Thank you for fix. It's no problem.
By the way, how do you define the effect in user-side TypeScript?

tetradice:
@diontools Under the current type definition, write:

```ts
const TimeoutRunner: EffectFunc<any, {delay: number, action: Dispatchable<any>}> = (dispatch, props) => {
    setTimeout(() => {
        dispatch(props.action);
    }, props.delay);
}

function timeout<S>(action: Dispatchable<S>, props: {delay: number}): Effect<S> {
    return [TimeoutRunner, {action: action, delay: props.delay}];
}

const IntervalRunner: SubscriptionFunc<any, {delay: number, action: Dispatchable<any, number>}, number> = (dispatch, props) => {
    let handle = setInterval(() => {
        dispatch(props.action, Date.now());
    }, props.delay);

    return () => clearInterval(handle);
}

function interval<S>(action: Dispatchable<S, number>, props: {delay: number}): Subscription<S> {
    return [IntervalRunner, { action: action, delay: props.delay }];
}
```

(This is a rewrite of `@hyperapp/time`)

This definition is mostly inherited from the PR #839. However, as I wrote the example above, I noticed that there were a few points to worry about ...

- `EffectFunc` and `SubscriptionFunc` should be renamed to `EffectRunner` and `SubscriptionRunner`.
- The above code is long and difficult to read. I might try to make it a little shorter. (Especially runner functions)

tetradice:
Supplement: another example

```ts
const TimeoutRunner = (dispatch: Dispatch<any, void>, props: { delay: number, action: Dispatchable<any> }) => {
    setTimeout(() => {
        dispatch(props.action);
    }, props.delay);
}

const IntervalRunner = (dispatch: Dispatch<any, number>, props: { delay: number, action: Dispatchable<any, number> }) => {
    let handle = setInterval(() => {
        dispatch(props.action, Date.now());
    }, props.delay);

    return () => clearInterval(handle);
}
```

tetradice:
> EffectFunc and SubscriptionFunc should be renamed to EffectRunner and SubscriptionRunner

I did this.

diontools:
@tetradice Thank you for comments. I like to separate Effect Creator and Effect Runner. 👍

I tried to write it in the original-like Higher-order function. However, Effect and Subscription cannot be handled by a single function.
This code works on [TypeScript 3.4](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#higher-order-type-inference-from-generic-functions).

```ts
// Add Payload/Props type argument to Effect/Subscription for more type-safe
type Effect<State = any, Payload = any, Props = any> = [EffectRunner<State, Payload, Props>, Props]
type Subscription<State = any, Payload = any, Props = any> = [SubscriptionRunner<State, Payload, Props>, Props]


type TimeProps<S, P = void> = { action: Dispatchable<S, P>, delay: number }

function timeFxForEffect<S, P>(fx: EffectRunner<S, P, TimeProps<S>>): (action: Dispatchable<S>, props: { delay: number }) => Effect<S, P, TimeProps<S>> {
    return (action, props) => [fx, { action: action, delay: props.delay }]
}

const timeout = timeFxForEffect(<S>(dispatch: Dispatch<S>, props: TimeProps<S>) => {
    setTimeout(() => dispatch(props.action), props.delay)
})


function timeFxForSubs<S, P>(fx: SubscriptionRunner<S, P, TimeProps<S, P>>): (action: Dispatchable<S, P>, props: { delay: number }) => Subscription<S, P, TimeProps<S, P>> {
    return (action, props) => [fx, { action: action, delay: props.delay }]
}

const interval = timeFxForSubs(<S>(dispatch: Dispatch<S, number>, props: TimeProps<S, number>) => {
    let id = setInterval(() => dispatch(props.action, Date.now()), props.delay)
    return () => clearInterval(id)
})
```

tetradice:
@diontools Thanks to comment!

> I tried to write it in the original-like Higher-order function. However, Effect and Subscription cannot be handled by a single function.

I've confirmed that this is possible with overload. (The return value is also correctly typed)

```ts
function timeFx<S, P>(fx: SubscriptionRunner<S, P, TimeProps<S, P>>): ((action: Dispatchable<S, P>, props: { delay: number }) => Subscription<S, P, TimeProps<S, P>>)
function timeFx<S, P>(fx: EffectRunner<S, P, TimeProps<S, P>>): ((action: Dispatchable<S, P>, props: { delay: number }) => Effect<S, P, TimeProps<S, P>>)
function timeFx<S, P>(fx: SubscriptionRunner<S, P, TimeProps<S, P>> | EffectRunner<S, P, TimeProps<S, P>> ) {
    return (action: Dispatchable<S, P>, props: TimeProps<S, P>) => [fx, { action: action, delay: props.delay }]
}

const timeout = timeFx(<S>(dispatch: Dispatch<S>, props: TimeProps<S>) => {
    setTimeout(() => dispatch(props.action), props.delay);
});

const interval = timeFx(<S>(dispatch: Dispatch<S, number>, props: TimeProps<S, number>) => {
    let id = setInterval(() => dispatch(props.action, Date.now()), props.delay)
    return () => clearInterval(id)
})
```

----

The topic is off, but I got an important inspiration from your comments.

I saw the [@hyperapp/time](https://gist.github.com/tetradice/86bd7d009199661f2a6002b1d8622b25) code and your comment. And I wanted to make the function like `timeFx` (receive Runner function and return Effect Constructor) more general and simple to use.

And I got there. I can make __Props Creator__ work like Payload Creator.


```js
// pure JS
const TimeoutRunner = (dispatch, props) => {
    setTimeout(() => dispatch(props.action), props.delay)
}
const timeoutPropsCreator = (action, props) => ({ action, ...props });

const timeout = fxConstructor(TimeoutRunner, timeoutPropsCreator);
// Created Function: timeout(action, props)
```

```ts
// TypeScript
type TimeProps<S> = { action: Dispatchable<S>, delay: number };

const TimeoutRunner = (dispatch: Dispatch<any>, props: TimeProps<any>) => {
    setTimeout(() => dispatch(props.action), props.delay)
}
const timeoutPropsCreator = (action: Dispatchable<any>, props: { delay: number }) => ({ action, ...props });

const timeout = fxConstructor(TimeoutRunner, timeoutPropsCreator);
// Created Function: timeout(action: Dispatchable<any>, props: { delay: number })
```


`fxConstructor` is a temporary name 🙄 

Implementing this function will allow Hyperapp users to easily create functions such as `timeout()`, `interval()` or `preventDefault()` themselves. I would like to send this as pull request or publish it as my npm package, so please let me know if you have any opinions.

([full test code](https://gist.github.com/tetradice/86bd7d009199661f2a6002b1d8622b25) is on gist) 
(If there is something difficult to understand, send a message in Japanese on Twitter.)

tetradice:
I'm sorry. The above description was deleted because it was found to be meaningless. 😅

r0skar:
Sorry to bump the thread, but since I very much prefer working with TS I was wondering if there is anything we can help with to get this PR merged?

jorgebucaran:
@r0skar You could review the PR -- that's one thing you could totally do. 💯

tetradice:
@r0skar Thank you! I would be happy if you could review.

I just pushed the latest updates, so I would like you to review them based on the latest commits. However, only the type definition of `@hyperapp/http` is not yet included. (Because it was added very recently)

jorgebucaran:
Thank you to everyone who started the original complot to add correct types to Hyperapp! ❤️ 

This PR is now superseded by https://github.com/jorgebucaran/hyperapp/pull/1016. 💁‍♂️

PR #869, opened 2019-08-06
Move non VDOM code to 2nd half of file

SteveALee:
Purely cosmetic but I think it's good to have the VDOM code in the 1st half and the framework code in the second half.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,10 +5,6 @@ var EMPTY_OBJ = {}
 var EMPTY_ARR = []
 var map = EMPTY_ARR.map
 var isArray = Array.isArray
-var defer =
-  typeof requestAnimationFrame !== "undefined"
-    ? requestAnimationFrame
-    : setTimeout
 
 var createClass = function(obj) {
   var out = ""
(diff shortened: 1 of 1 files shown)


jorgebucaran:
@SteveALee Believe it or not, this can affect the bundle size. I'll check if it hurts (or not) and decide whether to merge here or not then. Thank you! 👋

SteveALee:
@jorgebucaran wow, really? Well, the order of dependencies have not changed so even a buggy tree shaking should be ok.

Mind you, hyperapps contribution to bundle size is minimal.

Fyi I’m using a local copy and browser modules as am assuming modern browsers.

Also FYI I notice you don’t really support global script for no build as all the functions clutter up the global namespace. No IEFE to hide them. I guess that would mess with bundle size too.

Happy to raise issues if you care about these.

jorgebucaran:
> @SteveALee wow, really? 

Yeah, it's crazy. But maybe this is fine. I'll check soon.

> Also FYI I notice you don’t really support global script for no build as all the functions clutter up the global namespace. No IEFE to hide them. I guess that would mess with bundle size too.

```html
<script src=https://unpkg.com/hyperapp/dist/hyperapp.js>
  const { h, app } = hyperapp
</script>
```

SteveALee:
@jorgebucaran haha - of course! :)

sesam:
The topic of this PR sounded interesting. Like as it could have load-time effects in what order code is placed in the file. But I guess the compression also has to be considered.

jorgebucaran:
I actually like VDOM first, then state management. The state is the part closer to the user, and most bugs are probably there, not in the patch algorithm. It's also been a long time and this has become seriously outdated. 😅

SteveALee:
A nice reminder to explore the latest version.

Issue #868, opened 2019-08-04
onAnimationFrame subscription can't be stopped


zaceno:
Brought this up on the slack but since no quick conclusion was reached, I'm moving here for tracking.

Simply put: it seems `onAnimationFrame` as implemented in @hyperapp/events@0.0.1 cannot be stopped.  (using hyperapp@2.0.0-beta.25)

See this example:

https://codepen.io/zaceno/pen/mNqbJK

A suspicion is that it is somehow due to timing/interference with other use of rAF in core hyperapp.

Another clue might be gleaned from the fact that this alternative implementation (not using `cancelAnimationFrame` or keeping references) seems to work flawlessly -- at least in the above example:

```
const onAnimationFrame = (f => a => [f, a])((dispatch, action) => {
    let noop = () => {}
    let nextFrame = () =>
        requestAnimationFrame(time => {
            dispatch(action, time)
            nextFrame()
        })
    nextFrame()
    return () => {
        nextFrame = noop
    }    
})

```
(to wit: https://codepen.io/zaceno/pen/ymPBKJ)

---

Edit: ping also @mrozbarry

zaceno:
@jorgebucaran 
Good, quick work! I'm surprised that simply switching the order of dispatch & rAF made such a difference. Would you mind explaining why that worked?

Other thing: Why 2.0.1? Did you make changes in core as well? (I couldn't find any)

jorgebucaran:
It's a "pre-release". I released it so that the docs matched the examples. Nothing discourages you more than official examples that don't work.

As to why rAF works now, I'm not sure, but it makes more sense to grab the next frame before dispatching an action that can change the state.

PR #870, opened 2019-08-08
make patchSubs async

zaceno:
Calling patchSubs directly in setState without deferring makes the `onAnimationFrame` subscription behave in strange ways. For example, this codepen doesn't work: https://codepen.io/zaceno/pen/vzMbLQ?editors=0010 Also, issue #868 is probably related. 

This PR moves `patchSubs`  back to where it used to be, inside `render` and thus makes it async. Merging this would make the codepen above work.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -439,9 +439,6 @@ export var app = function(props) {
   var setState = function(newState) {
     if (state !== newState) {
       state = newState
-      if (subscriptions) {
-        subs = patchSubs(subs, batch([subscriptions(state)]), dispatch)
-      }
       if (view && !lock) defer(render, (lock = true))
     }
     return state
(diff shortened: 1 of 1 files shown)


zaceno:
Sorry, I just now realized my editor added a bunch of formatting junk to the diff. Completely unintentional! I will fix it ASAP (tomorrow ... it’s bed-time now :) )

zaceno:
I fixed PR now so it just moves the lines I intended. I don’t know what, if anything, I can do to make the codecov/patch check pass though.

zaceno:
making patchSubs async only appears to fix the problem. The subs are started and restarted still, due to patchSubs expecting subscription props to be objects. Hence: closing this PR

PR #872, opened 2019-08-12
Fix events subs

zaceno:
`patchSubs` expects subscriptions to be of the form `[subFn, {...props...}]` because it wants to merge props-objects

In lib/events/src/index.js, subscriptions were being generated as `[subFn, actionFn]`, which caused behavior where subscriptions were being continually restarted. 

This PR fixes those subscriptions to generate `[subFn, {action: actionFn}]` (and for corresponding subscription functions to expect an object, not an action, as second argument).

At this time, as far as I can tell, only subscriptions in events need this treatement.

### DIFF
--- a/lib/events/src/index.js
+++ b/lib/events/src/index.js
@@ -1,15 +1,15 @@
-var fx = function(a) {
-  return function(b) {
-    return [a, b]
+var sub = function(fn) {
+  return function(action) {
+    return [fn, {action: action}]
   }
 }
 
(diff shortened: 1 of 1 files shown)


jorgebucaran:
Thanks, @zaceno! 👍  

I'll make some cosmetic fixes and release a new version soon.

zaceno:
🎉 Super! Look forward to it :)

PR #873, opened 2019-08-15
Make app return kill func

zaceno:
Main idea: make it possible for apps to die like things usually do: the corpse (the DOM nodes) are still there, but do nothing. Subscriptions are shut off as well.

usage:

```
const kill = app({...})
setTimeout(kill, 5000) //the app just stops doing anything after five seconds
```

There was a thought to return an object, where `kill` would be the only prop, as a way of future-proofing (in case we want to add other things to app's return in the future, without breaking existing code).

I think it's a good idea but could use some discussion, so for this initial submission I decided to just return the kill function, sidestepping the naming discussion and keeping the bundle-size addition minimal.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -481,4 +481,8 @@ export var app = function(props) {
   }
 
   dispatch(props.init)
+
+  return function () {
+    patchSubs(subs, EMPTY_ARR, dispatch = Function)
+  }
 }


src/index.js
```diff
@@ -481,4 +481,9 @@ export var app = function(props) {
   }
 
   dispatch(props.init)
+
+  return function () {
+    dispatch = function () {}
+    patchSubs(subs, [])
+  }
```
frenzzy: probably it also should remove event listeners from DOM nodes
and probably it also should remove created DOM nodes
don't you think?

zaceno: I don't think kill should remove created dom-nodes because it's easy enough to do in userland, if you want -- and because there are cases where you *don't* want the created nodes removed. 

As for removing event listeners, `dispatch = function () {}` is effectively equivalent. We don't remove any event listeners, but all event handlers become no-ops. I don't think it would be worthwhile to explictly remove event listeners, as long as interacting with the dead DOM can't *do* anything.

zaceno:
Added `node=null` to make sure we don't keep the detatched DOM in memory longer than needed. But we return it from the kill function so post-processing can be done if needed.

zaceno:
Actually, let’s forget about those last two commits. Sorry! I’ll revert them first thing tomorrow. Short story: memory is only kept for as long as reference to kill function is kept. So no need to null the node.

src/index.js
jorgebucaran: I don't think we also need to return the `node`, could you get rid of this? If the user wants the node, they can add an id and use `getElementById`.

zaceno: @jorgebucaran agreed, and done

jorgebucaran: @frenzzy @zaceno Do you think this could work? 😄 

```js
patchSubs(subs, EMPTY_ARR, dispatch = Function)
```

zaceno: @jorgebucaran Yes I don't see why not! 🏌

jorgebucaran:
@zaceno Could you elaborate on how and why this work?

Also, can we always cut subscriptions off like this or would we benefit by having something like:

```js
const Goodbye = (state) => ({ ...state, isRunning: false })

const kill = app({
  node,
  init,
  done: Goodbye,
  subscriptions: (state) => [
    state.isRunning && [
      ping,
      {
        host,
        gateway,
        timeout,
        onRequestResponse,
      },
    ],
  ],
})
```

Finally, what does `dispatch = Function` do?

zaceno:
@jorgebucaran 
> @zaceno Could you elaborate on how and why this works?

Sure!

Ok, so what the point of this is, is to basically make an app stop _doing anything_ and to make it stop being interactive. We do want to leave the nodes as they are, because we might want to use them for something (like perhaps resurrecting the app).

If we want to remove the nodes that's easy enough in userland. The stuff this PR does is not doable in userland. 

First, we want to make sure that we can't interact with the elements in the app. Rather than rerendering or removing all the event handlers we simply set `dispatch=Function`, effectively making dispatch a no-op. Click buttons all you want and nothing will happen.

Furthermore this makes sure that effects and subscriptions that get the idea to dispatch something after the app has been killed, won't cause runtime errors.

But subscriptions also typically hold on to their event-listeners. If you are starting and stopping apps a lot like in a customElement scenario, that's a memory leak. So we also want to make sure to unsubscribe everuthing (even though nothing can be dispatched).

That's what `patchSubs(subs, EMPTY_ARR)` does.

As for the `done` prop, I don't know. I don't really see why that would be needed. Maybe you can explain?

zaceno:
Oh wait I see. You want to do something when the app stops? In principle, shouldn't that be an effect (=output) rather than a subscription (=input) ? At least if you're doing it from inside the app. If you're doing it from outside, you could do it with vanilla js

```js




```

jorgebucaran:
How important is it we `dispatch=Function` and could we get away with this entirely in userland?

Here are some ideas. No changes to core, but we need an id to identify so that my `end` and `kill` sub/fx pair can communicate:

```js
import { end, kill } from "./fx"
import { initialState } from "./state"

const Done = (state) => ({ ...state, isRunning: false })

app({
  init: (defaultProps = {}) => ({ ...defaultProps, ...initialState }),
  subscriptions: (state) => [
    [end, { id: 666, Done }],
    state.isRunning &&
      [
        /*...*/
      ],
  ],
})

// Moments later...
kill(666)
```

Or with minimal changes to core, `app` could return this "app `id`" that we also want to pass to `init` via props:

```js
const { end, kill } = "./fx"
import { initialState } from "./state"

const Done = (state) => ({ ...state, isRunning: false })

const id = app({
  init: (defaultProps = {}, id) => ({ ...defaultProps, ...initialState, id }),
  subscriptions: (state) => [
    [end, { id, Done }],
    state.isRunning &&
      [
        /*...*/
      ],
  ],
})

// Moments later...

kill(id)
```

Or even simpler, but more risky, just return `dispatch` from the `app`:

```js
import { initialState } from "./state"

const Done = (state) => ({ ...state, isRunning: false })

const dispatch = app({
  init: (defaultProps = {}) => ({ ...defaultProps, ...initialState }),
  subscriptions: (state) => [
    [end, { Done }],
    state.isRunning &&
      [
        /*...*/
      ],
  ],
})

// Moments later...

dispatch(Done)
```

zaceno:
> How important is it we dispatch=Function and could we get away without it?

dispatch=Function Is the “cause of death” that is how we effectively kill the app. Even any outstanding fetch-calls that come back with data will not cause any “death spasms” with that.

Your approach doing it all in userland all requires the app to work through one last render which doesn’t work in the scenario of customElements (which is what I was working on when I came up with this). There we can’t let the app render one final time because the node has been removed from the dom and weirdness ensues. That’s why dispatch=Function. To make absolutely sure nothing tries to modify the DOM once the kill function is called.

zaceno:
Can I bump this for consideration, since there aren't any plans (afaik?) to return anything from `app()`?

Being able to kill an app cleanly and immediately is useful when using a multi-app architecture, and when implementing custom elements.

Yes it is possible to use middleware + HOA emulate this _but_ it requires an extra render (since we can't get att the `patchSubs` function from the ouside). That makes any DIY solutions less elegant and potentially bug-inducing. Also it breaks the custom element use case since the node will be gone when we want to kill the app.

Those things might be considered edge cases but whatever:  I feel that Hyperapp is incomplete if it can't clean up after itself without relying on the DOM, basically.

jorgebucaran:
@zaceno Do you have a link to a code example that would rely on a `kill` function, it could be that custom elements pen you created at some point. I know I've seen it, but I can't find it anywhere now.

zaceno:
@jorgebucaran sure! That custom element pen isn't working anyway anymore, but I stripped down some code I was working on at the moment, to some examples that demonstrate the problems with stopping apps currently. Right now I'm working on an update to my sevenguis examples, and I wanted to make it like an app gallery (also as a proof of concept of how to do that in general). 

First have a look at [this example][app gallery - no stopping]. It is a simple gallery which lets you run a counter or a clock. The clock has a subscription to `every`, and on each tick it console logs 'TICK'. Notice that each time you switch over to the clock, it will start logging a new series of 'TICKS' – more and more, never stopping.

Clearly we need the ability to stop an app when we switch a way from it. So in [this example][app gallery - stopping with extra render errors] I use a custom `app` function to run the apps that returns a stop function (best I can do from the outside). This does indeed make the "TICK"ing stop.

_But_ now I am getting errors each time I switch (`null is not an object (evaluating 'state.count')`) – because it is trying to render one extra time with the "stopped" state – which does not match what the view expects.

I attempt to fix that in [this third example][app gallery - unsuccessful attempt at empty render] by making my custom app function handle the stopped state, and render a single, empty `div` instead of the declared view. Somehow that causes inscrutable runtime errors in hyperapp.

I can only conclude that properly stopping an app "from the outside" is not feasible, we need hyperapp to support a way to do that without causing an extra render.


[app gallery - no stopping]: https://flems.io/#0=N4Igxg9gdgzhA2BTEAucD4EMAONEBMQAaEAMwEskZUBtUKTAW2TQDoALAF0fmPSk6IBqECAC+RekxboArgMQAnVgCtqJSAuFpyjbBEWcABMBzYxR0ooiMjAci6dcKAPQv52ANYBzVpEYu7ACe2EpmdgA6ULr6hkZcPJbWtg6cTjCu7lBevv6BIUrw5JyRUFGaMMYAklBgiojMAkYAvEaVmIItAHxGABTArIPtgkRGkPKcKG2cHYh+EBNGANRGAIxiAJTl0JVGACKIdQ1Cxq3DiN19A0MzI2MLAlPn84sAtGubUVGIAB6xxvhEKRMLJ4MYoBBAZczP0okYjBDAUQ4UZyNFJiZxo8jAAGCQogBu5EQAHcnrcLs0egl4AADFHwgA8+HIBK6DPhRkZ7FWXQAJMBnljOGJGYFeRymQAjWRpaBGaBgIpgTzNAUHI6NEVdV5imVyqDsqCczmM-WceWK5WqgU1TUnMRdJZ62UWw0csUstm05FQTZ8PBIMCccg7EQAJhQOPEkhADGYIiVEBVqnU-EE2hAMQMxlM2HMSRs9kczjcHh88wCwVCinCUWzcRphZSJYyZeyFby1cKxVKDdziAJSiCFisRdS6Uy5dyNhcAAFu7X8y4Q8xSttYMZSAZGB0ACq6SlGVeUnq9KCk-azXonjYbVgWg-MADKnEUaO8vXvMFkUsqil6HFRgADi2WodmMZ9EE4J8j16c5RlvS4KgQOZ4AgT87D3KoAGEAGk7A2IwAB9iIAAv6QZWAQ49D0+MoLz+HMjEBYFQXBSFTyMGFgBRRFEF9eE0WKKZgBPKY9lmVgIRJL98WNIwiVJclZkuGlaW5XlYKmAVt0UXcYMPeCKQfQ8NlFcUuh9FEfz-OpyGwEMdhUzoqSMGgjEHYdenDHEgKMKDDOYDYAF1fX9EhA0OJzYBEVZI3DGMpATNg1D4TQM04ER+xMMxR2SYs0lLLIckrfIazraI9GYpsxxbIq2xKztZ0XIoSnraq4mFJRm3sVgXG65Q1D7TrjCTFVersfrxs8VN13AzcjG8aCAEF8wAOU4lo+iItz8GTWQtVYKVISCVgAEdZGHKCgwtAC7F3NEjB6AAqFAKEUSpXjAdhKHwQivgW3YACV5G24zZlGMxdp6GgaJoL9oXzXpls4NbsE2wEvw2UZgDEEKQsBni+M4qZ9rAQ6TmO07BNRdFRPk+ElLJIwAH01O4OkPRO-AgiNE0mWwfmBfhAAZCBMHwbioClxQwcwY0zCmSVTXNS1amtNVgBoUGoFGQaQsdIwcIeQRFC5Fw1fdBSRbNV11aVcgVS1nX5H19CVUNnocI9zwLat4WTTFIWVe5c2XED01HsNZlWSMCOxWjyO9VOqzwo2dKbGwSglFQKBQXgSLEFu0NYrQABWFBVgATiSuNpBEPwYDTDKThEcRDaAA

[app gallery - stopping with extra render errors]: https://flems.io/#0=N4IgzgpgNhDGAuEAmIBcICGAHLA6AVmCADQgBmAljEagNqgB2GAthGiLgBbzNQkiwA9g0Qj2IAL7FGLNumx5C-ISIhj0FZlkEAneAAJgC-RjD6A+gon6yOwc30BybvCxhUAeg8BXBlgDWAOa4QswenACeWBA6Co4AOgya2nr63Lw2dg7O8K7uXr4BwaHhUTFQFPAJDIleJjj6AO6xODH68IL6OhDw3joM+mAdWDa+CBTCtR64M-q+8FT6ABJlLSMARt5USGYUA1h20Tr6-lRQALRkYwuTDCpD9SMAvPoAFIJY8GAAlPpPAHyGRLxeAwAxDD7RJB-GwYKCQYGgnr6JAUMBYDDwWCcRGWHCvYCIkEzXAfL7EIlfbzrMCwHQUT4TBjuQbwTEQP6A16UkFDdn6ACETxeDG8UCg+gAZJL9GSwLgwNTafTGcIwDz4AB+WWfeWKml0hk3ZmvPmIb4a1D6WgAXUp3wpNXgIOYFCQSBgjQw3St0IBb1R6Mx2JhrxJ3sCP05QKdIJBFDIbwhrSQv26vX6GvTfQGSDDMwjP0pEgtTpLiOz-Tev39wH0lMDGKxnFeovFpbjX2GUJh8B03ggiIkiWHNTHEAAHikDEYcMQ0jwoBJ+JAYOM1ewAIwABlQ28k0hATFY7CE8xiBCIpBUong7GSuhnCnn6SXmXsTlwHgUl+qiXuBgAJJ3N0rAiDCZocv6BIkpB85niIVqQSEgjzPoADU+ibuWdxqgYAAicCgWoBgvJB0YwTMcH6Ah8BIWyiAoWh5xYThiSTtOKIQGQGBigYDCCEgUGAgoBKJPo+gCUJjoSXslRWsAtFWtuUjifoABuFAQI09H8v6r4AAZqRJAA8qLqf8xkSfoJmcJu-wACTAMhtESCZ4T2VZpmbLkwiyncFSwP4TxOYRdIQGB8ASP85zuT5HQMJZAzWd53i+QMwiwIFwVOcB4WRdF6FxWlCVJSl7nmf8BmOiWK7QHAxo0CAO6oAATAAzAeMgnugWWCEFl7KMIt73loj6GM+C68NYtgfo4X4-oQ1QPqkwAQOpMQRDNWROC4bieD4fhBChYQAAKREcCgeAsrB-rhzIGGQujMJiAAqmhQe0H0UQw2n6Ph7KvDdEDfN8uAdO9rAAMp9nsgSvGD+pDDorzbvOAAcpYAfoUM9JDn2mgxEDzsDNaAvcggwLgUCCPDjivYBADCADSji-AAPuzAAFlEKkTJMfWxv1TuNQk8XxkmCcJjxiclUnE2pcl0YYwNWgDjECY0COqclmnabpiDRoZtn2fjVpOU9OgvfA+OE+y4Mfd8bkeVVMmDEqhqqsyBvS7Q+jrZtrytduaM43jjs2jV3x1WujXsAAnKgm6dVI3VyBwSjXsNJGjZxs5YNtc17fkh1FCdpSXTgy1jakr7vtkxcHYUx0lBd5SVNXnG0W0s3ZF+3c6L+iQrQYfVBfXn4eGP-hD2O2OBD0ACCOAAHJS6GZMov13iRbg6yCREuAAI4DjoES47HuivI4L17PogIAFSoJQOhDOc2LbGzwL3Q8ABKvihmogoTetBqK0FeI2YMnBN6EmSshZMUoZTwOGAjNSkDmyvFxvAGGHx5yiQXvAZeWA15CQRqDEc85gASBtHaOeeEw7YOGIA-mrIPib15tRZMQtRKwIkvLX029d77yQBEN2SsFLJitG2Jcbs9Y6QsEbRcRlkreQPmVFKNksDqI0foAAMoIDA0IMAMGhP2AYxjHhWi8tZEy8U-KZWyiFYAtB-4MHgqhVQOgbTRX0IzDxiBjjFXStojRtiSr2IChQIKTiXG+HgjTIK3jASMwSf4GyHg7GJWse5LR2TODHA8CE0yt9EpmQoOpfQhT3IlJCXFNR1URzRyzloKgMQ0DSNIKuBqTImqbk3HuLqR5ZDsBCGAK8Ahs7qBAJIbxQA

[app gallery - unsuccessful attempt at empty render]: https://flems.io/#0=N4IgzgpgNhDGAuEAmIBcICGAHLA6AVmCADQgBmAljEagNqgB2GAthGiLgBbzNQkiwA9g0Qj2IAL7FGLNumx5C-ISIhj0FZlkEAneAAJgC-RjD6A+gon6yOwc30BybvCxhUAeg8BXBlgDWAOa4QswenACeWBA6Co4AOgya2nr63Lw2dg7O8K7uXr4BwaHhUTFQFPAJDIleJjj6AO6xODH68IL6OhDw3joM+mAdWDa+CBTCtR64M-q+8FT6ABJlLSMARt5USGYUA1h20Tr6-lRQALRkYwuTDCpD9SMAvPoAFIJY8GAAlPpPAHyGRLxeAwAxDD7RJB-GwYKCQYGgnr6JAUMBYDDwWCcRGWHCvYCIkEzXAfL7ERL6Kn6ABuFAgjVQg3gmIgf0BQ1Z+gAhE8XgxvFAoPoAPz6MlgXB0hmvTmIX5M9JQAAGAB5UTT9B5-sqKTV4CCwN51mBYDoKJ8Jgx3MyuQC3kSDV8WYgeXz9AKhfoAGTe8WfSVGk1mi03a2OkFiiW4IOm82W4RgWUuiDfCPwJm0AC6ju+epBIOYFCQSBgjQw3SZ0Ptr1R6Mx2JhrxJFcCP3ZQP1BYWZDeENaSF+3V6-XTw76AyQzZmrZ+jokaf1C8R4-6b1+9uA+kddYxWM4r09UEX3f7UJh8B03ggiIkiTvNUfEAAHikDEYcMQ0jwoBJ+JAYHGRN2AARhA1AACYAHZJGkEAmFYdghHmGICCIUgVFEeB2GSXR3wUL8lWsWx7CcXAPAUNDqkSe4DAASTubpWBEGE5TZGtgBJNiv2QkQmTYkJBHmfQAGp9BA5c7kTAwABE4CYtQDBeNiOwJLiUx4oS+NtRBBOE85xMkxIXzfFEIDIDBBQMBhBCQdjAQUAlKQ9WyID1Kk9kqJlgF4jN9AABikZzpUZHT7O-XhlWcql1QoGl-mi6lVU4ED-gAEmAATfIkVVwlSxKYs2XJhHFO4KlgfwngyuSzQgZj4Akf5zlyoqOgYBKBmpQrvGKgZhFgcrKoyhjavqxqRJanq2o6rrco1HU9QXf9oDgMMaBAcD-POECABZINgmREPQAbBAqtDlGELCcK0PDDAIiLf0yUjHHIyjCGqXDUmACAaRiCJiKyJwXDcTwfD8IJBLCAABSIjgUDwFlYaipOtAwyF0ZhMQAFU0dj2lx1SGAZfQZNZV5EdTb5cA6HHWAAZUvPZAleKnY0vV5-K-AAORdaP0OmelpvHk1ZL8KY3QF7kEGBcCgQRmccLG6IAYQAaUcX4AB9NYAArUmZuPx1gjKJ19brsiyrJcuyO0cwlOpsuz3P0Ty-OACmmVJ3SbMaFmgs6kL+JTDslTVFL-iFpkMvRnRMfgIWRd08WcryhbnNjEME2tIO7UBWh9B+v7Xgg-zOf5wXce+LNFu+ZbALW9goNQGCpEOuQOCUDDLsU67TI-LAAee4H8jBopIdKOGcA+m7UiVJ7smH0HCghkpYfKSpp9M3y2hI7JyO3nQqMST6DBOir57Ijwz-8I-Hz5wIegAQRwAA5VymwllFTu8ercHWWyIi4AAI7Xh0BEAW9ddCvEcJjPY+hAQACpUCUB0EMc42Jtga2BCjB4AAlXwTZDYKE-rQQ2tBaxoj3NiT+9sqQCX7D6P09Dhgs2cruBsB4BbwAZh8L8jkH7wGflgN+dkWbfEXFIQwEgsw5jvtJcu3DhiEI0syD4n99YxhUf2E2dtnKOwgFWb+v9-5IAiM7V23l+xMiPP7KkgcLAhx-FFTqhUAEzS6voVUWB3EeP0AAGUEBgaEGAGDQivAMEJjwmQFSSq1Eq-VBpVWALQfBDBNIoR0FmRq+hlZaUQMcSavUfEeNVHEvqZUKAVSSSk3wPE5YVSyYCZW9T-CeI8GU4pMUPDeJiclY42pemwParFTU2pcpDOKS1Nxup7y1y7loKgMQ0BHlIABVaVp1pgVQP5A68FZDsBCGAdCAhu7qBAJILJQA

jorgebucaran:
A variation of this proposal would be to modify `dispatch` so that `dispatch()` works like your kill function. 

```diff
+    : action !== null 
+      ? patchSubs(subs, EMPTY_ARR, (dispatch = id))
       : setState(action)
)
```

I've always felt like we should probably return `dispatch` or at least something more useful than `undefined`. This was always on my radar. We could achieve that and still have a way to terminate apps at once.

With `dispatch` in hand, you have absolute flexibility over the termination process, which may be useful. Heck, there may be a situation where you want subscriptions to run one more time before terminating the app. You may want to do one more thing, before passing on.

Userland `dispatch` would be a dev-feature we could also use for tooling, automation testing, integrating tests, etc.

zaceno:
CLosing in favor of #1018

PR #879, opened 2019-09-09
Example of named effect for debugger

### LINKED ISSUE #878: Ensure effects have names
For the hyperapp debugger I'm developing, I want to track any effect (including subs) and display them in a list. Right now, `@hyperapp/time` (and maybe others in the future) use unnamed `function`s, and it would be great if they had meaningful names, like `interval` or `timeout`.

Currently, @okwolf has been doing that in [hyperapp-fx#Interval](https://github.com/okwolf/hyperapp-fx/blob/HAv2/src/subs/Time.js) so I can get the name of the effect, but that's not the current behaviour in `@hyperapp/time`.

mrozbarry:
Names the effect functions so a debugger can display the effect names.

Closes #878

### DIFF
--- a/lib/time/src/index.js
+++ b/lib/time/src/index.js
@@ -4,13 +4,13 @@ var timeFx = function(fx) {
   }
 }
 
-export var timeout = timeFx(function(dispatch, props) {
+export var timeout = timeFx(function timeout(dispatch, props) {
   setTimeout(function() {
     dispatch(props.action)
   }, props.delay)
 })
 
-export var interval = timeFx(function(dispatch, props) {
+export var interval = timeFx(function interval(dispatch, props) {
   var id = setInterval(function() {
     dispatch(props.action, Date.now())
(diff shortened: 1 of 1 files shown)


jorgebucaran:
@mrozbarry We could rename timeoutFx/intervalFx to timeout/interval right?

mrozbarry:
@jorgebucaran _probably_, as long as we use `window.timeout`/`window.interval` in the implementation. Never tried overriding those.

mrozbarry:
Done

jorgebucaran:
@mrozbarry I think this is probably okay. Would you prefer a different name/capitalization or how do you plan to use the name in your debugger? Are you going to display it as is, append something to it, etc.?

mrozbarry:
@jorgebucaran I'm not picky on the capitalization. Typically, I've made all of my actions and effects class-case, but I can transform the text into whatever in the debug code.

PR #880, opened 2019-09-10
Propose less-cute README example

brandonros:
I think the example in the README is pretty stupid because in the real world, nobody is going to have just an integer as their state. They are most likely going to have an object.

I feel this proposal, while kind of dumb (I guess you could get rid of the `increments` and `decrements` properties) makes it much more clear for a beginning user what the heck is going on.

Also, I am kind of surprised we are cool with loading the scripts in `<head>`. I thought they belong in `<body>` as to not block the browser rendering?

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,12 +40,12 @@ Here's the first example to get you started: a counter that can go up or down. Y
       import { h, app } from "https://unpkg.com/hyperapp"
 
       app({
-        init: 0,
+        init: { total: 0, increments: 0, decrements: 0 },
         view: state =>
           h("div", {}, [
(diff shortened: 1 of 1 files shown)


jorgebucaran:
> I think the example in the README is pretty stupid because in the real world.

I agree that the example is not incredibly enlightening, but our goal is to get you started quickly, not teach you everything you need to know about Hyperapp. I think of it as an appetizer; too much of it, and you risk ruining the rest of the meal. 

>  I am kind of surprised we are cool with loading the scripts in <head>

Module scripts behave like `defer` by default. That means it doesn't block the HTML parser.

Issue #877, opened 2019-08-26
API for fetch-effect


zaceno:
In my humble opinion, there needs to be an official fetch effect (or http, or ajax or whatever) for the final release of Hyperapp 2. Otherwise, it will be the first thing people ask for, and telling them to make their own would be inconsistent with the message that "you should not need to write your own effects most of the time".

The question is only what the API should be. I opened this issue to have that discussion, as it isn't as easy as one might think at first.

My first suggestion would be something like this:

Like:

```js
Fetch({
  url: ... 
    //  the URL in a native fetch(url) call, mandatory
  
  options: {...}, 
  //the options in a fetch(url, options) call. 
  // Optional, default: undefined or {},
  
  onResponse: ... ,
  // fetch(...).then(resp => dispatch(props.onResponse, resp)
  // Optional, default: no dispatch

  onFailure: ...,
  //fetch(...).catch(err => dispatch(props.onFailure, err)
  // Optional, default: no dispatch

  bodyType: ...
  //only relevant if using onBody callback
  // can be 'text', 'json', 'blob', 'formData', 'arrayBuffer'
  // Optional, default 'text',

  onBody: ...
  // fetch(...)
  //    .then(response => response[props.bodyType]())
  //    .then(body => dispatch(props.onBody, body)
  // Optional, default: no dispatch

  onErrorParsingBody: ...
  // fetch(...)
  //    .then(response => response[props.bodyType]())
  //    .catch(error => dispatch(props.onErrorParsingBody, error)
  // Optional. Default: no dispatch
})
```

The thing that concerns me with this, is that `fetch` will resolve for any response from the remote. Including 404's et c. That means you'll get `onBody` callbacks even when there is no body, or the body is not of the `bodyType` you specified (in which case `onErrorParsingBody` will be dispatched).

So I was thinking, we could limit the body-parsing routine to be done only for 200 OK responses, and only if the request was using the GET method. But perhaps that's to restrictive?

*If* we limit body parsing to GET requests with 200 OK responses, then we'll need another option: `onNotOK` (maybe?) for the case when we got a response but decided not to parse the body. Or is that getting too complicated?

zaceno:
Clearly the complicating issue here is that what we most often want is the body (most often parsed JSON). And that requires a two-step promise-chain, with some logic in between.

A more "close-to-native" approach would be to have simply:

```
Http.fetch({
  url: ...,
  options: {...},
  onResolve: GotResponse,
[...]

PR #881, opened 2019-09-16
Documentation: add tutorial 

zaceno:
Please proofread and suggest better wordings.

There are still some technical things not quite done but I think it's ok to fix them later:

- The css file and json files for the "api" are currently hosted on my personal github-pages, because I'm not sure how this tutorial will be hosted later (and probably couldn't put the data there)
- The screenshots are just named `tut1.png`, `tut2.png` in the `docs` folder. We might want them in a folder.
- For the `Http` effect I'm importing from @okwolf 's `hyperapp-fx`. After a discussion in #877 I concluded (and was not opposed) that an official hyperapp effect for http should (at least initially) be very similar. When the official effect is out, we can edit this.
- I have empty links to "API Reference" and "Examples" because I wasn't sure how to link to them.


zaceno:
Quick link to read the tutorial https://github.com/jorgebucaran/hyperapp/blob/6acf5ba7dcbec4c78fb3f95c7a81966d5a768dab/docs/tutorial.md

docs/tutorial.md
```diff
@@ -1,5 +1,910 @@
...
+})
+    </script>
+  </head> 
+  <body>
+    <div id="app"></div>
+  </body>
+</html>
+```
+
+> The sections demarked by comments are not important. They're
```
mrozbarry: I've never heard of `demarked` - but a similar word would be `denoted`. Looks like demarked is a word, just not super common in Canada. Your call.

docs/tutorial.md
```diff
@@ -1,5 +1,910 @@
...
+  <i>World!</i>
+</h1>
+```
+
+### Rendering to the DOM
+
+`app` is the function that runs our app. It is called with a single argument - an object
+which can take several properties. For now we're just concerned with `view` and `node. `
+
+Hyperapp calls the `view` function to find out how we want the DOM. We express it in the form of virtual nodes.
```
mrozbarry: > ...to find out how we want the DOM

The phrasing seems a little weird, and may not translate well to non-native English speakers.

What about:

> ..., which describes what DOM elements we want to render.

docs/tutorial.md
```diff
@@ -1,5 +1,910 @@
...
+State
+-------------------------------
+
+With all that view logic broken out in separate functions, `view` is starting to look like
+plain _data_. The next step is to fully separate data from the view.
+
+Add an `init` property to your app, with this pure data:
+
+```js
+  init: {
```
mrozbarry: I'm curious if you want to change this to `init: () => ({ ...stateStuffHere })` so it's closer to an Action. I think it's fine to use a plain object, but I think we've discussed that behaviour before as a "happy accident," so I'm not certain we want to advertise it.

mrozbarry: I think the pay-off here will be when a user wants to fire an effect immediately, that will be easier to understand if `init` is an action, rather than describing this one-off direct call to dispatch.

mrozbarry: The downside is actions are the next section.

docs/tutorial.md
```diff
@@ -1,5 +1,910 @@
...
+```
+
+This will dispatch the `SetFilter` action everytime someone types in the input. Implement the action like this:
+
+```js
+const SetFilter = (state, event) => ({...state, filter: event.target.value})
+```
+
+The second argument to an action is known as the _payload_. Actions
+dispatched in response to an events on DOM elements receive the [event object](https://developer.mozilla.org/en-US/docs/Web/API/Event) for a payload. `event.target` refers to the input element in the DOM, and
```
mrozbarry: As I recently re-discovered, `event.currentTarget` might be the right thing more often than not.

mrozbarry:
Did a brief skim over, I'll read over it again later, because I haven't had enough ☕️ coffee yet.

docs/tutorial.md
zaceno: I actually think "demarked" is more appropriate than "denoted" here, because I'm using comments to create a separating boundary between different sections. (https://www.lexico.com/en/definition/demarcate). 

But if it's a strange term, we could perhaps say: "The sections outlined by comment headings ..."?

docs/tutorial.md
zaceno: Yeah and also, if I understand @jorgebucaran correctly, he doesn't want us to think of `init` as an action even though it works like one. (I could be wrong)

docs/tutorial.md
zaceno: AFAIK `target` and `currentTarget` are equivalent in this case because the handler is bound to the element on which the event we're interested in will occur. Assuming that is true, I prefer `event.target` because it's short. 

I think we'll have to assume users know the difference between target and currentTarget (we can't teach javascript in the tutorial), but perhaps we should make it more clear that the event is bound to the element on which we declare `onClick`. That is perhaps not obvious to someone coming from React where (afaik) that's not how things work.

docs/tutorial.md
mrozbarry: Yeah, that's fair.

zaceno:
Proofreading is great and definitely needed, but it struck me: it would be good with some feedback on how helpful (or not) the tutorial is. Does it do the job of introducing a newcomer to Hyperapp. I guess it’s a long shot, but if anyone knows any newcomers who would be willing to try the tutorial, their feedback would be quite valuable!

zaceno:
Tweaked some minor things after @mrozbarry 's comments (thanks!) New link for reading:

https://github.com/jorgebucaran/hyperapp/blob/c10aa036fd5359b6d5f6aec291cfd34b153dd7df/docs/tutorial.md

docs/tutorial.md
```diff
@@ -1,5 +1,909 @@
...
+and implement the `ToggleAutoUpdate` action:
+
+```js
+const ToggleAutoUpdate = state => ({...state, autoUpdate: !state.autoUpdate})
+```
+
+Now we've got `autoUpdate` in the state tracking the checkbox. All we need now, is to set up `FetchStories`
+to be dispatched every five seconds when `autoUpdate` is `true`.
+
+Imoport the `interval` _subscription creator_:
```
skanne: *Import

docs/tutorial.md
```diff
@@ -1,5 +1,909 @@
...
+> The section structure outlined in the comments is not important. It's
+> just a suggestion for how to organize the code we'll be
+> adding throughout the tutorial.
+
+Open it in a browser, and you'll be greeted with an optimistic **Hello _World!_**.
+
+View
+------------------------------------
+
+Lets step through what just happened.
```
skanne: *Let's

docs/tutorial.md
```diff
@@ -1,5 +1,909 @@
...
+  "Auto update: ",
+  h("input", {type: "checkbox"})
+])
+
+const Container = content => h("div", {class: "container"}, content)
+
+```
+
+
+With those, the view can be written as:
```
skanne: no comma

docs/tutorial.md
```diff
@@ -1,5 +1,909 @@
...
+This makes Hyperapp bind a click-event handler on the button element, so
+that when the button is clicked, an action named `StartEditingFilter` is
+_dispatched_. Create the action in the "ACTIONS" section:
+
+```js
+const StartEditingFilter = state => ({...state, editingFilter: true})
+```
+
+Actions are just functions describing transformations of the state.
+This one keep everything the same except that it sets `editingFilter` to `true`.
```
skanne: *This action basically keeps the same state, except for `editingFilter` which it sets to `true`.

docs/tutorial.md
```diff
@@ -1,5 +1,909 @@
...
+  
+  props.editingFilter                               // <---
+  ? h("input", {type: "text", value: props.filter}) // <---
+  : h("span", {class: "filter-word"}, props.filter),
+  
+  h("button", { onClick: StartEditingFilter }, "\u270E")
+])
+```
+
+Now when you click the pencil button the text input appears.  But we still need to add
```
skanne: *Now, when you...

docs/tutorial.md
```diff
@@ -1,5 +1,909 @@
...
+be _effects_. Effects are executed by Hyperapp as part of processing the action's return value.
+
+> Hyperapp provides effect creators for many common situations. If you've got an unusual case or are working
+> with less common APIs you may need to implement your own effects. Don't worry - it's easy! See the 
+> [API reference]() for more information.
+
+### Effects can dispatch actions
+
+One of the options we passed to `Http` was `action: GotStories`. The way this effect works is that when the response comes
+back from the api, an action named `GotStories` (yet to be implemented) will be dispatched, with the response-body as the payload.
```
skanne: *response body 
(without hyphen, see next sentence)

docs/tutorial.md
```diff
@@ -1,5 +1,909 @@
...
+      author: props.stories[id].author,
+      unread: !props.stories[id].seen,
+      reading: props.reading === id,
+      filter: props.filter
+    })
+  ))
+])
+```
+
+When the app loads, and when you change filter, you should see the spinner appear until the stories are loaded.
```
skanne: *change the filter

zaceno:
Thanks @skanne for the proofreading! 

New link for reading: https://github.com/jorgebucaran/hyperapp/blob/e47a0399a6ae4083972df8726898aebf9f2d05f0/docs/tutorial.md

jorgebucaran:
This is absolutely awesome! Thanks. ❤️

zaceno:
Sweet! 🎉 I’m honored 🙇‍♂️

skanne:
Kudos go to Zacharias!

Von meinem iPhone gesendet

Am 02.10.2019 um 18:59 schrieb Zacharias Enochsson <notifications@github.com>:

﻿

Sweet! 🎉 I’m honored 🙇‍♂️

PR #882, opened 2019-09-30
Allow chaining payload filters

zaceno:
### What this is: 

Today when you dispatch `([FooAction, 'baz'], 'bar')` it resolves to `dispatch(FooAction, 'baz')`, and when you dispatch `([FooAction, fn], 'bar')` it resolves to `dispatch(FooAction, fn('bar'))`. 

However, that _only_ works if `FooAction` is a function. If `FooAction` is an array, you get an error thrown at you. 

This teensy fix allows you to stack multiple payload processors on top of eachother, so that 

`([[[FooAction, fn1], fn2], fn3], 'bar')`

will resolve first to `dispatch([[FooAction, fn1], fn2], fn3('bar'))`

then to `dispatch([FooAction, fn1], fn2(fn3('bar')))`

and then (as it works today) `dispatch(FooAction, fn1(fn2(fn3('bar'))))`

### Motivation:

Wherever an action is asked for in the view, effects or subscriptions, you can provide an action bound to a payload or payload filter by `[someaction, payload]`. It is elegant if the action _in_ the tuple can itself be substituted with a tuple for further processing of the payload.

But when do you _need_ this? Why not just combine the payload filters all at once? It comes up when you want to compose views from intentionally generic and reusable components.

Lets say you have some fancy widget for adjusting a value:

```jsx
const SetFoo = (state, foo) => ({...state, foo})
...
const FooSliderStrip = ({state}) => (
  <div class="slider-strip" ontouchstart={[
    SetFoo,
    event => event.touches[0].clientX
  ]}>/* something with state.foo */</div>
```

You like it so much you want to use it for all your adjustable things in your app, so you turn it into a generic and reusable widget. Now, you can't assume the action is `SetFoo` anymore. You need to take the action as a prop.

```jsx
const SliderStrip = props => (
  <div class="slider-strip" ontouchstart={[
    props.onSet,
    event => event.touches[0].clientX
  ]} > /* something with props.value */ </div>

const FooSliderStrip = ({state}) => <SliderStrip onSet={SetFoo} value={state.foo} />
```

Now besides foo, you want to use the slider to adjust the `.red`, `.green`, and `.blue` values in the `.color` object in the state. Rather than making identical actions `SetColorRed`, `SetColorBlue` et c, you just make one that takes the name as a parameter. 

```js
const SetColorDim = (state, {name, value}) => ({...state, color: {...state.color, [name]: value}})
```

So naturally, you make your color picker like this:

```
const ColorPicker = ({state}} => (
  <div class="color-picker">
    <SliderStrip onSet={[SetColorDim, value=> ({name: 'red', value})]} value={state.color.red} />
    <SliderStrip onSet={[SetColorDim, value=> ({name: 'blue', value})]} value={state.color.blue} />
    <SliderStrip onSet={[SetColorDim, value=> ({name: 'green', value})]} value={state.color.green} />
 </div>
)
```

Very natural and straightforward. Except it won't work, because the Sliders will be trying to dispatch actions like this:
`[[SetColorDim, value => ({name: 'red', value})], event => event.touches[0].clientX]`

which breaks today.

With this fix, the behavior will be the expected, namely to dispatch an action with composed payload: 

`(SetColorDim, {name: 'red', value: event.touches[0].clientX})`

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -454,7 +454,7 @@ export var app = function(props) {
     return typeof action === "function"
       ? dispatch(action(state, props))
       : isArray(action)
-      ? typeof action[0] === "function"
+      ? typeof action[0] === "function" || isArray(action[0])
         ? dispatch(
             action[0],
             typeof action[1] === "function" ? action[1](props) : action[1]


zaceno:
From the slack chats, and thumbs-up here, it seems this change is untypically uncontroversial.
I don't mean to rush things of course, but if there's nothing else we're waiting for, I could use this change in a project I'm working on. So... merge?

jorgebucaran:
Yes! 🎉

zaceno:
Hooray! And thank _you_!

jorgebucaran:
Here's a minimal example to help people (like me) visualize this change quickly. Study the code below to see how payload functions can now stack, then read @zaceno's [issue](https://github.com/jorgebucaran/hyperapp/pull/882#issue-322741587) again for a real use case and understand why we added this. 

Click the button to see the payload function call order.

```js
const one = x => `${x}→1`
const two = x => `${x}→2`
const three = x => `${x}→3/`

const Grow = (state, x) => state + x

const subView = action => h("button", { onClick: [action, one] }, "Click")
const mainView = action => subView([action, two])

app({
  init: "State→",
  view: state => h("div", {}, h("h1", {}, state), mainView([Grow, three])),
})
```
https://codesandbox.io/s/hyperapp-payload-stack-iv537

sergey-shpak:
@jorgebucaran looks like this PR also allows nice action helpers/wrappers:
```javascript
import { targetValue } from '@hyperapp/events'

const Send = (state, msg) => 
  [state, http({ /* some props */})]

const onEnterKey = action => (state, e) =>
  e.key === 'Enter' ? [action, e] : state

// ... a few minutes later
<input onkeypress={ onEnterKey([Send, targetValue]) }  />
```
@zaceno am I right?

zaceno:
@sergey-shpak , yeah I think you are right!

PR #885, opened 2019-10-08
Fix issue #884

### LINKED ISSUE #884: Adding Lazy during application execution causes JS error: Uncaught TypeError
If Lazy is not displayed at the initial launch of the app and Lazy is added when state is updated, the following error occurs.

```
index.js:357 Uncaught TypeError: Cannot read property 'view' of undefined
    at propsChanged (index.js:357)
    at getVNode (index.js:362)
    at patch (index.js:273)
    at patch (index.js]:289)
    at render (index.js:471)
```

- [CodeSandbox Example](https://codesandbox.io/s/hyperapp-lazy-bug-rm18c?fontsize=14&module=%2Fsrc%2Findex.jsx)

### Example code (JSX)

```jsx
import { h, app, Lazy } from "hyperapp";

const Label = ({ name }) => {
  return `Hello, ${name}!`;
};

const LazyLabel = () => <Lazy view={Label} name="Lazy hyperapp" />;

const LazyFlagFlip = state => !state;

app({
  init: false,
  view: state => {
    return (
      <div>
        <p>{state ? <LazyLabel /> : <Label name="hyperapp" />}</p>
        <p>
          <input type="checkbox" onchange={LazyFlagFlip} /> Lazy ON
        </p>
      </div>
    );
  },
});
```

sergey-shpak:
- allows to replace nodes with 'Lazy' nodes
- creates textVNode in case 'Lazy' view returns text


jorgebucaran:
And also, @sergey-shpak why does this PR appear to be doing two different things?

sergey-shpak:
@jorgebucaran one more issue was discovered while fixing the general one:
issue#1: replacing vNode with 'Lazy' (general)
issue#2: Lazy view can return 'string' not 'vNode'

jorgebucaran:
> issue#2: Lazy view can return 'string' not 'vNode'

A lazy view that consists of a "string" doesn't make sense to me? What do you think? :)

sergey-shpak:
@jorgebucaran 'Lazy' was designed for performance optimizations, so looking from this point of view - yes, you are right, there is no much benefit if lazy view returns a string. But if you look from 'consistency' in general, we should not differentiate any component result, even if it's rendered with 'Lazy'. Moreover, there are real use cases when under some circumstances Lazy view can return a string:

```javascript
const List = ({ list }) => list.length 
  ? <ul>list.map(item => <li>{ item }</li> )</ul>
  : 'List is empty'
...
<Lazy view={List} list={ list } >
```

PR #889, opened 2019-10-15
Adds `dispatch` tests

sergey-shpak:
This PR adds basic dispatch tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - 6
   - 8
   - 10
   - 12
--- a/test/dispatch.test.js
+++ b/test/dispatch.test.js
@@ -69,20 +69,20 @@ export default {
       expected: { params: true }
     },
     {
-      test: "payload is decoded",
+      test: "payload filter",
       assert: deepEqual,
       actual: dispatch([
         function action(state, payload){
           return { ...state, ...payload }
         },
-        function decoder(){
-          return { decoded: true }
+        function filter(){
+          return { filtered: true }
         }
       ]),
-      expected: { decoded: true }
+      expected: { filtered: true }
     },
     {
-      test: "chaining payload decoders",
+      test: "chaining payload filters",
       assert: deepEqual,
       actual: dispatch([
         [
@@ -90,15 +90,15 @@ export default {
             function action(state, payload){
               return { ...state, ...payload }
             },
-            function decoderC(obj){
+            function filterC(obj){
               return { ...obj, C: true }
             }
           ],
-          function decoderB(obj){
+          function filterB(obj){
             return { ...obj, B: true }
           }
         ],
-        function decoderA(){
+        function filterA(){
           return { A: true }
         }
       ]),
--- a/test/dispatch.test.js
+++ b/test/dispatch.test.js
@@ -1,0 +1,107 @@
+import { app } from ".."
+import { deepEqual } from "testmatrix"
+
+const dispatch = action => {
+  let state
+  app({
+    init: action,
+    subscriptions: s => (state = s, [])
+  })
+  return state
+}
+
+export default {
+  "dispatch state(object)": [
+    {
+      test: "sets new app state",
+      assert: deepEqual,
+      actual: dispatch({ state: true }),
+      expected: { state: true }
+    }
+  ],
+  "dispatch action(function)": [
+    {
+      test: "returns new app state(object) to dispatch",
+      assert: deepEqual,
+      actual: dispatch(state => ({
+        ...state, property: true
+      })),
+      expected: { property: true }
+    },
+    {
+      test: "returns action(function) to dispatch",
+      assert: deepEqual,
+      actual: dispatch(state => function action(state){
+        return { ...state, action: true }
+      }),
+      expected: { action: true }
+    },
+    {
+      test: "returns parameterized action(tuple) to dispatch",
+      assert: deepEqual,
+      actual: dispatch(state => [
+        function action(state){
+          return { ...state, actionTuple: true }
+        }
+      ]),
+      expected: { actionTuple: true }
+    },
+    {
+      test: "returns effect(tuple) to dispatch",
+      assert: deepEqual,
+      actual: done => dispatch(state => [
+        { ...state },
+        [function effect(){ done({ effect: true }) }, {}]
+      ]),
+      expected: { effect: true }
+    }
+  ],
+  "dispatch parameterized action(tuple)": [
+    {
+      test: "payload is passed into action",
+      assert: deepEqual,
+      actual: dispatch([
+        function action(state, payload){
+          return { ...state, ...payload }
+        },
+        { params: true }
+      ]),
+      expected: { params: true }
+    },
+    {
+      test: "payload is decoded",
+      assert: deepEqual,
+      actual: dispatch([
+        function action(state, payload){
+          return { ...state, ...payload }
+        },
+        function decoder(){
+          return { decoded: true }
+        }
+      ]),
+      expected: { decoded: true }
+    },
+  ],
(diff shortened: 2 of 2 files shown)


sergey-shpak:
@jorgebucaran 
looks like Node v6 does not understand some 'destructing syntax' and travis-ci fails, 
should I downgrade tests to ES5 or we can update Node version?

test/dispatch.test.js
```diff
@@ -0,0 +1,107 @@
...
+      ]),
+      expected: { effect: true }
+    },
+    {
+      test: "runs effect with params",
+      assert: deepEqual,
+      actual: dispatch([
+        { effect: true },
+        [function effect(dsptch, props){
+          dsptch(state => ({ ...state, ...props}))
```
jorgebucaran: @sergey-shpak I think you can rename `dsptch` to `dispatch` without trouble.

jorgebucaran:
> @sergey-shpak looks like Node v6 does not understand some 'destructing syntax' and travis-ci fails,

We can upgrade to 8 if that's the really case, but can you point out exactly what destructuring features are not supported by Node 6? See: https://node.green

sergey-shpak:
@jorgebucaran sorry my bad, I thought about 'spread syntax' but wrote 'destruct',
[node@6 doesn't understand spread syntax](https://node.green/#ES2018-features-object-rest-spread-properties) like '...', as result build job fails:
https://travis-ci.org/jorgebucaran/hyperapp/jobs/598138597#L249

test/dispatch.test.js
sergey-shpak: @jorgebucaran didn't want to shadow 'dispatch', but since 'state' is also shadowed that is ok )

jorgebucaran:
@sergey-shpak Let's bump node to the first even version that supports the spread syntax, so 8?

sergey-shpak:
@jorgebucaran looks like minimal node version which supports the spread syntax is **8.6**

sergey-shpak:
@zaceno thanks, you're right, everything should be consistent, I'm ok with any variant

.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - 6
```
jorgebucaran: :+1:

test/dispatch.test.js
```diff
@@ -0,0 +1,130 @@
...
+import { deepEqual } from "testmatrix"
+
+const dispatch = action => {
+  let state
+  app({
+    init: action,
+    subscriptions: s => (state = s, [])
+  })
+  return state
+}
```
jorgebucaran: What's up with `s => (state = s, [])`?

sergey-shpak: this allows to `catch and return state` and return any value into subscriptions (it's shorter than `s => (state = s, undefined)`)
* hm, it's possible `s => (state = s,)` to return undefined

jorgebucaran: Haha, nice hack. Now, what would be a more readable way to write this?

sergey-shpak: I believe the most balanced short/readable is existing version `s => (state = s, [])`, but if you worry about comma operator, I can change it to `s => { state = s; return }` or even more readable:
```javascript
s => {
  state = s
  return
}
```

jorgebucaran: @zaceno Thoughts?

jorgebucaran: @mrozbarry @zaceno Do you have an opinion on this?

frenzzy: maybe hyperapp should expose internal state variable via api to be able to get current app state at any given time?

sergey-shpak: @frenzzy you can find discussion about exposing app state https://github.com/jorgebucaran/hyperapp/issues/863

mrozbarry: I don't think that's a bad solution. Initially I thought that using a promise would be better, since it could be more hyperapp-ish, and that looked like:

```js
let resolve = null;
const promise = new Promise((done) => { resolve = done});
const ResolverFX = (dispatch, state) => {
  resolve(state);
  return () => {};
});
const Resolver = state => [ResolverFX, state];

// App with resolver sub

const state = await promise;
```

There's pros and cons to both. The proposed code technically makes the app impure, but it's simpler code. Mine moves the impurity to a sub, but it's obviously more code and less manageable.

That's too say the proposed code should be fine.

zaceno: I have no contrary opinion. Looks fine to me.

sergey-shpak: @jorgebucaran I thought you were worrying about 'comma operator' :smile:  
OK, how about that?
```javascript
const dispatch = (init, view) => 
  app({ init, view })

// and test
{
  test: "test name",
  assert: deepEqual,
  actual: done => 
    dispatch({ some: true }, done),
  expected: { effect: true }
}
```

jirutka:
@jorgebucaran, why this PR is still not merged? It’s quite a shame that Hyperapp has only one test case at this moment. :slightly_frowning_face:

test/dispatch.test.js
sergey-shpak: @jorgebucaran dispatch wrapper update (see latest commit)

sergey-shpak:
@zaceno @mrozbarry @frenzzy @jorgebucaran 
need your thoughts about new updated 'dispatch' wrapper impl

mrozbarry:
I'm going to have to read through this some more over the weekend, and have a better understanding of how testmatrix works, and what should be tested as public interfaces. I think the hard problem of testing hyperapp, is the public interface is intentionally difficult to observe. That doesn't mean it's wrong, but it means I think we need to be careful how we test things.

I guess I'm also not sure what this tests - that dispatch as a unit or dispatch as end-to-end.

In fact, how do we want to test Hyperapp? I'd think doing more like Cypress tests would be way more holistic, but I also understand that it would mean ditching testmatrix, but I'd argue that we can't test dispatch in isolation. Hyperapp is state+effects+ui, and it's hard to really write meaningful tests that don't touch all those pieces together.

My opinion is that we should have these tests:

 1. Set an initial state, test that we can output a consistent vdom that uses that state.
 2. Set an initial state with an effect, test that we get two updates, the initial render, and the result of the effect.
 3. Set an initial state that enables a subscription, observe a render, effect, action (to stop subscription), and final result
 4. Small counter app, simulate click on button, observe new vdom
 5. Small text input app, simulate typing, observe that event decoders work

That covers all major use cases, and then we wait for bugs to reproduce specific use cases that the general flow doesn't cover (maybe chained actions/effects or whatever).

I guess my basic point is that the public interface that Hyperapp gives us isn't a good fit for unit tests. There's nothing wrong with that, it just means we'd be aiming for more feature/end-to-end/acceptance tests.

mrozbarry:
So, I've been thinking about what sort of app would handle holistic tests for hyperapp, and I think I have it. Our vdom should look something like:

```jsx
<div>
  <input type="number" value={state.number} oninput={[SetNumber, (event) => event.target.value]} />
  <button onPress={IncrementNumber}>increment</button>
  <button onPress={[DelayIncrementNumber, 5]}>increment after 5 seconds</button>
  <button onPress={StartIncrementEverySecond}>start increment every second</button>
  <button onPress={StopIncrementEverySecond}>stop increment every second</button>
</div>
```

That would give us the following test surface area:

 1. Test action with event decoder (change input value)
 2. Test action with no props (increment button)
 3. Test action with effect (increment after 5 seconds button)
 4. Test action with subscription trigger (start/stop increment every second button)

Is there any mission critical test that's missing from this mini app?

jorgebucaran:
We're no longer using Testmatrix, but [Twist](https://github.com/jorgebucaran/twist) for testing. We should definitely add `dispatch` tests soon, though.

PR #891, opened 2019-10-23
Request animation frame focus effect

cdrubin:
doing .focus() within a requestAnimationFrame callback allows the DOM element time to exist when that call is executed


frenzzy:
As far as I know on mobile devices focus effect will work only in the same tick when actual tap (click) triggered by user happened. So, this change possibly will break focus effect on mobile devices.

zaceno:
@frenzzy Hmm yeah I think you're right. I've noticed that behavior but never really thought about it. 

I guess then maybe it's best to leave this as is.

The only way I can think of to solve the problem this adresses, which works in desktop and mobile, is to use lifecycle events (which would have to be hacked in. Here's an example: https://codepen.io/zaceno/pen/ZEEeXRr?editors=0010)

jorgebucaran:
> @frenzzy As far as I know on mobile devices focus effect will work only in the same tick when actual tap (click) triggered by user happened. So, this change possibly will break focus effect on mobile devices.

Can you make a pen that shows this problem?

zaceno:
@jorgebucaran This pen demonstrates the four combinations of direct vs deferred focus effect and applying it to an existing element vs en element that is rendered after the action was dispatched.

You can try it on desktop and mobile to observe the differences and similarities in behavior.

https://codepen.io/zaceno/pen/PoopEpy

jorgebucaran:
@zaceno Any DOM operations must take place after we have a valid DOM, which is after Hyperapp's render loop is finished. When is this "guaranteed"?  Inside a rAF, so this change is arguably the correct implementation of this effect.

Mobile presents a curious challenge, though. We could provide (1) sync/async versions of `focus`, a different API (2) `focus("id", true)` (or `false`) to specify sync/async-ness or (3) dare to try to detect whether we're on a mobile device and decide to use the sync/async focus version accordingly.

Here's a working implementation of (3).

https://codepen.io/jorgebucaran/pen/BaaWPdw?editors=0010

```js
const isMobile = /* see live example for a possible implementation */

export const focus = (f => id => [f, id])((_, id) =>
  isMobile
    ? document.getElementById(id).focus()
    : requestAnimationFrame(_ => document.getElementById(id).focus())
)
```
```js
const SetFocusDirect = (state, id) => [state, focus(id)]
const SetFocusDeferred = (state, id) => [state, focus(id)]
const ShowAndFocusDirect = (state, id) => [{ ...state, [id]: true }, focus(id)]
const ShowAndFocusDeferred = (state, id) => [
  { ...state, [id]: true },
  focus(id)
]

app({
  init: {
    "input-c": false,
    "input-d": false
  },
  view: state =>
    h("body", {}, [
      h("p", {}, [
        h("button", { onclick: [SetFocusDirect, "input-a"] }, "Focus A"),
        h("input", {
          id: "input-a",
          type: "text",
          value: "input A"
        })
      ]),
      h("p", {}, [
        h("button", { onclick: [SetFocusDeferred, "input-b"] }, "Focus B"),
        h("input", {
          id: "input-b",
          type: "text",
          value: "input B"
        })
      ]),
      h("p", {}, [
        h(
          "button",
          { onclick: [ShowAndFocusDirect, "input-c"] },
          "Show and Focus C"
        ),
        state["input-c"] &&
          h("input", {
            id: "input-c",
            type: "text",
            value: "input C"
          })
      ]),
      h("p", {}, [
        h(
          "button",
          { onclick: [ShowAndFocusDeferred, "input-d"] },
          "Show and Focus D"
        ),
        state["input-d"] &&
          h("input", {
            id: "input-d",
            type: "text",
            value: "input D"
          })
      ]),
      h(
        "p",
        {},
        `
            Both Focus A and Focus B work fine on desktop
            But Focus B does not work on mobile, because it
            uses the deferred (with rAF) effect.
        `
      ),
      h(
        "p",
        {},
        `
            "Show and Focus C" fails everywhere, because
            it uses the direct effect, before the input exists. "Show and Focus D" uses
            the deferred effect, and hence works on desktop, 
            but not on mobile.
        `
      )
    ]),
  node: document.body
})
```

zaceno:
@jorgebucaran I agree, the rAF-wrapper is probably the most correct implementation, and you could merge this PR.

It's true this will break working focus implementations on mobile. But I'd argue that focus on mobile is already broken, since I think the need to show and focus an input in the same action is more common than the need to focus on an existing element, and there is no way to do that with just hyperapp (you need to resort to DOM-hacking on the side).

jorgebucaran:
@zaceno Can Elm do this?

zaceno:
@jorgebucaran 

> Can Elm do this? Can React/Preact do this?

Honestly I don't know. Someone should try it out. The trick is, like @frenzzy said, that the `.focus()` call needs to come in the same tick as a DOM event. The reason my lifecycle hack works is because when elements are created I dispatch a `create` event on them. Possibly something similar is used in react, but probably not Elm. 

I don't think we should remove rAF, that seems unsafe. I think we should merge this PR, and accept that focus on mobile is a weird known issue for now.

sergey-shpak:
imo, small clarification, - mobile browsers allow dispatching events only from call stack initiated by user event (click/touch etc). Since **'async' functions** (rAF/timeout/interval) **started from different(main) stack** any dispatched event is ignored (and yes, since they async they run in different tick)

sergey-shpak:
as result, in theory elm/react or any other framework can't async focus element on mobile devices without hacks and tricks (but you know this has to be doublechecked :) )

frenzzy:
In React.js both cases (focus on mount, focus on event) works just fine, open this link on mobile:
https://cdpn.io/frenzzy/debug/ExxmmEE ([CodePen](https://codepen.io/frenzzy/pen/ExxmmEE/left?editors=0010))

sergey-shpak:
@frenzzy you're example starting each call stack with user event (by click) which as I mentioned before can properly dispatch element events. (btw, you can update your example with `setTimeout`, and it will work - meaning that dispatching actions doesn't depend on execution 'tick')
  
  Changed your example with 'focusing' without any user action: https://codepen.io/sergey-shpak/pen/VwwbzLp (it doesn't focusing without user action even on my desktop browser but works great with button click)

sergey-shpak:
my summary: 
- dispatching element events allowed only with stack initiated by user action (click/touch etc)
- dispatching any element event has to be done when the element is rendered into dom (created/appended). 

This pr is trying to solve custom case, because element is not always unavailable at the moment of dispatching event.

In general this issue can be workarounded with more scalable pattern
```
import { focus } from '@hyperapp/dom'
import { rAF } from '@hyperapp/time' // 'rAF' effect should be added

const focusElement = id => focus(id)
const focusDefferedElement = id => rAF(focus(id))
```
@jorgebucaran @zaceno @frenzzy @cdrubin

zaceno:
@sergey-shpak it may be a custom case but it sure is a very common case! In almost every case I want to use .focus() it is for a newly created input. Because users can just click to focus the existing ones, it’s rare that I need to focus them programmatically.

sergey-shpak:
@zaceno my point that we shouldn't wrap 'focus' with rAF inside of core, instead we should export `focus` and export `requestAnimationFrame` separately, this gives several benefits in userland:
- possibility to wrap focus with rAF where it's needed
- to use rAF for any other purpose

in other words, core shouldn't export 'house', it should export bricks, which gives dev possibility to build a house they want

zaceno:
@sergey-shpak Personally I'd be totally fine with this approach. I often use a `nextFrame` custom effect anyway (in fact I nearly always use only custom effects so I don't have a stake in this at all really)

But I get the feeling that exporting "bricks" is not quite the philosophy behind the offcial effect-offering – or at least it hasn't been the philosophy so far. So far it seems to me the philosophy has been to export "kitchen", "bathroom", "living-room" et c. Each fully working on its own, with only a _little_ bit of wiring needed to put together a fully livable home.

To get back to the concrete ;) If Jorge wants to export a `nextFrame` or `rAF` effect, then the current focus is fine. But until then, the `focus` effect should be implemented in such a way it will work as reasonably expected in any common case.

sergey-shpak:
@zaceno I'm ok with this approach - it will work for me either, but in general we have to provide more simple/scalable patterns everywhere we can, and this pull request is not exception for me, if we can achieve the same functionality and gain some other benefits

sergey-shpak:
@zaceno for example, it is possible that 'focus' functionality has to be tested with `phantomjs` or `puppeteer` where 'requestAnimationFrame' event is never fired because of headless chrome.
Providing 'requestAnimationFrame' and 'focus' separately allows developer to make composition where it is needed, or use each effect separately. Win win.

zaceno:
@sergey-shpak I haven’t used hyperapp with phantomjs, but wouldn’t rAF need to be hacked anyway, to make hyperapp work at all?

sergey-shpak:
@zaceno since we have fallback to `[setTimeout](https://github.com/jorgebucaran/hyperapp/blob/master/src/index.js#L8)` hyperapp core should work as expected

zaceno:
Ah right, core falls back on its own, but this PR does not. In that case this PR should also include a fallback IMO. (Or if we break out a separate “nextFrame” effect, it should fall back in the same way.

jorgebucaran:
@cdrubin I went ahead and added this myself, but thank you for the PR and motivating this discussion anyway. 💯✨
 
[Try it here](https://codepen.io/jorgebucaran/pen/wjvEBj?editors=1100). Tested on Chrome, Firefox, Safari, and iOS. ✅

PR #892, opened 2019-10-23
Enable support for DOM level 3+ case sensitive event types

websemantics:
Enable support for DOM level 3+ case sensitive event types #892

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -106,7 +106,7 @@ var patchProperty = function(node, key, oldValue, newValue, listener, isSvg) {
   } else if (key[0] === "o" && key[1] === "n") {
     if (
       !((node.actions || (node.actions = {}))[
-        (key = key.slice(2).toLowerCase())
+        (key = key.slice(2))
       ] = newValue)
     ) {
       node.removeEventListener(key, listener)


jorgebucaran:
@frenzzy @sergey-shpak @Swizz Are we happy to merge here?

sergey-shpak:
@websemantics @jorgebucaran 
is it backward compatible? I mean If users have already used 'onClick' with hyperapp#2 apps/pens, this change will affect them.
* or this should be released as minor/major, not patch

sergey-shpak:
@jorgebucaran ok, then I'm ok)

Jesspu:
Is this getting merged in?

okwolf:
I don't think we should be making breaking changes at this point.

zaceno:
I'm curious: what are some examples of events we can't handle unless we merge this?

jorgebucaran:
Forgot to merge this. We could since Hyperapp is still not "officially" released (poor excuse).

jimfranke:
I think we should go with the following, which still allows for `onClick` instead of `onclick`, but also supports case-sensitive custom events.

```javascript
const keyLower = key.toLowerCase();
    if (
      !((node.actions || (node.actions = {}))[
	(keyLower in node ? keyLower : key).slice(2)
      ] = newValue)
    ) {
```

What do you think, @jorgebucaran?

jorgebucaran:
@jimfranke I'm going to merge here first, then we can consider bringing that in the future! Thank you, @websemantics.

jorgebucaran:
@jimfranke Revisiting this. Could you clarify:

> but also supports case-sensitive custom events.

jorgebucaran:
@websemantics @jimfranke I'd also like to know the answer to:

> @zaceno I'm curious: what are some examples of events we can't handle unless we merge this?

websemantics:
Thanks @jorgebucaran, take the [Alert](https://ionicframework.com/docs/api/alert) element in [Ionic Framework](https://ionicframework.com) for example `ionAlertDidDismiss` is case-sensitive. Also check the type attribute in [Changes to DOM Level 2 Events interfaces](https://www.w3.org/TR/uievents/#changes-DOMLevel2to3Changes).

Jesspu:
I ran into the same issue. I tried using ionic web components with hyper
app for a hybrid mobile app and could not get some of ionics events to
work. If this was resolved i would love to be able to use ionics (or other)
web components with case sensitive events with hyper app.

jorgebucaran:
@websemantics Would [this solution](https://github.com/jorgebucaran/hyperapp/pull/892#issuecomment-598660865) allow both to work? 

@Jesspu We merged this a while ago, so it's how Hyperapp works right now, but I was revisiting some old issues and wanted to fully understand why we dropped support for onClick, etc.

websemantics:
@jorgebucaran Yes, just tested it and it works nicely. Also, with a bit of tweaking we can get `onIonAlertDidDismiss` work too.

PR #893, opened 2019-10-24
Update CSS for the tutorial

loteoo:
## Updated CSS for the tutorial

Preview here:
https://hyperapp-tutorial-demo-1.netlify.com/

hyperapp-tutorial-demo-1 netlify com_(iPad)

The text of the tutorial itself might need a small tweaking for the "Filter" component with this PR, but I want to get some peer review first.

I'd like to know what @zaceno thinks as well :)


jorgebucaran:
@zaceno @loteoo Can we change the filter icon to a search icon or something else? 

A few other tweaks I'd like to ask if possible:

- The checkbox looks too small compared to the rest of the UI -- is it just me?
- Can spacing, margin and line-height be improved?

jorgebucaran:
@loteoo @zaceno I'm working on a "style guide" for Hyperapp that should eventually be the basis for the website. Do you think we should use that to design the tutorial or should the tutorial have its own unique styles for some reason?

loteoo:
Thanks for the insightful feedback @zaceno 

@jorgebucaran If the tutorial could inherit the same look as the site using a shared style guide, I think that'd be great!

If someone can create a design for it and update the tutorial text to match it (or vice-versa), I can definitively handle the CSS !!

Let's keep this PR open to discuss the topic, I will update it as we develop the style guide :)

jorgebucaran:
@loteoo I think the checkbox is not using the one from the style guide.

---

I feel the UX on this app is problematic. The problem is one has to click the search button to search, but that could be implemented without a button -- what do you think? On the other hand, this affects what's written in the tutorial. So, we're between a rock and a hard place. 

cc @zaceno

zaceno:
@jorgebucaran @loteoo 

Yup changing to this style would require a (non-trivial) change time the tutorial. And while I do like the look of this, I’m not keen on making those changes myself, since I wrote the tutorial with a thought of a particular progression, and then shoehorned an example in to fit my outline, rather than the other way around. 

That’s why the UX is not great either. This is not how I would make this app if my ambition was designed to make a great app. Rather the interface was designed to let me talk about things in the order I wanted, and the style was made to produce good screenshots. All the while not straying too far from being at least close to realistic.

loteoo:
Should we close this now?

jorgebucaran:
@loteoo How hard would it be to update the CSS to the beautiful version with @zaceno's recent edits now merged in?

PR #894, opened 2019-10-25
More realistic, less cute README

brandonros:
Showing this wonderful framework/piece of software (seriously, I mean it, I reference it all of the time) to somebody without intermediate/deep JavaScript knowledge is confusing because all of the trickery/cleverness that goes on in the example.

I know we've discussed this before, but just today I went to show this software to a Scala developer looking to get more interested in frontend development through JavaScript and was reminded how complicated/subtle the README example is.

I know you didn't like it before, but I'd really ask you reconsider having a more "real world, common" example showcase your fine work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,14 +39,35 @@ Here's the first example to get you started: a counter that can go up or down. Y
     <script type="module">
       import { h, app } from "https://unpkg.com/hyperapp"
 
+      const cloneObject = (obj) => JSON.parse(JSON.stringify(obj))
+
+      const decrement = function (state) {
+        return function () {
+          const newState = cloneObject(state)
+          newState.counter -= 1
+          return newState
+        }
+      }
+
+      const increment = function (state) {
+        return function () {
+          const newState = cloneObject(state)
+          newState.counter += 1
+          return newState
+        }
+      }
+
       app({
-        init: 0,
-        view: state =>
-          h("main", {}, [
-            h("h1", {}, state),
-            h("button", { onClick: state => state - 1 }, "-"),
-            h("button", { onClick: state => state + 1 }, "+")
-          ]),
+        init: {
+          counter: 0
+        },
+        view: function (state) {
+          return h("main", {}, [
+            h("h1", {}, state.counter),
+            h("button", { onClick: decrement(state) }, "-"),
+            h("button", { onClick: increment(state) }, "+")
+          ])
+        },
         node: document.getElementById("app")
       })
     </script>
--- a/README.md
+++ b/README.md
@@ -39,7 +39,9 @@ Here's the first example to get you started: a counter that can go up or down. Y
     <script type="module">
       import { h, app } from "https://unpkg.com/hyperapp"
 
-      const cloneObject = (obj) => JSON.parse(JSON.stringify(obj))
+      const cloneObject = function (obj) {
+        return JSON.parse(JSON.stringify(obj))
+      }
 
       const decrement = function (state) {
         return function () {
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Here's the first example to get you started: a counter that can go up or down. Y
         },
         view: function (state) {
           return h("main", {}, [
-            h("h1", {}, state.counter),
+            h("h1", {}, `Value: ${state.counter}`),
             h("button", { onClick: decrement(state) }, "-"),
             h("button", { onClick: increment(state) }, "+")
           ])


brandonros:
@ChristopherDavenport

jorgebucaran:
> is confusing because all of the trickery/cleverness that goes on in the example.

I'm more than happy to spend some time discussing what that trickery is one by one. 

Since you're the one pointing this out, please state what are those things that you find confusing about the example?

README.md
```diff
@@ -39,14 +39,37 @@ Here's the first example to get you started: a counter that can go up or down. Y
     <script type="module">
       import { h, app } from "https://unpkg.com/hyperapp"
 
+      const cloneObject = function (obj) {
+        return JSON.parse(JSON.stringify(obj))
```
jorgebucaran: This far from realistic. No one writes this.

README.md
```diff
@@ -39,14 +39,37 @@ Here's the first example to get you started: a counter that can go up or down. Y
...
+      }
+
+      const decrement = function (state) {
+        return function () {
+          const newState = cloneObject(state)
+          newState.counter -= 1
+          return newState
+        }
+      }
```
jorgebucaran: This is how you'd write this action:

```js
const decrement = state => ({ ...state, counter: state.counter - 1 })
```

README.md
```diff
@@ -39,14 +39,37 @@ Here's the first example to get you started: a counter that can go up or down. Y
...
+
+      const increment = function (state) {
+        return function () {
+          const newState = cloneObject(state)
+          newState.counter += 1
+          return newState
+        }
+      }
```
jorgebucaran: This is how you'd write this action:

```js
const increment = state => ({ ...state, counter: state.counter + 1 })
```

README.md
```diff
@@ -39,14 +39,37 @@ Here's the first example to get you started: a counter that can go up or down. Y
...
-        view: state =>
-          h("main", {}, [
-            h("h1", {}, state),
-            h("button", { onClick: state => state - 1 }, "-"),
-            h("button", { onClick: state => state + 1 }, "+")
-          ]),
+        init: {
+          counter: 0
+        },
+        view: function (state) {
```
jorgebucaran: This doesn't make sense.

jorgebucaran:
Sorry, but no thanks. I don't see what's wrong with the example. I'm still more than glad to go over what you think are tricks, but mostly to explain to you that they aren't.

README.md
brandonros: Is this actually what happens though? aka, is the state actually getting cloned and returned new instead of mutated?

README.md
brandonros: This is very confusing to somebody who doesn't know all of the tricks/syntactic sugar of ES6. I get that you don't really care about those people (and you shouldn't have to)

but... say a newcomer from Java/Scala wants to come dive into your project. My example is much more verbose/clear on what is happening.

jorgebucaran: I don't know if you're being serious or trolling me. 😉  

Your example creates a closure for no reason. Then clones the state inefficiently, mutates it, and finally returns it. Unless you're completely new to programming, `state => state - 1` should be readable enough.

README.md
jorgebucaran: I don't understand what you're asking.

README.md
brandonros: Examples are supposed to show case real world usage.

In what real world usage is the entire state for a component simply an `integer`? An object is much more realistic.

Also, arrow functions didn't exist in JavaScript until semi-recently (and you won't find them in other languages).

I'm not trolling you (and I'm not questioning how long you've been programming, like you are to me). I'm trying to collaborate on making a real world example that doesn't use clever tricks.

README.md
brandonros: You said "no one writes like this"

Nobody writes cloneObject functions? Your library requires the `state` object be cloned because mutating the variable will not work/cause the changes to be flushed to the UI.

README.md
zaceno: True, no real world example will just have a number as the state. However, I disagree that all examples are for showcasing real-world use cases. The point of this first example is to give an idea at a quick glance, how the framework works.

There are plenty of other examples (official and community) that are more realistic.

One thing I would suggest though, @jorgebucaran, is linking to the tutorial and examples from the readme.

README.md
zaceno: @brandonros  Writing a cloneObject function is a very inefficient way to update the state, as you will have to process everything in a potentially large and deep state tree in order to alter a single value. Every single time a value changes (even in, say, an animation at 60fps...)

The way Jorge demonstrated is actually _does_ make the top level a new instance, and thus triggers a rerender. If your state is

```
{ foo: {flip: 1, flap: 2, flop: 3}, bar: {...}, baz: {...}}
```

and you say: 

```
{
  ...state,
  foo: {
    ...state.foo,
    flap: 42
  }
}
```

... then the state object itself will be new. `state.foo` will be new. `state.foo.flap` will have a new value. _But_ `state.bar` is the _same_ as before, and so is `state.foo.flip` (whatever it was).

Using the spread operator to compose new state not only saves you the trouble of cloning stuff -- because `bar` is still the same instance, it allows you to optimize rendering (basically skip diffing entire blocks of the vdom) any vnodes which depend on bar. 

That is the reason Jorges implementation of the `Increment` function is "official best practice".  It's unfortunate if people coming from other languages find it weird, but it is how it is done in all the official and unofficial examples I have seen anywhere. It's just something a newcomer to the framework will have to learn.

brandonros: @zaceno Let me try my example without the cloneObject function. I'm *pretty sure* it doesn't work, which makes me wonder... does the `spread` operator actually do some kind of shallow clone behind the scenes?

brandonros: yeah, take a look at this:

```
const decrement = function (event, state) {
  state.counter -= 1
  return state
}

const increment = function (event, state) {
  state.counter += 1
  return state
}

app({
  init: {
    counter: 0
  },
  view: function (state) {
    return h("main", {}, [
      h("h1", {}, `Value: ${state.counter}`),
      h("button", { onClick: function (event) { return decrement(event, state) } }, "-"),
      h("button", { onClick: function (event) { return increment(event, state) } }, "+")
    ])
  },
})
```

this does not update the view on change

README.md
brandonros: > There are plenty of other examples (official and community) that are more realistic.

Why have an unrealistic example in the README (one of the first things people see when they land on your project)?

README.md
brandonros: I got it. Mutation is gross and causing weird things to subtly happen.

```
const decrement = function (event, state) {
  return Object.assign({}, state, {
    counter: state.counter - 1
  })
}

const increment = function (event, state) {
  return Object.assign({}, state, {
    counter: state.counter + 1
  })
}

app({
  init: {
    counter: 0
  },
  view: function (state) {
    return h("main", {}, [
      h("h1", {}, `Value: ${state.counter}`),
      h("button", { onClick: function (event) { return decrement(event, state) } }, "-"),
      h("button", { onClick: function (event) { return increment(event, state) } }, "+")
    ])
  },
})
```

This is a better example :)

zaceno: You got it! Spread operator is just a shorter, more idiomatic `Object.assign({}, ...)` 👍

README.md
jorgebucaran: @zaceno Definitely a good idea! I just haven't found a way to do it that stays in harmony with the rest of the file, but I'll see what I can do! 😄

README.md
jorgebucaran: Thank you, @zaceno!

PR #898, opened 2019-11-01
adding back support for oncreate

ibudiallo:
Upgrading 1.2.10 broke the oncreate event. 

We used to be able to do this:

```
h("div", {
  oncreate: el => {
    console.log(el, "element created");
  }
}, "Test");
```

This PR adds it back.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -111,7 +111,11 @@ var patchProperty = function(node, key, oldValue, newValue, listener, isSvg) {
     ) {
       node.removeEventListener(key, listener)
     } else if (!oldValue) {
-      node.addEventListener(key, listener)
+      if (key === "create") {
+        newValue.apply(node, [node])
+      } else {
+        node.addEventListener(key, listener)
+      }
     }
   } else if (!isSvg && key !== "list" && key in node) {
     node[key] = newValue == null ? "" : newValue


zaceno:
@ibudiallo Lifecycle events were removed for v2 after long, intense debating. They may come back at some point, but I suspect only after more long debating. In the mean time, there are ways to get the functionality back without support in core. See for instance https://github.com/websemantics/hyperapp-lifecycle

jorgebucaran:
Hey, @ibudiallo! Basically what @zaceno said, but I have to admit: this is a clever trick to implement `oncreate`. 

Can `onremove` be implemented the same way? 😂

ibudiallo:
Thank you @zaceno and @jorgebucaran I was not aware that there had been a whole debate about Lifecycle events. I'm a fairly new user of the library and recently I accidentally added the latest version from a new CDN and saw that all my `oncreate` events stopped working. 

I understand that these won't be added back soon, but for my own case, I'll be using my modified version. Thanks again for this awesome framework.

jorgebucaran:
@ibudiallo Please be aware that V1 will be deprecated soon. I suggest you look into V2 if you're interested in functional programming. If you're just interested in a small VDOM library, check out [Superfine](https://github.com/jorgebucaran/superfine) instead! 👋

Issue #896, opened 2019-10-29
Modules in V2


zaceno:
###  💁‍♂ [Check out the current proposal here](https://github.com/jorgebucaran/hyperapp/issues/896#issuecomment-551552323).

---

Consider this minimal example of a modular Elm app (copied from [here](https://discourse.elm-lang.org/t/a-simple-example-of-splitting-view-update-into-a-separate-module/3107/2)):

<details>
<summary>Button.elm</summary>

```elm
module Button exposing (Model, Msg(..), init, update, view)

import Html exposing (Html, button, div, h1, text)
import Html.Events exposing (onClick)



-- MODEL


type alias Model =
    Int


init : Model
init =
    0



-- UPDATE


type Msg
    = Increment
    | Decrement


update : Msg -> Model -> Model
update msg model =
    case Debug.log "Button.update msg: " msg of
        Increment ->
            model + 1

        Decrement ->
            model - 1


view : Model -> Html Msg
view model =
    div []
        [ button [ onClick Decrement ] [ text "-" ]
        , div [] [ text (String.fromInt model) ]
        , button [ onClick Increment ] [ text "+" ]
        ]
```

</details>

<details>
<summary>Main.elm</summary>

```elm
type alias Model =
    { button : Button.Model }


init : Model
init =
    { button = Button.init
    }



-- UPDATE


type Msg
    = ButtonMsg Button.Msg


update : Msg -> Model -> Model
update msg model =
    case Debug.log "update msg" msg of
        ButtonMsg buttonMsg ->
            { model | button = Button.update buttonMsg model.button }



-- VIEW


view : Model -> Html Msg
view model =
    div []
        [ h1 [] [ text "elm button example" ]
        , map ButtonMsg (Button.view model.button)
        ]

```

</details>

Notice two things:

A) `Increment` and `Decrement` are completely unaware of the shape of the global app state. Their definitions are completely self contained.

B) `Main.elm` never once explicitly references `Increment` or `Decrement`.

As far as I've been able to tell, this is not possible to achieve using Hyperapp 2. You can have A or B but not both. This has frightening implications for large scale apps. 

I would like to see a convention, add-on library or core change -- or some combination of all three -- which enabled us to replicate the example above in Hyperapp v2. 

That essentially means being able to use a module which defines its actions in a self-contained, app-agnostic way, without being required to export them. They should not need to be explicitly referenced anywhere else. 

[...]

PR #899, opened 2019-11-01
allow mapping actions in vdom and subs

zaceno:
Proof of concept solution for https://github.com/jorgebucaran/hyperapp/issues/896

Basically, it works by giving `dispatch` a possible third argument, which, if present, tells it how to transform an action before applying it. 

But how do we get an actionMap _to_ dispatch? 

If the actions you want to map are all dispatched from a part of the vdom, use the new exported function: `mapEvents(actionMap, vdom)`: it returns the given vnode with a new `actionMap` property. This action map is propagated through all recursive `patch` calls until it is finally added to the real node alongside its actions. Then `listener` will dispatch with the `actionMap`.

If the actions you want to map are on an effect, use the new exported `mapEffect(actionMap, effect)`. The action map is added as the third element in the effect tuple. The `dispatch` passed to efffects and subs will not be the normal dispatch, but one bound with the actionMap as third argument. 

To map actions from subscriptions, use `mapSubs`, which essentially just maps the array and applies `mapEffect` to each.

I don't presume this is the final form of this, but I submit it as a proof of concept to be iterated upon and refactored.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -4,7 +4,6 @@ var TEXT_NODE = 3
 var EMPTY_OBJ = {}
 var EMPTY_ARR = []
 var isArray = Array.isArray
-var isFunc = function(x) { return typeof x === "function" }
 var defer =
   typeof requestAnimationFrame !== "undefined"
     ? requestAnimationFrame
@@ -46,7 +45,7 @@ var batch = function(list) {
     return out.concat(
       !item || item === true
         ? 0
-        : isFunc(item[0])
+        : typeof item[0] === "function"
         ? [item]
         : batch(item)
     )
@@ -67,16 +66,16 @@ var makeActionMap = function(extract, merge) {
   return function actionMap(action) {
     return function(state, data) {
       return (function resolve(x, state, data) {
-        return isFunc(x)
+        return typeof x === "function"
           ? resolve(x(extract(state), data), state)
           : !isArray(x)
             ? merge(state, x)
-            : !isArray(x[0]) && !isFunc(x[0])
+            : !isArray(x[0]) && typeof x[0] !== "function"
               ? [merge(state, x[0])].concat(mapEffects(x.slice(1), actionMap))
               : resolve(
                 x[0],
                 state,
-                isFunc(x[1]) ? x[1](data) : x[1]
+                typeof x[1] === "function" ? x[1](data) : x[1]
               )
       })(action, state, data)
     }
@@ -92,13 +91,13 @@ var mappedDispatch = function(dispatch, map) {
 export var map = function(extract, merge, x, actionMap = makeActionMap(extract, merge)) {
   return isArray(x)
     ? mapEffects(x, actionMap)
-    : isFunc(x)
+    : typeof x === "function"
       ? actionMap(x)
-      : (x.actionMap = actionMap, x)
+      : (x.actionMap = compose(actionMap, x.actionMap), x)
 }
 
 var isSameAction = function(a, b) {
-  return isArray(a) && isArray(b) && a[0] === b[0] && isFunc(a[0])
+  return isArray(a) && isArray(b) && a[0] === b[0] && typeof a[0] === "function"
 }
 
 var shouldRestart = function(a, b) {
@@ -148,6 +147,7 @@ var patchProperty = function(node, key, oldValue, newValue, listener, actionMap,
       }
     }
   } else if (key[0] === "o" && key[1] === "n") {
+    node.actionMap = actionMap
     if (
       !((node.actions || (node.actions = {}))[
         (key = key.slice(2).toLowerCase())
@@ -155,7 +155,6 @@ var patchProperty = function(node, key, oldValue, newValue, listener, actionMap,
     ) {
       node.removeEventListener(key, listener)
     } else if (!oldValue) {
-      node.actionMap = actionMap
       node.addEventListener(key, listener)
     }
   } else if (!isSvg && key !== "list" && key in node) {
@@ -480,7 +479,7 @@ export var h = function(name, props) {
 
   props = props || EMPTY_OBJ
 
-  return isFunc(name)
+  return typeof name === "function"
     ? name(props, children)
     : createVNode(name, props, children, undefined, props.key)
 }
@@ -513,13 +512,13 @@ export var app = function(props) {
     function(obj) {
       return obj
     })(function(action, props, map) {
-      return isFunc(action)
+      return typeof action === "function"
         ? dispatch((map ? map(action) : action)(state, props))
         : isArray(action)
-          ? isFunc(action[0]) || isArray(action[0])
+          ? typeof action[0] === "function" || isArray(action[0])
             ? dispatch(
               action[0],
-              isFunc(action[1]) ? action[1](props) : action[1],
+              typeof action[1] === "function" ? action[1](props) : action[1],
               map
             )
             : (batch(action.slice(1)).map(function(fx) {
--- a/src/index.js
+++ b/src/index.js
@@ -53,6 +53,22 @@ var batch = function(list) {
   }, EMPTY_ARR)
 }
 
+var composeActionMaps = function(a, b) {
+  return b
+    ? function(x) {
+        return a(b(x))
+      }
+    : a
+}
+
+var mappedDispatch = function(dispatch, map) {
+  return map
+    ? function(action, props) {
+        return dispatch(action, props, map)
+      }
+    : dispatch
+}
+
 var isSameAction = function(a, b) {
   return isArray(a) && isArray(b) && a[0] === b[0] && typeof a[0] === "function"
 }
@@ -82,7 +98,7 @@ var patchSubs = function(oldSubs, newSubs, dispatch) {
           ? [
               newSub[0],
               newSub[1],
-              newSub[0](dispatch, newSub[1]),
+              newSub[0](mappedDispatch(dispatch, newSub[2]), newSub[1]),
               oldSub && oldSub[2]()
             ]
           : oldSub
@@ -92,7 +108,15 @@ var patchSubs = function(oldSubs, newSubs, dispatch) {
   return subs
 }
 
-var patchProperty = function(node, key, oldValue, newValue, listener, isSvg) {
+var patchProperty = function(
+  node,
+  key,
+  oldValue,
+  newValue,
+  listener,
+  actionMap,
+  isSvg
+) {
   if (key === "key") {
   } else if (key === "style") {
     for (var k in merge(oldValue, newValue)) {
@@ -111,6 +135,7 @@ var patchProperty = function(node, key, oldValue, newValue, listener, isSvg) {
     ) {
       node.removeEventListener(key, listener)
     } else if (!oldValue) {
+      node.actionMap = actionMap
       node.addEventListener(key, listener)
     }
   } else if (!isSvg && key !== "list" && key in node) {
@@ -126,7 +151,7 @@ var patchProperty = function(node, key, oldValue, newValue, listener, isSvg) {
   }
 }
 
-var createNode = function(vdom, listener, isSvg) {
+var createNode = function(vdom, listener, actionMap, isSvg) {
   var ns = "http://www.w3.org/2000/svg"
   var props = vdom.props
   var node =
@@ -137,7 +162,7 @@ var createNode = function(vdom, listener, isSvg) {
       : document.createElement(vdom.name, { is: props.is })
 
   for (var k in props) {
-    patchProperty(node, k, null, props[k], listener, isSvg)
+    patchProperty(node, k, null, props[k], listener, actionMap, isSvg)
   }
 
   for (var i = 0, len = vdom.children.length; i < len; i++) {
@@ -145,6 +170,7 @@ var createNode = function(vdom, listener, isSvg) {
       createNode(
         (vdom.children[i] = getVNode(vdom.children[i])),
         listener,
+        composeActionMaps(actionMap, vdom.children[i].actionMap),
         isSvg
       )
     )
@@ -157,7 +183,17 @@ var getKey = function(vdom) {
   return vdom == null ? null : vdom.key
 }
 
-var patch = function(parent, node, oldVNode, newVNode, listener, isSvg) {
+var patch = function(
+  parent,
+  node,
+  oldVNode,
+  newVNode,
+  listener,
+  actionMap,
+  isSvg
+) {
+  actionMap = composeActionMaps(actionMap, newVNode.actionMap)
+
   if (oldVNode === newVNode) {
   } else if (
     oldVNode != null &&
@@ -167,7 +203,7 @@ var patch = function(parent, node, oldVNode, newVNode, listener, isSvg) {
     if (oldVNode.name !== newVNode.name) node.nodeValue = newVNode.name
   } else if (oldVNode == null || oldVNode.name !== newVNode.name) {
     node = parent.insertBefore(
-      createNode((newVNode = getVNode(newVNode)), listener, isSvg),
+      createNode((newVNode = getVNode(newVNode)), listener, actionMap, isSvg),
       node
     )
     if (oldVNode != null) {
@@ -199,7 +235,15 @@ var patch = function(parent, node, oldVNode, newVNode, listener, isSvg) {
           ? node[i]
           : oldVProps[i]) !== newVProps[i]
       ) {
-        patchProperty(node, i, oldVProps[i], newVProps[i], listener, isSvg)
+        patchProperty(
+          node,
+          i,
+          oldVProps[i],
+          newVProps[i],
+          listener,
+          actionMap,
+          isSvg
+        )
       }
     }
 
@@ -220,6 +264,7 @@ var patch = function(parent, node, oldVNode, newVNode, listener, isSvg) {
           oldVKids[oldHead++]
         )),
         listener,
+        actionMap,
         isSvg
       )
     }
@@ -241,6 +286,7 @@ var patch = function(parent, node, oldVNode, newVNode, listener, isSvg) {
           oldVKids[oldTail--]
         )),
         listener,
+        actionMap,
         isSvg
       )
     }
@@ -251,6 +297,7 @@ var patch = function(parent, node, oldVNode, newVNode, listener, isSvg) {
           createNode(
             (newVKids[newHead] = getVNode(newVKids[newHead++])),
             listener,
+            actionMap,
             isSvg
           ),
           (oldVKid = oldVKids[oldHead]) && oldVKid.node
@@ -292,6 +339,7 @@ var patch = function(parent, node, oldVNode, newVNode, listener, isSvg) {
               oldVKid,
               newVKids[newHead],
               listener,
+              actionMap,
               isSvg
             )
             newHead++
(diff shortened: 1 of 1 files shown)


zaceno:
Since this is a deep change and there aren't a lot of tests I can understand if this is scary ;) 

But at least here's this demo to prove that it works for many cases (nested action maps in subs, effects and views)

https://codesandbox.io/s/hyperapp-action-mapping-demo-z3ydz

jorgebucaran:
@zaceno Can we use an object's key? People are likely to perceive it as "safer" than a string.

e.g.

```js
sliceView({ main: Main.view(state) })
```

or

```js
const main = sliceView({ main: Main.view })
// ...
main.view(state.main)
```

zaceno:
@jorgebucaran sure! This PR isn’t so much about sliceView - that’s just my own implementation, which you could chose to leave in userland or publish as a @/package. The thing we should decide on first though, is api/implementation for mapEvents and mapEffect (and perhaps mapSubs), in which my or any other slice would rely.

I guess we could make sliceView/sliceSubs the only public api to this, but there is more power in mapEvents. On the other hand I can’t think of any other use case atm.

jorgebucaran:
@zaceno I think we should expose this through a single function that handles all cases.

zaceno:
Hmm. Not sure what you mean there. So far I’ve only ever managed to get away with two, and that was presupposing the “slice” use case only

jorgebucaran:
@zaceno Get away with two what?

zaceno:
I guess it would be ok to expose just mapEvents and mapEffect. But I don’t think you can squash them to one... unless you do detection inside.

zaceno:
Two functions

jorgebucaran:
> @zaceno ...but there is more power in mapEvents. On the other hand I can’t think of any other use case atm.

Hyperapp is not about "flexibility". This API should be as hardcoded and limiting as possible. I just want to support the original use case.

zaceno:
> I just want to support the original use case.

Ok gotcha. In that case I could move more of my `slice.js` transform into core. I doubt it would be much less code though.

zaceno:
And it would still have to be two functions we export (corresponding to my `sliceSubs`  and `sliceView` impl.

jorgebucaran:
@zaceno What about a function that returns an array of "apps", each of which exposes props like the ones we pass into the `app` function:

```js
import { gizmo } from "./gizmo.js"
import { h, app, map } from "hyperapp"

const [A, B] = map({ A: Counter, B: Counter })

app({
  init: {
    ...A.init,
    ...B.init
  },
  view: state =>
    h("div", {}, [
      h("h1", {}, "100% independent gizmos:"),
      A.view(state.A),
      B.view(state.B)
    ]),
  subscriptions: state => [
    A.subscriptions(state.A),
    B.subscriptions(state.B)
  ],
})
```

zaceno:
@jorgebucaran That could work. But is it really declarative? It doesn't look very different to https://github.com/jorgebucaran/hyperapp/issues/896#issuecomment-547662315 (which requires no change in core)

I'm not entirely sold on that yet, though, for a couple of reasons:

1 - I like the explicitness of `mapEvents(slice('A'), Counter.view(state.A))`. It's not so much about the flexibility as the clear and concise intent. You understand that it is a counter that will be rendered, you understand it will you will use `state.A` to render it, and you understand that events will be scoped to state.A as well (less obvious, but still pretty clear).

2 - I understand that there is some redundancy in `mapEvents(slice('A'), Counter.view(state.A))`, and it could be more clearly expressed with something like `sliceView({A: Counter.view})`.

However, we don't really want the counter to be 100% independent from the rest of the app. While modules should encapsulate a bunch of related logic, we need to be able to get data in & out of them via some interface they expose.

While I haven't looked in to that use case much yet, I think we'll need to be able to say something like: `mapEvents(slice('A', {onmax: addAnotherCounter}), Counter.view(state.A, {max: 5)))`

jorgebucaran:
> @zaceno That could work. But is it really declarative? 

Of course it is.


> It doesn't look very different to #896 (comment) (which requires no change in core)

You're right, does that support subscriptions too?

zaceno:
> @jorgebucaran does that support subscriptions too?

It does, because the actions are defined with the action transform, straight in the module. Ie:

```js
export actionMap => {
  ...
  const increment = actionMap(state => state + 1)
  ...
  const subs = state => [ keydown({key: 'k', action: increment}) ]
  return {init, view, subs}
}
```

It works, but I seem to recall there being some issues with it. I just don't remember exactly what. I guess it makes me a bit uneasy to export a function which you must call with an `actionMap`  to get something out of it. It feels a bit "dependency-injection-ish" and thus not very declarative.

jorgebucaran:
> I just don't remember exactly what. I guess it makes me a bit uneasy to export a function which you must call with an actionMap to get something out of it. It feels a bit "dependency-injection-ish" and thus not very declarative.

Either approach is declarative (your solution and my proposal). 

> It does, because the actions are defined with the action transform, straight in the module. Ie:

Still declarative. Now, I don't like this approach. I prefer my own proposal [here](https://github.com/jorgebucaran/hyperapp/pull/899#issuecomment-549055231) where you don't have to use `mapScope` inside "Counter".

> Generally I feel a bit uneasy about making "components" or "modules" a built-in part of the framework without thinking much more deeply about how they should work.

Remember `lazy`. It was possible to emulate in userland, but we now offer it as a built-in feature because it's easier to implement in core and performs better than any userland solution. We shouldn't go by "gut" here. If modules are better served to you as a built-in feature, that's what we should do.

> how Elm solves the "module" question I think they got it quite right leaving it at Html.map and Cmd.map -- not defining a new type of "subprogram" or something.

[This](https://github.com/jorgebucaran/hyperapp/pull/899#issuecomment-549055231) is how Elm's modules look like in idiomatic Hyperapp. Elm has messages and message types, but we only have actions. We can only adapt this feature, not copy it.

zaceno:
> This is how Elm's modules would look like in idiomatic Hyperapp. Elm has messages and message types, but we only have actions. We can only adapt from this feature, not copy it.

Well I don't entirely agree (yet). So far we've just talked about 100% independent modules, but that's unrealistic. There needs to be provisions for parent-child communication, or else there's no reason to use the module in the first place.   

But how the api ends up looking in the end is of course your call. Moreover it's a question to continue in  https://github.com/jorgebucaran/hyperapp/issues/896 This PR is really just about the internal plumbing (mapping actions == Elm's "tagging messages") that is necessary to support whatever api is preferrable. If you don't want anyone to use the low level `mapEvents` et c, we can just not export them and export whatever API you prefer.

jorgebucaran:
> @zaceno There needs to be provisions for parent-child communication, or else there's no reason to use the module in the first place.

How are you going to send a message from a view that is mapped to some specific state slice? What does Elm do?

zaceno:
> How are you going to send a message from a view that is mapped to some specific state slice?  What does Elm do?

There are a couple patterns Elmers use. I opted to adapt the so called "translator pattern" in my "slice" PoC (see https://github.com/jorgebucaran/hyperapp/issues/896#issuecomment-549135564) because it didn't require changing the signatures of actions.

jorgebucaran:
@zaceno Looking at https://codesandbox.io/s/hyperapp-action-mapping-demo-z3ydz:

The amount of work that you need to do in slice.js isn't trivial. This solution offers excellent flexibility, but at the cost of complexity. Hyperapp is usually the opposite, little flexibility, but more satisfaction out of the box.

My thinking is: if I can't give you a simpler primitive, then I'd prefer a working, albeit unflexible module solution out of the box (or nothing).

johnkazer:
I had very little time to think about this, but the concept seems related to [XState actors](https://xstate.js.org/docs/guides/actors.html)? A parent can generate (dynamically) a set of children and set up communication links - the children are then able to carry out specific actions in parallel.

zaceno:
@jorgebucaran 
> My thinking is: if I can't give you a simpler primitive, then I'd prefer a working, albeit unflexible module solution out of the box (or nothing).

Sorry, I must have misunderstood what you meant by "simplest primitive". I thought you were asking basically: "what is the minimal addition I can make to hyperapp, to _enable_  modules such as we have discussed". 

To that, the answer is: this PR. 

My `slice` function that I use in the demos isn't included it in this PR because I just wanted to add the necessary plumbing. I made the demos with `slice` to ensure that the plumbing is sufficent to support the module use case. With my latest version (using the translator pattern), I feel satisfied that is the case. A generally useful and workable module-system could be built on top of this PR. 

But as you point out, that will require quite some work on the user's part, or a third-party library.

Now, I think I understand you're asking: "what is the most straightforward, developer friendly way I can offer modularity, with near-zero userland boilerplate or helpers needed?" 

To that, my answer is: I don't know yet. The options I see (besides "forget the whole thing") are:

0) 

Merge this PR as is

pro:
  - Modules are now possible
  - it keeps core conceptually clean (just plumbing with no opinion on how to use it)

con:
  - It is _far_ from obvious how `mapEvents`, `mapEffects` and `mapSubs` work, and how they enable modules.
  - Users will need to write their own action transforms, which is non-trivial.

1)
We can export my `slice` function alongside `mapEvents/Effect/Subs`. Either from core or a @/ package.

pro:
  - The general power of transforming actions for other cases than modules (???) is still there
  - users don't have to write their own slice function. And virtually everyone will want the same one anyway

con:
  - There will be a lot of `mapEvents(slice(...` repeated everywhere in the code, and you will probably never use the `map*` functions with anything _but_  slice

2)
We can combine my `slice` implementation with `map*`, and _instead_ export `sliceView`, `sliceSubs` and `sliceAction` (either from hyperapp directly or a @/ package)

The signatures would be something like:

```
sliceView(
  'user',
  User.statusMenu(
    state.user,
    {theme: 'dark'}
  ),
  {onLogOut: sliceAction('messages', Messages.notifyLoggedOut)},
)
```

```
sliceSubs(
  'messages',
  Messages.subs(
    state.messages,
    {doNotDisturb: User.getSetting(state.user, 'doNotDisturb')}
  ),
  {onNewMessage: PopUpNotification },
)
```

pro:
- It eliminates the unnecessary flexibility of the `mapXYZ` functions (since modules based on named branches in the state tree is the only use case for them I can even come up with)
- It eliminates the repetiton of `mapXYZ(slice(...`
- Conceptually, it targets the module-use-case we're after more clearly and directly

con:
  - Using string as name is felt to be "unsafe".
  - Perhaps perceived as redundant to have both the slice name `messages`, and then later in the same expression refer to `state.messages` (you will never put `state.foobars` there)
  - The explanation of how  `sliceView/Subs/Action` functions work is a little more involved than `mapEvents/Subs/Effect`.

3) 
Instead of `sliceView/Subs/Action`, we could export a single, unified and more elaborate abstraction. It could let you specify slice names using object-keys. We could pre-bind functions in modules so you don't need to explicitly write `sliceView('messages', ...` and:  you wouldn't have redundant expression of slice name and substate in same expression.

pro:
- All the pros of (2)
- Fixes all but the last cons of (2)

con:
- Using views et c that "bound" to to a slice this way may be perceived as too implicit. Perhaps even "magical" or "confusing".
- It will likely need to rely on conventions in how modules are written (are subs named "subs" or "subscriptions"?)
- There will likely be much disagreement about the API, and attempts to improve on it, due to the level of abstraction.
- Describing how this works and fits into the TEA will be an even greater challenge than (2)


For now, my money is on (2). But as you see, all these solutions build upon eachother. It's just a question of which level of abstraction we export, and from which package.  Would you like me to update this PR to (2)?  If so would you like `sliceAction/View/Subs` exported from a `lib/...` or `src/index.js` ?

zaceno:
@jorgebucaran

The api, if any, that we chose is of course up to you.

If you chose nothing (rejecting this PR), then we all still have the option of mapping actions in userland (either by processing vnodes, or by dynamically defining actions using the method:

```
export default actionMap => {

  const Increment = actionMap(x => x + 1)
...
``` 
)

Since you're planning to make a Hyperapp version of the RealWorld app anyway, you could look into that to figure out what api is most appealing to you.

zaceno:
@johnkazer I don't know enough about XState to say, but I think what we're going for here is more than that. It's not just that we want (mostly-)independent sub-state machines: it's mainly about "routing" the actions dispatched from views, effects and subs to the correct "sub-state-machine".

jorgebucaran:
>  it's mainly about "routing" the actions dispatched from views

I think this is a very simple way to put it. 👍

jorgebucaran:
@zaceno We need a simpler solution for Hyperapp. As hard as it may be, let's try to define an API, then come back here. 

#896 talks too much about implementation details, so we should create another issue to just talk about the  API.

zaceno:
@jorgebucaran 
> We need a simpler solution for Hyperapp.  

I don't disagree, but I also don't understand what you mean by "simpler". I understand it probably means _less of something_, but what is _something_?  Less code for the user to write? Lesser cognitive load when users are using it? Fewer exports from hyperapp? Less plumbing and core-changes to work?  

> As hard as it may be, let's try to define an API, then come back here.

Sounds good, I'm happy to help think about good APIs. But in order to help, please elaborate on what qualities it should have -- What "simpler" means  in this case.

zaceno:
So I'm working on implementing this api (more seriously this time, with TDD so I will be able to optimize the implementation with full confidence all cases work) and I ran into a little dilemma/issue with subs.

basically if you have:

```
mapSlice = (name, x) => map(s => s[name], (s, y) => ({...s, [name]: y}), x)
...
subscriptions: state => [
  state.listenToA
  ? mapSlice('A', Messages.subs(state.A))
  : mapSlice('B', Messages.subs(state.B))
]
```

Then there is no way I can figure out how the subscription-patcher will be able notice that the mapper has changed and stop the A-subs and start the B-subs when it should. Not without having subs restarted all the time.

It's probably just a minor issue, since quite likely some of the props will have changed between `Messages.subs(state.A)` and `Messages.subs(state.B)`, so the subs will be restarted properly anyway. But in the unlikely event that `Messages.subs(whatever)` always returns the exact same props, we will not be able to know that it has changed.

The workaround is to introduce an artificial property. Like define `subs` in `messages.js` as

```
export const subs = (state, instance) => [
  SomeSub({foo: 'always the same', instance}),
  ...
]
```

`SomeSub` doesn't need to care about it, just as long as it changes, then subs will restart ok.

So your subs would have to look like this:

```
subscriptions: state => [
  state.listenToA
  ? mapSlice('A', Messages.subs(state.A, 'A'))
  : mapSlice('B', Messages.subs(state.B, 'B'))
]
```

... and it will all work fine.

Like I said, I find the scenario highly unlikely, so I don't think we need to shut everything down because of it. But it's good to be aware of this gotcha in case it comes up.

zaceno:
Another thing I haven't yet considered is how this will will work with `Lazy`... probably similarly to subs, in that we shouldn't consider new action-maps to invalidate the memo (because that would break Lazy completely)

jorgebucaran:
@zaceno I still trying to understand the problem, but why:

> incidentally, if we had gone with "strings" to identify slices, this would have not been an issue

?

zaceno:
@jorgebucaran Not saying we should have gone with "strings" -- I much prefer the current api proposal. But if we *had*, we would have no problem identifying that the same subFn, with the same props, only mapped with a different map-definition should be restarted. If we try to check against a map defined with functions, it will be different each time (and subs would be restarted each time) because:

`(state => state.foo) !== (state => state.foo)` but `"foo" === "foo"`

zaceno:
Latest update is basically same implementation as before, but now all we export is `map` with the signature `map(extractFn, mergeFn, action|vnode|subs)`  where: 

- `extractFn` could be `state => state.foo` (returns the slice of full state action should operate on)
- `mergeFn` could be `(state, slice) => ({...state, foo: slice})` (how to put the returned new slice value  into the right place of the total state)

I have tried to get some very light golfing in, but this implementation still adds  222 bytes to the gzipped bundle. Hope we can improve upon that. I've noticed  apparent duplication between the `resolve` function I added, and `dispatch`, but I don't see how to reduce it

On the other hand, I'm quite confident the solution is robust, as I've run it against a battery of tests continually during development. I can't include the tests in the PR, sadly, as they are ava+undom based - not testmatrix (not familiar enough yet). But they exist in case anyone wants them :)

src/index.js
```diff
@@ -393,7 +448,7 @@ var recycleNode = function(node) {
     : createVNode(
         node.nodeName.toLowerCase(),
         EMPTY_OBJ,
-        map.call(node.childNodes, recycleNode),
+        EMPTY_ARR.map.call(node.childNodes, recycleNode),
```
jorgebucaran: @zaceno Why?

src/index.js
```diff
@@ -157,7 +203,9 @@ var getKey = function(vdom) {
   return vdom == null ? null : vdom.key
 }
 
-var patch = function(parent, node, oldVNode, newVNode, listener, isSvg) {
+var patch = function(parent, node, oldVNode, newVNode, listener, actionMap, isSvg) {
+  actionMap = compose(actionMap, newVNode.actionMap)
```
jorgebucaran: @zaceno This could affect performance. Can we avoid this?

src/index.js
```diff
@@ -3,8 +3,8 @@ var LAZY_NODE = 2
 var TEXT_NODE = 3
 var EMPTY_OBJ = {}
 var EMPTY_ARR = []
-var map = EMPTY_ARR.map
 var isArray = Array.isArray
+var isFunc = function(x) { return typeof x === "function" }
```
jorgebucaran: @zaceno I'd prefer to avoid this function and simply inline every `typeof x === "function"` for now.

src/index.js
```diff
@@ -46,15 +46,59 @@ var batch = function(list) {
...
+}
+
+export var map = function(extract, merge, x, actionMap = makeActionMap(extract, merge)) {
+  return isArray(x)
+    ? mapEffects(x, actionMap)
+    : isFunc(x)
+      ? actionMap(x)
+      : (x.actionMap = actionMap, x)
+}
+
```
jorgebucaran: @zaceno Can this live in a separate module? `@hyperapp/x` or whatever?

src/index.js
zaceno: I don't think this will affect performance badly, because if you look at how `compose` is defined, there will only be a new function created when necessary (i e only once for each time `map` is used in the vdom).

We can't avoid it unfortunately, without taking the entire implemation of `map` out of core. I already demonstrated before that it is possible to implement the entire thing in userland. The problem with that is that it is quite inefficient since the vdom needs to be processed once to compose action maps, and then sent into `patch` for processing again.

src/index.js
zaceno: There was already a function called `map` in `src/index.js` which was just a proxy to `EMPTY_ARR.map`. That `map` was only used once. So when I was going to define my own function named `map`, I got rid of the old definition and replaced the usage with `EMPTY_ARR.map` directly.

src/index.js
zaceno: Can do!

src/index.js
zaceno: possibly yes. I think we could move out `map` and at least the bulk of `makeActionMap`. But where should I move it? 

Also, if it is going in a separate module, perhaps we should rethink the idea of exporting just a single function?

At least, the current `map` is a little funky since we overload it with different behavior depending on type of argument in, and putting it in a separate module would give us more freedom to export multiple functions for specific purposes.

jorgebucaran:
I'd love to revisit this idea, just not at this time. Maybe for V3! ✨ 

Let's continue the conversation on a discussion? 👋

PR #901, opened 2019-12-12
use MC instead of rAF

yisar:
requestAnimationFrame will be paused in when running in background tabs, Message channel will not.
This may be the best choice.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,10 +5,15 @@ var EMPTY_OBJ = {}
 var EMPTY_ARR = []
 var map = EMPTY_ARR.map
 var isArray = Array.isArray
-var defer =
-  typeof requestAnimationFrame !== "undefined"
-    ? requestAnimationFrame
-    : setTimeout
+var defer =(() => {
+  if (typeof MessageChannel !== 'undefined') {
+    const channel = new MessageChannel()
+    const port = channel.port2
+    channel.port1.onmessage = render
+    return () => port.postMessage(null)
+  }
+  return () => setTimeout(render)
+})()
 
 var createClass = function(obj) {
   var out = ""
@@ -449,7 +454,7 @@ export var app = function(props) {
       if (subscriptions) {
         subs = patchSubs(subs, batch([subscriptions(state)]), dispatch)
       }
-      if (view && !lock) defer(render, (lock = true))
+      if (view && !lock) defer((lock = true))
     }
     return state
   }


jorgebucaran:
@yisar Interesting alternative!

sergey-shpak:
@yisar 
There is no need to redraw any tab while it's in background (that is why 'requestAnimationFrame' is paused by browser), therefore you shouldn't try to redraw it with your app view. (As soon as tab becomes active 'requestAnimationFrame' is triggered and app view is rendered)

App view is only for reflecting state into dom, that's it. If you need to process something while your app is in background tab, you have to use subscriptions/effects to update state, as result re-render view. 

*Main purpose of `requestAnimationFrame` defer is to sync 'view' render with browser screen redrawing, as result getting better app performance. Btw, there is no any benefit of MC defer example comparing to simple 'setTimeout' defer.

yisar:
> Btw, there is no any benefit of MC defer example comparing to simple 'setTimeout' defer.

It can reduce time waste as much as possible, compared with setTimeout or PostMessage.

> Main purpose of requestAnimationFrame defer is to sync 'view' render with browser screen redrawing, as result getting better app performance. 

I don't think performance is the key point? The key is that we should operate DOM less later and more safely?

sergey-shpak:
@yisar performance is the key point, otherwise we wouldn't add 'requestAnimationFrame' defer.
> It can reduce time waste as much as possible, compared with setTimeout or PostMessage.
We don't want reduce time waste, we want to get rid of unnecessary view renders, please read my prev message once more

src/index.js
```diff
@@ -5,10 +5,15 @@ var EMPTY_OBJ = {}
 var EMPTY_ARR = []
 var map = EMPTY_ARR.map
 var isArray = Array.isArray
-var defer =
-  typeof requestAnimationFrame !== "undefined"
-    ? requestAnimationFrame
-    : setTimeout
+var defer =(() => {
```
okwolf: Using an arrow function here would drop support for some of the older browsers.

yisar: > 在此处使用箭头功能会放弃对某些旧版浏览器的支持。

Yeap, it should be function(), but this PR seems can not merged

jorgebucaran:
@yisar Please read @sergey-shpak's comment again https://github.com/jorgebucaran/hyperapp/pull/901#issuecomment-564982282.

yisar:
@jorgebucaran I see, so can I closed this PR?

sergey-shpak:
@jorgebucaran let's close this

jorgebucaran:
Yup! Thank you @yisar @sergey-shpak! 👋

PR #904, opened 2019-12-22
Fix bug  https://github.com/jorgebucaran/hyperapp/issues/903

gbagan:
### Motivation

Like I said in the bug report https://github.com/jorgebucaran/hyperapp/issues/903 ,
the render function of Hyperapp has problems with subtrees shared into a vdom. There is also a problem for a part of vdom keeped between two renders.
I don't see immediately realistic programs that can provoke the bug.
One of the problems I see is that it breaks referential transparency.
Referential transparency is an important concept in FP.
See for example: https://www.sitepoint.com/what-is-referential-transparency/
That means that if we replace somewhere in the execution the call of a function by its resullt, the behaviour of the program remains the same.
I also think that if someone in the future will be confronted to this bug, he/she will hard time to find it.

my PR proposes to solve it but keeps optimisation by reference equality (see below)

### Implementation details

I think the only way to permit referential transparency is to not mutate the vdom returned by the function view. It is must be immutable.
So in this PR, I don't store nodes into vnodes. Instead I create a structure ptc (the name is temporary) that makes the link between nodes and vnodes.
The structure is an array of 3 elements [node, ptc children, vnode]
I have modified the algorithm patch and other functions such that they use this structure.
They can mutate the structure but does not mutate the vdom.
This structure is kept for comparaison in the next render instead of vdom.
Some functions (createVDom, patch) have been modified to return a ptc instead of node.

### Changes
- the app now stores a ptc instead of a vdom
- the signature of patch is now  function(parent, node, oldPtc, newVNode, listener, isSvg)
      the third argument is a ptc instead of a vnode
      it returns a ptc instead of a node
- createNode returns a ptc
- recycleNode returns a ptc

### Remarks about the PR
- I think it totally solves the problem of shared nodes and thus the problem of referential transparency
- I think all optimisations (subtree skipping) remain valid
- the diff is quite big but this is mainly a straighforward adaptation of the existing algorithms.
- the total number of lines is nearly the same.
    I have splitted some lines of code but it can be merged again.
- I think the performance is nearly the same as the current version of hyperapp (unless I have written a bug)
- Elm uses a similar approach but a lot more tricky.

### alternative approachs
make a shallow/deep copy of the vdom
It can be done in different ways
- in the function h (it may be not sufficient if the shared nodes are deep in the tree)
- in the function getVNode
- a deep copy before call the render
That approach solves the problem but the implementation can be tricky if we want to keep optimisation by reference equality.
And the problem to decide what we keep as previous vdom has not a straighforward answer
- if we keep the original vdom, all node informations are lost
- if we keep the copied vdom, we lost optimisation by reference equality and thus tree skiping.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -179,6 +179,7 @@ var patch = function(parent, node, oldPtc, newVNode, listener, isSvg) {
 	return oldPtc
   } else {
     var tmpPtcKid
+    var tmpVKid
     var oldPtcKid
 
     var oldKey
@@ -276,11 +277,9 @@ var patch = function(parent, node, oldPtc, newVNode, listener, isSvg) {
       }
 
       while (newHead <= newTail) {
-		oldPtcKid = oldPtcKids[oldHead]
+        oldPtcKid = oldPtcKids[oldHead]
         oldKey = getKey(oldPtcKid && oldPtcKid[2])
-        newKey = getKey(newVKids[newHead])
-		  //(ptcKids[newHead] = getVNode(, oldPtcKid))[2] // todo
-        //)
+        newKey = getKey(tmpVKid = getVNode(newVKids[newHead], oldPtcKid && oldPtcKid[2]))
 
         if (
           newKeyed[oldKey] ||
@@ -299,7 +298,7 @@ var patch = function(parent, node, oldPtc, newVNode, listener, isSvg) {
               node,
               oldPtcKid && oldPtcKid[0],
               oldPtcKid,
-              newVKids[newHead],
+              tmpVKid,
               listener,
               isSvg
             )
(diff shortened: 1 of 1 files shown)


gbagan:
The unit test because I have removed node to createVNode since it does not make sense with this PR but I can llet it as a dummy variable if you want.

src/index.js
```diff
@@ -362,42 +369,44 @@ var getTextVNode = function(node) {
 }
 
 var getVNode = function(newVNode, oldVNode) {
-  return newVNode.type === LAZY_NODE
```
gbagan: not really. I have accenditally change the indentation

PR #906, opened 2019-12-22
Lazy view fix

sergey-shpak:
https://github.com/jorgebucaran/hyperapp/issues/905
Since `oldVNode` is the only result of lazy view and not the lazy node itself, `oldVNode.type !== LAZY_NODE` check leads to constant lazy view re-rendering. Changing check to '!oldVNode.lazy'

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -363,10 +363,8 @@ var getTextVNode = function(node) {
 
 var getVNode = function(newVNode, oldVNode) {
   return newVNode.type === LAZY_NODE
-    ? ((!oldVNode ||
-        (!oldVNode.lazy ||
-          propsChanged(oldVNode.lazy, newVNode.lazy))) &&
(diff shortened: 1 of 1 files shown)


okwolf:
Performance results look good for this change:
Screen Shot 2019-12-22 at 20 39 01

Columns are:

- code from this PR
- code from before the `Lazy` regression
- code from `master` without `Lazy` for comparison

jorgebucaran:
This is awesome, thank you @okwolf! To clarify, did you run all the tests (the 3 different tests I see above)?

Also, what are the results for this PR vs what's currently on master (using lazy)?

okwolf:
@jorgebucaran added results for using `<Lazy>` against master. All test results are on my machine without anything else running.
Screen Shot 2019-12-24 at 8 10 49

okwolf:
@jorgebucaran any more benchmarks or analysis needed before merging this?

jirutka:
@jorgebucaran, can you please release a new version with this fix?

PR #907, opened 2019-12-28
Fix link from tutorial to api reference

danprince:
Can see that the API reference is in progress, so might be intentionally busted. Feel free to close, if so.


jorgebucaran:
Thanks! 👍

PR #913, opened 2020-01-29
Adds CNAME to site

sergey-shpak:
SInce CNAME is [excluded at docs action deployment step](https://github.com/jorgebucaran/hyperapp/pull/909#issuecomment-579709033), 
CNAME should be added manually

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+hyperapp.dev


sergey-shpak:
missed `site` branch

PR #914, opened 2020-01-29
Adds CNAME

sergey-shpak:
SInce CNAME is [excluded at docs action deployment step](https://github.com/jorgebucaran/hyperapp/pull/909#issuecomment-579709033),
CNAME should be added manually
*this is pull request to 'site' branch

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+hyperapp.dev


jorgebucaran:
Do we need to manually add this every time we want to deploy the site? 😨

sergey-shpak:
@jorgebucaran good question, only once, but action should be reconfigured (do not to clean branch before deployment)

PR #921, opened 2020-02-11
Adds docs folder

sergey-shpak:
Fix for docs actions

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -19,3 +19,4 @@ jobs:
       with:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         BRANCH: gh-pages
+        FOLDER: docs/dist


jorgebucaran:
@sergey-shpak Where is this `dist` directory?

sergey-shpak:
`dist` directory created by docs script `build` (which is triggered by `npm run build-site`):
https://github.com/jorgebucaran/hyperapp/blob/master/docs/package.json#L9

jorgebucaran:
@sergey-shpak So there's actually never a `dist` directory in any branch, right?

sergey-shpak:
yes, you are right, there is no any branch with `dist` directory (you can browse it only on local env, when building docs)

jorgebucaran:
@sergey-shpak But it exists in GitHub servers when they build the page, and before they copy the files to `gh-pages`, correct?

sergey-shpak:
@jorgebucaran yes, correct

PR #928, opened 2020-02-29
fix #919

### LINKED ISSUE #919: Website doesn’t work in Firefox 60.8.0 ESR
I see just a blank page and the following error in the console:

```
TypeError: Object.fromEntries is not a function[Learn More]
```

jaeh:
replace Object.fromEntries with an entries function,
fixes #919 

this should reduce required firefox version down to 47 according to caniuse:
https://caniuse.com/#search=entries

### DIFF
--- a/docs/src/utils.js
+++ b/docs/src/utils.js
@@ -1,3 +1,17 @@
+export const entries = (params) => {
+  if (typeof Object.fromEntries === 'function') {
+    return Object.fromEntries(params)
+  }
+
+  const obj = {}
+
+  params.forEach(([key, val]) => {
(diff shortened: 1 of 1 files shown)


jorgebucaran:
This is great! Thank you, @jaeh! 🙌

jirutka:
It doesn’t work, now I’m getting: `TypeError: r.forEach is not a function` in Firefox 60. And since you don’t even deploy source maps, I cannot provide more details without extra effort digging in minified code.

jorgebucaran:
Would you like to help with this, @jirutka?

PR #931, opened 2020-03-03
Use for of instead of forEach

loteoo:
Fix rendering bug in Edge.

Apparently the params object was an Iterator in edge, but not an array? Not sure why, but this works!

Tested with Microsoft Edge 44

### DIFF
--- a/docs/src/utils.js
+++ b/docs/src/utils.js
@@ -5,9 +5,9 @@ export const entries = (params) => {
 
   const obj = {}
 
-  params.forEach(([key, val]) => {
+  for (const [key, val] of params) {
     obj[key] = val
-  })
+  }
 
   return obj
 }


zaceno:
Does this mean we're finally dropping IE support? 🤞 😁

johnkazer:
Will the [ie support in edge](https://docs.microsoft.com/en-us/deployedge/edge-ie-mode) be compatible?

zaceno:
@johnkazer Sorry to cause confusion -- I was actually only joking (but my joke was based on an initial misunderstanding on my part what this PR is for). This pr will only break the hyperapp website (https://hyperapp.dev) on IE11. Regular hyperapps will continue to work in IE. 

The website will begin to work in Edge, without IE-compatibility mode. Turning on IE-compatibility mode for the site will _probably_ also work but why do that, if it woks on "regular" Edge anyway.

jorgebucaran:
What is it that we are trying to do here that can't run on IE? Think about it, Hyperapp itself can run in IE10, but we need at least Edge to run this function? Sounds like challenge.

zaceno:
@jorgebucaran `for of` is not supported in IE, that's all. It only affects the website and I highly doubt anyone will be visiting the website using IE, even if they are a developer who cares about IE-support. So we're all good here.

loteoo:
I can do some more testing on IE tonight before merging this, what bothers me is that parcel should do some transpilation by default for all this.

https://parceljs.org/transforms.html

Actually I'll play around with adding a .babelrc file and see what happens

zaceno:
@loteoo Ah, right. Parcel - totally forgot about that.

What if you just got rid of the code change and also the `.babelrc` (parcel already knows how to turn jsx into hyperapp's `h` without config). And just add a `"browserslist": ">1%, ie11"` to `package.json`?

jirutka:
It’s not broken only in IE, but also in Firefox 60.8.0 ESR.

@zaceno, once again https://github.com/jorgebucaran/hyperapp/issues/919#issuecomment-585366611:
> The problem is that `parcel-bundler 1.12.4` (the latest stable) depends on outdated `core-js ^2.6.5` that doesn’t support `Object.fromEntries` yet (it’s supported since 3.0.0). That’s a quite good example of why these “magic” all-in-one tools are better to avoid…

And also https://github.com/jorgebucaran/hyperapp/issues/919#issuecomment-581434199.

It’s really unbelievable that authors of such great web framework are not able to create a website that actually works in quite recent web browsers… :roll_eyes:

jorgebucaran:
Sorry, I haven't had time to look into the problem. 

I think we should just stop using `for of` or any of that rubbish and simply write our own `entries` function.

Would you like to help with that @jirutka?

> It’s really unbelievable that authors of such great web framework are not able to create a website that actually works in quite recent web browsers… 🙄

You'd be surprised! 😄

selfup:
Wouldn't [Array.from](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) fix this?

`Array.from(params).forEach(...)`

It's a shallow copy but should work for these purposes.

jaeh:
i introduced this. my linked pull request above fixes this.

jorgebucaran:
Do we still need this? @jaeh @loteoo

jaeh:
this should be obsolete after #934, 
i passed the searchParams.entries() into the entries function which caused all of this,
now i just pass the searchParams and call .entries in Object.fromEntries(params.entries()) instead, which leads to params being an array if fromEntries does not exist.

PR #934, opened 2020-03-09
only use entries if we are in a modern browser.

jaeh:
i used searchParams.entries() in the entries function call.

this made the argument passed to entries an iterator.

fixed this by only using the entries when we need them because Object.fromEntries exists.

fixes: #919, makes #931 obsolete.

btw, this time i downloaded firefox 60 and actually tested it.


jorgebucaran:
Yay! Thank you, @jaeh!

PR #938, opened 2020-04-04
Remove lib (@hyperapp/*) scoped packages

jorgebucaran:
Now it's up to you to create your own effects and subscriptions.

The intention is to give you more freedom in exchange for some initial user-friendliness and make Hyperapp easier to maintain in the long term.

We still encourage you to create reusable effects and subscriptions, rather than ad-hoc solutions to solve specific problems.


cpitclaudel:
Hi @jorgebucaran.  Thanks a lot for your work.

Given that there are lots of tutorials on the web that point to `lib/` (including https://medium.com/hyperapp/a-walk-through-hyperapp-2-b1f642fca172 , this bug tracker), and preciously few examples of how to wire events yourself, it would be nice to resurrect `lib/`, at least as examples.  The current situation is pretty confusing.

One way could be to have `lib/` contain just a README and no files; the README would point to an `examples/` directory containing the old `lib/` code.

jorgebucaran:
@cpitclaudel You are right. What are currently considering is bringing them back, but slowly and experimentally. While Hyperapp itself is essentially done, getting right each lib/module API is going to take a while.

See the notes under _experimental_ in https://github.com/jorgebucaran/hyperapp/pulls.

cpitclaudel:
:+1: Thanks a lot for your hard work.

jorgebucaran:
Thank you so much! 💯

jarednorman:
Why were the packages unpublished from npm as well? We had a project using the `events` package and while it was trivial to resolve ourselves, it was super annoying to have all our builds broken by the npm package getting unpublished.

jorgebucaran:
You are right, I should've deprecated it instead. Glad to know you resolved the issue, though.

jorgebucaran:
You are right, I should've deprecated it instead. But I'm glad to know you resolved the issue!

jorgebucaran:
You are right, I should've deprecated it instead. But I'm glad to hear the issue has been resolved!

PR #946, opened 2020-05-05
Read-only form attribute fix

kwasniew:
HTMLButtonElement.form and HTMLInputElement.form are read-only but can be set using setAttribute. 

Example of a failing case:
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <script type="module">
        import {h, app} from "https://unpkg.com/hyperapp?module"

        app({
            init: 0,
            view: state =>
                h("button", {form: "failing"}, "irrelevant"),
        })
    </script>
</head>
<body>
<main id="app"></main>
</body>
</html>
```

I found a similar solution in preact: https://github.com/preactjs/preact/blob/master/src/diff/props.js#L113

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -113,7 +113,7 @@ var patchProperty = function(node, key, oldValue, newValue, listener, isSvg) {
     } else if (!oldValue) {
       node.addEventListener(key, listener)
     }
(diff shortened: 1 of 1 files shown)


jorgebucaran:
Thank you, @kwasniew! 🎉

PR #953, opened 2020-05-19
The tiny framework for building hypertext applications

jorgebucaran:
## tl;dr

Prepare for the official V2 release.

## What's changed?

> This PR is still missing tests. They're coming. Please hold! 😉

### Performance

- Bytes: down to ~~1779~~ **1697** gzipped (1528 brotlified).
- Render performance should be about the same, but memory usage and startup metrics should improve a little.

### Hot

- Use modern JavaScript, arrow functions, blazing fast spread merge, ES modules, etc.
    - Drop out-of-the-box IE support (you may still compile Hyperapp to ES3 so it works on >=IE10).
- Make Hyperapp an ES module by default and drop common JS support.

### New

- Introduce new `text` function to create text nodes. Wherever you previously had a string like "hello world", you need to change it to `text("hello world")`.
  - This drops built-in JSX support. If you were JSX user, you'll need a dedicated babel plugin to compile JSX to Hyperapp's hyperscript signature or use a tagged template literal function like @zaceno's https://github.com/zaceno/hyperlit!
- Rename `Lazy` function to `memo`.
- Initial state is now `undefined` (it was an empty object by default).
  - This seemingly trivial feature enables you to set initial props inside `init` using [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) to initialize the state elegantly.

```js
app({
  init: (
    dimensions = {
      width: window.innerWidth,
      height: window.innerHeight,
    }
  ) => [
    {
      dimensions,
      flipX: false,
      flipY: false,
      ...defaultState,
    },
    loadSprite(TextureLoaded, {
      magnify: dimensions.width / 2,
      minify: dimensions.height / 2,
    }),
  ],
})
```

### Payloads

- Remove payload creators.
  - Payload creators were generally used to create a new payload to include the original user payload and the event object from an event handler or subscription.
  - However, that since actions are allowed to return other actions, it's possible to solve the above problem using a function like so:

```js
const withTargetValue = (fn) => (_, event) => fn(event.target.value)

// ..

withTargetValue((value) => [UpdateItem, { id, value }])
```

<a name="experimental"></a>
### Experimental

- Experimental: return `dispatch` from `app`.
    - Why? Makes it easier to implement kill in userland [https://github.com/jorgebucaran/hyperapp/pull/873]
- Tentatively bring back `lib`.
  - @hyperapp scoped packages will take time to develop, so create your own effects and subscriptions or use @okwolf's [FX](https://github.com/okwolf/hyperapp-fx) library.
  - Current priority here is shipping `@hyperapp/html` with HTML utility functions (`div`, `section`, `p`, etc.).


icylace:
In `patchSubs` the loop's OR conditional seems to consistently reference the same length value because, to my understanding, neither `oldSubs` nor `newSubs` changes their length during the loop:

https://github.com/jorgebucaran/hyperapp/blob/ff09446ec85b8090f52791f31c103e51377e525f/src/index.js#L59

Wouldn't you be able to optimize away redundant checking? So, instead of

```js
for (
  var subs = [], i = 0, oldSub, newSub;
  i < oldSubs.length || i < newSubs.length;
  i++
) {
```

how about this

```js
var length = Math.min(oldSubs.length, newSubs.length)

for (
  var subs = [], i = 0, oldSub, newSub;
  i < length;
  i++
) {
```

or maybe just

```js
var length = oldSubs.length < newSubs.length ? oldSubs.length : newSubs.length

for (
  var subs = [], i = 0, oldSub, newSub;
  i < length;
  i++
) {
```

Disclaimer: I have not tested this and am no performance expert.

icylace:
Since the new dispatcher doesn't return a value:

https://github.com/jorgebucaran/hyperapp/blob/ff09446ec85b8090f52791f31c103e51377e525f/src/index.js#L426

compared to its 2.0.4 version:

https://github.com/jorgebucaran/hyperapp/blob/2ba417b14765ba38214a89b3b514a06cdfb53258/src/index.js#L459

maybe the default "no middleware" function can get rid of the `return` keyword to save a few bytes:

https://github.com/jorgebucaran/hyperapp/blob/ff09446ec85b8090f52791f31c103e51377e525f/src/index.js#L423

jorgebucaran:
@icylace I tried golfing that in various ways, but we still end up gaining +10 bytes or so. Thanks for the suggestion, though!

Perf-wise, it doesn't make much difference, so I'm going to favor bytes here.

icylace:
Payload creator functions were removed too, correct ?

src/index.js
```diff
@@ -352,135 +323,106 @@ var patch = function(parent, node, oldVNode, newVNode, listener, isSvg) {
...
-      children.push(getTextVNode(vdom))
-    }
-  }
+var h = (type, props, children) =>
+  createVNode(
+    type,
+    props,
+    isArray(children) ? children : children == null ? EMPTY_ARR : [children],
+    null,
+    props.key
```
diontools, line 374: In Babel's JSX, the `props` may be null, so a null error occurs.

jorgebucaran: Thanks for pointing that out. We're dropping out-of-the-box JSX support. So, you'll need a dedicated babel plugin to compile JSX to Hyperapp's hyperscript or use a tagged template literal function like @zaceno's https://github.com/zaceno/hyperlit.

zaceno: @jorgebucaran @diontools 

Technically, you could still continue to use `babel-plugin-transform-react-jsx` for your jsx. You just can't use hyperapp´s built in `h` pragma any more. You need a custom pragma as a go-between what JSX compiles to, and what the new `h` will accept

jorgebucaran: Ah, yes, that'd work too!

diontools: @jorgebucaran @zaceno Thanks!

selfup:
Yea from a perf perspective `.length` shouldn't make much difference since it is a computed property and not evaluated when called. It is updated as the array changes, not calculated on the fly. Good to know that holds true in testing!

jimfranke:
@jorgebucaran Now that all vnodes are lazy evaluated, where previously the `h` function eagerly removed boolean and null vnodes, possibly a lot of empty text nodes will be inserted in case of boolean or null, correct?

jorgebucaran:
> @jimfranke Now that all vnodes are lazy evaluated... where previously the h function eagerly removed boolean and null vnodes

I guess you could say boolean and null nodes are lazily evaluated, yes, but it isn't like we're going to lazy/memo everything by default now (we have no plans to do that), which is how I read your comment. Just clarifying for everyone.

> @jimfranke ...possibly a lot of empty text nodes will be inserted in case of boolean or null, correct?

Correct.

jorgebucaran:
I'd like to merge here, create a pre-release and work on adding tests on another PR. Also missing is revising the tutorial and API reference.

PR #954, opened 2020-05-31
TypeScript definitions for the upcoming release

icylace:
I've updated and adapted my original TypeScript definitions to now align with Hyperapp's next release.

I've not done much testing since I completed these types very recently but so far it works.

Any suggestions welcome !

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -70,8 +70,16 @@ declare module "hyperapp" {
   }
 
   // Virtual DOM properties will often correspond to HTML attributes.
-  type PropList = Record<string, Prop> & Readonly<{ key?: Key }> & ElementCreationOptions
   type Prop = bigint | boolean | null | number | string | symbol | undefined | Function | ClassProp | StyleProp
+  type PropList = ElementCreationOptions & {
+    [k: string]: Prop;
+    readonly class?: ClassProp;
+    readonly key?: Key;
+    readonly style?: StyleProp;
+  }
+
+  // A key can uniquely associate a virtual DOM node with a certain DOM element.
+  type Key = null | string | undefined
 
   // The `class` property represents an HTML class attribute string.
   type ClassProp = string | Record<string, boolean> | ClassProp[]
@@ -99,9 +107,6 @@ declare module "hyperapp" {
     Lazy = 2,
     Text = 3,
   }
-
-  // A key can uniquely associate a virtual DOM node with certain a DOM element.
-  type Key = null | string | undefined
 
   // -----------------------------------------------------------------------------
 
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,0 +1,121 @@
+declare module "hyperapp" {
+  // A Hyperapp application instance has an initial state and a base view.
+  // It must also be mounted over an available DOM element.
+  type App<S, P, D> = Readonly<{
+    init: Action<S, P, D>;
+    view: View;
+    node: Node;
+    subscriptions?: Subscription;
+    middleware?: Middleware;
+  }>
+
+  // A view builds a virtual DOM node representation of the application state.
+  type View = <S>(state: State<S>) => VDOM
+
+  // Application state is accessible in every view, action, and subscription.
+  type State<S> = S
+
+  // A subscription is a set of recurring effects.
+  type Subscription = <S>(state: State<S>) => Subscriber[]
+
+  // A subscriber reacts to subscription updates.
+  type Subscriber = boolean | void | Effect | Unsubscribe
+
+  // A subscriber ideally provides a function that cancels itself properly.
+  type Unsubscribe = () => void
+
+  // Middleware allows for custom processing during dispatching.
+  type Middleware = (f: Dispatch) => Dispatch
+
+  // -----------------------------------------------------------------------------
+
+  // A dispatch is a response to an event within the context of the current state.
+  type Dispatch = <S, P, D>(action: Action<S, P, D>, props?: Payload<P>) => void
+
+  // An action transforms existing state while possibly invoking effects.
+  type Action<S, P, D>
+    // An action can wrap another action.
+    = ((state: State<S>, props?: Payload<P>) => Action<S, P, D>)
+    // An action can be given a payload.
+    | [Action<S, P, D>, Payload<P>]
+    // State can be set directly along with any effects to run.
+    // This is particularly when setting the initial state.
+    | [State<S>, ...EffectDescriptor<D>[]]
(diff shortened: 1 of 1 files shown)


src/index.d.ts
```diff
@@ -0,0 +1,120 @@
...
+    readonly children: VNode[];
+    node: MaybeNode;
+    readonly tag?: Tag;
+    readonly key: Key;
+    memo?: PropList;
+  }
+
+  // Virtual DOM properties will often correspond to HTML attributes.
+  type PropList = Record<string, Prop> & Readonly<{ key?: Key }> & ElementCreationOptions
+  type Prop = bigint | boolean | null | number | string | symbol | undefined | Function | ClassProp | StyleProp
```
Hammster, line 74: Couldn't there also be Arrays or Object of any signature be a prop value ? For example `{foo: string: bar: number}[]`, and shouldn't the component dictate what explicit type the property has ?

src/index.d.ts
```diff
@@ -0,0 +1,120 @@
...
+  type Key = null | string | undefined
+
+  // -----------------------------------------------------------------------------
+
+  // The `app` function initiates a Hyperapp application. `app` along with effects
+  // should be the only places you need to worry about side effects.
+  function app<S, P, D>(props: App<S, P, D>): Dispatch
+
+  // The `h` function builds a virtual DOM node.
+  function h(type: string, props: PropList, children?: VNode | readonly VNode[]): VDOM
```
Hammster, line 113: Alternatively i would allow to type the `h` with a generic

```
function h<T = {}>(type: string, props: T & PropList, children?: VNode | readonly VNode[]): VDOM
```

icylace:
> It would be beneficial to have some `Component<PropType>` this way you can reduce boilerplate code when defining a Component.

That sounds tempting but what would that look like specifically? I think we should be careful to provide accurate types that don't make too many assumptions. That way people are free to be opinionated in their own components libraries without conflict.

src/index.d.ts
icylace: So, `Prop` exists primarily to describe the properties `h` would be able to work with. As far as I'm aware the only prop that can have an array value or object value is `class` while `style` must have an object value.

I'm hesitant to broaden the scope of `Prop` to let it be directly relied upon in user-created components. I want `Prop` to maintain a level of precision that doesn't make it a good fit for that purpose. It would be overkill in some ways and too restrictive in others.

Instead, I encourage creating component-level prop types tailored to your components.

src/index.d.ts
icylace: Making `h` generic in that way would defeat the purpose of `PropList` if you could just do something like `h<any>`. Please see what I wrote about `Prop` above.

Hammster:
I would take an example on preact in regard to this, it does not limit the user in the definition of property types, but it makes the process of working with types easier.

- You only need to provide a single interface and mark optional props
- The props are also available in JSX
- You can create a union type as part of the generic handling adding for example:
  - HTML Attributes to the intellisense
  - typed `children` property

https://github.com/preactjs/preact/blob/59c9abaee16dfeb1411f197ef071c94525087bc6/src/index.d.ts#L40-L175

This is one example from a preact app I have

```ts
import { h } from 'preact'
import { Indicator } from '../indicator' // Indicator is a Enum with Ok, Warning, info, Error
import { FunctionComponent } from 'preact'
import { JSXInternal } from 'preact/src/jsx'

export interface ButtonProps extends JSXInternal.HTMLAttributes {
  indicator?: Indicator,
  disabled?: boolean
}

export const Button: FunctionComponent<ButtonProps> = ({
  indicator = Indicator.Default,
  children,
  ...attr
}) => <button class={`button ${indicator}`} {...attr}>
    {children}
  </button>

export default Button
```

icylace:
Is the following close to what you would like to see happen? :

```ts
// in Hyperapp:

export type Component = <T>(props: T & PropList) => VDOM

// in your component file:

import type { Component } from "hyperapp"
import { Indicator } from '../indicator'

export interface ButtonProps {
  indicator?: Indicator
}

export const Button: Component<ButtonProps> = ({
  indicator = Indicator.Default,
  children,
  ...attr
}) => <button class={`button ${indicator}`} {...attr}>
    {children}
  </button>

export default Button
```

Hammster:
Yeah, this is pretty much it. didn't knew about the `import type` syntax yet though haha.

icylace:
Yeah, `import type` is great !

Now, while I see the appeal of that definition for `Component` I still think it's not Hyperapp's place to prescribe it.

I say this because I use a different approach to components and component composition that's oriented around functional lenses. Long explanation short I'd rather use this definition for `Component`:

```ts
export type Component = <T>(props: T & PropList) => (path: string[]) => <S>(state: State<S>) => VDOM
```

src/index.d.ts
```diff
@@ -0,0 +1,125 @@
+declare module "hyperapp" {
+  // A Hyperapp application instance has an initial state and a base view.
+  // It must also be mounted over an available DOM element.
+  type App<S, P, D> = Readonly<{
+    init: Action<S, P, D>;
```
gamebox, line 5: It seems that `init` can take the above as well as `S`

icylace: Yes, are you referring to line 44 ?
https://github.com/jorgebucaran/hyperapp/pull/954/files/b36ac7fb6a21fdf340ecebb6ae1290cb50871cad#diff-b739f95ace6bbf2f0fd5e8cf1dab97d8R44

gamebox: My apologies, yes.  I think that the name Action for that Union is not really intuitive.  Can you really provide a state object to a event handler?

jorgebucaran: @gamebox Yes, you can, but we don't want people to do that, so is it possible to have TypeScript complain if you try that?

gamebox: We just need to adjust the types so that Action<...> for event handlers excludes that case.

icylace: Just to be clear, something like
```js
h("button", { onclick: "Mario Mario" }, "It's a me, Mario!")
```
should be frowned upon, correct ?

gamebox: @icylace I believe that is correct.

@jorgebucaran It also seems that the first type in the Action<..> union doesn't make much sense for init either:

`(state: State<S>, props?: Payload<P>) => Action<S, P, D>`

icylace: @gamebox I've made some updates and put them into a new pull request:  https://github.com/jorgebucaran/hyperapp/pull/969

PR #958, opened 2020-06-30
Use more common "subtract" instead of obsolete "substract"

mabhub:


jorgebucaran:
Thank you, @mabhub! 🎉

PR #968, opened 2020-07-19
Alternative module format

gamebox:
This rewrite is smaller and compresses better at a small hit to developer ergonomics.

Full size 2.1kB, Gzips to 768B

### DIFF
--- a/pkg/html/index.js
+++ b/pkg/html/index.js
@@ -1,6 +1,4 @@
-const EMPTY_ARR = []
-
-const h = (type) => (props, children = EMPTY_ARR, node, tag) => ({
+const h = (type) => (props, children = [], node, tag) => ({
   type,
   props,
   children: Array.isArray(children) ? children : [children],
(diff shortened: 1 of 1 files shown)


jorgebucaran:
Thank you, @gamebox, do you know if this would cause tree-shaking issues because the module now has a `default` export?

jorgebucaran:
@gamebox So, this means we'd have to:

```js
import html from "@hyperapp/html"
const { main, h1, button } = html
```

instead of:

```js
import { main, h1, button } from "@hyperapp/html"
```

Also, I'm not sure how bundlers should handle default exports, but I suspect they will not be tree-shakable, whereas the current solution is, so you'd be only using what you actually import anyway! 

Sorry, but I'd really like to keep the current usage in favor of ergonomics, so I am afraid I can't accept this. Thank you!

gamebox:
Totally understand.  This was optimized for tool-free usecases, not for bundlers.  This is useful for me to understand your priorities

PR #969, opened 2020-07-21
TypeScript definitions for 2.0.8

icylace:
Here is an update to the TypeScript definitions I've made for Hyperapp.

Suggestions always welcome !


index.d.ts
```diff
@@ -0,0 +1,125 @@
...
+  function app<S, P, D>(props: App<S, P, D>): Dispatch
+
+  // The `h` function builds a virtual DOM node.
+  function h(type: string, props: PropList, children?: VNode | readonly VNode[]): VDOM
+
+  // The `memo` function stores a view along with properties for it.
+  function memo(view: View, props: PropList): Partial<VDOM>
+
+  // The `text` function creates a virtual DOM node representing plain text.
+  function text(value: string | number, node?: Node): VDOM
```
hlibco, line 124: @icylace +1 to the type definition! 

Small suggestion, the `value` argument on the text() function also accepts an array of strings, numbers, booleans, null (but not undefined) and it will convert this array into a comma-separated string.

index.d.ts
```diff
@@ -0,0 +1,125 @@
...
+    readonly children: VNode[];
+    node: MaybeNode;
+    readonly tag?: Tag;
+    readonly key: Key;
+    memo?: PropList;
+  }
+
+  // Virtual DOM properties will often correspond to HTML attributes.
+  type Prop = bigint | boolean | null | number | string | symbol | undefined | Function | ClassProp | StyleProp
+  type PropList = Readonly<ElementCreationOptions & {
```
gamebox, line 78: Where does `ElementCreationOptions` come from?

index.d.ts
```diff
@@ -0,0 +1,125 @@
...
+    [k: string]: Prop;
+    class?: ClassProp;
+    key?: Key;
+    style?: StyleProp;
```
gamebox, line 82: Maybe add the on* handlers here with a signature of maybe `type EventHandler<E> = (event: E) => void`?  Typing the event handlers shouldn't be necessary as TypeScript has those types already

index.d.ts
icylace: It comes from the definitions TypeScript provides itself.  Specifically, `lib.dom.d.ts`: https://github.com/microsoft/TypeScript/blob/56865f722d6985e7e446402ac57e2f20349cdbde/lib/lib.dom.d.ts#L459

index.d.ts
icylace: TypeScript does provides a `GlobalEventHandlers` interface which, as far as I know, defines all the handlers:
https://github.com/microsoft/TypeScript/blob/56865f722d6985e7e446402ac57e2f20349cdbde/lib/lib.dom.d.ts#L5772

However, those would not be able to be used directly because in Hyperapp you assign actions as event handlers, not ordinary functions like in vanilla JS.  What I mean is in Hyperapp the function you assign as, let's say, the `onclick` handler must accept the state as it's first parameter, event data as the second, and then it must eventually return a new state (e.g. `onclick: (state, event) => state`).

But in vanilla, events handlers just get one parameter, the event data, and they are not required to return anything.

I'm open to the possibility of adapting `GlobalEventHandlers` for Hyperapp usage through mapped types but I don't know if it's possible or not.

index.d.ts
icylace: I explored `text` a bit more and found that only symbols give it trouble:
https://codepen.io/icylace/pen/XWXoLYy?editors=0010

Arrays of undefined are acceptable in terms of whether or not a crash happens.

So, it becomes a matter of intent.  Should we think of `text` as a glorified `toString` or do we intend to define the purpose of `text` in a more curated way?  @hlibco @jorgebucaran

hlibco: @icylace That's a good point. If `text` abstracts a native method that performs a `toString` under the hood, I think this feature should be exposed upstream.

Here's an experiment using `innerText` to observe how it converts non-strings to strings.
https://codepen.io/hlibco/pen/NWxeQmo

As a reference: https://dom.spec.whatwg.org/#text

index.d.ts
gamebox: Sorry, I meant just using the Event subtype to parameterize the handler, not use the event handler type

icylace: I think I know what you mean but could you elaborate ?

index.d.ts
```diff
@@ -0,0 +1,125 @@
...
+  type Key = null | string | undefined
+
+  // The `class` property represents an HTML class attribute string.
+  type ClassProp = string | Record<string, boolean> | ClassProp[]
+
+  // The `style` property represents inline CSS.
+  type StyleProp = Record<string, null | number | string>
+
+  // A virtual node is a convenience layer over a virtual DOM node.
+  type VNode = null | undefined | VDOM
```
hlibco, line 95: @icylace `VNode` also accepts `boolean` which are filtered out down the road. Here's an example:

```
view: () => h('div', {}, [false])
```
and
```
view: () => h('div', {}, false)
```

icylace: What happens with `true` ?

hlibco: Nothing happens. Either `true` or `false` the node (in this case `div`) is not created.

index.d.ts
```diff
@@ -0,0 +1,125 @@
...
+  type Key = null | string | undefined
+
+  // The `class` property represents an HTML class attribute string.
+  type ClassProp = string | Record<string, boolean> | ClassProp[]
+
+  // The `style` property represents inline CSS.
+  type StyleProp = Record<string, null | number | string>
+
+  // A virtual node is a convenience layer over a virtual DOM node.
+  type VNode = boolean | null | undefined | VDOM
```
icylace, line 95: @hlibco Okay, I added `boolean`.

index.d.ts
```diff
@@ -0,0 +1,125 @@
...
+    class?: ClassProp;
+    key?: Key;
+    style?: StyleProp;
+  }>
+
+  // A key can uniquely associate a virtual DOM node with a certain DOM element.
+  type Key = null | string | undefined
+
+  // The `class` property represents an HTML class attribute string.
+  type ClassProp = string | Record<string, boolean> | ClassProp[]
```
hlibco, line 89: @icylace `ClassProp` can also be boolean. In this case, if it's either True or False, the class is not added to the element.

Some valid use-cases:

```
h('div', { class: ['menu', true] }, 'Hello') // <div class="menu"></div>
h('div', { class: ['menu', false] }, 'Hello') // <div class="menu"></div>

const isOpen = false
h('div', { class: ['menu', isOpen && 'open'] }, 'Hello') // <div class="menu"></div>
```

icylace: For sake of completeness, the following result is true?:

```js
const isOpen = true
h('div', { class: ['menu', isOpen && 'open'] }, text('Hello'))

// <div class="menu open">Hello</div>
```

hlibco:
That's correct and it's also the expected result.

index.d.ts
icylace: I updated `ClassProp` to now accept `false`.  I was tempted to use `boolean` but didn't see a point to allowing `true`.

zaceno: @icylace I might have misunderstood, but when you say you don't see a point in allowing `true`, does that mean you think this shouldn't work:

```js
h('div', {class: {navigation: true, active: url === state.currentUrl}}, [...])
//--> <div class="navigation active">...</div>
```

Or is it specifically about arrays?

icylace: @zaceno What you wrote should work. My update was about situations like:
```js
const sayingHi = true
h("div", { class: ["greeting", sayingHi && "hi"] }, text("Hi"))
h("div", { class: !sayingHi && "bye" }, text("Bye"))
```

index.d.ts
icylace: @hlibco I think no changes to `text` are required for the time being.  Would you agree?

index.d.ts
icylace: @gamebox Today, I was able to make an update which includes event definitions!

index.d.ts
```diff
@@ -119,5 +122,234 @@ declare module "hyperapp" {
...
+
+  // ---------------------------------------------------------------------------
+
+  // Due to current limitations with TypeScript (which should get resolved in
+  // the future: https://github.com/microsoft/TypeScript/pull/40336), here is
+  // a collection of modified copies of relevant event maps from TypeScript's
+  // "lib.dom.d.ts" definition file to assist with defining `EventActions`:
+
+  type OnElementEventMap = {
+    "onfullscreenchange": Event
```
jorgebucaran, line 135: @icylace Where does `Event` come from?

icylace: @jorgebucaran It comes from TypeScript itself: https://github.com/microsoft/TypeScript/blob/d89635381a65229c5ab5b0c73de15347b86235a2/lib/lib.dom.d.ts#L5285

icylace:
I'm currently working on adding unit testing for my type definitions using [dtslint](https://github.com/microsoft/dtslint).

If anyone has ideas for clever test cases please share !

jameswilddev:
Been trying this locally, looks good!  Only place I'm not sure it's right is the return value of app(); I'm seeing undefined rather than a dispatch callback, and there doesn't seem to be a return in the JS source: https://github.com/jorgebucaran/hyperapp/blob/bfa02ba45227d939a5a711bea5355362dc6c3bc9/index.js#L373

icylace:
@jameswilddev Nice catch! I'll make a correction when I get a chance, hopefully soon.

icylace:
@jameswilddev Okay, `app` should be better now.

jameswilddev:
Thanks!

thequailman:
Is there anything more to do with this?  Would love to see this merged.

zaceno:
+1 I'd like to see this merged (or uploaded to @types) 

I'm just learning typescript, and would like to be able to experiment with HA+Typescript. I think strong TS support will be a good selling point. At least the devs where I work, I suspect, are more willing to give up React than typescript.

hlibco:
+1 Devs that I know who work with Typescript have a strong good feelings
about it.

jorgebucaran:
We're definitely planning on merging this but only after the documentation lands.

asabatba:
Should array-based Subscribers be taken into account?

What I'm trying to do is based on the tutorial (pretty new to hyperapp, excuse me if anything's off). When adding subscriptions to the app:

```
...
subscriptions: (state) => [
        [keySub, {
            keys: ['w', 'a', 's', 'd'],
            action: IncrementCounter,
        }],
], ...
```

Typescript fails. Extending Subscriber like this works for me, but I'm probably not looking at the whole picture and there's a better way to do this:

`type Subscriber<S, D = any> = boolean | undefined | Effect<S, D> | Unsubscribe | [Effect<S, D>, Payload<D>]`

icylace:
@asabatba Interesting, I'll have to try that out myself !

types/tslint.json
```diff
@@ -0,0 +1,17 @@
+{
```
talentlessguy, line 1: TSLint is deprecated, see: https://github.com/palantir/tslint/issues/4534

ESLint with `@typescript-eslint` should be used instead

zaceno:
@icylace With the following definition of `h` we can throw an error if someone accidentally omits the properties argument

```
    function h<S, T>(
      type: string,
      props: T extends VDOM<S> ? never : T & PropList<S>,
      children?: VNode<S> | readonly VNode<S>[],
    ): VDOM<S>
```

demo:

Screenshot 2021-01-15 at 12 09 45

zaceno:
@icylace I discovered another issue with the types as they are at the moment. It seems that `init` is not assignable to anything that is dispactchable (which it shoudl be). For instance, if you want to run an effect at that start of your app, you will typically have:

```js
app({
  init: [
    {initial: 'state'},
    initalEffect('http://foo.com')
  ],
  ...
})
```

But it looks like that is not allowed by the types  even though it is a fully legal thing to dispatch.

Screenshot 2021-01-16 at 21 14 43

icylace:
@zaceno Your suggestion for `h` fixes one problem but somehow fails an existing test:

```js
h("p", { style: { clor: "hi" } })          // $ExpectError
```

`style` shouldn't allow nonexistent CSS properties.

I'm trying to address this in accordance with your suggested change but it's tricky. If you have any ideas, that'd be great !

zaceno:
@icylace 

Sorry I don't know TS well enough yet. I don't even understand how that error could have come up, let alone how to fix it 😅 

I guess it  means `T & PropList<S>` is somehow _less_ restrictive than just `PropList<S>`  – but how can that be?!

jorgebucaran:
Isn't the point of types to help you catch errors before they happen? Why not also forbid going against best practice? 

What are types `init` can be but shouldn't?

zaceno:
@jorgebucaran I think enforcing "best practice" with types is going to be bad for people who use typescript to write and compile their apps.  Maybe that's what a linter is for (warnings without actually breaking your code). 

> What are types init can be but shouldn't?

I can't think of anything off-hand. Are you saying having an effect immediately on start up is a bad idea?

jorgebucaran:
Agreed. 👍 

As for `init`, the ideal would be an action such as Initialize or anything that makes sense to you. Init is an action itself, so an anonymous function is good too.

icylace:
@zaceno I just updated the types to fix false positives for `PropList`.  It was really tricky for a while !

types/index.d.ts
```diff
@@ -91,6 +91,10 @@ declare module "hyperapp" {
     readonly tag?: Tag<S>
     readonly key: Key
     memo?: PropList<S>
+
+    // `_VDOM` is a guard property which gives us a way to tell `VDOM` objects
+    // apart from `PropList` object.
+    _VDOM: true
```
jorgebucaran, line 97: Can we name this guard prop whatever we want? What about `ours`? Bad idea?

icylace: We can name it whatever we want. I chose `_VDOM` because it seemed the most straightforward thing at the time.

Let's say for the sake of argument we wanted to generalize it so we can use it elsewhere like in `State` or whatever.  We could use something like `_h: "VDOM"` for VDOMs and `_h: "blahbiddyblah"` for other things.

I'm open to the idea but I have yet to think of where else it could be useful and worth the trade-off in type complexity.

types/tslint.json
icylace: You know come to think of it I probably don't need that file to be part of this PR anyway. I was mainly using it for my own convenience. I'll see about omitting it. Anyway, great point about the deprecation.

icylace: Oh, I actually just tried removing it.  Looks like `dtslint` complains more without it.

This is really more of a `dtslint` issue as described here: https://github.com/microsoft/dtslint/issues/214

icylace:
New and improved PR coming soon !

icylace:
Please go here for the new PR: https://github.com/jorgebucaran/hyperapp/pull/1016

PR #973, opened 2020-07-28
Use `text(...)` instead of plain `string` in tutorial.

xieyuheng:
I am not sure this is your intention,

if it is not, please feel no pressure to close my PR.

------

Also a question here.

Why do we not convert `string` to `TEXT_NODE` `VNode` implicitly anymore?


jorgebucaran:
@xieyuheng This is a bug in the documentation, whoops! I need to rewrite the tutorial, but we can merge here right away.

> Why do we not convert string to TEXT_NODE VNode implicitly anymore?

I want to reduce complexity. text("Why go out of your way when it's just as easy as this?")

See also: https://github.com/jorgebucaran/hyperapp/issues/971#issuecomment-664531474

Thanks!

Issue #965, opened 2020-07-08
Remove reference to payload filters from tutorial


dwknippers:
Currently the tutorial on the website references **payload filters**.
This was _experimental_ functionality and will be removed in the future (v2).

Ref: https://github.com/jorgebucaran/hyperapp/blob/minimizations/docs/tutorial.md#payloadfilters

zaiste:
@dwknippers what's the proper alternative in this scenario?

jorgebucaran:
@zaiste The trick is that an action can return an action too (`=> [MyAction, newPayload]`), so I usually define a function to help me with that.

```js
const withPayload = (filter) => (_, x) => filter(x)
```

and use it like this for example:

```js
{
  oninput: withPayload((event) => [
    MyAction,
    { id: state.currentId, value: event.target.value },
  ]),
}
```

The action that carries the payload is the anonymous function in the middle `(_, x)` of `withPayload`.


We could (and probably should) export that from **@hyperapp/events**, as I also want to include higher level event utilities. e.g:

```js
const targetValue = (action) =>
  withPayload(({ target: { value } }) => [action, value])
```

zaceno:
This has been adressed in #994 We can close here I think

Seanmcdon:
This should be in the docs somewhere? This is a high use-case. Payloads that can stopPropagation and send ids, etc.

PR #986, opened 2020-10-18
Fix function as custom payload not executed

### LINKED ISSUE #985: Update tutorial docs -> function as custom payload
> When we give a function as the custom payload, Hyperapp considers it a payload filter and passes the default payload through it, providing the returned value as payload to the action.

```
oninput: [SetFilter, (event) => event.target.value], 
```

This does not work since function is not executed just returned.

VictorWinberg:
Fixes #985

### DIFF
--- a/index.js
+++ b/index.js
@@ -394,7 +394,7 @@ export var app = (props) => {
       ? dispatch(action(state, props))
       : isArray(action)
       ? typeof action[0] === "function"
-        ? dispatch(action[0], action[1])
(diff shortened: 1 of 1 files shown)


isabo:
@VictorWinberg please see #965 -- it seems payload filters were an experimental feature that was deliberately removed.

VictorWinberg:
Oh, okay! Then I would recommend updating the code on `tutorial.md` on codesandbox.io, since it is used there.

PR #987, opened 2020-10-19
Fix cannot read property X of null/undefined

### LINKED ISSUE #983: Cannot read property X of undefined
Tutorial with newest hyperapp version gives: `Cannot read property 'is' of undefined`

E.g. https://codesandbox.io/s/hyperapp-tutorial-step-2-forked-k0bmx

This is due to that the `props` parameter to `h` function can be `null`.

VictorWinberg:
Fixes #983

Add missing spread operator on children, when using `jsx`.

Fix `Cannot read property 'key' of null` when using a `jsx` with a node that does not have any attributes.


VictorWinberg:
Since JSX is dropped this issue is no more.

PR #990, opened 2020-10-19
Fix missing children with spread operator

### LINKED ISSUE #989: Missing children in JSX
When using JSX, only the first child of every element is visible.

Because `h` function is not using the spread operator, or the `arguments` method

VictorWinberg:
Fixes #989


VictorWinberg:
Issue is closed, hence PR closed.

PR #991, opened 2020-10-22
update hyperapp link in import module

flopanzer:
With the updated link the hello world example in the `tutorial.md` is working locally for me


jorgebucaran:
https://unpkg.com/hyperapp gives you the latest release already, so we don't really need to specify the version. 😁

PR #992, opened 2020-10-31
Fixes subscriptions restart

sergey-shpak:
Subscriptions props comparing ('shouldRestart' func): 
since `isArray((b[k] = a[k]))` is an assignment, next condition check `if (a[k] !== b[k]) return true` never happens, which brakes subscription restart (actually never restarts). Fixed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,7 +34,9 @@ var createClass = (obj) => {
 
 var shouldRestart = (a, b) => {
   for (var k in { ...a, ...b }) {
-    if ([].concat(a[k])[0] !== [].concat(b[k])[0]) return true
+    if (typeof (isArray(a[k]) ? a[k][0] : a[k]) === "function") {
+      b[k] = a[k]
+    } else if (a[k] !== b[k]) return true
   }
 }
 
--- a/index.js
+++ b/index.js
@@ -34,8 +34,7 @@ var createClass = (obj) => {
 
 var shouldRestart = (a, b) => {
   for (var k in { ...a, ...b }) {
-    if (typeof (isArray((b[k] = a[k])) ? b[k][0] : b[k]) === "function") {
-    } else if (a[k] !== b[k]) return true
+    if ([].concat(a[k])[0] !== [].concat(b[k])[0]) return true
   }
 }
 


jorgebucaran:
Thank you, @sergey-shpak. You found the Easter egg! We need that assignment, though. I think we can do that inside the if branch like this:

```js 
var shouldRestart = (a, b) => {
  for (var k in { ...a, ...b }) {
    if (typeof (isArray(a[k]) ? a[k][0] : a[k]) === "function") {
      b[k] = a[k]
    } else if (a[k] !== b[k]) return true
  }
}
```

Make sense to you?

sergey-shpak:
@jorgebucaran, why do we need this assignment?

zaceno:
@sergey-shpak short answer: in order to support anonymous actions in subscription props.

sergey-shpak:
@jorgebucaran, @zaceno, ok, ok, could you explain? I need more details, is there any related issue with this assignment?))

zaceno:
@sergey-shpak Sorry for the long delay. Here's some more info:

[Have a look at this example][example] – in particular line 38. Notice how, in order to make a payload for `UpdateTimer` that contains _both_ the timer id, as well as the new timestamp (which is the payload provided from the subscription), we need to make an anonymous action that returns a new action with the combined payload.

Now, if subscriptions would do a strict diff to check if any props have changed, it would detect a diff _every interval_ because the actions we provide are anonymous, defined inline. That means the subscriptions would be restarted each time which, in principle works most of the time – probably – but is bad in principle (I think you agree with that)

In order to solve that problem, hyperapp currently ignores diffs on functions. But that means it will ignore diffs on functions/actions even when they are _supposed_ to change. So how does the subscription find out about changed actions? You guessed it: they are reassigned on the props object. So the subscription will not be restarted, but it will begin using the new actions.

This means it is unsafe to destructure the subscriptions props in your subscription implementation. 


[example]: https://flems.io/#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAOgAsAXLKEAGhAGMB7NYmBvEAXwvW10RGIhwBO+AFYIqdBk2J5+AB1oDiAAmAwAbjAEBPVgAKwA2liUByEsVlxEAemsBXNLIDWAc3x0s1gAKEtszRiystZ8OCYAOmhyCsokZEoGRqbmljb2jq7uRta+-gJQEMQRUVjyikqyhi4C8HAARhgCCYbGJvjWlbTVtQC0DUKixZEScMoAIgCqAEoAggAqAJIA8gBySgC8SgCsAAx7kZEwAB5lyiPKEFF8GFAbKkpotADuiEo7FEqjjYwAJq-vSgEDiiaBcrzANzgMCUrAOaGOpyU5yUdlkPwwjDuAApQpoPo8ngBKDYAPhUkSUSggYCUWIAhLihEC0CCXMSasQ7AI0EpGRSqTScfxNPgvooYD8lAAyKUPZ5KHq84VCMW-JRkyazRardkwTncqlXCA3fnU2kM5Wi4jfCW6-U84Byp4fVUS14Ej7M1mvYhA6GwnmAvVcnlY4D4COM-HPViEyIBw4nGJI+ijT7W8qbLHE9ZksNKCP4S6FY1QT3Ay5g3l+mFxtCJxHI-rY4CM2OkipVGpweqNMP8yBQWB-JQAWQxhHwhgcPyxQsE+AJCqVC9dP0J1k182WawAVEoAIx7QkUflQDB1GBQV7j4iT6gwaBzrfataKxmL+UAahXIrXxOsQ89h2YkfxMOBiljOEEWTJs7GIYh6BbKMlAAZQzYh21zJQ4igAADfkAB46ngxCeXoagCmoJx1gAEmAdDvlYJQfggOAL2HOjW0tL1K1YEl+UpVC5hmKY5iI6wSIQ+gSQI+t4STcpkTgOw6jgZDlQ+CY0QxGAsLJZBfyZCtQWlWV1E0LQsSPLSdMYYkAF04ThSgQChWBqD4VM8AADkQAB2NgOBATAcDwTpuh7PpGhEMQaHoRhmB4aJylw5okjMBDUlsBxnDcDwcj8TQCiKaDFOUH4YAhOwoGUTpLA7XC5MpQjWLUJFzx7dZwhACLuzgHoJGtS5NB6gTAxatrBMpDqMC6nq+t6QcoB66bKVGLRYC44AnggH471eeq4HwZaJSUMCAFITFYAMZqUaxxruwjZFm+beq7Xpz0vFaQBJeijvwL6r1YQiOkepRQba2TnLQVz3JgTyIG8ngDwAJkQHYgs4MKeFi1yhqkGRSmTYBAlkZjElaFIrByjJ8uyXIAiCYoUtiUhbkp5Ispp9I8qyTxGfyQoWeJ8p9zmoz0tadoP0GGHkW09FGDmZVsS+RgPm4nAvlKD49v02lgH5Qt1ZgU9A2QPaHJ9S1USVmAsVNy2fgcj5cR12Q6yg+TkUYxQVcENXrQ1ql1w7ftAxN4Ozf5Z3raMq1vkd6O469uslDhMmI8pYtiFeI2JqUGYbYXXPS3Nu6ACES5FMubgrmaAGEa6EOuy2msYW6LI16+mgBRLu24b9h+TUCAYBedNdMa9nmpm4jaB+LRwbu+iTBmSvG7GPu2jgWQSqxEwTEJfAsECLE9pnsg57ulr6I-fpmMZLinat5iHrW+f78tKSyKf5UX4pzfmhDCXFkB+2IAHPEocHLvxXrfQihABAPUpJ-PChJbrz0kovZeeEG4qTUtQAQEBZBeTQFYKemJczTXXpvbexQ7qin3oUQ+x9pqn3PpfbCH4CFwGzrfRkrxX4uwbndRWulXhYgAPpu2FB7HMBlxHK00ioPasjtbWlKKwJyhdYzpxmidc8xBswN0eJVV4PxaDUDsDgBg+A6g4PjISfGRhZDQE0HgBo304ZXgRmQhAPAADM-lECBKxiFLgeB3A9hcZIJKBBFpRSGhgEaTQC6UgcQISqAhXgHlkEcJQdRzzUU+LQAoPwADc-JWJ73PFoV4lwCjwj6FAKxTgqmBkII+FwJBXgABYdj5I6ZSXa+1CCvECXsIZ-J5BwBLPQV4NQjEQA0MMpQtANACDAK0yehA9qVTQGshxRwehwEIBgSxk9LhQmUDsJQqN8nbEeQIFwDQsTvA+RQHY+A+mEiOQobJPQBAXIgHYChfTpmBgaNRaotAZyDTKQoV4ABiGAaKOkJjQPgRJA1lrkkDLM+ZaBXgXjgGU+CMA1mIVkP8NZsAwB5zeP86SWBaX8mha4acaAfgItaTkpQyKdj+TAP8rJmggUgrBa8LYkKMm0BOWci5zwGnkL1G8JQgTHkPIKS8t5qMthbAoPqw1xqvk-L+fGOE2KPpRSBrcdJCQEo9AhFgaA9TPgYHIaczQ1I1lgCdXMgAXjAXJPlZWOoYD0J43TemFLKZUs8I0ehdIgD0xlB4w1HDWYSshJK1LksYGss+LzLhssDNS3Jmqs38lGQdICOwLpUuOMQHoNxU3EqRFITQGKfEeX8XgQJ6MDxsFgUAA

sergey-shpak:
@zaceno ok, now it's more clear. So, instead of following hyperapp patterns, we are creating hacks inside of hyperapp core) 
If we want to properly use dynamic actions (without restarting sub each time) you can use following pattern:
```js
const hideProps = (props, target) => 
  props.reduce((acc, prop) => Object.defineProperty(acc, prop, {
    enumerable: false,
  }), target)

// some app definition
subscriptions: state => 
  [someSub, hideProps(['action'], { 
    // this won't cause any restart anymore
    action(){ /* do your stuff here */ } 
  } )]
```

jorgebucaran:
@sergey-shpak Remember that creating your own subscriptions is a stretch you probably don't need. Of course, we don't really have any official subscriptions at the moment, so you don't really much of a choice right now, but it will not be like that forever. 🙏

sergey-shpak:
@jorgebucaran sure, we are discussing more internal core (`shouldRestart`) and why it has to do this weird(at least to me) assignment hack, I have shown how to avoid restarting subs from userland (just for example) and I cannot find any reason why `b[k] = a[k]` cannot be removed

jorgebucaran:
@sergey-shpak Ah, thanks for the clarification. You're not alone, it's a bit weird for me too.

jorgebucaran:
@sergey-shpak I'll look more into this when I start writing tests for this part.

sergey-shpak:
@jorgebucaran please keep in mind this is not only improvement but also a fix for the issue described at the start of the conversation, so PR can't wait for a long time

jorgebucaran:
@sergey-shpak Yes, but removing the inline assignment breaks the behavior [described](https://github.com/jorgebucaran/hyperapp/pull/992#issuecomment-734410991) by @zaceno above. So, we need to keep it.

sergey-shpak:
@jorgebucaran behavior described by @zaceno has to be done in userland, not in core (and I [just showed how to do it properly](https://github.com/jorgebucaran/hyperapp/pull/992#issuecomment-739311936))

jorgebucaran:
No, it doesn't have to. It certainly could be, but that's one thing I wouldn't want to impose on users. Also, I don't see why we couldn't deal with this in core.

sergey-shpak:
ok, after some thoughts I came to that 'ignoring' props won't help (forget [that example](https://github.com/jorgebucaran/hyperapp/pull/992#issuecomment-739311936)), we need this assignment to mutate props which are used inside of subscription

fsrc:
My team found the same bug some time ago and corrected it with this code 

```
var shouldRestart = (a, b) => {
  for (var k in { ...a, ...b }) {
    if (typeof (isArray(a[k]) ? a[k][0] : a[k]) === "function") {
      b[k] = a[k]
    } else if (a[k] !== b[k]) return true
  }
}
```

fsrc:
Just noticed that it was the same code as @jorgebucaran posted.

zaceno:
Bump! Please change the PR to the code snippet already posted here (https://github.com/jorgebucaran/hyperapp/pull/992#issuecomment-754036508)

and then let us merge this asap. I ran in to this issue as well and would like to see it fixed.

jorgebucaran:
Won't this break anonymous actions in subscription props?

zaceno:
> @jorgebucaran Won't this break anonymous actions in subscription props?

Yes the commit by @sergey-shpak in this PR would break anonymous actions in subscription props.

However the code you suggested:

```js
var shouldRestart = (a, b) => {
  for (var k in { ...a, ...b }) {
    if (typeof (isArray(a[k]) ? a[k][0] : a[k]) === "function") {
      b[k] = a[k]
    } else if (a[k] !== b[k]) return true
  }
}
```

– this is the code I would like to have in a new update asap. I have verified (to my own satisfaction at least) that it both 1) fixes the bug and 2) allows anonymous actions in subscription props.

jorgebucaran:
@sergey-shpak Could you update the code so we could have your PR committed instead? 👊

zaceno:
@jorgebucaran It looks like @sergey-shpak hasn't been active on github for the last six months (according to his profile). I hope he's fine, but maybe we shouldn't wait too long for his input before we fix this bug.

sergey-shpak:
@zaceno hey, I'm fine, just too many activities outside github these days, thanks) I'm going update the PR ASAP

zaceno:
@sergey-shpak sweet, good to see you again man! 😅

sergey-shpak:
@zaceno good to see you too man! :beers: :beers: :beers: 
Updated PR with proper support of anonymous actions in subscription props)

jorgebucaran (approved):

jorgebucaran:
Thank you, @sergey-shpak, and everyone else here for helping with this. Get your `@2.0.10` while is still hot! 🔥

jorgebucaran:
Thank you, @sergey-shpak, and everyone else here for helping with this. Get your `@2.0.10` while it's still hot! 🔥

jorgebucaran:
Thank you, @sergey-shpak, and everyone else here for helping with this. Get your `@2.0.10` while it's hot! 🌶

PR #1011, opened 2021-01-21
Add @hyperapp/jsx and docs for JSX runtime

loteoo:
This PR is more of a suggestion to both:
- Show how supporting JSX could look like with the new [Automatic JSX runtime](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#manual-babel-setup) that has been introduced last year
- See what are the thoughts are these days about hyperapp + JSX

I think it would be great to have official support for this, especially with https://github.com/jorgebucaran/hyperapp/pull/969 that will eventually be merged.

JSX + TypeScript work really well together. We will be able to have type safety with attributes like `onclick` that will expect values of type `Action`, etc...

**Edit:**
Fragments are also supported!


pkg/jsx/jsx-runtime.js
```diff
@@ -0,0 +1,16 @@
...
+        type,
+        props || {},
+        []
+          .concat(children)
+          .map((any) =>
+            typeof any === 'string' || typeof any === 'number' ? text(any) : any
+          )
+      )
+
+export { jsx, jsx as jsxs, jsx as jsxDEV }
```
jorgebucaran, line 16: What is `jsxDEV`?

pkg/jsx/jsx-runtime.js
```diff
@@ -0,0 +1,16 @@
+import { h, text } from 'hyperapp'
+
+const jsx = (type, { children = [], ...props }) =>
+  typeof type === 'function'
+    ? type(props || [], children)
```
jorgebucaran, line 5: Why `props || []`? Props should be an object, no?

pkg/jsx/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@hyperapp/jsx",
+  "description": "Hyperapp JSX runtime",
+  "version": "0.0.1",
+  "type": "module",
+  "main": "jsx-runtime.js",
```
jorgebucaran, line 6: Let's rename the file to `index.js` and use `index.js` as `main`.

pkg/jsx/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@hyperapp/jsx",
+  "description": "Hyperapp JSX runtime",
+  "version": "0.0.1",
```
jorgebucaran, line 4: Please make the version `0.0.0`.

pkg/jsx/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@hyperapp/jsx",
+  "description": "Hyperapp JSX runtime",
```
jorgebucaran, line 3: Let's go with something like: "JSX meets Hyperapp" for now. 😄

pkg/jsx/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "version": "0.0.1",
+  "type": "module",
+  "main": "jsx-runtime.js",
+  "license": "MIT",
+  "repository": "https://github.com/jorgebucaran/hyperapp/tree/master/pkg/jsx",
```
jorgebucaran, line 8: Should be `main/packages` after you rebase with main. ☝️

pkg/jsx/jsx-runtime.js
jorgebucaran, line 16: What is `jsxs`?

pkg/jsx/jsx-runtime.js
```diff
@@ -0,0 +1,16 @@
+import { h, text } from 'hyperapp'
+
+const jsx = (type, { children = [], ...props }) =>
```
jorgebucaran, line 3: Please rename `type` to `tag`.

pkg/jsx/jsx-runtime.js
```diff
@@ -0,0 +1,16 @@
+import { h, text } from 'hyperapp'
```
jorgebucaran, line 1: Please use double quotes.

jorgebucaran:
I was thinking, could you create a PR in https://github.com/zaceno/hyperapp-jsx-pragma too? I think there would be merged faster than here and also gives us a chance to try this out before it becomes an official Hyperapp package.

pkg/jsx/jsx-runtime.js
loteoo: Makes sense, not sure how I ended up with the array 🙊. I changed it to default to an object via parameter destructuring 👍

pkg/jsx/package.json
loteoo: Updated 👍

pkg/jsx/package.json
loteoo: Updated

pkg/jsx/jsx-runtime.js
loteoo: Switched to double quotes everywhere 👍

pkg/jsx/package.json
loteoo: Thanks for the heads up, rebased 👍

pkg/jsx/jsx-runtime.js
loteoo: So from my understanding, the new transform will use the `jsx` export for single elements with no children, `jsxs` when the element has children (I think this is so we can avoid doing the concat and the "default to array" or conditional logic) and `jsxDEV` so we can add development utilities. Because the function in the PR handles both cases (single and with children), we can just use the same function exported with the 2 names. [Same thing as preact does](https://github.com/preactjs/preact/blob/master/jsx-runtime/src/index.js#L67-L72).

loteoo: I think it's so we can hook in extra stuff for development usage (devtools and what not).

loteoo: Since we're not changing anything in the implementation for jsxDEV, I'll remove it from the exports in this PR. (it needs to be explicitly enabled anyways)

pkg/jsx/package.json
loteoo: I named it that way because for a `importSource` of `my-lib-name`, the transform will be `my-lib-name/jsx-runtime`. 
Idea was that we would set `@hyperapp/jsx` as importSource, and `@hyperapp/jsx/jsx-runtime` would be imported.
Not sure this is the best way to do it tho 🤔 .

loteoo:
> I was thinking, could you create a PR in https://github.com/zaceno/hyperapp-jsx-pragma too? I think there would be merged faster than here and also gives us a chance to try this out before it becomes an official Hyperapp package.

I would have to change zac's h function to support the `children` via props like the new transform expects (instead of via 3rd param). Should I publish a `hyperapp-jsx-transform` so we can more easily try it out?

zaceno:
@loteoo As long as `hyperapp-jsx-pragma` still works with `@babep/plugin-transform-react-jsx` I see no problem with it.

It could use the children prop if it exists, otherwise defer to the 3d-Nth arguments as children.

loteoo:
> @loteoo As long as `hyperapp-jsx-pragma` still works with `@babep/plugin-transform-react-jsx` I see no problem with it.
> 
> It could use the children prop if it exists, otherwise defer to the 3d-Nth arguments as children.

Sounds good! 👍  I'll do a PR soon and test with both old and new runtime :)

pkg/jsx/jsx-runtime.js
loteoo: Turns out `jsx` is for 0 or 1 child, `jsxs` is for many children. Updated the PR to take advantage of it (eliminated the `.concat` which should help perf)

pkg/jsx/package.json
jorgebucaran: Third-party project needs can't really impose how we name files in Hyperapp, so this should be `index.js` in principle.

loteoo:
@zaceno @jorgebucaran Here's the PR on zac's repo, I updated the PR here as well to match it:
https://github.com/zaceno/hyperapp-jsx-pragma/pull/2

pkg/jsx/jsx-runtime.js
jorgebucaran: Is there any documentation of `jsxs`?

loteoo: Best I could find:
https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#react-automatic-runtime
(second example block shows with jsxs)

Also: https://github.com/preactjs/preact/blob/master/jsx-runtime/src/index.js#L7-L10

And: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#whats-different-in-the-new-transform

pkg/jsx/package.json
loteoo: We can rename to index.js here but it would have to be moved to a file named `jsx-runtime` in a publishing step or something? If you look at https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#whats-different-in-the-new-transform, 
image

loteoo: It's published via @zaceno's hyperapp-jsx-pragma now and seems to work fine so far

jorgebucaran: Thanks, I see now.

loteoo:
Thanks @zaceno for the merge and deploy, much appreciated ! 🙏 

@jorgebucaran We can now try it out! Check out the readme here: https://github.com/loteoo/hyperapp/tree/jsx-runtime/packages/jsx#hyperappjsx (readme from this PR). Just replace `@hyperapp/jsx` with `hyperapp-jsx-pragma`.

Final bit that needs to be decided is what "standard" do we want to promote for receiving children via JSX components (view functions).

Right now, the children is passed via the 2nd argument on the "components". ex:

```jsx
const Foo = (props, children) => (
  <div>
    {children}
  </div>
)

// <Foo>Test</Foo>
```

And that's something we can change. It is determined by this bit: https://github.com/zaceno/hyperapp-jsx-pragma/blob/main/jsx-runtime/index.js#L7

We could also have children passed via `props.children`, so it would look like this instead:

```jsx
const Foo = ({ children, ...props }) => (
  <div>
    {children}
  </div>
)

// <Foo>Test</Foo>
```

What do you think? @jorgebucaran @zaceno

packages/jsx/jsx-runtime.js
```diff
@@ -0,0 +1,23 @@
+import { h, text } from "hyperapp"
+
+const childNode = (child) => typeof child === "string" || typeof child === "number" ? text(child) : child;
```
mshgh, line 3: Just my 2c
I use similar condition in my private project and have to check on `typeof child === "bigint"` too.

jorgebucaran: Wow, true. What about `boolean` also? What about dates?

What would you expect this render:
```js
<ul>
  <li>{true}</li>
  <li>{false}</li>
  <li>{Date.now()}</li>
</li>
```

Just playing 😈's advocate, really.

jorgebucaran: Wow, true. What about `boolean` also? What about dates?

What would you expect this to render?
```js
<ul>
  <li>{true}</li>
  <li>{false}</li>
  <li>{Date.now()}</li>
</li>
```

Just playing 😈's advocate here.

jorgebucaran: Wow, true! And what about `boolean` types? What about dates? Anything else? 🤔 

What would you expect this to render?

```js
<ul>
  <li>{true}</li>
  <li>{false}</li>
  <li>{(new Date)}</li>
</li>
```

Just playing 😈's advocate here.

zaceno: It's sensible reasoning. Maybe a better approach would be:

```js
child => ('node' in child) ? child : text(child.toString())
```

loteoo: @zaceno Hmm not sure why but I was having some issues with the above function. I ended with this instead:

```js
const textNodes = ["string", "number", "bigint"];
const childNode = (child) => textNodes.includes(typeof child) ? text(child) : child;
```

mshgh: I do not think we can include `false` aka `boolean` as it would prevent usage of `<div>{condition && element}</div>`.
Not sure about `Date`. Anyway, `typeof new Date` is "object". There had to be added `instanceof` base check.

loteoo: @mshgh @zaceno @jorgebucaran Sorry for the late reply, here's new solution that seems to cover to everything:

```js
const isVNode = (child) => typeof child === "object" && ('node' in child)

const childNode = (child) => {
  if (child) {
    return isVNode(child) ? child : text(child)
  }
}
```

I updated the PR as well, let me know what you think!

jorgebucaran:
Using flatMap() is an option too:

```js
const jsx = (type, props, ...children) =>
  typeof type === "function"
    ? type(props, children)
    : h(type, props || {}, children.flatMap((any) =>
        typeof any === "string" || typeof any === "number" ? text(any) : any
      )
    )
```

jorgebucaran:
Since Hyperapp's core doesn't know about JSX, some things may or not work.

> I still don't know how flatMap could be advantageous...

Using `flatMap()` allowed me to get rid of `.concat(...children)`.

```diff
h(type, props || {}, 
-  [].concat(...children)
-  .map((any) =>
+ children.flatMap((any) =>
    typeof any === "string" || typeof any === "number" ? text(any) : any
  )
)
```

jorgebucaran:
I've deprecated all `@hyperapp/*` packages and I will not be adding any new ones at this time. I'm happy to consider `@hyperapp/jsx` down the line, but only after we've tested it out and make sure it makes sense for Hyperapp. 👋

loteoo:
> I've deprecated all `@hyperapp/*` packages...

Makes sense! For now I'll add my JSX related improvements in @zaceno's hyperapp-jsx-pragma repo to centralize everything there :) 

Thanks @zaceno for maintaining that repo by the way!

PR #1012, opened 2021-01-22
Cleanup old sub before starting new sub on restart

### LINKED ISSUE #944: Restarted subs run new before unsubscribe
I ran into an unexpected behavior when a subscription restarts where the subscription function is called with the new props _before_ the old subscription cleanup/unsubscribe function is run. This order can cause issues if the setup/teardown of the subscription operates on a shared resource where running the cleanup after subscribing breaks things.

Here's the line of code causing the behavior: https://github.com/jorgebucaran/hyperapp/blob/4121de9db41a3e98d8407b00b2a26f0eb410d96d/src/index.js#L85

I would have expected the line after to execute first:
https://github.com/jorgebucaran/hyperapp/blob/4121de9db41a3e98d8407b00b2a26f0eb410d96d/src/index.js#L86

Is this intentional/expected?

okwolf:
This reverses the order of execution for `newSub[0](dispatch, newSub[1])` and `oldSub && oldSub[2]()` while still storing the results of the former in the sub array for later cleanup execution.

Closes #944.

### DIFF
--- a/index.js
+++ b/index.js
@@ -57,8 +57,7 @@ var patchSubs = (oldSubs, newSubs, dispatch) => {
           ? [
               newSub[0],
               newSub[1],
-              newSub[0](dispatch, newSub[1]),
-              oldSub && oldSub[2](),
+              (oldSub && oldSub[2](), newSub[0](dispatch, newSub[1]))
             ]
           : oldSub
         : oldSub && oldSub[2]()


okwolf:
Or, if you want to keep the number of bytes the same, we could still store both in the array and swap the order like this:
```js
[
  newSub[0],
  newSub[1],
  oldSub && oldSub[3](),
  newSub[0](dispatch, newSub[1])
]
```

jorgebucaran:
I prefer the original commit, as it eliminates the potential of storing something in `subs[3]`. The unsubscribe function could technically (possibly by mistake) return something and we don't want to keep track of that anywhere.

jorgebucaran:
I prefer the original commit, as it eliminates the potential of storing something in `subs[3]`. The unsubscribe function could technically (possibly by mistake) return something and we don't want to keep track of that anywhere.

I'm not 100% certain that this is a bug, as reversing the order may break your counter app in a parallel universe, but I couldn't detect any issues and the idea is perfectly logical (cleanup before restart) so I'm going to tag this as such and merge here.

PR #1014, opened 2021-01-24
Search up the tree to find which A was clicked on

shish:
Given the code `<a href="/blah"><span>foo</span></a>`, the original onclick handler would see `event.target = <span>`, and then `href = undefined`. With this change, we iterate up from the `<span>` to the `<a>` and then get the `href` from there.

### DIFF
--- a/packages/navigation/index.js
+++ b/packages/navigation/index.js
@@ -27,14 +27,16 @@ export const onUrlRequest = fx(
   ([action]) => ({ action }),
   (dispatch, { action }) => {
     const clicks = (event) => {
+      let target = event.target;
+      while (target && !target.matches("a")) target = target.parentElement;
       if (
         !event.ctrlKey &&
         !event.metaKey &&
         !event.shiftKey &&
-        event.target.matches("a")
+        target
       ) {
         event.preventDefault()
-        const href = event.target.getAttribute("href")
+        const href = target.getAttribute("href")
         dispatch(action, { pathname: href })
       }
     }


jorgebucaran:
So, is the current behavior incorrect? Is this an improvement? Do you have a small example, pseudo-code even, to demonstrate what the issue is about? Thanks for working on this. 🙌

shish:
I'd say the current behaviour is inconsistent - I'd *personally* consider it "incorrect", but I guess it depends what the intention is :)

As a maybe clearer example, `<a href="/blah">one <b>two</b> three</a>` - my expectation is that whether I click on "one", "two", or "three", they're all part of the same link, so clicking on any of them should have the same behaviour. Right now, "one" and "three" trigger `pushState()`, but "two" triggers an HTTP request. I can't think of any other case in webdev where an individual link will act differently depending on which part of it gets clicked, hence the vague feeling of wrong-ness.

Aside from text, I think this also means that `<a href="/blah"></a>` will never trigger a `pushState`?

Interactive example:
https://codesandbox.io/s/hyperapp-minimal-navigation-example-2-forked-w4j9z?file=/src/index.js

jorgebucaran:
The example works. I need an example that doesn't.

shish:
I am confused - for me at least (trying in firefox 84 and chrome 88), that example is showing the "bug" (if you click on the plain part of the link you get one behaviour, if you click on the bold part you get different behaviour)

An example of the "fix" after this PR is implemented (the whole link acts the same):
https://codesandbox.io/s/hyperapp-minimal-navigation-example-2-forked-hhr7e?file=/src/navigation/index.js

(Both examples now include counters which measure number of clicks vs page loads)

jorgebucaran:
I didn't know I had to click on the bold text—duh! I see it now, great, we should add this too. 💯

jorgebucaran (approved):

PR #1015, opened 2021-01-24
Consistent use of Location object

shish:
I'm not sure how this worked for the example at https://codesandbox.io/s/hyperapp-minimal-navigation-example-2-g0mr7 - when I tried it with my real app, I had issues because sometimes `state.location` was a `Location` object and sometimes it was a `string`. When it's consistently a `Location`, everything seems to work :)

### DIFF
--- a/packages/navigation/index.js
+++ b/packages/navigation/index.js
@@ -11,7 +11,7 @@ export const pushUrl = fx(
 export const onUrlChange = fx(
   ([action]) => ({ action }),
   (dispatch, { action }) => {
-    const popstate = (_) => dispatch(action, location.href)
+    const popstate = (_) => dispatch(action, location)
 
     addEventListener("popstate", popstate)
     addEventListener("hyperapp-pushstate", popstate)


jorgebucaran:
We need to define what location is going to mean for us, so it may be a good idea to create our pure [Location](https://developer.mozilla.org/en-US/docs/Web/API/Location) wrapper.

Do you have any ideas?

shish:
I have no strong feelings one way or another - For my current use-case I only care about `location.pathname`, so my first attempt to make it work was to have everything be a `string` which only contained the path - but then I thought maaaaybe some people might want to do certain actions based on `location.search` or `location.hash`, so might as well go for the whole object.

(I did note though that it's a security violation to `pushState()` some value which changes the current protocol or domain name; not sure if that affects the decision at all)

jorgebucaran:
We can do it little by little, starting with just `pathname`, and keep adding to it as the need arises.

jorgebucaran:
We can do it little by little, starting with just `pathname` and keep adding to it as the need arises.

jorgebucaran:
Closing as per https://github.com/jorgebucaran/hyperapp/issues/1033#issuecomment-811646682.

PR #1016, opened 2021-01-24
TypeScript definitions for 2.0.10

icylace:
A new and improved TypeScript definitions PR !


zaceno:
Here is a test of a valid use of `app`, which unfortunately fails with the current types. 

```js

// -------------------------------------
// app with initial effect

import { delay } from '../../packages/time/index.js'

// $ExpectType void
app<Test>({
  init: [
    {foo: 2}
    delay(2000, {foo: 3})
  ]
})


```

zaceno:
```ts
type StateWithEffects<S, D = any> = [State<S>, ...(Effect<S, D> | RunnerDescriptor<S, D>)[]]
```
appears to permit the following usage

```ts
import {delay} from @hyperapp/time
const MyAction = state => [state, delay]
```

But as far as I know hyperapp does not support this, and will encounter a runtime error.

zaceno:
The following tests represent valid uses of hyperapp, but fail with the current types. 

```ts

// -------------------------------------
// app with initial effect

import { delay } from '../../packages/time/index.js'
const IncrementFoo: Action<Test> = state => ({ ...state, foo: state.foo + 1 })

// $ExpectType void
app<Test>({
  init: [{ foo: 2 }, delay(2000, { foo: 3 })],
  view: state => h('main', {}, []),
  node: document.body,
})

// $ExpectType void
app<Test>({
  init: [{ foo: 2 }, delay(2000, IncrementFoo)],
  view: state => h('main', {}, []),
  node: document.body,
})

// $ExpectType void
app<Test>({
  init: (state: Test) => [{ foo: 2 }, delay(200, IncrementFoo)],
  view: state => h('main', {}, []),
  node: document.body,
})
```

As far as I can tell, the reason they fail is because `delay(200, action | nextstate)` is not assignable to `RunnerDescriptor<Test>`  but I could be mistaken.

The error messages are: 

```
> hyperapp@2.0.10 dts /Users/zache/Code/hyperapp
> dtslint types

Error: /Users/zache/Code/hyperapp/types/test/app.test.ts:126:12
ERROR: 126:12  expect  TypeScript@4.2 compile error: 
Type '{ foo: number; }' is not assignable to type 'Action<Test, any>'.
  Object literal may only specify known properties, and 'foo' does not exist in type 'Action<Test, any>'.
ERROR: 133:12  expect  TypeScript@4.2 compile error: 
Type '{ foo: number; }' is not assignable to type 'Action<Test, any>'.
  Object literal may only specify known properties, and 'foo' does not exist in type 'Action<Test, any>'.
ERROR: 140:3   expect  TypeScript@4.2 compile error: 
Type '(state: Test) => [{ foo: number; }, (((dispatch: any, props: any) => NodeJS.Timeout) | { delay: any; action: any; })[]]' is not assignable to type 'StateFormat<Test> | Action<Test, any>'.
  Type '(state: Test) => [{ foo: number; }, (((dispatch: any, props: any) => NodeJS.Timeout) | { delay: any; action: any; })[]]' is not assignable to type 'ActionTransform<Test, any>'.
    Type '[{ foo: number; }, (((dispatch: any, props: any) => Timeout) | { delay: any; action: any; })[]]' is not assignable to type 'StateFormat<Test> | Action<Test, any>'.
      Type '[{ foo: number; }, (((dispatch: any, props: any) => Timeout) | { delay: any; action: any; })[]]' is not assignable to type 'ActionDescriptor<Test, any>'.
        Type '{ foo: number; }' is not assignable to type 'Action<Test, any>'.
          Type '{ foo: number; }' is not assignable to type 'ActionDescriptor<Test, any>'.

    at /Users/zache/Code/hyperapp/node_modules/dtslint/bin/index.js:206:19
    at Generator.next (<anonymous>)
    at fulfilled (/Users/zache/Code/hyperapp/node_modules/dtslint/bin/index.js:6:58)
```

icylace:
@zaceno Nice work!  I'll look at these later.

icylace:
@asabatba Your subscriptions example from https://github.com/jorgebucaran/hyperapp/pull/969#issuecomment-750034695 seems to expose a bug in my types. An important find indeed! I'll see what I can do.

icylace:
@zaceno 

> ```ts
> import {delay} from @hyperapp/time
> const MyAction = state => [state, delay]
> ```
> 
> But as far as I know hyperapp does not support this, and will encounter a runtime error.

Well,

```ts
const MyAction = state => [state, delay]
```

is, in Hyperapp's eyes, equivalent to

```ts
const MyAction = state => [(state) => state, delay]
```

which is covered by the type definitions

```ts
type Action<S, P = any> = ActionTransform<S, P> | ActionDescriptor<S, P>
// ...
type ActionDescriptor<S, P> = [Action<S, P>, Payload<P>]
```

zaceno:
@icylace You're mistaken I'm afraid. 

```
state => [state, delay]
```

and

```
state => [(state) => state, delay]
```

are _not_ equivalent to hyperapp.  The former causes a runtime error while the latter does not, as evidenced by [this live example][example]


[example]: https://flems.io/#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAOgAsAXLKEAGhAGMB7NYmBvEAXwvW10QICsEqdBk2J4IWAA60ATsQAEwQhTmMAHsWUYJE1nLDTaWOQHISxCXEQB6KwFc0EgNYBzfHSxXCATwkxpWiQABACZ8AAZ8AEYw4wAdNHEpWQUAExgoDC9dfUMTMwtrOwcXN0MrQO9ff20rYnEYOLR4oTh5AFkvACEMFIBBajr6OQBeOQAKVoxGAEoRgD45ZEnGZTSMrwBdZvpWuQ6AcVpaPoGIIdGJ4imYWeGF5Evr24XlmFX0zK3MbTHgeLk5BAEsREAp2P85AA3CAwADuoLGzzkhDGxgARscvMZlMB2IsIQCUejbMRiPRsQp6NQoBBqI5QR1uidBmg8WpiKiAMI0ulyJnGabTCgE5GotEksloCnAKk8+l7LyHY79FlsmDqLlyuRKlICoUQjb6tAAtDHGCglK0ai2HAMfAYlJeeKsabxSggODpGCnHZ4ACciAAzME2BwQJgcHg3HABDR6IxmDw2BtWEA

zaceno:
> @zaceno I love test cases !

 Sorry I think I misunderstood you before. Like you were approving of my flems.io example.😅  I now think you meant: "please give me a test case that I can make work, or reject as being an invalid test case" ? If so, here's a failing test case. I believe the types should be altered so that it does not fail:

```ts

import { Dispatch, Payload, Effect, Action } from "hyperapp"

type Test = { x: number; y: number }

const runJustEcho = (
  _dispatch: Dispatch<Test>,
  data?: Payload<string>
): void => {
  console.log(data)
}

const justEcho: Effect<Test, string> = (x) => [runJustEcho, x]

/*
  The following action definition is not valid 
  hyperapp. We would expect a type error.

  However, the types as defined allow this
  action definition. Hence this test
  case fails
*/
// $ExpectError
const IncrXAndEcho: Action<Test> = (state: Test) => [
  { ...state, x: state.x + 1 },
  justEcho
]

```

zaceno:
@icylace Here's another test case that I would not expect to fail but does:

```ts

// $ExpectType VDOM<unknown>
h("input", {oninput: (state, event) => event.target.value})

```


Error output:

```
ERROR: 228:40  expect                  TypeScript@4.2 compile error: 
Object is possibly 'undefined'.
ERROR: 228:40  expect                  TypeScript@4.2 compile error: 
Object is possibly 'null'.
ERROR: 228:53  expect                  TypeScript@4.2 compile error: 
Property 'value' does not exist on type 'EventTarget'.
```

I'm afraid I'm not good enough with TS yet to know wether the types should be improved, loosened or if the test itself needs to have some type annotations to pass.

zaceno:
Further findings about https://github.com/jorgebucaran/hyperapp/pull/1016#issuecomment-766322451 :

It is because RunnerDescriptor is typed as a tuple: 

```ts
 type RunnerDescriptor<S, D> = [Runner<S, D>, Payload<D>]
```

However, the type-checker is unable to infer that `delay(2000, { foo: 3 })` matches `RunnerDescriptor`, as TS interprets the return value `[fn, {foo: number}]` at best as `(Runner | Payload)[]` – that is to say: an array of _any_ length where each of the items could be _either_ `Runner` or `Payload` type. – which of course doesn't match.

The following is how the test could be made to work:

```ts

import { h, app, Action, RunnerDescriptor } from "hyperapp"
import { delay } from "../../packages/time/index.js"

type Test = { foo: number; bar?: number }

const IncrementFoo: Action<Test> = state => ({ ...state, foo: state.foo + 1 })

// $ExpectType void
app<Test>({
    init: [{ foo: 2 }, delay(200, IncrementFoo) as RunnerDescriptor<Test, any>], // <--- note "as RunnerD..."
    view: state => h("main", {}, []),
    node: document.body,
})
```

The same problem, and same approach would go for any action/event handler where an action with payload is needed: It would have to be cast to `ActionDescriptor` using `as`.  

Seems like it could end up getting annoying... especially when using plain javascript as it requires using  `/** @type {import('hyperapp').ActionDescriptor} */([MyAction, 'foo'])` (I think. Have not yet verified this). Wonder if there's a way to cast types by force inside the declaration file?

jorgebucaran:
Are we missing a lot here or can I merge this now? 🤔

jorgebucaran:
I want to wait for this before going official, but just how close are we to merging here? Left you some comments, @icylace.

package.json
```diff
@@ -21,10 +25,13 @@
     "release": "tag=$(npm run --silent which) npm run deploy && cd ./${pkg:+packages/$pkg} && npm publish --access public",
     "deploy": "npm test && git commit --all --message $tag && git tag --sign $tag --message $tag && git push && git push --tags",
     "which": "node --print \"('$pkg' ? '@$npm_package_name/$pkg@' : '') + require('./${pkg:+packages/$pkg/}package').version\"",
-    "test": "c8 twist tests/*.js"
+    "test": "c8 twist tests/*.js",
+    "dts": "dtslint types"
   },
   "devDependencies": {
+    "dtslint": "*",
```
jorgebucaran, line 32: Can we do without these two dev deps?

jorgebucaran (changes requested):
I'd like to remove all non-essential type tests, and put just the bare minimum in a single file.

icylace:
> I'd like to remove all non-essential type tests, and put just the bare minimum in a single file.

I would agree that any redundant tests should be removed but what is a "non-essential" test?  Why would you assume the bare minimum fits appropriately in a single file?

jorgebucaran:
I just one to merge `index.d.ts`, so perhaps the comments could include some examples? 😄

types/index.d.ts
```diff
@@ -0,0 +1,166 @@
...
+  // ---------------------------------------------------------------------------
+
+  // A Hyperapp application instance has an initial state and a base view.
+  // It must also be mounted over an available DOM element.
+  type App<S> = Readonly<{
+    init: StateFormat<S> | Action<S>
+    view: View<S>
+    node: Node
+    subscriptions?: Subscriptions<S>
+    middleware?: Middleware<S>
```
jorgebucaran, line 42: Should be `dispatch` now.

types/index.d.ts
```diff
@@ -0,0 +1,166 @@
...
+  type View<S> = (state: State<S>) => VDOM<S>
+
+  // The subscriptions function manages a set of subscriptions.
+  type Subscriptions<S> = (state: State<S>) => Subscription<S>[]
+
+  // A subscription represents subscriber activity.
+  type Subscription<S, D = any> = boolean | undefined | SubscriberDescriptor<S, D> | Unsubscribe
+
+  // A subscriber reacts to subscription updates.
+  type SubscriberDescriptor<S, D> = [Subscriber<S, D>, Payload<D>]
```
jorgebucaran, line 55: Do we need this `SubscriberDescriptor ` type? Why not just inline `[Subscriber<S, D>, Payload<D>]` in the `Subscription` definition?

types/index.d.ts
```diff
@@ -0,0 +1,166 @@
...
+  // A dispatched action handles an event in the context of the current state.
+  type Dispatch<S> = (action: Action<S>, props?: Payload<any>) => void
+
+  // An action transforms existing state and/or wraps another action.
+  type Action<S, P = any> = ActionTransform<S, P> | ActionDescriptor<S, P>
+  type ActionTransform<S, P = any> = (state: State<S>, props?: Payload<P>) => StateFormat<S> | Action<S>
+  type ActionDescriptor<S, P> = [ActionTransform<S, P>, Payload<P>]
+
+  // A transform carries out the transition from one state to another.
+  type Transform<S, P = any> = (state: StateFormat<S>, props?: Payload<P>) => StateFormat<S>
```
jorgebucaran, line 75: What's a `Transform`? 💭

types/index.d.ts
```diff
@@ -0,0 +1,166 @@
...
+  type Middleware<S> = (dispatch: Dispatch<S>) => Dispatch<S>
+
+  // ---------------------------------------------------------------------------
+
+  // A dispatched action handles an event in the context of the current state.
+  type Dispatch<S> = (action: Action<S>, props?: Payload<any>) => void
+
+  // An action transforms existing state and/or wraps another action.
+  type Action<S, P = any> = ActionTransform<S, P> | ActionDescriptor<S, P>
+  type ActionTransform<S, P = any> = (state: State<S>, props?: Payload<P>) => StateFormat<S> | Action<S>
```
jorgebucaran, line 71: `StateFormat` is used only once (twice if we count `Transform`, but I'm not sure that will make it to the types). 

Could we just define the return type as `State<S> | StateWithEffects<S>`, instead of defining a new type?

types/index.d.ts
```diff
@@ -0,0 +1,166 @@
...
+  type State<S> = S
+
+  // State can be associated with a list of effects to run.
+  type StateWithEffects<S, D = any> = [State<S>, ...(Effect<S, D> | RunnerDescriptor<S, D>)[]]
+
+  // An effect is an abstraction over an impure process.
+  type Effect<S, D = any> = (..._: any[]) => RunnerDescriptor<S, D>
+
+  // A runner is where side effects and any additional dispatching may occur.
+  type RunnerDescriptor<S, D = any> = [Runner<S, D>, Payload<D>]
```
jorgebucaran, line 90: Same as with the `SubscriberDescriptor`, do we need a `RunnerDescriptor`? I'd favor inlining it instead.

types/index.d.ts
```diff
@@ -0,0 +1,166 @@
...
+    readonly children: VNode<S>[]
+    node: MaybeNode
+    readonly tag: Tag<S>
+    readonly key: Key
+    memo?: PropList<S>
+    events?: Record<string, Action<S>>
+
+    // `_VDOM` is a guard property which gives us a way to tell `VDOM` objects
+    // apart from `PropList` object.
+    _VDOM: true
```
jorgebucaran, line 111: Could you explain how this works? As you know, vnodes don't actually have a `_VDOM` property, but somehow this helps the TypeScript compiler infer something?

jorgebucaran (changes requested):

types/index.d.ts
```diff
@@ -0,0 +1,166 @@
+// Minimum TypeScript Version: 4.2
+
+// NOTE: `dtslint` needs 4.2 even though these definitions should work with 4.1.
+
+declare module "hyperapp" {
+  // `app()` initiates a Hyperapp application. `app()` along with runners and
+  // subscribers are the only places where side effects are allowed.
+  function app<S>(props: App<S>): void
```
zaceno, line 8: App now returns dispatch! Not void

types/index.d.ts
```diff
@@ -0,0 +1,166 @@
...
+    value: T extends (symbol | ((..._: any[]) => any)) ? never : T,
+    node?: Node
+  ): VDOM<S>
+
+  // ---------------------------------------------------------------------------
+
+  // A Hyperapp application instance has an initial state and a base view.
+  // It must also be mounted over an available DOM element.
+  type App<S> = Readonly<{
+    init: StateFormat<S> | Action<S>
```
zaceno, line 38: init is optional. All props are optional, actually. The only rule is that if you define `view` you must also define `node`.   I'd suggest using two different sets of `App` and overloading the `app` declaration

types/index.d.ts
```diff
@@ -0,0 +1,166 @@
...
+  // Middleware allows for custom processing during dispatching.
+  type Middleware<S> = (dispatch: Dispatch<S>) => Dispatch<S>
+
+  // ---------------------------------------------------------------------------
+
+  // A dispatched action handles an event in the context of the current state.
+  type Dispatch<S> = (action: Action<S>, props?: Payload<any>) => void
+
+  // An action transforms existing state and/or wraps another action.
+  type Action<S, P = any> = ActionTransform<S, P> | ActionDescriptor<S, P>
```
zaceno, line 70: Incorrect, considering later down you bind all `onsomething` events to `Action<S, WhateverPayloadFromEvent>`

This definition of action means: `onclick: [(state, x) => ({...state, foo: state.foo + x}), 42]`  would be invalid when it is clearly not. 

I would suggest changing it to `Action<S, P> = ActionTransform<S,P>| ActionDescriptior<S>`  and then use a validator function to ensure that payloads given in action descriptors match the required type for the ActionTransform. (Like I did in my set of types)

types/index.d.ts
```diff
@@ -0,0 +1,166 @@
...
+  type Transform<S, P = any> = (state: StateFormat<S>, props?: Payload<P>) => StateFormat<S>
+
+  // State can either be on its own or associated with effects.
+  type StateFormat<S> = State<S> | StateWithEffects<S>
+
+  // Application state is accessible in every view, action, and subscription.
+  type State<S> = S
+
+  // State can be associated with a list of effects to run.
+  type StateWithEffects<S, D = any> = [State<S>, ...(Effect<S, D> | RunnerDescriptor<S, D>)[]]
```
zaceno, line 84: Incorrect, because Effect<S, D> is a _function_ and if an action returns `[someState, anyFunction]` it will immediately be a runtime error. 

It should be just:

```
type StateWithEffects<S, D = any> = [State<S>, ...RunnerDescriptor<S, D>[]]
```

I would prefer not having the `Effect` definition in the types at all, and instead use the _name_ "Effect" for what currently is called "RunnerDescriptor"

zaceno (changes requested):

types/index.d.ts
icylace: Wow, I thought I had fixed that, LOL !

types/index.d.ts
icylace: Ah, great catch !

types/index.d.ts
icylace: Regarding the current `Effect` type and its usage in `StateWithEffects`, good point !

I'll need to ponder the rest.

types/index.d.ts
icylace: Could you expand your `onclick` example into a test case ?

types/index.d.ts
icylace: Yeah, so TypeScript does structural typing, which means the props an object has acts as a distinguishing factor rather than what the name of the object type is, as would be the case in nominal type systems, like Java, C++, etc.

types/index.d.ts
icylace: D'oh !!!   Haha, good catch.

types/index.d.ts
icylace: `Transform` would apply to the functions that more complicated actions would leverage.

types/index.d.ts
icylace: The original intent of `StateFormat` was to act as a convenience for actions that may or may not return effects depending on what happens within them.

types/index.d.ts
icylace: I'll have to ponder it.

types/index.d.ts
jorgebucaran: So, this is only a trick, I see, since vnodes don't have a `_VDOM` prop. Okay!

types/index.d.ts
jorgebucaran: What do you mean by a  small action? Actions can't be called, only given to Hyperapp or returned from other actions.

types/index.d.ts
zaceno, line 71: props should not be optional. With the current definition if I try to define an action like this:

```ts
type State = {foo: number}
const AddSome : ActionTransform<State, number> = (state, amount) => ({
  ...state, 
  foo: state.foo + amount
})
```

I get an error squiggle underneath ` +  amount` with the error message saying that `amount` might be undefined.

zaceno (changes requested):

types/index.d.ts
zaceno: I tried, but I got stuck on not being able to define an action. Gonna see if I can work around it

types/index.d.ts
zaceno: I don't see much value to defining this type tbh. I mean sure, it exists as a concept but it could easily be defined by a user who needs it.

types/index.d.ts
zaceno: Ok here's the test:

```ts
import { app, h, text, ActionTransform } from "hyperapp"

type State = { foo: number }

const AddSome: ActionTransform<State, number> = (state, amount) => ({
    ...state,
    foo: state.foo + (amount || 0),
})

//$ExpectError
h("button", { onclick: AddSome }, text("add event?!"))

//$ExpectType VNode
h("button", { onclick: [AddSome, 32] }, text("add 32"))

//$ExpectError
h("button", { onclick: [AddSome, "foo"] }, text("add string?!"))
```

zaceno: The part that fails now is the middle button, because `AddSome` does not take `MouseEvent` for a payload.

types/index.d.ts
zaceno: To clarify what I mean:

Users define their own actions, and should have full control over what types they will allow for payloads in their actions. So if I am prepared to accept that the payload is either a number or undefined, then I could define my action as:

```ts

const Add:ActionTransform<MyState, number | undefined> = (state, amount) =>
  amount === undefined ? state :  ({
    ...state, foo: state.foo + amount
 })
 ```
 
 If, however, I do not want anyone ever to dispatch the action with undefined payload, I should be able to define the action as simply:
 
 ```ts
 
 const Add:ActionTransform<MyState, number> = (state, amount) => ({...state, foo: state.foo + amount})
```

And then I would expect the type checker to warn me elsewhere, whenever I do something that might cause the payload to become anything but a number.

types/index.d.ts
icylace: @jorgebucaran In the docs [here](https://github.com/jorgebucaran/hyperapp/blob/main/docs/architecture/actions.md#transforms), the `Liokaiser` action calls functions that would, in a TypeScript project, be defined in terms of `Transform`.  A small action would be small enough to not have to be split up.

@zaceno There are was a Slack discussion about `Transform` that you had at the time agreed with me on:  https://hyperapp.slack.com/archives/C41ECC0V6/p1600930431006000?thread_ts=1600929373.003100&cid=C41ECC0V6

My defense for having `Transform` ready "out-of-the-box" is that I had found it to be a useful type that would show up often in larger projects.

types/index.d.ts
icylace: Exactly!  `_VDOM` takes advantage of how TypeScript works without forcing the user to do anything special with their code.

types/index.d.ts
icylace: This custom payload issue is a great catch !

jorgebucaran:
Let's just merge `index.d.ts` and take it from there. @icylace Can you update the PR? 🙏

icylace:
I'm making updates soon which includes bug fixes and some renaming.

icylace:
Massive changes were made !!!

Here's my changelog:

- I moved the types test suite into their own dedicated repo! For those interested, they're now available at: https://github.com/icylace/hyperapp-types-tests
  - This was done to help @jorgebucaran focus his long-term resources.
- `EventActions` now recognizes when custom payloads are being used with event handlers.
  - However, it's not perfect. The types test repo contains the relevant test cases, graciously provided by @zaceno.
- Updated `App` in the following ways:
  - `Dispatch` is returned instead of `void`.
  - `init:` was made optional.
  - Renamed `middleware:` to `dispatch:`.
  - Refactored/expanded `App` as a whole to treat `app({})` as an error.
  - Other obscure edge cases are currently not handled correctly. See the types test repo for details.
- In discussions about the recently merged docs PR @zaceno suggested that VNodes were redundant and ultimately I agree, so I replaced `VNode` with the new `MaybeVDOM`. Also, I removed talk of VNodes from the docs.
- Inlined `MaybeNode` in the only placed it was used and then removed it because it seemed unnecessary.
- Inlined `StateFormat` where its used in the type definitions and then removed it.
  - I created `StateFormat` to handle situations where an action may or may not invoke effects depending on what happens within its processing. However, I later realized `StateWithEffects` can cover this use case pretty well.
- The original `Effect` got dropped from `StateWithEffects` as per @zaceno's suggestion.
- Renamed the `props` parameter of `ActionTransform` to `payload`.
- Renamed the `props` parameter of `Transform` to `payload`.
- Renamed the `props` parameter of `Runner` to `payload`.
- Renamed `Effect` to `EffectCreator`.
- Renamed `RunnerDescriptor` to `Effect`.
- Renamed `Runner` to `Effecter`.
- Inlined `Subscription` into `Subscriptions`.
- Removed the original `Subscription`.
- Renamed `SubscriberDescriptor` to `Subscription`.
- The docs were updated to reflect the new terminology.
- Some comments are either new, removed, reordered, or slightly updated.
- Test cases were updated accordingly.

For now, I've decided to leave in certain "superfluous" but useful types, namely `EffectCreator` and `Transform`. I believe their inclusion in the core types is worthy of debate but for the moment I think the focus should be on having the types as a whole in a releaseable state.

types/index.d.ts
icylace: I decided to remove `StateFormat`.

types/index.d.ts
icylace: My latest changes affect these greatly !

types/index.d.ts
icylace: Got this in the latest changes.

types/index.d.ts
icylace: Should be fixed now, hopefully !

package.json
icylace: My latest changes includes off-loading the types tests into their own brand new repo !

types/index.d.ts
icylace: My latest changes improves `App` but there's still room for improvement.  Perhaps you're right about overloading.

types/index.d.ts
icylace: My latest changes includes an updated `StateWithEffects` definition as well as some relevant renaming.

types/index.d.ts
icylace: I've yet to solve that third case with `[AddSome, "foo"]`.  Were you able to solve that in your types ?

zaceno:
Wow @icylace, massive work! Look forward to digging in to the news!

jorgebucaran (approved):

jorgebucaran:
Thank you, @icylace! 💯

types/index.d.ts
icylace: Addressed in https://github.com/jorgebucaran/hyperapp/pull/1047.

types/index.d.ts
icylace: Addressed in https://github.com/jorgebucaran/hyperapp/pull/1047.

types/index.d.ts
icylace: For posterity, this was addressed in https://github.com/jorgebucaran/hyperapp/pull/1045.

PR #1019, opened 2021-01-26
Adds effects compositions

sergey-shpak:
As update to https://github.com/jorgebucaran/hyperapp/pull/955

This PR makes dispatching effects recursive, which allows to dispatch multiple effect compositions, for example:

```js
const EffectA = () => { /* some effect */ }
const EffectB = () => { /* some effect */ }

const action = state => 
  [[ state, [ EffectA, {} ] ], [ EffectB, {} ] ]
```
or
```js
const ActionA = (state, paramA) => { /* some action */ }
const EffectB = () => { /* some effect */ }

const action = state => 
  [[ actionA, paramA ], [ EffectB, {} ] ]
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -394,7 +394,7 @@ export var app = (props) => {
             .slice(1)
             .map(
               (fx) => fx && fx !== true && fx[0](dispatch, fx[1]),
-              setState(action[0])
+              dispatch(action[0])
             )
       : action == null
       ? patchSubs(subs, EMPTY_ARR, (dispatch = id))


jorgebucaran (approved):

jorgebucaran:
I like how this extends the recursive nature of `dispatch`, but it makes it too difficult to tell apart actions with payloads and effects. If you really wish to dispatch an action and then run an effect, you could create an effect just for that:

```js
const runAction = action => [dispatch => dispatch(action)]
```

and used it like so:

```js
[state, runAction(SomeAction), ..., someEffect]
```

PR #1022, opened 2021-01-29
Explain the conditions which cause a subscription to be restarted

shish:
I spent a long time being very confused about subscriptions, because sometimes they would hang around and sometimes they would get restarted on every state change - now I think I understand how they work, so adding some documentation for the next person to be in my situation :)

~

(Incidentally, even the example subscription suffers from the constant re-subscribing; one needs to extract `["w", "a", "s", "d"]` into a module-level constant to avoid it. Not such a big deal for a keyboard listener where we're only wasting a few nanoseconds of CPU time with frequent addEventListener / removeEventListener - but more of a pain for network stuff. Which makes me think it would be nice if `shouldRestart` was more forgiving, but that's a big discussion for a different time :) )

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -377,6 +377,8 @@ They are used for both adding and removing connections to events outside hyperap
 
 Binds **dispatch** to an external event. Returns a cleanup function that removes the binding.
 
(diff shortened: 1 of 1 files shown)


jorgebucaran:
Thank you, @shish. This section was in dire need of help. I took bits and pieces from your patch and decided to write something myself instead to avoid creating a second commit. 

> it would be nice if `shouldRestart` was more forgiving, but that's a big discussion for a different time 

Want to [start a discussion](https://github.com/jorgebucaran/hyperapp/discussions/new)? The example we have right now is not great anyway. Fortunately, it's going away soon. ✂️

PR #1023, opened 2021-01-30
Fix typo on API reference

sthobis:


docs/reference.md
```diff
@@ -91,7 +91,7 @@ The `class` prop can be either a string of classes or an object of classes. For
 
 <!-- prettier-ignore -->
 ```js
-const variableProfileCardView = (user, useBorders, variant) =>
+const variableProfileCardView = (user, userBorders, variant) =>
```
jorgebucaran, line 94: This should be `useBorders` here, and in line 99

```
useBorders,
```

instead of

```js
useBorders: userBorders
```

jorgebucaran (changes requested):

docs/reference.md
jorgebucaran: Fixed and commited. 💯

jorgebucaran:
Thank you, @sthobis! 🙆‍♂️

PR #1031, opened 2021-02-13
Dispatching action as null destroys the state.

diasbruno:
When action is `null`, it tries to `update(action)`, destroying the state.
We can pass the current `state` to the update function, so it works as a no-op.

### DIFF
--- a/index.js
+++ b/index.js
@@ -412,7 +412,7 @@ export var app = ({
                 (fx) => fx && fx !== true && fx[0](dispatch, fx[1]),
                 update(action[0])
               )
-        : update(action)
+        : update(action || state)
     ))(init),
     dispatch
   )


jorgebucaran:
What's your use case?

jorgebucaran:
Returning `null` or `undefined` from an action is how we stop an app (#873, #1018).

jorgebucaran:
Ping @diasbruno

diasbruno:
Got it. Problably something I'm missing.
I'll close this if I fix the problem on my side.
Thanks.

jorgebucaran:
@diasbruno What problem did you run into?

diasbruno:
I'm using `hyperapp-router` and it uses an effect to dispatch a `CustomEvent`. 
But this effect doesn't call the dispatch function, so the dispatch is called with no arguments destroying the state.

jorgebucaran:
That router should probably check if the action it's trying to dispatch is not `null` or `undefined` first. 😄

diasbruno:
I'll figure something...maybe the library needs to be updated or something...
Thanks.

jorgebucaran:
@diasbruno Where is that library? Link.

diasbruno:
[hyperapp-router](https://github.com/mrozbarry/hyperapp-router)

jorgebucaran:
I browsed through the source, but couldn't find any immediately obvious issues. Can you show me the code where you encountered the problem?

diasbruno:
Here is where I think the problem is (if is in the library): https://github.com/mrozbarry/hyperapp-router/blob/master/src/withRouter/effects.js#L3 

I spent some time debugging yesterday, on the Hyperapp side, and the dispatch received no arguments. 
I thought the problem was in the dispatch mechanism.

If it's not on hypperapp, it must be the library that need an update or I'm doing something wrong 😆

jorgebucaran:
That `navigate` effect only runs `dispatchEvent()` behind the scenes. It's not even using `dispatch`. The problem must be somewhere else, perhaps in your application code.

Make sure you are not returning `null` or `undefined` from an action (which is a Hyperapp idiom meaning to stop the app). 

```js
const StopTheWorld = () => void Infinity
```

diasbruno:
It's not even using dispatch

If there is no problem running an effect that doesn't call the dispatch, then it's on the app.

I'll look for something in the app. If the problem persists, I'll try to make a reproducible example.

PR #1034, opened 2021-02-27
Only use pushState for same-origin URLs

### LINKED ISSUE #1033: @hyperapp/navigation can't handle external links
`history.pushState({}, "", url)` is a security error if `url` has a different origin than the current document.

I'm not sure what the "correct" solution is here, but my feeling is that the nicest thing to do would be to detect links to an external domain and fall back to the browser's default onclick handler instead of trying to handle the click for ourselves (ie don't call `event.preventDefault()`, don't call `dispatch()`)

shish:
Now links within the app will be handled by local JS, but links to external sites will be handled by the browser

Fixes #1033

Test plan: Running on https://scuba.shish.io/ - links within the app work (instantly switching screens), links on the "about" page also work (taking the user to a different website)

### DIFF
--- a/packages/navigation/index.js
+++ b/packages/navigation/index.js
@@ -35,9 +35,12 @@ export const onUrlRequest = fx(
         !event.shiftKey &&
         target
       ) {
-        event.preventDefault()
         const href = target.getAttribute("href")
-        dispatch(action, { pathname: href })
+        const dest = new URL(href, location.href);
+        if (dest.protocol == location.protocol && dest.hostname == location.hostname) {
+          event.preventDefault()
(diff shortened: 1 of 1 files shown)


jorgebucaran:
Do you happen to know what other navigation packages do in similar situations?

shish:
No idea TBH ^^;

jorgebucaran:
Thanks. Well then, I guess I'll look into it when I get to work on this package.

jorgebucaran:
Closing as per https://github.com/jorgebucaran/hyperapp/issues/1033#issuecomment-811646682.

PR #1037, opened 2021-02-28
Use node.append(child) instead of node.appendChild(child)

skanne:
Hi Jorge,

similar to [PR 1036](https://github.com/jorgebucaran/hyperapp/pull/1036) I suggest replacing node.appendChild(child) with node.append(child). Difference explained here: https://dev.to/ibn_abubakre/append-vs-appendchild-a4m.

Benefit: 5 characters/bytes of code less.

Regards,
Sven

### DIFF
--- a/index.js
+++ b/index.js
@@ -114,7 +114,7 @@ var createNode = (vdom, listener, isSvg) => {
   }
 
   for (var i = 0; i < vdom.children.length; i++) {
-    node.appendChild(
+    node.append(
       createNode(
         (vdom.children[i] = maybeVNode(vdom.children[i])),
         listener,


jorgebucaran:
Since `append()` accepts plain DOMStrings, I wonder if it's possible to take advantage of that around here:

https://github.com/jorgebucaran/hyperapp/blob/9df1acafc0d5ac488e52ad92df4bdb47c4a40fcb/index.js#L131-L136

jorgebucaran:
Since [`append()`](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/append) accepts plain DOMStrings, I wonder if it's possible to take advantage of it around here:

https://github.com/jorgebucaran/hyperapp/blob/9df1acafc0d5ac488e52ad92df4bdb47c4a40fcb/index.js#L131-L136

skanne:
Possibly. I'll spend some time on this and find out. I do see some potential around https://github.com/jorgebucaran/hyperapp/blob/9df1acafc0d5ac488e52ad92df4bdb47c4a40fcb/index.js#L105-L110 and https://github.com/jorgebucaran/hyperapp/blob/9df1acafc0d5ac488e52ad92df4bdb47c4a40fcb/index.js#L117-L123

But I need to wrap my head around it. High noon for me to try to understand all inner workings of Hyperapp's core. (Code comments would help, but are non-existent. Are you planning on documenting the core?)

jorgebucaran:
> High noon for me to try to understand all inner workings of Hyperapp's core.

What do you need help with?

skanne:
Well, I need to walk through the code and see what happens. Never bothered to fully understand the concept of patching.

lawrence-dol:
It would seem to me that here, as with #1036 the more generic function is likely to be less performant. At the least, it must inspect argument types and branch, whereas the dedicated function for Nodes expects a Node.

skanne:
The performance difference really is neglible IMO. See https://www.measurethat.net/Benchmarks/Show/9176/0/js-append-vs-appendchild

sergey-shpak:
@skanne performance difference is about 12%, which is not critical for custom usage, but not for hyperapp(superfine)

PR #1038, opened 2021-03-05
Replace .map for .forEach in packages/navigation

loteoo:
Very small change but it bothered me enough to open a PR 🙊


jorgebucaran:
I actually prefer `map()` over `forEach()`. So, no thanks! 🙉

PR #1045, opened 2021-04-05
Types: enable definite payloads

zaceno:
The current types require my actions to handle the eventuality that the payload might be undefined. Even when that can't happen. For example the following example is perfectly valid and would run just fine:

```ts
import { h, text, app } from "hyperapp"

type State = { inputText: string }

const HandleInput = (state: State, event: Event) => ({
    ...state,
    inputText: (event.target as HTMLInputElement).value,
})

let node = document.getElementById("app")
node &&
    app({
        node,
        init: { inputText: "" },
        view: state =>
            h("main", {}, [
                h("input", {
                    type: "text",
                    value: state.inputText,
                    oninput: HandleInput,
                }),
            ]),
    })

```

but typescript complains about it:

Screenshot 2021-04-05 at 20 15 35

This pull request changes it so that there is no error.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -74,7 +74,7 @@ declare module "hyperapp" {
   type Action<S, P = any> = ActionTransform<S, P> | ActionWithPayload<S, P>
   type ActionTransform<S, P = any> = (
     state: State<S>,
-    payload?: Payload<P>
+    payload: Payload<P>
   ) => State<S> | StateWithEffects<S> | Action<S>
   type ActionWithPayload<S, P> = [ActionTransform<S, P>, Payload<P>]
 


jorgebucaran (approved):

jorgebucaran:
Unrelated, but why are you doing weird things like:

```js
let node = document.getElementById("app")
node &&
```

icylace:
@jorgebucaran @zaceno What about using actions in `init:` or direct dispatching from within effecters and subscribers?  Event handlers aren't the only places actions get used.

zaceno:
@icylace sure that's true. Actions could get dispatched in a variety of ways with a variety of payload types (or none). But that's no different than how any typed function _could_ be called with any kind of arguments –– even in ways it doesn't support –– and that's what the types are for.  So I guess I don't really see what you're warning about. Could you give a concrete example?

icylace:
If payload is required, then `init: (state = "hi") => state` is invalid.

icylace:
What a discussion on Discord!  I guess TypeScript acts in funny ways !

jorgebucaran:
TL;DR

I like to think about `init` as a _dispatchable_. That's exactly how Hyperapp uses it to start your app. It's essentially `dispatch(init)`.

zaceno:
To summarize the parts of the discord discussion (those parts relevant to this PR) for posterity:

this

```ts
app<string>({
  init: (state = "hi") => state,
  view: text,
  node: document.body,
})
```

Actually works with no errors. Even though the action does not take a payload, it is still recognized as an action by typescript because: https://github.com/Microsoft/TypeScript/wiki/FAQ#why-are-functions-with-fewer-parameters-assignable-to-functions-that-take-more-parameters

There's more to be said about the typing of `init` (I like the "dispatchable" concept too but I think there's even more to it, that is special for `init` maybe). _But_ I think we're in agreement that this PR is ok.

PR #1046, opened 2021-04-06
Improved Dispatch docs

icylace:
I've made some improvements to the Dispatch documentation in an effort to address the points made by @zaceno [in the now merged overhauled docs PR](https://github.com/jorgebucaran/hyperapp/pull/1042#discussion_r599309868).

### DIFF
--- a/docs/architecture/dispatch.md
+++ b/docs/architecture/dispatch.md
@@ -1,6 +1,6 @@
 # Dispatch
 
-The **dispatch** function controls Hyperapp's core dispatching process which executes [actions](actions.md), applies state transitions, runs [effects](actions.md#effects), patches the DOM, and starts/stops [subscriptions](subscriptions.md) that need it.
+The **dispatch** function controls Hyperapp's core dispatching process which executes [actions](actions.md), applies state transitions, runs [effects](actions.md#effects), and starts/stops [subscriptions](subscriptions.md) that need it.
 
 You can augment the dispatcher to tap into the dispatching process for debugging/instrumentation purposes. Such augmentation is loosely comparable to middleware used in other frameworks.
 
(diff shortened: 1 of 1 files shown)


docs/architecture/dispatch.md
```diff
@@ -1,8 +1,8 @@
 # Dispatch
 
-The **dispatch** function controls how to execute [actions](actions.md) and apply the state transitions and effect running that they request to be done.
+The **dispatch** function controls Hyperapp's core dispatching process which executes [actions](actions.md), applies state transitions, runs [effects](actions.md#effects), patches the DOM, and starts/stops [subscriptions](subscriptions.md) that need it.
```
jorgebucaran, line 3: > patches the DOM,

`dispatch` ultimately causes the DOM to be patched, but very indirectly. What motivated you to add this part?

icylace: Hyperapp has to ultimately patch the DOM at some point during its processing, otherwise the screen would be blank. The mention of it here is just an acknowledgement of that.

Perhaps better wording is needed to show that the patching part is clearly internal functionality ?

jorgebucaran: Or just remove the "patches the DOM," part for now.

jorgebucaran: > Perhaps better wording is needed to show that the patching part is clearly internal functionality?

Sure, but it may be tricky to come up with the right wording, so until then, can we just go without "patches the DOM"?

icylace: Upon reflection I'll take it out.

jorgebucaran (approved):

jorgebucaran:
Thank you, @icylace! 🙌

PR #1047, opened 2021-04-06
Update Types

icylace:
This PR was originally about `App<S>` but has evolved into a pervasive update.

Changelog:

- Added `TypedH<S>`. It lets you define a version of `h` that has your app's state type "baked in" so you don't need to worry about whether or not you need to explicitly assign the generic type parameter for `h` every time you call it.
  - Originally from: https://github.com/jorgebucaran/hyperapp/pull/1049
  - Addresses: https://github.com/jorgebucaran/hyperapp/issues/1048-
- Added the utility types `AllOrNothing<T>`, `AtLeastOne<T>`, and `IndexableByKey`. They assist the implementation of some of the other types which are refactored accordingly.
  - These utility types might be useful to the user in their own right.
- Improved `App<S>` to validate all combinations of `app()` props usage.
- Improved validation of using event actions with custom payloads.
  - For instance, this should be treated correctly now: `h("p", { onclick: [MyAction, 42] }, text("hi"))`.
- Made the `payload` parameter of `Effecter<S, P>` and `Subscriber<S, P>` definite.
  - This follows from discoveries found during discussions of: https://github.com/jorgebucaran/hyperapp/pull/1045
- Removed `EffectCreator<S, D>`. I don't believe it adds enough value even on a purely self-documenting code basis.
- Removed `Payload<P>` because in the places where it could appear either a parameter name or named tuple could express the same level of detail.
- Removed `State<S>` for partially the same reason as removing `Payload<P>`. Another reason was I wanted to prevent potential confusion on the user's end about whether `State<S>` is required like how `StateWithEffects<S>` is.
- Added `Dispatchable<S>`. It simplifies a couple type definitions while adding more meaning. It can also be handy to the user when defining effecters and subscribers.
  - Based off of @zaceno's [idea for a dispatchable type](https://discord.com/channels/804672552348680192/805746611467583499/829100737967030364).
- Added names to tuple types. This helped with the rationale for removing `Payload<P>` and `State<S>`.
- Added a `types` reference to `index.d.ts` in `package.json` per @talentlessguy's [suggestion](https://discord.com/channels/804672552348680192/805746611467583499/830532470122217472).
- Renamed some generic type parameters.
- Added a little more commentary.

---

Original PR title and text for posterity:

### Improve the `App` type

This fixes certain cases when using `App`. I've also added a little more commentary in some parts of the type definitions.

Here are the relevant lines in the types test suite if you're curious:

- https://github.com/icylace/hyperapp-types-tests/blob/4ebe6e928945184bfb96321d40cfa6a88b058686/test/api/app.test.ts#L26
- https://github.com/icylace/hyperapp-types-tests/blob/4ebe6e928945184bfb96321d40cfa6a88b058686/test/api/app.test.ts#L36-L73

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -6,10 +6,10 @@ declare module "hyperapp" {
   function app<S>(props: App<S>): Dispatch<S>
 
   // `h()` builds a virtual DOM node.
-  function h<S, T extends string = string>(
+  function h<S, C = unknown, T extends string = string>(
     // Tags cannot be empty strings.
     tag: T extends "" ? never : T,
-    props: PropList<S>,
+    props: PropList<S, C>,
     children?: MaybeVDOM<S> | readonly MaybeVDOM<S>[]
   ): VDOM<S>
 
@@ -28,15 +28,30 @@ declare module "hyperapp" {
 
   // ---------------------------------------------------------------------------
 
-  // This utility type ensures at least one property in an object is present.
-  // `App` uses this to conveniently mark `app({})` as invalid because that
-  // is basically a fancy no-op.
-  type AtLeastOne<T> = { [K in keyof T]: Pick<T, K> }[keyof T]
-  // Credit: https://stackoverflow.com/a/59987826/1935675
+  // This lets you use a version of `h` which assumes your particular app state.
+  interface TypedH<S, C = unknown> {
+    <_ extends never, T extends string = string>(
+      tag: T extends "" ? never : T,
+      props: PropList<S, C>,
+      children?: MaybeVDOM<S> | readonly MaybeVDOM<S>[]
+    ): VDOM<S>
+  }
+
+  // ---------------------------------------------------------------------------
 
   // This utility type requires every property of an object or none at all.
   // `App` uses this to make sure `view:` always appears alongside `node:`.
   type AllOrNothing<T> = T | { [K in keyof T]?: never }
+
+  // This utility type ensures at least one property in an object to be present.
+  // `App` uses this to conveniently invalidate `app({})` because that is
+  // basically a fancy no-op.
+  type AtLeastOne<T> = { [K in keyof T]: Pick<T, K> }[keyof T]
+  // Credit: https://stackoverflow.com/a/59987826/1935675
+
+  // This utility type validates plain objects while invalidating array objects
+  // and string objects by disallowing numerical indexing. Used by `StyleProp`.
+  type IndexableByKey = Record<number, never>
 
   // ---------------------------------------------------------------------------
 
@@ -48,7 +63,7 @@ declare module "hyperapp" {
       node: Node
     } |
     AtLeastOne<{
-      init: State<S> | StateWithEffects<S> | Action<S>
+      init: Dispatchable<S>
       subscriptions: Subscriptions<S>
       dispatch: DispatchInitializer<S>
     }> &
@@ -59,21 +74,18 @@ declare module "hyperapp" {
   >
 
   // A view builds a virtual DOM node representation of the application state.
-  type View<S> = (state: State<S>) => VDOM<S>
+  type View<S> = (state: S) => VDOM<S>
 
   // The subscriptions function manages a set of subscriptions.
   type Subscriptions<S> = (
-    state: State<S>
+    state: S
   ) => (boolean | undefined | Subscription<S> | Unsubscribe)[]
 
   // A subscription represents subscriber activity.
-  type Subscription<S, D = any> = [Subscriber<S, D>, Payload<D>]
+  type Subscription<S, P = any> = [subscriber: Subscriber<S, P>, payload: P]
 
   // A subscriber reacts to subscription updates.
-  type Subscriber<S, D> = (
-    dispatch: Dispatch<S>,
-    payload?: Payload<D>
-  ) => void | Unsubscribe
+  type Subscriber<S, P> = (dispatch: Dispatch<S>, payload: P) => void | Unsubscribe
 
   // An unsubscribe function cleans up a canceled subscription.
   type Unsubscribe = () => void
@@ -84,60 +96,51 @@ declare module "hyperapp" {
   // ---------------------------------------------------------------------------
 
   // A dispatched action handles an event in the context of the current state.
-  type Dispatch<S> = (action: Action<S>, payload?: Payload<any>) => void
+  type Dispatch<S> = (action: Action<S>, payload?: any) => void
+
+  // A dispatchable entity, when processed, causes a state transition.
+  type Dispatchable<S> = S | StateWithEffects<S> | Action<S>
 
   // An action transforms existing state and/or wraps another action.
   type Action<S, P = any> = ActionTransform<S, P> | ActionWithPayload<S, P>
-  type ActionTransform<S, P = any> = (
-    state: State<S>,
-    payload: Payload<P>
-  ) => State<S> | StateWithEffects<S> | Action<S>
-  type ActionWithPayload<S, P> = [ActionTransform<S, P>, Payload<P>]
+  type ActionTransform<S, P = any> = (state: S, payload: P) => Dispatchable<S>
+  type ActionWithPayload<S, P> = [action: ActionTransform<S, P>, payload: P]
 
   // A transform carries out the transition from one state to another.
   type Transform<S, P = any> = (
-    state: State<S> | StateWithEffects<S>,
-    payload?: Payload<P>
-  ) => State<S> | StateWithEffects<S>
-
-  // Application state is accessible in every view, action, and subscription.
-  type State<S> = S
+    state: S | StateWithEffects<S>,
+    payload: P
+  ) => S | StateWithEffects<S>
 
   // State can be associated with a list of effects to run.
-  type StateWithEffects<S, D = any> = [State<S>, ...Effect<S, D>[]]
-
-  // It is often convenient to abstract away effect preparation.
-  type EffectCreator<S, D = any> = (..._: any[]) => Effect<S, D>
+  type StateWithEffects<S, P = any> = [state: S, ...effects: Effect<S, P>[]]
 
   // An effect is an abstraction over an impure process.
-  type Effect<S, D = any> = [Effecter<S, D>, Payload<D>]
+  type Effect<S, P = any> = [effecter: Effecter<S, P>, payload: P]
 
   // An effecter is where side effects and any additional dispatching may occur.
-  type Effecter<S, D> = (
+  type Effecter<S, P> = (
     dispatch: Dispatch<S>,
-    payload?: Payload<D>
+    payload: P
   ) => void | Promise<void>
-
-  // A payload is data given to an action, effect, or subscription.
-  type Payload<P> = P
 
   // ---------------------------------------------------------------------------
 
   // A virtual DOM node represents an actual DOM element.
-  type VDOM<S> = {
+  type VDOM<S, C = unknown> = {
     readonly type: VDOMNodeType
-    readonly props: PropList<S>
+    readonly props: PropList<S, C>
     readonly children: MaybeVDOM<S>[]
     node: null | undefined | Node
     readonly tag: Tag<S>
     readonly key: Key
-    memo?: PropList<S>
+    memo?: PropList<S, C>
     events?: Record<string, Action<S>>
 
     // `_VDOM` is a guard property which gives us a way to tell `VDOM` objects
-    // apart from `PropList` objects. Since users are not expected to manually
-    // create their own VDOM elements, we can take advantage of this
-    // TypeScript-specific trick without forcing the user to do
+    // apart from `PropList` objects. Since we don't expect users to manually
+    // create their own VNodes, we can take advantage of this trick that's
+    // specific to TypeScript without forcing the user to do
     // anything different.
     _VDOM: true
   }
@@ -160,18 +163,18 @@ declare module "hyperapp" {
   type Key = string | null | undefined
 
   // Virtual DOM properties will often correspond to HTML attributes.
-  type PropList<S> = Readonly<
+  type PropList<S, C> = Readonly<
     ElementCreationOptions &
-      EventActions<S> & {
-        [_: string]: unknown
-        class?: ClassProp
-        key?: Key
-        style?: StyleProp
+    EventActions<S, C> & {
+      [_: string]: unknown
+      class?: ClassProp
+      key?: Key
+      style?: StyleProp
 
-        // By disallowing `_VDOM` we ensure that values having the `VDOM` type are
-        // not mistaken for also having `PropList`.
-        _VDOM?: never
-      }
+      // By disallowing `_VDOM` we ensure that values having the `VDOM` type
+      // are not mistaken for also having `PropList`.
+      _VDOM?: never
+    }
   >
 
   // The `class` property represents an HTML class attribute string.
@@ -191,19 +194,18 @@ declare module "hyperapp" {
   // TypeScript's CSS property definitions. The trade-off doesn't
   // seem worth it given the chances of using such properties.
   // However, you can use type casting if you want to them.
-  type StyleProp = {
+  type StyleProp = IndexableByKey & {
     [K in keyof CSSStyleDeclaration]?: CSSStyleDeclaration[K] | null
-  } & // Since strings are indexable we can avoid them by preventing indexing.
-  { [_: number]: never }
+  }
 
   // Event handlers are implemented using actions.
-  type EventActions<S> = {
-    [K in keyof EventsMap]?: Action<S, EventsMap[K]> | ActionWithPayload<S, any>
+  type EventActions<S, C> = {
+    [K in keyof EventsMap]?: Action<S, EventsMap[K]> | ActionWithPayload<S, C>
   }
-  type EventsMap = {
-    [K in keyof HTMLElementEventMap as `on${K}`]: HTMLElementEventMap[K]
-  } &
-    { [K in keyof WindowEventMap as `on${K}`]: WindowEventMap[K] } & {
-      onsearch: Event
-    }
+
+  // Most event typings are provided by TypeScript itself.
+  type EventsMap
+    = { [K in keyof HTMLElementEventMap as `on${K}`]: HTMLElementEventMap[K] }
+    & { [K in keyof WindowEventMap as `on${K}`]: WindowEventMap[K] }
+    & { onsearch: Event }
 }
--- a/index.d.ts
+++ b/index.d.ts
@@ -102,12 +102,15 @@ declare module "hyperapp" {
   ) => void
 
   // A dispatchable entity, when processed, causes a state transition.
-  type Dispatchable<S> = S | StateWithEffects<S> | Reaction<S>
+  type Dispatchable<S, P = any> =
+    | S
+    | StateWithEffects<S>
+    | Action<S, P>
+    | ActionWithPayload<S, P>
 
   // An action transforms existing state and/or wraps another action.
-  type Reaction<S, P = any> = Action<S, P> | ActionWithPayload<S, P>
   type Action<S, P = any> = (state: S, payload: P) => Dispatchable<S>
-  type ActionWithPayload<S, P> = [action: Action<S, P>, payload: P]
+  type ActionWithPayload<S, P = any> = [action: Action<S, P>, payload: P]
 
   // A transform carries out the transition from one state to another.
   type Transform<S, P = any> = (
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -28,36 +28,64 @@ declare module "hyperapp" {
...
+  type App<S> = Readonly<
+    {
+      view: View<S>
+      node: Node
+    } |
+    AtLeastOne<{
+      init: Dispatchable<S>
+      subscriptions: Subscriptions<S>
+      dispatch: DispatchInitializer<S>
+    }> &
```
jorgebucaran, line 69: Actually, you can create an app without `init`, `subscriptions`, or `dispatch `. Why is this necessary?

jorgebucaran (changes requested):

index.d.ts
icylace: `app({ view, node })` works but `app({})` has no redeeming value.

index.d.ts
```diff
@@ -28,36 +28,64 @@ declare module "hyperapp" {
...
+    AllOrNothing<{
+      view: View<S>
+      node: Node
+    }>
+  >
```
jorgebucaran, line 74: This is interesting because it's true that `view` and `node` must be present for Hyperapp to render your view.

icylace: Yes, and while `app({ node })` causes no runtime error I felt it was pointless and should be treated as invalid.

index.d.ts
```diff
@@ -68,57 +96,52 @@ declare module "hyperapp" {
...
+  type Dispatchable<S> = S | StateWithEffects<S> | Action<S>
 
   // An action transforms existing state and/or wraps another action.
   type Action<S, P = any> = ActionTransform<S, P> | ActionWithPayload<S, P>
-  type ActionTransform<S, P = any> = (
-    state: State<S>,
-    payload: Payload<P>
-  ) => State<S> | StateWithEffects<S> | Action<S>
-  type ActionWithPayload<S, P> = [ActionTransform<S, P>, Payload<P>]
+  type ActionTransform<S, P = any> = (state: S, payload: P) => Dispatchable<S>
```
jorgebucaran, line 106: There's something redundant in ActionsTransform and Dispatchable. `Dispatchable` could very well be an `(state: S, payload: P)` too. I think there's room for simplification here.

jorgebucaran (changes requested):

index.d.ts
jorgebucaran: `app({})` is not an error, so the types shouldn't disallow that.

icylace: The types disallow many things that are not errors. For example, `text()`. Part of the role of types is to eliminate obviously useless code.  What does `app({})` do better than `() => {}` besides waste resources?

Also, consider scenarios where Hyperapp instances are dynamically generated.  If I have `app(generatedAppProps)` somewhere in my setup code I would like to know if there's a regression that's causing `generatedAppProps` to be empty for some reason while editing/compiling.

jorgebucaran: Okay, maybe create an issue (or even better, start a discussion) and let's talk about it? 

Until then, let's not error on `app({})`?

index.d.ts
icylace: Well, `type Dispatchable<S> = S | StateWithEffects<S> | Action<S>`.  So, it can simplify into `(state: S, payload: P) => Dispatchable<S>` but of course there's more going on.

If there's a way to simplify even further that would be great but I don't see it.

index.d.ts
icylace: Sorry, I don't see the problem.  It's already, IMHO, an obscure edge case and I've mentioned two reasons to treat it as bad.

icylace: If someone really really wanted a "no-op" `app()` call then `app({ init: void this })` is at least catchy.

jorgebucaran: I want to defer making this decision after other things have been solved, like making the types more ergonomic and user-friendly. Right now, erroring on `app({})` seems is not a priority and I am not 100% sure I want to have this. If I accept this, I'll be unable to cut a new release until I make that decision.

jorgebucaran: I want to defer this after other things have been worked on, like making the types more ergonomic and user-friendly. Right now, erroring on `app({})` is not a priority and I am not even 100% convinced we want to have this. If I accept this now, I'll be unable to cut a new release until I'm convinced.

jorgebucaran: I want to defer this after other things have been worked on, like making the types more ergonomic and user-friendly. Right now, erroring on `app({})` is not a priority and I am not even 100% convinced we want this. If I accept this now, I'll be unable to cut a new release.

icylace: I hear what you're saying but you're asking me to put in more effort to undo the protections I believe in without giving me an example of where `app({})` is the preferred choice and also without refuting the issues I've brought up about allowing it.

Besides, disallowing `app({})` increases the ergonomics of the embedded Hyperapp instance case because I can be rest assured that the types will let me know if my app configuration generation code is doing nothing.

icylace: Also, it's not like we're talking about a complex situation here. Again, it seems like an open-and-shut case to me.

icylace: And if there truly is somehow a use case for `app({})` then that can be addressed in the future.

jorgebucaran: I want to take the reverse approach. Start simple, and if there's a necessity to forbid `app({})`, do it only then.

icylace: Okay, but I already told you why it's beneficial to forbid it.

icylace: Consider this scenario:

Let's say I'm working on a larger project with a team of developers and we have this relevant code in the project:

```ts
// farawaySetup.ts
const configureProperly = () => ({ init: 42 })
```

```ts
// embedded.ts
import { configureProperly } from "../somewhere/else/farawaySetup.ts"
app(configureProperly())
```

So far, so good.

Now, let's say I pull in a future change from a new team member who committed this:

```ts
// farawaySetup.ts
const configureProperly = () => ({})
```

Obviously, that's contrived but I'm using that for sake of brevity. In a real-word scenario there would of course be more going on that would ultimately result in the same way.

Anyway, without having to check or know to check `farawaySetup.ts` **pre-emptively**, I'll know immediately that something went wrong somewhere during my colleague's commit process. For instance, was the githook that should have been setup to run a type/compile check on code being committed somehow get misconfigured?

This benefit is thanks to `app({})` being disallowed.

However, if `app({})` _were_ allowed then I wouldn't be privy to the issue until testing time or runtime. Therefore wasting time.

index.d.ts
icylace: @jorgebucaran Mind if I resolve this one? I don't think a change can be made here without introducing redundancy.

jorgebucaran: Let's give @zaceno a chance to review the PR too. 👌

zaceno: I see no problem with the definition of `ActionTransform`. 

_Technically_  I think `Dispatch` should probably be:

```ts
type Dispatch<S> = (action: Dispatchable<S>,  payload?: any) => void
```

because it's fully possible to `dispatch({foo: 'bar'})`, or `dispatch([{foo: 'bar'}, http('example.com')])`

But if we want to discourage that useage then the current definition is fine (I don't feel strongly about this – just pointing it out)

zaceno (commented):
Overall I really like this update! 👍 👍  The utility types I could take or leave but the rest is great. I haven't actually tried it out in practice yet but I trust @icylace ;) 

There is one thing I've been thinking about though. And it is the _names_ of `Action`, `ActionTransform` and `ActionWithPayload`.

As I see it, the one that users will most commonly import and use in their own code will be `ActionTransform`.    The others will mainly be used when users implement their own effects & subs I think.

So, compare:

```js
const Add : ActionTransform<State, number> = (state, amount) => ({...
```

vs

```js
const Add: Action<State, number> = (state, amount) => ({...

```

I would prefer the latter. I know - the latter already actually works, probably anyway, but I feel like there is something wrong with defining a const as some type which is actually a union type. 

So I'm thinking what if we could rename:

`ActionTransform` -> `Action`
`Action` -> ??? maybe `Handler`?  or `ActionType`

Or  just get rid of the current `Action` and inline `Action | ActionWithPayload` where it was. 

... or of course we could leave things as they are for now and think more about this. But I wanted to raise the question anyway.

jorgebucaran:
> Or just get rid of the current Action and inline Action | ActionWithPayload where it was.

Works for me! 👍

index.d.ts
icylace: Fascinating, I had not even considered using dispatch like that. I'm not sure if there's a use case for not using an action with dispatch but I'm open to that possibility. So, I wouldn't be opposed to expanding the definition for `Dispatch` in this way.

icylace:
I'd rather keep `Action` or rename it if necessary. I'm using it a lot in a UI library I'm working on. It's used in configuration objects for my components that need to reference actions.

Also, there's another option:

```ts
const Add = (state: State, amount: number): State => ({...
```

index.d.ts
jorgebucaran: Well, the most obvious use case is `init`. It's usually not an action, but just the initial state or a state+effect tuple.

jorgebucaran: The most obvious use case for that would be `init`, usually not an action, but just the state or a state+effect tuple.

icylace: Oh, yeah, haha !

icylace: I updated the PR to use an expanded definition of `Dispatch` similar  to the one suggested by @zaceno.

zaceno:
Yeah if you're using the current `Action` type a lot then let's keep it of course. 

But I do think we should keep considering renaming it to something else to free up the name "Action" to be used in place of "ActionTransform". The more I think about it the more important I think this is.  

At least, when I say "action" I _always_ mean the function. I never think of or explain `[DoFoo, 32]` as a special _kind_ of action. Rather it is a way to customize the payload the actual action gets dispatched with.  Actions are such an important concept I think we need to be very specific how we communicate precisely what an action is and isn't. But also I can see how that question is beyond the scope of this PR. Let's keep it going in Discord or Github Discussions.

jorgebucaran:
>  I do think we should keep considering renaming it to something else to free up the name "Action" to be used in place of "ActionTransform". 

Makes sense to me. 👍

icylace:
Just renamed `Action` to `Reaction` and `ActionTransform` to `Action`.

index.d.ts
```diff
@@ -138,7 +138,7 @@ declare module "hyperapp" {
     readonly tag: Tag<S>
     readonly key: Key
     memo?: PropList<S, C>
-    events?: Record<string, Action<S>>
+    events?: Record<string, Reaction<S>>
```
jorgebucaran, line 141: 😅

index.d.ts
```diff
@@ -68,57 +96,55 @@ declare module "hyperapp" {
...
+  type Dispatchable<S> = S | StateWithEffects<S> | Reaction<S>
 
   // An action transforms existing state and/or wraps another action.
-  type Action<S, P = any> = ActionTransform<S, P> | ActionWithPayload<S, P>
-  type ActionTransform<S, P = any> = (
-    state: State<S>,
-    payload: Payload<P>
-  ) => State<S> | StateWithEffects<S> | Action<S>
-  type ActionWithPayload<S, P> = [ActionTransform<S, P>, Payload<P>]
+  type Reaction<S, P = any> = Action<S, P> | ActionWithPayload<S, P>
```
jorgebucaran, line 108: Can just inline this type? `Reaction` feels out of place, as it is not a formal (or even informal) Hyperapp thing.

jorgebucaran (changes requested):

index.d.ts
icylace: As per the discussion on Discord, yes `Reaction` will get removed.

index.d.ts
icylace: `Reaction` has now been inlined and removed.

jorgebucaran (approved):

icylace:
Latest change removes `Transform` because it ultimately isn't essential and is arguably too specialized.

index.d.ts
```diff
@@ -191,19 +194,18 @@ declare module "hyperapp" {
...
     [K in keyof CSSStyleDeclaration]?: CSSStyleDeclaration[K] | null
-  } & // Since strings are indexable we can avoid them by preventing indexing.
-  { [_: number]: never }
+  }
 
   // Event handlers are implemented using actions.
-  type EventActions<S> = {
-    [K in keyof EventsMap]?: Action<S, EventsMap[K]> | ActionWithPayload<S, any>
+  type EventActions<S, C> = {
+    [K in keyof EventsMap]?: Action<S, EventsMap[K]> | ActionWithPayload<S, C>
```
zaceno: Are we sure this will work for more events on a single node? I mean for something like this (maybe contrived but still, the principle ;) ):

```ts
const AddFoo:Action<State, number> = (state, amount) => ({...state, foo: state.foo + amount})
const SetName:Action<State, string> = (state, name) => ({...state, name})
//...
h('div', {
   onmousedown: [AddFoo, 32],
   onmouseup: [SetName, 'unknown']
})
```

I don't see how `C` could be properly inferred in that case.

For the record, the way that I (in my own types-attempt) handled making sure that payloads in ActionWithPayloads matches their action, is with a "validator type" that I intersect with in the `h` call. Also not 100% perfect but might be interesting to consider: https://github.com/zaceno/zx-typed-todomvc-hyperapp/blob/ed64b9d2ea5e61a2c7b1677b1639f0914fd73509/typings/hyperapp.d.ts#L64

icylace: Ah, of course, haha. Seems so obvious in hindsight. Yeah this case fails currently.

icylace: If I understand your types correctly, you get this to not error out by basically not checking the payload: https://github.com/zaceno/zx-typed-todomvc-hyperapp/blob/ed64b9d2ea5e61a2c7b1677b1639f0914fd73509/typings/hyperapp.d.ts#L89

That was my original answer, too. If worse comes to worst then we'll go with that.

However, I'll try a little more to see if there's a more type-safe way to do this that works in the general case.

zaceno: I think you misread that. Notice I'm using `extends` and `infer` keywords! I'm saying "if `X` looks like an ActionWithPayload (while inferring what `S` and `P` actually should be) then X must be assignable to `readonly [Action<S, P>, P]`.

And I do that process for each prop individually. So the problem with this approach is not that it allows too much. On the contrary, the problem is there are edge cases that should be allowed which are not. Like for instance:

```ts

h('div', {
  notAnEventProp: [AddFoo, 'what?! a string?!']
  })
  ```
  
  so my approach would forbid that in the same way it would forbid it if we were assigning the action-with-payload to `onclick`. But since `notAnEventProp` is ... well, not an event prop, it should be allowed to be anything it wants. 
  
  This is such a weird edge case I don't consider it a huge problem. And besides this sort of case my approach works well.

Comment (zaceno) [reply index.d.ts]: And now that I think about it, I could probably amend this problem too. Just not worth it imo.

Comment (zaceno) [reply index.d.ts]: If there is another solution that solves type checking actions with custom payloads it would blow my mind. I think I spent over a month trying to work out this solution so I'm actually kind of proud of it 😅

Comment (zaceno) [reply index.d.ts]: And I had a lot of help from the kind folks on the typescript gitter as well (credit where credit is due)

Comment (icylace) [reply index.d.ts]: @zaceno You're right, I read too quickly.  I'm exploring more on this with ideas borrowed from your attempt.

**Comment by icylace** [comment]
Just updated the PR with the following changes:

- Added a variation of @zaceno's `ValidateCustomPayloads` type to help solve custom payloads.
  - Optimized away some type variables accordingly.
- Added some more commentary and reformatting.

File: index.d.ts (hunk 6 above)
Comment (icylace) [reply index.d.ts]: Just updated the PR with a hopefully complete solution.

**Comment by jorgebucaran** [comment]
Do you think you can make a CodeSandbox example or share a snippet to can see how you're using these types?

**Comment by icylace** [comment]
> Do you think you can make a CodeSandbox example or share a snippet to can see how you're using these types?

Yeah, I'll be doing something along those lines to show how to use these types.

Up until now I've been relying on my test suite to work them out but that code is not necessarily approachable.

File: index.d.ts [hunk 7]
```diff
@@ -68,41 +104,27 @@ declare module "hyperapp" {
...
 
   // An effecter is where side effects and any additional dispatching may occur.
-  type Effecter<S, D> = (
-    dispatch: Dispatch<S>,
-    payload?: Payload<D>
-  ) => void | Promise<void>
-
-  // A payload is data given to an action, effect, or subscription.
-  type Payload<P> = P
+  type Effecter<S, P> = (dispatch: Dispatch<S>, payload: P) => void | Promise<void>
```
jorgebucaran, line 127: Why is this `void | Promise<void>`? Hyperapp doesn't know anything about promises.

icylace: Because it's reasonable to let effecters use async code.  For example, usage of the browser's `fetch()` API.

jorgebucaran: I see. You may want to declare the function as `async` in order to use `await` in it, causing the function to return a Promise implicitly.

icylace: Yep, exactly !

jorgebucaran, line 127: Let's remove the `| Promise<void>` for now. I don't know if I'll want to support this right now.

jorgebucaran (changes requested):

index.d.ts
icylace: Without `| Promise<void>` using stuff `fetch()` is impossible.  I'm using async effects to implement an app-specific API I have.

icylace: How would you do async code in a Hyperapp app?  Delegate that stuff to an external layer of custom code outside of the  Hyperapp app ?

jorgebucaran: You can use `fetch()`, what you can't do is declare the function as `async`, that's all.

jorgebucaran: You can use `fetch()`, what you won't be able to do is declare the function as `async`, that's it.

icylace: Umm... use `fetch()` _where exactly_?  If I press a button in my app and that triggers a fetch to some resource, what does that code look like if effecters cannot return `Promise<void>`?

jorgebucaran: A function doesn't need to be declared as `async` for you to use `fetch()`. You just use `fetch().then(...)`. 

We could definitely allow `async` effect implementations in the future, but don't have to right now.

icylace: Let's say for the sake of argument what you say is correct, that using `fetch().then(...)` is an alternative.

Well, then you're basically dictating to the user how they should do their async coding which I believe is outside of Hyperapp's purview.

And for what gain, exactly ?

icylace: If effecters and subscribers are sanctioned areas where the user can do their messy imperative stuff, then let them do it the way they want.

jorgebucaran: That's a great point, which is exactly why I think we can consider this again in the future.

icylace: Could you clarify? Which one to consider in the future?

jorgebucaran: Allowing `async`-`await` effect implementations.

icylace: The types already allow that.  What is the benefit to removing this now when it's not known to cause any issues while allowing a modern style of async coding ?

zaceno: I think Promise<void> needs to stay on the basis that anything that is possible in plain JS should be possible in TS (at least if it adds some value, however slight). 

Due to the interconnected nature of the types, it won’t be possible to add your own version of Effect that supports this, without redefining all the types. So I guess technically it will be possible but in a quite obnoxious way.

I also don’t prefer using async/await over raw Promises, but it is popular (especially among the TS crowd it seems). It won’t be long before people beg to have it put in (or are alienated for the lack)

jorgebucaran: > I think Promise needs to stay on the basis that anything that is possible in plain JS should be possible in TS (at least if it adds some value, however slight).

Hmm, I see what you mean. What's already possible in JS should not be made impossible in TS. Got it. 

In this case, the types just present the world the way it is. Hyperapp is not necessarily advocating for `async` effects, but no one can stop you from using `async`/`await` if that's what you want.

zaceno: > the types just present the world the way it is
☝️ exactly!

zaceno: Actually... if your issue is with using the word `Promise` in our type defs @jorgebucaran, we could just have:

```js
type Effecter<S, P> = (dispatch: Dispatch<S>, payload: P) => any
```

... and in fact maybe we _should_ have that. Because hyperapp doesn't care what an effecter returns, right?

jorgebucaran: Makes sense too! More, actually.

icylace: Interesting, so Hyperapp simply does nothing with the return type, right ?

icylace: Same goes for subscribers I'd imagine ?

zaceno: Subscribers are supposed to return unsubscribers. But yeah, Hyperapp ignores the returns value of unsubscribers.

icylace: Oh, right, duh !

icylace: Carrying on this discussion here: https://github.com/jorgebucaran/hyperapp/discussions/1052

This particular topic is at the top of the list.

PR #1049, opened 2021-04-08
Add the new `TypedH<S>` type.

icylace:
Introducing the `TypedH<S>` type.  It lets you define a version of `h` that has your app's state type "baked in" so you don't need to worry about whether or not you need to explicitly assign the type variable for `h` every time you call it.

```ts
import { h as ha } from "hyperapp"

type YourAppState = { youGetTheIdea: boolean }

const h: TypedH<YourAppState> = ha

// ... continue using `h` like you normally would.
```

Addresses https://github.com/jorgebucaran/hyperapp/issues/1048.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -28,6 +28,17 @@ declare module "hyperapp" {
 
   // ---------------------------------------------------------------------------
 
+  // This lets you use a version of `h` which assumes your particular app state.
+  interface TypedH<S> {
+    <_, T extends string = string>(
+      tag: T extends "" ? never : T,
+      props: PropList<S>,
+      children?: MaybeVDOM<S> | readonly MaybeVDOM<S>[]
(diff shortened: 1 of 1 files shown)


jorgebucaran:
This doesn't seem very user-friendly to me. What do you think?

icylace:
@jorgebucaran I modeled `TypedH` after `TypedUseSelectorHook` from `react-redux`. To paraphrase https://react-redux.js.org/using-react-redux/usage-with-typescript#define-typed-hooks:

```ts
// Use throughout your app instead of plain ... `useSelector`
...
export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector
```

icylace:
Closed in favor of https://github.com/jorgebucaran/hyperapp/pull/1047.

PR #1059, opened 2021-04-27
Proofreading

skanne:
1) Actual proofreading (spelling, grammar, punctuation, typography, style, accessibility).

2) Formalizing the introduction to the conceptual topics (especially the API functions): 
  - *Definition*: A concise "one-liner" defining the discussed concept or API function.
  - *Signature*: An Elm-ish (or FP-style in general, mixed with JS-syntax) type annotation outlining the "pattern" of input and output parameters (of the discussed function or tuple). I ~~believe~~ _know_ that these annotations help grasp the concepts and input/output types faster, even without them being exact science. We've successfully been using such annotations in our dev projects.
  - *Import & Usage*: A (not really usable) code sample at the beginning of the page showing – with all due brevity – how to import and call API functions.
  - *Parameter Overview*: Tables that list not only the input parameters, their types and whether required, but also the return parameter and its type.
  - *Naming Recommendation*: Suggestions how to name actions, effects, subscriptions and components are given to help organize and recognize code better.

3) Corrections and enhancements of code samples. Some code samples have gained signature comments (see above), too, where applicable/beneficial.

4) Removed the paragraph mentioning `@hyperapp/` packages from `subscriptions.md` as they were removed (=deprecated) by Jorge.

5) Moved “Effects” into their own markdown file, separating them from “Actions”, and thus treating them like an architectural pillar in their own right.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,10 +3,10 @@
 > The tiny framework for building hypertext applications.
 
 - **Do more with less**—We have minimized the concepts you need to learn to get stuff done. Views, actions, effects, and subscriptions are all pretty easy to get to grips with and work together seamlessly.
-- **Write what, not how**—With a declarative API that's easy to read and fun to write, Hyperapp is the best way to build purely functional, feature-rich, browser-based apps using idiomatic JavaScript.
+- **Write what, not how**—With a declarative API that’s easy to read and fun to write, Hyperapp is the best way to build purely functional, feature-rich, browser-based apps using idiomatic JavaScript.
 - **Smaller than a favicon**—1 kB, give or take. Hyperapp is an ultra-lightweight Virtual DOM, [highly-optimized diff algorithm](https://javascript.plainenglish.io/javascript-frameworks-performance-comparison-2020-cd881ac21fce), and state management library obsessed with minimalism.
 
-Here's the first example to get you started. [Try it here](https://codepen.io/jorgebucaran/pen/zNxZLP?editors=1000)—no build step required!
+Here’s the first example to get you started. [Try it here](https://codepen.io/jorgebucaran/pen/zNxZLP?editors=1000)—no build step required!
 
 <!-- prettier-ignore -->
 ```html
@@ -46,7 +46,7 @@ Here's the first example to get you started. [Try it here](https://codepen.io/jo
 
 The app starts by setting the initial state and rendering the view on the page. User input flows into actions, whose function is to update the state, causing Hyperapp to re-render the view.
 
-When describing how a page looks in Hyperapp, we don't write markup. Instead, we use `h()` and `text()` to create a lightweight representation of the DOM (or virtual DOM for short), and Hyperapp takes care of updating the real DOM efficiently.
+When describing how a page looks in Hyperapp, we don’t write markup. Instead, we use `h()` and `text()` to create a lightweight representation of the DOM (or virtual DOM for short), and Hyperapp takes care of updating the real DOM efficiently.
 
 ## Installation
 
@@ -60,9 +60,9 @@ Ready to dive in? Learn the basics in the [Tutorial](docs/tutorial.md), check ou
 
 To access [Web Platform APIs](https://platform.html5.org) (like `fetch` or `addEventListener`) in a way that makes sense for Hyperapp, learn [how to create your own effects and subscriptions](/docs/architecture/subscriptions.md). For everything else, from third-party packages to real-world examples, browse the [Hyperawesome](https://github.com/jorgebucaran/hyperawesome) collection.
 
-## Help, I'm stuck!
+## Help, I’m stuck!
 
-If you've hit a stumbling block, hop on our [Discord](https://discord.gg/5CtfCYEq8V) server to get help, and if you remain stuck, [please file an issue](https://github.com/jorgebucaran/hyperapp/issues/new), and we'll help you figure it out.
+If you’ve hit a stumbling block, hop on our [Discord](https://discord.gg/5CtfCYEq8V) server to get help, and if you remain stuck, [please file an issue](https://github.com/jorgebucaran/hyperapp/issues/new), and we’ll help you figure it out.
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -3,10 +3,10 @@
 > The tiny framework for building hypertext applications.
 
 - **Do more with less**—We have minimized the concepts you need to learn to get stuff done. Views, actions, effects, and subscriptions are all pretty easy to get to grips with and work together seamlessly.
-- **Write what, not how**—With a declarative API that’s easy to read and fun to write, Hyperapp is the best way to build purely functional, feature-rich, browser-based apps using idiomatic JavaScript.
+- **Write what, not how**—With a declarative API that's easy to read and fun to write, Hyperapp is the best way to build purely functional, feature-rich, browser-based apps using idiomatic JavaScript.
 - **Smaller than a favicon**—1 kB, give or take. Hyperapp is an ultra-lightweight Virtual DOM, [highly-optimized diff algorithm](https://javascript.plainenglish.io/javascript-frameworks-performance-comparison-2020-cd881ac21fce), and state management library obsessed with minimalism.
 
-Here’s the first example to get you started. [Try it here](https://codepen.io/jorgebucaran/pen/zNxZLP?editors=1000)—no build step required!
+Here's the first example to get you started. [Try it here](https://codepen.io/jorgebucaran/pen/zNxZLP?editors=1000)—no build step required!
 
 <!-- prettier-ignore -->
 ```html
@@ -46,7 +46,7 @@ Here’s the first example to get you started. [Try it here](https://codepen.io/
 
 The app starts by setting the initial state and rendering the view on the page. User input flows into actions, whose function is to update the state, causing Hyperapp to re-render the view.
 
-When describing how a page looks in Hyperapp, we don’t write markup. Instead, we use `h()` and `text()` to create a lightweight representation of the DOM (or virtual DOM for short), and Hyperapp takes care of updating the real DOM efficiently.
+When describing how a page looks in Hyperapp, we don't write markup. Instead, we use `h()` and `text()` to create a lightweight representation of the DOM (or virtual DOM for short), and Hyperapp takes care of updating the real DOM efficiently.
 
 ## Installation
 
@@ -60,9 +60,9 @@ Ready to dive in? Learn the basics in the [Tutorial](docs/tutorial.md), check ou
 
 To access [Web Platform APIs](https://platform.html5.org) (like `fetch` or `addEventListener`) in a way that makes sense for Hyperapp, learn [how to create your own effects and subscriptions](/docs/architecture/subscriptions.md). For everything else, from third-party packages to real-world examples, browse the [Hyperawesome](https://github.com/jorgebucaran/hyperawesome) collection.
 
-## Help, I’m stuck!
+## Help, I'm stuck!
 
-If you’ve hit a stumbling block, hop on our [Discord](https://discord.gg/5CtfCYEq8V) server to get help, and if you remain stuck, [please file an issue](https://github.com/jorgebucaran/hyperapp/issues/new), and we’ll help you figure it out.
+If you've hit a stumbling block, hop on our [Discord](https://discord.gg/5CtfCYEq8V) server to get help, and if you remain stuck, [please file an issue](https://github.com/jorgebucaran/hyperapp/issues/new), and we'll help you figure it out.
 
 ## Contributing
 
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -3,10 +3,10 @@
 > The tiny framework for building hypertext applications.
 
 - **Do more with less**—We have minimized the concepts you need to learn to get stuff done. Views, actions, effects, and subscriptions are all pretty easy to get to grips with and work together seamlessly.
-- **Write what, not how**—With a declarative API that's easy to read and fun to write, Hyperapp is the best way to build purely functional, feature-rich, browser-based apps using idiomatic JavaScript.
+- **Write what, not how**—With a declarative API that’s easy to read and fun to write, Hyperapp is the best way to build purely functional, feature-rich, browser-based apps using idiomatic JavaScript.
```
jorgebucaran, line 6: Whoopsie: `’`. Can we go back to my good old `'`?

README.md
```diff
@@ -3,10 +3,10 @@
...
 - **Smaller than a favicon**—1 kB, give or take. Hyperapp is an ultra-lightweight Virtual DOM, [highly-optimized diff algorithm](https://javascript.plainenglish.io/javascript-frameworks-performance-comparison-2020-cd881ac21fce), and state management library obsessed with minimalism.
 
-Here's the first example to get you started. [Try it here](https://codepen.io/jorgebucaran/pen/zNxZLP?editors=1000)—no build step required!
+Here’s the first example to get you started. [Try it here](https://codepen.io/jorgebucaran/pen/zNxZLP?editors=1000)—no build step required!
```
jorgebucaran, line 9: Same here and everywhere! 🙏

docs/api/app.md
```diff
@@ -22,13 +38,13 @@ app(props)
...
-| [init](#init)                   | [State](../architecture/state.md) or [Action](../architecture/actions.md) | :100:
-| [view](#view)                   | [View](../architecture/views.md)                                          | :100:
-| [node](#node)                   | DOM element                                                               | :100:
-| [subscriptions](#subscriptions) | function                                                                  | no
-| [dispatch](#dispatch)           | [Dispatch](../architecture/dispatch.md)                                   | no
+| Prop                            | Type                                                                      | Required? |
+| ------------------------------- | ------------------------------------------------------------------------- | --------- |
+| [init](#init)                   | [State](../architecture/state.md) or [Action](../architecture/actions.md) | yes       |
+| [view](#view)                   | [View](../architecture/views.md)                                          | yes       |
```
jorgebucaran, line 44: I'd like to keep the 💯.

docs/api/app.md
```diff
@@ -147,18 +166,19 @@ A dispatch initializer that can create a [custom dispatch function](../architect
 ```js
 app({
   init: { message: "Hello, World!" },
-  view: (state) => h("p", {}, text(state.message)),
+  view: (state) => h("body", {}, h("p", {}, text(state.message))),
```
jorgebucaran, line 169: Let's use `main` instead of `body`.

docs/api/app.md
```diff
@@ -147,18 +166,19 @@ A dispatch initializer that can create a [custom dispatch function](../architect
...
+
+<!-- A “Hello, World!” program is traditionally the first program you would write when learning a new programming language. -->
 
 ### Full Usage
 
 ```js
 app({
   init: { message: "Hello, World!" },
-  view: (state) => h("p", {}, text(state.message)),
+  view: (state) => h("body", {}, h("p", {}, text(state.message))),
```
jorgebucaran, line 181: Same here and everywhere else. 🙏

skanne: I don't quite get your comment. Did I miss something (and need to correct) or do you like my change?

README.md
skanne: If you wish... 🧞‍♂️

docs/api/app.md
skanne: Not really! It's not accessible to just have an emoji. Can we have at least "Yes 💯"?

docs/api/app.md
skanne: with `node: document.querySelector('main')`

docs/api/app.md
```diff
@@ -1,6 +1,16 @@
...
 
-A function that initializes and mounts a Hyperapp application.
+**_Definition:_**
+
+> A function that initializes and mounts a Hyperapp application.
+
+**_Signature:_**
+
+```elm
+app : ({ Init, View, Node, Subscriptions?, Dispatch? }) -> DispatchFn
```
jorgebucaran, line 10: Do these match the type definitions? We should be careful not to duplicate work.

skanne: Not (yet). At the time of writing I clearly wasn't aware of the final (?) names for types. Besides that, the intention of these type annotations/definitions is to quickly give the reader an overview of what goes in and what comes out, without paying too much intention to the _exact math and inner workings and dependencies_. 

So, if you insist on having the _exact same_ names as in the type definitions (https://github.com/jorgebucaran/hyperapp/blob/main/index.d.ts) then I suggest to wait, till _those_ are finalized. Then we can retrofit them into the reference. But again: please not _too_ specific like in the index.d.ts, otherwise the reader will tend to ignore the annotation or be more confused than informed.

Or, we just keep them _light_ and just try to match them to the names used in the text.

In either case, we should be consistent. And, I would like to keep them (in one form or the other). I can be as insistive as you with :100: ;-).

jorgebucaran: I don't know if they should be exactly the same, maybe a simplified, less accurate, subset of them would do, but we definitely don't want to create _new_ definitions.

jorgebucaran: I don't know if they should be exactly the same, maybe a simplified, less accurate, subset of them would do, but we don't want to create _new_ definitions.

Basically, this means, that we are allowed to annotate only what has already been annotated using the names and symbols in the type definitions, but cannot add something new.

jorgebucaran: I don't know if they should be exactly the same, maybe a simplified, less accurate, subset of them would do, but we don't want to create _new_ definitions.

Basically, this means that we are allowed to include only what has already been annotated, using the names and symbols in the type definitions, but not new stuff.

skanne:
I was too eager with replacing typographic quotation marks. Hence the new commit.

jorgebucaran:
@icylace Could you have a look?

docs/api/app.md
icylace: Paraphrasing the type definitions where convenient is probably the way to go. While I did try keeping the type definitions readable as much as possible, they are first and foremost geared towards precision.

icylace:
@jorgebucaran @skanne I should have time this weekend to review all the changes. Looking forward to it !

icylace (changes requested):
@skanne Solid work overall! In addition to my review of specific files, here are my...

General Impressions:

- Showing simplified versions of type definitions seems helpful for setting the reader's expectations.
- Explicitly stating return values and return types makes a lot of sense. I'm mildly embarrassed I neglected that originally 😅 
- I like the type definition comments that precede some code.
- The docs I wrote had more typos than I thought !

General Critiques & Suggestions:

- I think "Import & Usage" should appear before "Signature". Importation is the first thing that needs to happen before the user would be concerned with how to use what they're importing. Also, the switch would visually group function signatures with their parameter info which is a natural pairing. Thematically, the progression becomes "Here's an important function I'm importing and this is how I need to write it and here's what each part I'm writing is all about."

- The parameters table having text content gets across the notion of an overview. I would rename **_Parameter Overview:_** to just **_Parameters:_**. Even better, I would consider combining the signature section with the parameters section and call it **_Signature & Parameters:_**.

- Speaking of the parameters sections, a couple things:
  - Rename "Input Parameter"/"Input Parameters" back to "Parameter". Parameters are always inputs, at least in JavaScript/TypeScript.
  - Rename "Output Parameter" to "Return Value". Output parameters and return values are different things. See: https://en.wikipedia.org/wiki/Parameter_(computer_programming)#Output_parameters). Also, the way people refer to output parameters reflects this. For instance: https://stackoverflow.com/questions/3175687/how-best-to-implement-out-params-in-javascript

- In several code examples, the trailing comma of the last item of multi-line arrays were removed. Why? We're not talking about tuples, which means the possibility of appending more items exists which is reasonable as projects change over time. Keeping the trailing comma lets diffs be cleaner and also makes item rearrangement less of a hassle. I know these code examples are simple and are focused on Hyperapp usage and not necessarily JavaScript usage but it's good to showcase code that we ourselves would be more comfortable maintaining. See also the first paragraph of: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas

- Additionally, a few nitpicks:
  - In several code examples some lines have trailing spaces. Please remove them for sake of tidiness.
    - Specific lines:
      - app.md: 133, 135
      - h.md: 116, 117, 190, 214
      - subscriptions.md: 121, 163, 178, 201
  - There are a few unnecessary blank lines I'd remove too.
    - Specific lines:
      - effects.md: 139, 140
      - subscriptions.md: 87, 88
      - views.md: 106, 107

docs/architecture/subscriptions.md
```diff
@@ -1,8 +1,20 @@
...
+
+**_Signature:_**
+
+```elm
+Subscription : [SubscriberFn, Payload?]
+```
+
+**_Naming Recommendation:_**
+
+Subscriptions are recommended to be named in `camelCase` prefixed by `on`, for instance `onEvery` or `onMouseEnter` in order to reflect their (app-external) event handling character.
```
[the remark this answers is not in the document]
jorgebucaran: Actually, I'd remove this entire section. IMO we're not ready to prescribe naming conventions yet. Maybe someday.

docs/reference.md
```diff
@@ -11,19 +11,20 @@
...
 - [State](architecture/state.md) represents your application's data.
 - [Views](architecture/views.md) represent your state visually.
-- [Actions](architecture/actions.md) cause state transitions and run effects.
+- [Actions](architecture/actions.md) cause state transitions and trigger effects.
+- [Effects](architecture/effects.md) are triggered by actions to interact with external processes.
 - [Subscriptions](architecture/subscriptions.md) dispatch actions in response to external events.
 - [Dispatch](architecture/dispatch.md) controls action dispatching.
 
-<h2 title="“I’d like to think of browsing the glossary as flipping pages in a book. I can go anywhere instantly and learn whatever suits my fancy.” ―@icylace">Glossary</h2>
+<h2 title="“I'd like to think of browsing the glossary as flipping pages in a book. I can go anywhere instantly and learn whatever suits my fancy.” ―@icylace">Glossary</h2>
```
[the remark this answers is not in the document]
jorgebucaran: Please just remove the stylized double quotes.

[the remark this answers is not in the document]
icylace: Fair enough. I don't feel too strongly about this.

docs/architecture/subscriptions.md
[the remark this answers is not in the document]
skanne: What's wrong with _suggesting_ a name pattern? We are certainly not _prescribing_ them. It does help the developer to keep all those functions apart by simply looking at how they are named. When I scroll up and down hundreds of lines of code (especially someone else's) I personally am glad when I can quickly make out the purpose and usage of a function by looking at both the name and the signature.

skanne:
I've incorporated the changes @icylace has pointed out.

docs/architecture/subscriptions.md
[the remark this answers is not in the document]
jorgebucaran: > What's wrong with suggesting a name pattern?

Nothing, I'm just not ready to make those suggestions.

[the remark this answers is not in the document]
jorgebucaran: > What's wrong with suggesting a name pattern?

Nothing, in fact, it's a great idea, and I'd like to do it eventually, but I'm just not ready to make those suggestions right now. Maybe after we build a big app or two.

docs/reference.md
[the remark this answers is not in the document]
skanne: Turns out that with embedded HTML in markdown you can't have double quotes for the attribute value *and* for the quote inside the attribute value.

title="\"...quote...\"" doesn't work

docs/architecture/subscriptions.md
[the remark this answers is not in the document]
skanne: I've built a big app with hyperapp (and we're going live with it this summer). And I can tell you that _this_ naming convention hints best at what subscriptions are about: reacting _on_ events from outside.

docs/reference.md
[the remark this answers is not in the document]
skanne: I've readded typographic quotation marks (and the typographic apostrophe for consistency). They appear only as tooltip text anyhow. So IMO no big deal.

jorgebucaran (approved):

PR #1063, opened 2021-05-09
Fix tutorial code

Seasonley:


zaceno:
Good catch @Seasonley ! This is indeed a typo in the tutorial and I endorse this change.

jorgebucaran:
Thanks, @Seasonley!

PR #1075, opened 2021-10-25
Allow bare effecter as effect

zaceno:
The idea behind this PR is to create some more symmetry in the API. 

An action can be dispatched on it's own (a bare function):

```js
onclick: SomeAction
```

or with a custom payload:

```js
onclick: [SomeAction, {foo: 'bar'}]
```

But when returning effects from actions, they need to be given as tuples (in square brackets), regardless if the effecter function needs a payload or not. I e:

```js
const doSomethingFx = () =>  {alert('I did something')}

const JustDoSomething = (state) => [state, [doSomethingFx]]
```

This PR would allow giving just the bare effecter when it does not require any parameters:

```js
const JustDoSomething = (state) => [state, doSomethingFx]
```

I admit this is not a common need and thus not a huge deal. But the change is small, and make the api more flexible and symmetrical (imo) without violating any principles or any breaking changes.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -120,16 +120,16 @@ declare module "hyperapp" {
     | Action<S, P>
     | readonly [action: Action<S, P>, payload: P]
 
-  // An Effecter is the function that runs an effect
+  // An effecter is the function that runs an effect.
   type Effecter<S, P = any> = (
     dispatch: Dispatch<S>,
-    payload?: P
+    payload: P
   ) => void | Promise<void>
 
   // An effect is where side effects and any additional dispatching may occur.
   type Effect<S, P = any> =
     | Effecter<S, P>
-    | readonly [effecter: Effecter<S, P>, payload?: P]
+    | readonly [effecter: Effecter<S, P>, payload: P]
 
   
   // Effects can be declared conditionally.
--- a/index.d.ts
+++ b/index.d.ts
@@ -120,11 +120,17 @@ declare module "hyperapp" {
     | Action<S, P>
     | readonly [action: Action<S, P>, payload: P]
 
+  // An Effecter is the function that runs an effect
+  type Effecter<S, P = any> = (
+    dispatch: Dispatch<S>,
+    payload?: P
+  ) => void | Promise<void>
+
   // An effect is where side effects and any additional dispatching may occur.
-  type Effect<S, P = any> = readonly [
-    effecter: (dispatch: Dispatch<S>, payload: P) => void | Promise<void>,
-    payload: P
-  ]
+  type Effect<S, P = any> =
+    | Effecter<S, P>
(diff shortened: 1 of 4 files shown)


index.d.ts
```diff
@@ -120,11 +120,17 @@ declare module "hyperapp" {
     | Action<S, P>
     | readonly [action: Action<S, P>, payload: P]
 
+  // An Effecter is the function that runs an effect
+  type Effecter<S, P = any> = (
+    dispatch: Dispatch<S>,
+    payload?: P
```
icylace, line 126: `payload?: P` doesn't pass my type tests but changing it back to `payload: P` will.

index.d.ts
```diff
@@ -120,11 +120,17 @@ declare module "hyperapp" {
...
+  ) => void | Promise<void>
+
   // An effect is where side effects and any additional dispatching may occur.
-  type Effect<S, P = any> = readonly [
-    effecter: (dispatch: Dispatch<S>, payload: P) => void | Promise<void>,
-    payload: P
-  ]
+  type Effect<S, P = any> =
+    | Effecter<S, P>
+    | readonly [effecter: Effecter<S, P>, payload?: P]
```
icylace, line 132: `payload?: P` doesn't pass my type tests but changing it back to `payload: P` will.

index.d.ts
```diff
@@ -120,11 +120,17 @@ declare module "hyperapp" {
     | Action<S, P>
     | readonly [action: Action<S, P>, payload: P]
 
+  // An Effecter is the function that runs an effect
```
icylace, line 123: Please change this line to: `// An effecter is the function that runs an effect.`

😅

icylace (changes requested):
Only a couple tweaks need to happen and then I think we would be good to go !

zaceno:
Thanks for the feedback @icylace ! Should be good now

icylace (approved):
I tested the corrections that were made and things look good !

jorgebucaran (approved):

index.js
```diff
@@ -409,7 +409,7 @@ export var app = ({
           : action
               .slice(1)
               .map(
-                (fx) => fx && fx !== true && fx[0](dispatch, fx[1]),
+                (fx) => fx && fx !== true && (fx[0] || fx)(dispatch, fx[1]),
```
jorgebucaran, line 412: 👍 `fx` can't be `false` when we check `fx[0] || fx`.

jorgebucaran: Once we're in `fx[0] || fx`, `fx` can't be `false`, yeah?

jorgebucaran:
Thanks, @zaceno! Sorry for the long delay here, `2.0.21` is out with this change. 💯

PR #1076, opened 2021-11-27
Fix create node options when is undefined

zaceno:
When using custom elements that define an `is` property, e.g. `h('custom-foo', {is: 'button'})`, then hyperapp will internally create it like this:

```js
document.createElement('button', {is: 'custom-button'})
```

_But_ whenever we don't use the `is` prop (as is generally the case), the call becomes

```js
document.createElement('button', {is: undefined})
```

Which is not correct as far as I can tell. When there is no `is`, there should be no second argument at all. This PR fixes the behavior so the call becomes:

```js
document.createElement('button', undefined)
```

Although browsers seem to handle the current behavior just as well, the current behavior becomes a problem with some custom element libraries that wrap the native `document.createElement` (don't ask me why... – I guess because of legacy cross-browser concerns?)

For example consider [this example][example] of using a-frame. It imports my modified hyperapp, but notice how it breaks when you import the latest published version of hyperapp from cdn.

[example]: https://flems.io/#0=N4Igxg9gdgzhA2BTEAucD4EMAONEBMQAaEAMwEskZUBtUKTAW2TQDoALAF0fmPSk6IBqEAB585AG4ACcvgC8AHRA5sygHyKoogPQTJ6vniRhO5aNTQBGFADYQAXyL0mLEOwCe2RACdVrACtqEkgBIU4RSUwfaQBlWIAlAH0AOQB5ABEAUWl5aSstKJiAFSyADWLUzJy8gGZC6OksgFkABWKATSS0gCEAKVzpYAcGmJb2roBBBITBmgBdUbiANQBxVNjB5S5ObBQdHQB3Y9ZD2tYIHwBzHQAmAAZHnRhJK+UtJblBgApMAEpcuppJglowcINxp0ktMEqwwWooEVZDBJj4-B5Bqj0axyCi0ZgPEshABHACuiHJuS00mknC8iAgpGkPkQZMQME4kyg5DBZmgADE-MxpABCeR5ZSkqD4RAUKAEd5QGk0gD8zNZ5I5XJ5mD5UEFrmpypQ0jwnGKPIZpM4H0RjTALN1iAAwlgYDAfhAAEYBAHyIHAI1IiDWrYgRVG8hM7507yM6TegK5cXSZQcnzkKBvEAAlmcUk+JWJ200qPSb64rEE76Jv4AwNK5WkS7lpEAa0G9yItMY2AA3NIO6IEz7WEgs5x2AO2wBqGf1o3K2TRmO9wYOxBO12Yd01n00NvzOsLxtLmkhzjSGd5GuhgBkd9TICfAJnPYRZ+kI1PX6NDmkiDwHgQyLtIzYxN87ayEWPonp+ZZ7gEB5HgmobXuWF7SA+T4vleg6gd+NLfkaeYFkW1paMRdoxDA7AhvA+AJOynDRJeN6YN2Xp+gGRrga2jQdpmQzSKwokcSJolel+cHKghsYMtGlb4h4vzIQCaqYMhND3PM0gmpph5+imyikFKpjmFAygyUuXrIYMBmLD+-6AcBCEOaKKa2YZ6r5oWtI+OSf6UbaSLYLqYDsLEpJeh6N4IPgUUxd28qHIlsVNG0UIwt2EgwGFnARdxIGNnx3ygUiMDRelCzduQnbdvFiXJYgqXRX2oF1cOjVVWOQhXJO0gAD6DbI0jDilaW9ROU4dXORrWQmDGJYM3UxTQ5COUuE3RYM21rRtJbKpVMWsNgpK0WVP40ntWGPjdYp5JwAWIKBS5qiKq1DYNr1nnt2m6Q9i0JdF-1fT9S60fRjHMax3x-VY8wNUtIMI384PKmqNDo5+f06UQ2O-S1iU0Aj+NXZ+yo1sjUnYatNC3PM3x-M1rW2Tp3y5flEUs8TqPMwTNKbRTNImqt6Oi9Tt1A8TDNM6BaONoRPlkaaVXBZZ1HSFciCcAA0ogGI3pI+AQIwRWQSbjDJtIUCkvA8DSGqxum3p0jO4wrBtgbCtLFz7CtD4EDeD4dI-FAEAyt2XseEj+DLJg8Dkiz8eJ4g3bwLigjyj4tUwLErxFQ2pbRtHyYSiA0dWcVREAUBiDLuWpfiuXHIeEgVdF02LaQQJ0HCaJFwMSnScSawKXD-XDgLeeQ8J5SeTj3P9cprb9tfTbLUT3ZK92w7arvM+JqL6nyE-Qhbag83T4ALQd+j4cygeBvzKwZoB0Hvh0t8baxxPCufs5Ou1cKYP0QE-Dw8xt5Awnj9JWRE-y11ciXA2l9jIgAgMoKW0cSa6SvsoSyOZgHF3LD9EU3w4YR0QKwRAkhwgemGuWUB1DaECHSsMOsNBv4G0GNHV+GcwCIG+LcOsuCN6HD-vLIhyomEskYBAWhWQWGcAADKZyEL4LhMdpAZw5Oonw-9lSAKQeWD6s9U7TxtpQ1gmB8D4EUeEVRujs6aPTmo7OBj4GK0QfXBCIpcT5yuFg7hgNlA6JtM+bCpcQlkEuIwTBkTuFCVAQtUB4DRHH3ngvXejsnyYKPpvJeCCXI+OjKBDJy8slr2GmUgpqcy5gQTsBapP5NH1OUGAN0MB4mPjIeU9cjpBDbl3OU4880pEyMQHI2hkxOBPXIF6a0gjo4GKMfXTu10rFmhmXMhZggXFiIkYrdWSwNxOhSJQn47tXFON8LnAJhcgyNGALibsA9sCB1wP+PI7tTofK6Y2JEoCqQ-h+fJeppQKhVGyD9NUJswCkmYAIVgpzBDFEQAAD04OcmUFtTasBYlcDxxpyz+NeIMUlgSGGgswIEvBIAXjZiJaqaQcKEXhGRQMxAWQkCIqxbEb4sQ1gbG7NSq4ucpbPJgFPH6JpWW8o5ZuQQ3LJnhFxR7Al4rsKSulRrGkpUoJCXeUHGAC0-bv2Dl-UBUdkq727Ea3AyFrlZ1ucie5QVdVgW7kiOqeR7gDk6m7S2yL2CUHwCyKAU1+pTlkHOFJVjVBCHwM6ENDFLqfhRYgbFgj0ZquDaG8N61RFgg8F6RAyws25oivmoQha6xk2FmEvR9aKYUp+kSlZh1SJ+VzUChelCFZUVCuFdgPwwrhs4MlShv8s3JxndotxLqKUPMbAheK5aLlX3HlmmSqyG5puVGurNHkbbZIfKBQ9ViwVXwhZUdI2Rbo1PEVm-F9JwXlFvdUMZ6y90XplPimlHkF6b2fQSgETDQET12sBy9NKilANXUPI9O8qkjV-VQglgGDkgZpXGmUgwx3sszHgEOPRZSXGzeTDNFa4bQbw3kYtpb104q3f25m86bk51dQXZtW1KHyw6tGNDx7V7wAsQRpFsj5EuhTfgKmcdn3JIInB4C36kTcGwMsXWchyqNDXVp-Ah0aTBgYvrQkILGgpVM4Zt2umGLmvSmh35xqdMxBSvZqDT6rH2v+S56B+mHOIasVWhi4bfPj38x5pjVDgthqENZ4z+AAASm58CdjCy1ZLNi0vmZiPFYomBKArSHv5yNA1r75HS4cfLhWgPiJK+OKN0hysFA9aWPOZK8gUvXixv9GG6UMojD+fVjQ6pCWAKPeK9nXmiTc386SUjiH7s-BWNpIAoip0wQwn1aDjCIFMAqZ8W3VsRT217Az4ZyZvUsY-A6l3iWTb+bWzDs3jWFrbQt5UZqPmfxUla2Qsd7OFpZoDjaTqm1ccJbApTrXpCHBTfXGjhxMupdELV6rDtaYMWR2NPIeWCuiY+whHN8VTODG1nrA2cnNNyBgDQeKyOjxGRPSh9GJPgkpnJ6ZxH1P8C05Sgzpl0gLE0i9I6NsUOYc0j9kt36U7WfFZp3TrHKWX5Wvl3HfzSuksq542ebnmv+cq8GAxstFaCbhcV4bmxc5EYEz04r+nKWbfozrejRt2dddLlbeTIlVElxw8oAjq3KO0f46lo7rLqOgbo4sUTu7cnSd5E55T+3vOtfo8Z9bETYM7ts4xIDZPv3N4G5ahnwXwvpCi83OL8mfuzzS-vnL3PCu0948oKrpvwtU+07b-AW38eLdp5Shn43BJGNm7u9dYvlvS-4+vtffvwsZ4a4dwxdH8+hafld3d93tz0be639DgTGFldZaBL3ixAekCMIyylnHYiY8fY2X+ojn9SPgQo0v6QVHKEy+FvrmfJHI3ejMfU3X-QfPnW-a3GcRnT3CmXfHOAWCHAmfmSfE-FfVLXHFvHvU-fAFCTHDAsePjO7QXOBL8bxPdYPaQIEYffHS-eHdA7HKPC-J-a7KhSTWhZNUNKnTXCPfAG3IgmUUggiCg79PVL1ASA2AgQYYYFmUzaQvIWQ0aLA7XGxf1e-Xvf1WNVghCBPbhFQnnWnDaT2b2YTXeCvGkaOAgLXUzURbvN7O7MgzxH6K-IPKAkPB-Og1gvPMnHWLnHgr4Aw3g3Axne+FqRPLWPwynAmAAofdwotUAqLWIyAoAmxRGPzOQOsF3azL3UpNAyzKQ-AGw5+HPL-RHUnMUZnDHO6cI-QjnKIlSewvgvCPmF3VgxbHw5DAnMQkBKxDg6Tbg1PQQxAQXJcJws8PguaNA0IMwW2F6RwnI2SaMAow2SpB2BhRzK9FMeIZIO9LICwn9EzOoqog4qXYdP-JfNXNApcVPcPFvYYuArvFvR4kBafOI1IvAl4z8BAr4s8A-L-IXJA4PKYr-cYm43AkEimXdHo3I9AxPFeWojwU46QBva46RTvAE5fHnB4pArE-TX4raN4lIhnAk5UH43E5AtE0YwmQ4eQoolY0RJ6QKNEyYmcAmaEpA3Q9THnHhQomgBkgESokTZEs4gqdgC4r-K4zE3jF-WAN-MjFkVcDTfTB4jI1LAg7E5JUkpcbk-EikqfOrQAkk-U9jZ1RAtEr3drN4C0wEm0lY6whkwYJk+YzEjkm01E6U9EyOE024jUlUqUz0qo7U6RIk-k+I4Mmkckm0trAJCk6ksY9k83dwyEgBCXFwhguTJglQx-GEvdQvchW4oI1fVQ-gmAusLPcw9otg1gforg1NIYxTBY2vRY4bGIUbJUKw-ACvBCe0oojaCs+2A4iZKZAY1NTswtYY6k8YsgpWOvLtJUbnBTDdNggdEKRobzZNTALMBQ8scSLiQEYBVswcPuf4PddyQGLyFCec-yZk8QiCA1JUfctyOyC8tSZWPyZ045TWE3KLMOWjNOaBbdQEI0HreuQGZ0qWUCzDUgRpeubCUC0CNUeSeMUC-9WlNBUyKAcyaARUT8NUchUxeTC5ZpL-RzZgORAc9Y76NAjc9gLc7WWTMiyZCAWdKxciiAcss9NAgswLOjLDGDK4Rcti5iusOEZiyLZ9di1Ap4oioQmVfimUUCE0QQTFMqcMHMNcmIFkMADwDpTNZcihIQ4CxscDShYoV9a9d9KFLIRC2kDFTgQyqhCDJeSdOSn8E0DNJIxvP9UBFIVwfFCAZRCAQ4XwZ0HcQRaSimSELoXoPoF4+EZFBOeARyvNBiLNGAbsbS3SpAbdF4nY6yl40BNtTS7-TlX8m8DVFEv5bsGLcNbseSVykYg874TuDVI0bzZtaOE0bzEwmOI0WqoQZtBqkCpvHVLQDFbAS4S8JEdin4SqqS5q8bea8Ssa+UdFSakOGzGIFStifiVORq7ifqsqitdbEeaK7ofobsc67KaQG9Aqlc20Caqara6QEdCqmlO1aq7-GTcNCETKKYGYQ6xsTyitSqjq5EKsFSAaqAdSb66tJUE0GgaGo8R69a56pEVQH4TuAMyQcgFqZtY6GAB0cgbAPUDKo0TmYdclfAZtTMcgXa862KsmKeA8zuJEd2XaC5bCLKvSitRsozRoQmuYIWCqFiQQR5GIBZGAMzCW6QUkbAfAJ0P81KMWpq-0KRBCDkJWwGCaVW2PFcLWwQDzWIPWyigESmsUwYQm4m0miydKcNGUGITrAzcmTWqqG2smgEIWvIP2NKb4Qm7sa2jMW2iwf2vWnKXEP2QXBCXGlqKWEUKWpEgCKANkckb4B2l1b4ROp056LIpyd1WWjOp28sIGpcRy-Dc4n6JhcTLFDEmUtOH6K5H6NVQYWOw4MOp0bfb4hdc05bbOvIGCuudtH2AFRoBAwYTC7ChcmhcIBaC2iKGMENGAZhOhGgGepFeSdI9ezgDtD1a8mXDmSOqmvIee8U34Kez641Uus8ZCpkTAKe1bSevUXCimWFI+sU8+vUDuwQS+3APO4WE0JSdET+iyakpC+keMe+vUVBcuJ+iyF+4WN+vKc4qBiyf6bsVB6AHBeM12TByXJfPhcgARb4KwHBpcOEHACUpfb4UgdFIqWhqWBh8C56Rh9Ff6Q+5BsU7sWh7BiMuWhWp0EBrBnSMh5Uakk0eWxWvZPB4eisbkHe3XU++aL8owQCPbMmkQAAFgeBQHuEcGcBAAYGYBEECGCH4CzgiDQB5A2svAAHIdhcB9gdAwB8AI0YA2wvB762xWAZRJAdBMBSAhREBbGtBrHnrgBVB-xAmXZbHWAdBPBg5-AghbHpADhpB7HZlHGDgXG3GPGwowBvHfH4n6Q-BsBsAQnuRexnquAeAwJA4rYMndgYAnGcnX48mvGfGaFing4M5OAKnVAWqRqZRZUIB4V5VydlVeUegPAABJWTWx1QWxyKtuk0JmA8mp+AAAA0XFEEwGviJvUU0B-F2evi9AgHRRRIgBgHposiUBAGvisGkHuFYAAFYmt6hnxA4xbbnlB7hpBNG3n7hMFIB4BLg7mABiTR50Z0WoDIAATg0F0D2bOfRSOaXBObynYF8Hrkmuub1Dub+asFYFuDeevhecwT8AkHOjuaJZJeBYQDBeUHBayH5FuAyBeZspAHUCRf2ewCxZZDReVBOZ0ozmlF8EubxZ+ZAEeeeYAHZSWPnmQbFyBqXfnXnMEsXyArguAaX1XnwQXGWQBwX+R+RnRbAXmMhEX-Hr4RXMxHbBWaQTnsAsB5QJWbnoACWnmmtNGKWIBvmPXlBr44W-mQ3MFDg5BJw7mfXnxNXtXOAo36XQWfAIXZWehnQAAOSYaN7l6151rcxAB1saPZ9xjEA15NplrIZ0Stytq14tjxwVnlg5+UdQbZqAFm1MDWVRkwDR6wFAWoewJwFwYxtgMAd0PgGY8IEQRweYBwIAA

### DIFF
--- a/index.js
+++ b/index.js
@@ -101,13 +101,13 @@ var patchProperty = (node, key, oldValue, newValue, listener, isSvg) => {
 }
 
 var createNode = (vdom, listener, isSvg) => {
-  var props = vdom.props
+  var {is, ...props} = vdom.props
   var node =
     vdom.type === TEXT_NODE
       ? document.createTextNode(vdom.tag)
       : (isSvg = isSvg || vdom.tag === "svg")
-      ? document.createElementNS(SVG_NS, vdom.tag, { is: props.is })
-      : document.createElement(vdom.tag, { is: props.is })
+      ? document.createElementNS(SVG_NS, vdom.tag, is && {is})
+      : document.createElement(vdom.tag, is && {is})
 
   for (var k in props) {
     patchProperty(node, k, null, props[k], listener, isSvg)
--- a/index.js
+++ b/index.js
@@ -101,13 +101,13 @@ var patchProperty = (node, key, oldValue, newValue, listener, isSvg) => {
 }
 
 var createNode = (vdom, listener, isSvg) => {
-  var {is, ...props} = vdom.props
+  var props = vdom.props
   var node =
     vdom.type === TEXT_NODE
       ? document.createTextNode(vdom.tag)
       : (isSvg = isSvg || vdom.tag === "svg")
-      ? document.createElementNS(SVG_NS, vdom.tag, is && {is})
-      : document.createElement(vdom.tag, is && {is})
+      ? document.createElementNS(SVG_NS, vdom.tag, props.is && props)
+      : document.createElement(vdom.tag, props.is && props)
 
   for (var k in props) {
     patchProperty(node, k, null, props[k], listener, isSvg)


jorgebucaran:
Do you have any idea what would happen if we just  createElement(..., props)? 😏

zaceno:
> Do you have any idea what would happen if we just createElement(..., props)? 😏

Not sure but I assume browsers would handle it ok, but custom-element libraries like https://github.com/dmarcos/document-register-element (which a-frame uses) would break the same way because if there is an object given as the second argument they do not expect `props.is` to be undefined. And they would be right, at least according to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement):

> An optional ElementCreationOptions object, containing a single property named is, whose value is the tag name of a custom element previously defined via customElements.define().

index.js
```diff
@@ -101,13 +101,13 @@ var patchProperty = (node, key, oldValue, newValue, listener, isSvg) => {
...
-  var props = vdom.props
+  var {is, ...props} = vdom.props
   var node =
     vdom.type === TEXT_NODE
       ? document.createTextNode(vdom.tag)
       : (isSvg = isSvg || vdom.tag === "svg")
-      ? document.createElementNS(SVG_NS, vdom.tag, { is: props.is })
-      : document.createElement(vdom.tag, { is: props.is })
+      ? document.createElementNS(SVG_NS, vdom.tag, is && {is})
+      : document.createElement(vdom.tag, is && {is})
```
jorgebucaran, line 110: Sorry for the late response. 

Perhaps `props.is && props` would be fewer bytes and work just as well. What do you think?

jorgebucaran (commented):
I usually prettify the code (via prettier). Do you mind doing that as well?

zaceno:
> I usually prettify the code (via prettier). Do you mind doing that as well?

No problem, I can do that. Would you mind sending your prettierrc (unless you are using their defaults?) I have prettier automatically deactivated in projects that don’t have a prettierrc file in them, because I rules differ and I always end up causing some unintentional diff. But I can apply your rules and just not submit the prettierrc file.

index.js
zaceno: I’ll give it a try!

jorgebucaran:
The only rule I have is no semicolons. 😅

index.js
zaceno: Good call – saves 8 bytes (minified and gzipped)!  :)

zaceno: And works just as well!

zaceno:
🎉

PR #1085, opened 2022-02-23
Flowchart

kofifus:
Added a flowchart showing a birds eye view of the different hyperapp architecture components, the relationships between them and the general flow of a hyperapp app.

Linked the flowchart from the reference page under "Architecture"

Flowchart is in mermaid.js markup and rendered by github automatically :)

The emphasis was on making a relatively simple flowchart showing a flow from the input (init, user events & external events) to output (new DOM, new subscriptions & effectors logic). The chart is intended as a learning aid for new users rather than a precise reference.

In the interest of clarity some things were left unclear or no included ie:
- `init: [state, ...effects]` does not technically generate an Action like the chart implies, however the intention is clear and having a precise representation would make the graph _much_ more complicated
- cleanup (`dispatch()`) and modifying `dispatch` are not included as they are separate issues and complicate the chart significantly

Hope this helps

### DIFF
--- a/docs/architecture/flowchart.md
+++ b/docs/architecture/flowchart.md
@@ -1,0 +1,32 @@
+```mermaid
+graph TD
+    userInteraction((user<br/>interaction)) -->|onclick| viewEvent[view<br/>event]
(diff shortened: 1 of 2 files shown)


jorgebucaran:
This is so cool! Thank you, @kofifus!

PR #1090, opened 2022-03-16
Less class patching

### LINKED ISSUE #1078: Injected classes gets removed when using object/array to define class props
There is an issue where if you use the object/array method to define your class prop it "overriders/patches away" other injected classes on state updates, which could come from consumers of the developed app.  ie. maybe there is an app loaded in as a microservice and someone would like to enhance some css and do this by injecting directly into the dom.

Here is a jsfiddle which shows the issue:
Description: modifying the class prop value by an "external" source adding the class `green`  
This is done on 3 components with object/array/string defining the class prop, a state update will cause a patching of the components were object/array is used and thus the externally added class is lost, while using a string will not cause the diff.
https://jsfiddle.net/azhvnyeg/2/

From discord (https://discord.com/channels/804672552348680192/804672553095528470/927872567518298152):

> zaceno — Idag 14:38
> @MrFluFFy OK I understand what you're asking now, and think I can explain what's going on.
> 
> Since you dispatch an action and change the state each time one of your buttons is focused, then after you add something to the classList,  hyperapp will diff the DOM. And if hyperapp detects that one of the button's class attributes has changed, it will reapply it, thereby removing the class you added using the event handler.
> 
> As your example shows, hyperapp thinks that your button's class prop has changed when you use the array- or object- method of defining a class prop. But not when you use a string. That is, simply because and inline-defined object will have a different reference each time hyperapp diffs the vdom. A plain string (being a primitive value) is seen as the same - hence no change detected – hence hyperapp doesn't fiddle with the classes.

zaceno:
Fixes #1078 

When defining classes in "object form" (i e: `h('div', {class: {foo: true, bar: false, baz: true}})`) hyperapp converts that form into the usual string form (`"foo baz"`)  _but_ not until _after_ it has compared the current class prop to the previous vnode's class prop. 

Since `{} !== {}` there will always be a diff when comparing. Meaning this conversion, and subsequent patching of the DOM node will occur on _every_ update wether it is actually needed or not. 

This PR fixes the situation by normalizing the class property to a string directly in the `h` function. This way class-props are only patched in the DOM when they have actually changed. 

(Also I snuck in a small improvement regarding how keys are handled to help me keep the byte-diff down)

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,7 +15,7 @@ var enqueue =
 var createClass = (obj) => {
   var out = ""
 
-  if (typeof obj === "string") return obj
+  if (!obj || typeof obj === "string") return obj
 
   if (isArray(obj)) {
     for (var k = 0, tmp; k < obj.length; k++) {
@@ -69,8 +69,7 @@ var patchSubs = (oldSubs, newSubs = EMPTY_ARR, dispatch) => {
 var getKey = (vdom) => (vdom == null ? vdom : vdom.key)
 
 var patchProperty = (node, key, oldValue, newValue, listener, isSvg) => {
-  if (key === "key") {
-  } else if (key === "style") {
+  if (key === "style") {
     for (var k in { ...oldValue, ...newValue }) {
       oldValue = newValue == null || newValue[k] == null ? "" : newValue[k]
       if (k[0] === "-") {
@@ -89,11 +88,7 @@ var patchProperty = (node, key, oldValue, newValue, listener, isSvg) => {
     }
   } else if (!isSvg && key !== "list" && key !== "form" && key in node) {
     node[key] = newValue == null ? "" : newValue
-  } else if (
-    newValue == null ||
-    newValue === false ||
-    (key === "class" && !(newValue = createClass(newValue)))
-  ) {
+  } else if (newValue == null || newValue === false) {
     node.removeAttribute(key)
   } else {
     node.setAttribute(key, newValue)
@@ -348,10 +343,10 @@ var recycleNode = (node) =>
         node
       )
 
-var createVNode = (tag, props, children, type, node) => ({
+var createVNode = (tag, { key, ...props }, children, type, node) => ({
   tag,
   props,
-  key: props.key,
+  key,
   children,
   type,
   node,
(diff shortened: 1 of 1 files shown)


zaceno:
Hmm something weird happened with my git history it seems. Let me fix that

index.js
```diff
@@ -15,7 +15,7 @@ var enqueue =
 var createClass = (obj) => {
   var out = ""
 
-  if (typeof obj === "string") return obj
+  if (!obj || typeof obj === "string") return obj
```
jorgebucaran, line 18: What is the purpose of `!obj`?

jorgebucaran: Why `!obj? Is this change necessary right now?

jorgebucaran, line 18: Why `!obj`? Do we need this now?

zaceno: Because how I use createClass further down, obj could be undefined (for vnodes that do not have a class prop)

jorgebucaran: And why didn't we have that issue before?

index.js
```diff
@@ -348,10 +343,10 @@ var recycleNode = (node) =>
         node
       )
 
-var createVNode = (tag, props, children, type, node) => ({
+var createVNode = (tag, { key, ...props }, children, type, node) => ({
   tag,
   props,
-  key: props.key,
```
jorgebucaran, line 354: Nice! I do wonder if destructuring arguments will be "slower" than regular function arguments.

index.js
zaceno: Because we had this function call inside an if-clause in patchProperty (see the diff in patchProperty), so it would always have some value.

index.js
zaceno: I don’t know tbh. I expect it will matter hardly at all because props isn’t monomorphic. But could be worth running the krause bench on this to check.

index.js
```diff
@@ -362,8 +357,12 @@ export var memo = (tag, memo) => ({ tag, memo })
 export var text = (value, node) =>
   createVNode(value, EMPTY_OBJ, EMPTY_ARR, TEXT_NODE, node)
 
-export var h = (tag, props, children = EMPTY_ARR) =>
-  createVNode(tag, props, isArray(children) ? children : [children])
+export var h = (tag, { class: c, ...p }, children = EMPTY_ARR) =>
```
icylace, line 360: Question: Does `{ ...p, class: c }` work as well ?

If so it might be worth considering using it to have some symmetry with line 363.

jorgebucaran (approved):

index.js
jorgebucaran: Can we use a default value for `c` to avoid the `!obj ||` inside `createClass` and would it be worth it?

icylace (approved):

index.js
zaceno: Yes I fixed that now In my own way of checking that means this diff is just +2B (minified and gzipped), So I'd say it looks worth it ;) 

Caveat is I still don't know how this affects performance. I feel pretty confident this does not affect perf at all but I haven't verified. I could run the krause bench test but since we memoize for that test any perf impact of this wouldn't show anyway. 

I'd be happy to do a comparison without memoization, but I don't have the bandwidth for it the next week. For myself I'm fine awaiting those results before merging.

jorgebucaran: Thanks, @zaceno. I say we merge this and _see_. If perf is affected, we can go back to no default arguments.

PR #1091, opened 2022-03-25
Prevent unnecssary patching of class attribute

zaceno:
A second more correct attempt that doesn't enforce a `class` property on every vnode at the cost of just a couple more bytes


jorgebucaran (approved):

jorgebucaran:
Thanks, @zaceno! 🎉

PR #1092, opened 2022-04-11
remove setTimeout

kofifus:
This PR does two things:
1. remove the fallback to `setTimeout` if RAF is not available. In such rare environments RAF can be replaced externally.
2. move the assignment `busy=true` to a separate line to increase code readability.

I tested on a few of the available codepen examples simply replacing the hyperapp link with https://raw.githack.com/kofifus/hyperapp/patch-1/index.js

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,10 +7,6 @@ var SVG_NS = "http://www.w3.org/2000/svg"
 var id = (a) => a
 var map = EMPTY_ARR.map
 var isArray = Array.isArray
-var enqueue =
-  typeof requestAnimationFrame !== "undefined"
-    ? requestAnimationFrame
-    : setTimeout
 
 var createClass = (obj) => {
   var out = ""
@@ -380,7 +376,10 @@ export var app = ({
     if (state !== newState) {
       if ((state = newState) == null) dispatch = subscriptions = render = id
       if (subscriptions) subs = patchSubs(subs, subscriptions(state), dispatch)
-      if (view && !busy) enqueue(render, (busy = true))
+      if (view && !busy) {
+        requestAnimationFrame(render)
+        busy = true;
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


jorgebucaran:
This will cause a runtime error if rAF is not defined in the environment e.g., Node.js. Or will it not?

kofifus:
Yes it will.

The thinking is that for environments where RAF is not available (which is very rate), it needs to be replaced outside of hyperapp (which is an easy one liner which should be there for other code that may need RAF). It is not catered for by hyperapp internals.

zaceno:
If I read this change correctly, the runtime error in e.g node will only occur when you try to run an app with a view (i.e. headless mode is still safe). If you are using node and want to render a view you already need a simulated browser environment anyway or else there are plenty of things that will throw errors. I’d argue it’s the responsibility of that environment to provide an implementation of rAF just as it provides an implementation of document.createElement et c

sergey-shpak:
@kofifus @zaceno @jorgebucaran please consider prev discussion: 
https://github.com/jorgebucaran/hyperapp/pull/866

kofifus:
OK seems like this is much more involved then I assume, I think it's better closed for now.

zaceno:
@kofifus @sergey-shpak  Hyperapp has changed since that discussion. Effects/subs are no longer deferred. Running headless apps in e.g. node.js does not require a mocked requestAnimationFrame to be available. 

I even verified that by using the version of hyperapp proposed in this PR, and running the following code in node with no  errors reported and correct output produced.

```js
import { app } from "./hyperapp.mjs"
const Incr = (state) => [state + 1, () => console.log("Incremented ", state)]
const dispatch = app({ init: 0 })
dispatch(Incr)
dispatch(Incr)
dispatch(Incr)
```

sergey-shpak:
@zaceno yes, you are right, since effects are no longer deferred it makes sense to remove the setTimeout fallback :+1:

PR #1096, opened 2022-06-01
Fix typo in api url

ankitshaw625:
The url should point to the users resource instead of posts


ankitshaw625:
Closing this since there's already an open PR for this.

PR #1097, opened 2022-07-19
init: of app() needs effectful array too

markfirmware:


docs/architecture/actions.md
```diff
@@ -191,7 +191,7 @@ An array returned from an action carries [special meaning as already mentioned e
 
 There are a couple of options available:
 
-- Wrap the return state within an [effectful state array](state.md#state-with-effects).
+- Wrap the return state within an [effectful state array](state.md#state-with-effects). Mention also that init: option of app() function must also be wrapped.
```
jorgebucaran, line 194: @markfirmware I'm not clear on this. What are you trying to say here?

jorgebucaran (changes requested):

docs/architecture/actions.md
markfirmware: I do not think that you mention anywhere that ```init: [state]``` is also needed if you want the state to be an array.

This is a draft pull request that needs to have the "mention ..." replaced by something appropriate.

jorgebucaran: But the state can't be an array, can it? Hyperapp will think you meant an effecter.

markfirmware: Actually the state can be an array if you always return ```[state, ...effecters]``` from your actions. see [Transition array state](https://github.com/jorgebucaran/hyperapp/blob/main/docs/architecture/actions.md#transitioning-array-state)

The PR here is to provide guidance that the same must be done in ```init:```

If you use ```init: state``` where state is an array, then you are right that the tail of the state will be taken to be effecters.

PR #1107, opened 2023-02-08
provide examples that work

zaceno:
The original example referenced dispatching an action within an augmented dispatch function, which is harmful. This proposal shows how augmented dispatch "middleware" can be used in a safe way to tap in to dispatched actions and state updates. Also demonstrates how multiple predefined middlewares can be combined.


jorgebucaran:
Thank you, @zaceno! 😁

PR #1108, opened 2023-02-22
docs: Fix a typo in Components Returning Multiple VNodes example section

dmylian:
I think I've found a small typo in the docs in a section where there's a view that leverages a function returning multiple nodes.
It reads as:
> However, to make use of such components you'll need to spread their result.

But the example lists no apparent spread of the result:

> `const view = () => h("div", {}, finishingMoveOptions())`

I assume the intended line is: `...finishingMoveOptions()`? I'm doing a tiny project of mine and that seems to be the case.

### DIFF
--- a/docs/architecture/views.md
+++ b/docs/architecture/views.md
@@ -120,7 +120,7 @@ const finishingMoveOptions = () => [
   h("button", { onclick: BefriendHim }, text("Friendship")),
 ]
 
-const view = () => h("div", {}, finishingMoveOptions())
+const view = () => h("div", {}, ...finishingMoveOptions())
 ```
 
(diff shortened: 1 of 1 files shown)


zaceno (commented):
Good catch! 

Unfortunately the fix is not quite this simple because the original is actually proper (h takes children as an array). 

So the original example is correct but fails to make the point being explained.

Going to see if I can make a bit more elaborate example.

zaceno:
See #1109 for what I think is a more appropriate fix.

jorgebucaran:
Fixed in #1109. Thank you, @dmylian and @zaceno! 💯

PR #1112, opened 2023-03-21
Enhance augmented dispatch examples

kofifus:
I propose enhancement to the augmented dispatch examples:

- changed example 1 to handle case where action is defined inline

- change example 2 to separate `stateMiddleware` which can be useful in other scenarios (ie immutability) and then uses it for logging in `logStateMiddleware`. (credit to @zaceno for `stateMiddleware`)

- added example 3 that uses `stateMiddleware` for an augmented dispatch for immutability which can be invaluable when starting in HA or during development

### DIFF
--- a/docs/architecture/dispatch.md
+++ b/docs/architecture/dispatch.md
@@ -93,7 +93,7 @@ const logStateMiddleware = stateMiddleware(state => {
 
 ## Example 3 - Immutable state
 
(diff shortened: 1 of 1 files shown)


jorgebucaran (approved):

jorgebucaran:
Thank you, @kofifus! 💯

PR #1113, opened 2023-05-06
Ignore return value of unsubscribe handler

mshgh:
It is silently assumed unsubscribe handler has not return value, however if it has it compromises subscriptions processing. To make the code more robust I propose to ignore the return value and push into `subs` always `false` _(it would work with `unsubscribe` too, but `false` is shorter ;)_


mshgh:
I was asked to cover this change by tests, but I am not sure how to approach it. https://github.com/jorgebucaran/hyperapp/issues/1105#issuecomment-1537121750

